PR #72, opened 2014-12-19
Interface changes in genseq

dmitriy-serdyuk:
Changed interface of SequenceGenerator and Emitter for beam search.


rizar:
What should with this PR, @dmitriy-serdyuk?

dmitriy-serdyuk:
I'm going to contunue the work and port to blocks the beam search I implemented for our inner project.

rizar:
Okay, I keep it open. 

I do not want to rush you, but this is the main missing part of the sequence generation framework (apart that it is badly documented, but I am working on that). So I will have to do it myself if you do not start soon.

rizar:
Sorry! Pressed wrong button.

PR #73, opened 2014-12-19
New selector interface

janchorowski:
Hi,

this is my first shot at a more feature-full selector interface. It is not meant to be merged, but please comment. 

It supports pretty much everything I could think of:
`/` to go down one level 
`//` to go down zero or more levels
`name` to match against names
`[regexp]` to regexp match against names
`@attribute==value` to check object attributes (to e.g. select bricks that have biases, that were initialized etc.)
`.container` to select stuff from the container (to select e.g. parameter)

The `selection` is always a `set` of selected objects or a `dict` mapping paths to selected objects. The `Selectors` take a selection and enumerate selected items. 

Usage examples are in the tests. the syntax is slightly changed wrt Dima's selectors, but this is much more flexible. In fact, I only assume that things have a `children` attribute and a `name` attribute.

I think it is an important part to get right, so feel free to comment


rizar:
Nice Jan, I really like your new selectors. I am sorry for ignoring this pull request for so long. 

I will try to find time for a proper code review, meanwhile you can rebase it and make tests pass.

janchorowski:
Thanks,

one thing I am thinking about is how to select individual attributes and 
how to recourse into attributes. So that you can write a selector "give 
me all paramaters with names..." or "give me all rngs that...".

I will write some examples in the test cases and ask you and Bart to see 
if you like the selector mini-language.

rizar:
Why can't the user just select all the parameters and then choose the ones he needs?

Anyways, I think you have more than enough functionality for the start.

janchorowski:
He can.

The idea is that when you write:
"//.parameters"
it creates you a set of all paramaters (there is nothing special in the 
name params, just as well you could have written "//.children" to select 
all children. The semantics for the ".member" selector was to return the 
contents of member, which has to be a container.

Now say a brick has the field "rng". If you write ".rng" it fails, as 
rng is not a container. I would like to somehow be able to write "get me 
the set of all rng members".

rizar:
I am still willing to finish this at some point though.

bartvm:
Surer, I just figured that it'd be better to make an issue out of instead of keeping a PR open which probably is so out of date by now that it needs to be rewritten anyway.

PR #86, opened 2014-12-24
Docstring test

bartvm:
Since other people are now starting to contribute it might save some time if we have some automatic checks to make sure that docstrings are correctly formatted, so that we don't have to point these things out for each PR. This PR does exactly that. It uses the [pep257](https://github.com/GreenSteam/pep257) library, for which I implemented NumPy support: https://github.com/bartvm/pep257/compare/numpy

The following checks are now performed, in line with NumPy docstring standards and PEP257 standards (when they don't conflict):
- Docstrings must start with a one-line summary ending with a period, followed by a blank line
- Docstring must be indented at the same level as the code
- Docstrings must use `"""` (e.g. not `'''`), and must use `r"""` if the docstring contains backslashes, `u"""` if it contains UTF-8 characters
- No blank lines before or after the docstring
- There must be a blank line before the closing quotes
- Lines cannot exceed 75 characters
- The sections must be in the right order, and a section header cannot be followed by a blank line

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -30,10 +30,11 @@ install:
       fi
   - |
       if [ $FORMAT ]; then
-        pip install -q flake8 pep257
+        pip install -q flake8
+        pip install -q git+git://github.com/bartvm/pep257.git@numpy
       fi
 script:
   - if [ $TESTS ]; then THEANO_FLAGS=floatX=$TESTS,blas.ldflags='-lblas -lgfortran' coverage run --source=blocks -m nose2.__main__ tests; fi
-  - if [ $FORMAT ]; then flake8 blocks tests; pep257 blocks --ignore=D100,D101,D102,D103,D203,D204,D401; fi
+  - if [ $FORMAT ]; then flake8 blocks tests; pep257 blocks --numpy --ignore=D100,D101,D102,D103; fi
 after_script:
   - if [ $TESTS ]; then coveralls; fi


rizar:
Hmm, I should say that I find the 75 characters restriction extremely inconvenient. I can not use automatic alignment  anymore because out of a sudden max. widthes are different for docs and for the rest of the code. Do we really need that?

bartvm:
I prefer just following the docstring standard instead of making exceptions, because if everyone makes exceptions it's not really a standard anymore :p I personally never found it too much of an issue once I knew which vim shortcuts to use, but if it really bothers you just make a PR to reverse it.

rizar:
Which vim shortcuts do you use then?

bartvm:
I seem to have lost the shortcut I was using during the re-install I did over Christmas, but it basically just did `set tw=75; gq; set tw=80` on the selected text. I guess that the clever thing to do though would be to write a script that basically does this on a dosctring so that you can format them in one go, but I'm not very good at Vim magic :)

```
set tw=75
for line in lines:
    if len(line) > 75:
        select lines until indentation changes
        gq on this selection
set tw=80
```

rizar:
Well, changing 'tw' every time works, but this quite a hassle.

bartvm:
You can just create a macro right? `qd`, `set tw=75`, `vip`, `gq`, `set tw=80`, `q` to create the macro and then just type `@d` to wrap a docstring paragraph correctly.

PR #89, opened 2015-01-07
Sparse init

bartvm:
Just separate this out from #61 and added unit tests. This introduces sparse initialization as an initialization scheme.


bartvm:
@rizar Let me know if you want to have a look before merging; it's a pretty straightforward PR and I've used it successfully in some experiments, so I think it should be okay.

rizar:
Just go ahead and merge! :)

PR #90, opened 2015-01-07
Monitoring channels

bartvm:
This is a cleaned up version of @janchorowski's code, with all of the compilation/readout code taken out, since there are lots of other plans on how to do that now.

Since the `ModelProperty` effectively covers any kind of Theano variable monitoring that doesn't need aggregation, the only other case is non-Theano monitoring i.e. functions that need to be called intermittently and its results logged. I added a class for that, but I don't see much point in going much further considering that the plan now seems to have the main loop and its extensions responsible for collecting the updates and storing them in a log, which is what @rizar said he would start working on.


blocks/utils.py
```diff
@@ -97,6 +100,26 @@ def shared_floatx(value, name=None, borrow=False, dtype=None):
                          borrow=borrow)
 
 
+def shared_expression(expression, name=None):
```
janchorowski: mabe in the spirito of numpy's zeros_like, change it to shared_like?

blocks/monitoring/__init__.py
```diff
@@ -0,0 +1,193 @@
...
+"""
+import logging
+from abc import ABCMeta, abstractmethod
+
+from blocks.utils import shared_expression, update_instance
+
+logger = logging.getLogger(__name__)
+
+
+class MonitorChannel(object):
```
janchorowski: Why introduce this dummy class?

blocks/monitoring/__init__.py
```diff
@@ -0,0 +1,193 @@
...
+    name : str
+        The name of this monitor channel, should be unique in order to
+        distinguish channels.
+
+    """
+    def __init__(self, name):
+        update_instance(self, locals())
+
+
+class VariableMonitorChannel(MonitorChannel):
```
janchorowski: Why remove the name property?

blocks/monitoring/__init__.py
```diff
@@ -0,0 +1,193 @@
...
+    object.
+
+    """
+    def __init__(self, function, **kwargs):
+        super(FunctionMonitorChannel, self).__init__(**kwargs)
+        assert callable(function)
+        update_instance(self, locals())
+
+
+class AggregationScheme(object):
```
janchorowski: The aggregation scheme should know the expression it aggregates e.g. to know its name.

blocks/monitoring/__init__.py
```diff
@@ -0,0 +1,193 @@
...
+        numerator_accumulated = shared_expression(self.numerator)
+        denominator_accumulated = shared_expression(self.denominator)
+        initialization_updates = [(numerator_accumulated, 0.0),
+                                  (denominator_accumulated, 0.0)]
+        accumulation_updates = [(numerator_accumulated,
+                                 numerator_accumulated + self.numerator),
+                                (denominator_accumulated,
+                                 denominator_accumulated + self.denominator)]
+        return VariableMonitorChannel(
+            aggregation_scheme=self,
```
janchorowski: Here you should pass a name. I didn't do it, because the name was the expression's name.

blocks/monitoring/__init__.py
bartvm: So that you can do `isinstance(channel, MonitorChannel)`

janchorowski:
Hi Bart,

thanks for cleaning mu stuff. I am OK with your changes, though I guess it will be easier if we:
do not introduce the name monitoring and channel in this file. The aggregators only tell how to evaluate a theano variable incrementally. They are not monitors by themselves.
Also, I don't see the rationale for introducing the named MonitorChannel instead of referring to the expression. Even in PL2, a channel is a variable evaluated on  some data (a channel holds reference to a dataset). So calling a variable a channel is a little confusing.

blocks/monitoring/__init__.py
bartvm: Because it depended on the `name` property of the `expression` attribute, but the `expression` attribute didn't necessarily exist (it was added manually by the `aggregated_div` function), so it would have crashed if you instantiated the `AggregatedDiv` class yourself.

blocks/utils.py
bartvm: SGTM

blocks/monitoring/__init__.py
janchorowski: No, it was an argument to the constructor and the AggregationScheme. It was always required and never added after construction.

janchorowski: And AggragatedDiv being a subclass of AggregationScheme coulnd;t have been built without an expression

blocks/monitoring/__init__.py
janchorowski: But the aggregator only adds semantics to a theano variable by telling how to deal with it on a large dataset. It should not be automatically monitored, nor imply to be.

blocks/monitoring/__init__.py
bartvm: Ah, I missed that it was positional argument. Either way, isn't it kind of strange to expect the user to pass `(a, b, a / b)` in the constructor, with the final result still being `name = None` by default?

janchorowski: The name is not None, it is whatever theano has set for the name of the 
expression.

I know it is strange to pass |(a, b, a / b), but this is the price for 
having the expresion as a constructor argument, rather than relying on 
it being set once the object os constructed. And the user wll never do 
it himself|

blocks/monitoring/__init__.py
bartvm: I'm not sure what the best alternative would be. `Aggregator` doesn't make any sense to me, because `ModelProperty` doesn't aggregate anything, and in many cases the variables monitored will be of this type. I think it will be very counter-intuitive for a user to be asked to use an `Aggregator` in order to monitor static model properties such as weight norms.

Something like *statistics* or *info* as suggested by @rizar could make sense. But taking the weight norm, `W_norm = tensor.mean(W)`, and then saying that we provide a "statistic" of this variable is slightly counter intuitive for me as well (it would suggest you are providing the mean, variance, etc. of this variable).

I find monitor channel a better option already, because it's simply a channel that *can* be monitored (which is what the single purpose of this entire class is after all, to provide the infrastructure that allows us to monitor things). The fact that it's called a monitor channel (as opposed to monitored channel, or monitored variable) doesn't necessarily imply that it *is* monitored, merely that it can be.

The only thing the monitoring extension now has to do is collect the aggregation schemes and apply them to different datasets to get a set of monitoring channels. These monitoring channels can then be monitored and their results added to the log. I don't think this would confuse anyone.

blocks/monitoring/__init__.py
bartvm: The name of an expression is always `None` by default. Try
```python
a = tensor.vector('a')
b = tensor.vector('b')
c = a / b
print(c.name)
```
```
None
```

janchorowski: OK, my bad. I thought theano assigns something more useful in this case.

janchorowski:
Bart,

I pushed most of your fixes to my PR. Why would I prefer that you merge it and then introduce the MonitorChannel class? Because the aggregators are a concept by themselves. 

Let the monitor channel depend on the aggregators like you did, but it is cleaner to keep the aggregators as a concept and build the monitor on top of it. I changed names accordingly. I also left the classes to do the evaluation on datasets and as updates. I can remove them (though I wanted to have a working commit with them because they are useful for testing and they show how the aggregators can be used.)

blocks/monitoring/__init__.py
bartvm: Sadly not. We could use `str(c)` instead, but I'm not sure whether `Elemwise{true_div,no_inplace}.0` is any better...

blocks/monitoring/__init__.py
janchorowski: On 1/7/2015 4:33 PM, Bart van Merriënboer wrote:
>
> In blocks/monitoring/__init__.py 
> <https://github.com/bartvm/blocks/pull/90#discussion-diff-22591977>:
>
> > +
> > +   Monitored Theano variables might depend on inputs beyond those given to
> > +   the model. There should be a way of providing these to the monitoring
> > +   channels, collecting them, and compiling them as part of the Theano
> > +   function that performs monitoring.
> > +
> > +"""
> > +import logging
> > +from abc import ABCMeta, abstractmethod
> > +
> > +from blocks.utils import shared_expression, update_instance
> > +
> > +logger = logging.getLogger(__name__)
> > +
> > +
> > +class MonitorChannel(object):
>
> I'm not sure what the best alternative would be. |Aggregator| doesn't 
> make any sense to me, because |ModelProperty| doesn't aggregate 
> anything, and in many cases the variables monitored will be of this 
> type. I think it will be very counter-intuitive for a user to be asked 
> to use an |Aggregator| in order to monitor static model properties 
> such as weight norms.
>
Please look into the evaluator classes. The aggregator only adds 
semantic information to an expression. The evaluators silently 
instantiate an aggregator - the user just says that he/she wants to 
monitor the value. The model property is just a convenience to make the 
implementation of the evaluators easier.

The user never has to do anything about the model properties by himself.

> Something like /statistics/ or /info/ as suggested by @rizar 
> <https://github.com/rizar> could make sense. But taking the weight 
> norm, |W_norm = tensor.mean(W)|, and then saying that we provide a 
> "statistic" of this variable is slightly counter intuitive for me as 
> well (it would suggest you are providing the mean, variance, etc. of 
> this variable).
>
> I find monitor channel a better option already, because it's simply a 
> channel that /can/ be monitored (which is what the single purpose of 
> this entire class is after all, to provide the infrastructure that 
> allows us to monitor things). The fact that it's called a monitor 
> channel (as opposed to monitored channel, or monitored variable) 
> doesn't necessarily imply that it /is/ monitored, merely that it can be.
>
Any theano variable can be monitored. The aggragators only say HOW to 
compute something on a large dataset.
>
> The only thing the monitoring extension now has to do is collect the 
> aggregation schemes and apply them to different datasets to get a set 
> of monitoring channels. These monitoring channels can then be 
> monitored and their results added to the log. I don't think this would 
> confuse anyone.
>
No, the monitoring extension has to select appropriate expressions. This 
is why I have written the new selectors - you can select based on tags, 
atttributes, names, brick membership etc.

The aggregator doesn;t say something is to be monitored, it says how to 
compute the value on a large dataset by small minibatches.

janchorowski:
Bart,

I guess I was very confusing with starting the aggregators as monitors. I am sorry for that.  So please read this with a clean mind. 

Ideally, the monitor would receive, among others, a list of theano expressions to monitor. We have selectors to select them. However, by itself a theano expression doesn't tell how to compute its value when the whole dataset doesn't fit into RAM/gpu.

Thus enter aggregators: all they do (for all the fuss I made about them) is tell how to compute an expression on a large dataset in small batches. That's it.

The DatasetEvaluator class (which is tasked with evaluating a set of expressions on a given dataset) tries to do the right thing:
- when the expression is tagged with the AggregationScheme, it is used
- when the expression doesn't depend on the data (which is determined by walking the graph), a ModelProperty (that's a misnomer, maybe it should just be called DataIndependent) is instantiated. This is just convenience for further code
- when the expression depends on the data, we have a choice of raising an expception, or taking per-batch mean and warning the user (current default).

Thus so far the only aggregator useful to the end-user is the `mean`. A user does not need to explicitly tag stuff with the ModelProperty.

Your MonitorChannel should be something above the expressions+aggregators, bridging together other stuff one wants to monitor, such as python functions. Maybe we should have an ExpressionChannel (an expression to be monitored, potentially tagged with an aggregator to know how to deal with large data) and a FunctionChannel (a function to be monitored)?

blocks/monitoring/__init__.py
bartvm: I didn't say that the monitoring extension needs to collect *all* monitoring channels. It can obviously choose a subset of them. A *channel* (or aggregator in your preferred wording) doesn't mean it *is* monitored, merely that it *can* be.

You say that aggregators refer to how variables are monitored on large datasets. The counter-example I mentioned before is the `ModelProperty` class you wrote, which suggests that "aggregators" aren't just used to monitor variables over multiple batches, but *also* to monitor other Theano variables which don't rely on data. The `ModelProperty` doesn't add any useful information on **how** to the variable should be monitored, the only thing it does it tagging a variable in order to signify that it *can* be monitored.

The way I see it we can do two things:

1. We can *get rid off* the `ModelProperty` class and treat variables which don't aggregation entirely differently, in which case `Aggregator` makes sense. In that case we need another way of attaching variables that can be monitored but don't need data to the application call.

2. The other solution is the one I tried to convey here: A single class (`VariableMonitorChannel`) is used to designate that a Theano variable is monitorable (not monitored!). These channels have "optional" aggregation schemes, the `ModelProperty` class being used for variables which don't need any. In this case I feel that we need to avoid the word `Aggregator` because for a large subset nothing is actually aggregated.

janchorowski: Bart,


On 1/7/2015 5:16 PM, Bart van Merriënboer wrote:
> I didn't say that the monitoring extension needs to collect /all/ 
> monitoring channels. It can obviously choose a subset of them. A 
> /channel/ (or aggregator in your preferred wording) doesn't mean it 
> /is/ monitored, merely that it /can/ be. 
Ok, sorry. The misunderstanding is because for me everything can be 
monitored. The aggregation is in a sense just a workaround for memory 
limitations.
>
> You say that aggregators refer to how variables are monitored on large 
> datasets. The counter-example I mentioned before is the 
> |ModelProperty| class you wrote, which suggests that "aggregators" 
> aren't just used to monitor variables over multiple batches, but 
> /also/ to monitor other Theano variables which don't rely on data. The 
> |ModelProperty| doesn't add any useful information on *how* to the 
> variable should be monitored, the only thing it does it tagging a 
> variable in order to signify that it /can/ be monitored.
>
the ModelProperty is just a convenience class, such that evaluation code 
is simpler. Instead of having ifs, I created a dummy aggregation scheme 
that wraps variables that don't need aggregation. It is quite common 
practice in programming. It can be a completely hidden, internal to the 
evaluator class. All its purpose is to make writing evaluation code easy.

Let me give you a totally unrelated example of the practice. Say you 
look for the index of something in an array. You append what you look 
for at the end of the array. Then you are guaranted to find it and you 
don't have to have ifs to detect the end of the array (people would do 
it for speed). In the same spirit, in my selectors there is the 
TerminatorSelector 
https://github.com/janchorowski/blocks/blob/selectors/blocks/select.py#L297 
which raises parse errors.
>
> The way I see it we can do two things:
>
> 1.
>
>     We can /get rid off/ the |ModelProperty| class and treat variables
>     which don't aggregation entirely differently, in which case
>     |Aggregator| makes sense. In that case we need another way of
>     attaching variables that can be monitored but don't need data to
>     the application call.
>
So lets completely remove the ModelProperty form the public interface 
and keep it for internal use only. If something does not depend on data, 
then there is no need to force the user to provide an aggregation 
scheme. If the evaluator is implemented like I did it, the ModelProperty 
can be a thing totaly internal to the evaluator, just a programming 
convenience.
>
> 1.
>
>     The other solution is the one I tried to convey here: A single
>     class (|VariableMonitorChannel|) is used to designate that a
>     Theano variable is monitorable (not monitored!). These channels
>     have "optional" aggregation schemes, the |ModelProperty| class
>     being used for variables which don't need any. In this case I feel
>     that we need to avoid the word |Aggregator| because for a large
>     subset nothing is actually aggregated.
>
I prefer to keep the aggregators to say how to evaluate variables 
separate from the concept of monitoring.

bartvm: Okay, I think we actually agree, I was just thoroughly confused by your use of the `ModelProperty` as an `AggregationScheme` and on how you expected people to use it, but if you just use it internally to simplify code and add it silently it makes more sense perhaps.

I'll change the names back and maybe write some documentation. Also, if your idea is just to add some information on the **how** to aggregate, we don't need to name aggregators, instead we can just name the resulting expressions using the `add_monitor` method.

janchorowski: I did all of it in my PR  - rabesed against current master and I'm 
working on fixing Travis complaints right now to save you some time. The 
ModelProperty is no longer there, I changed it to _DataIndependent, to 
shoiw users that it is not useful to them.

Jan

bartvm: Okay, I'll work of that then. I don't want to merge it yet because your PR (unlike this one) contains a bit more code which interacts with other parts of the library (which don't necessarily exist yet) e.g. as far as I can see functions such as `_process_batch` assume that data sources have names and that they are the same as the Theano variables.

bartvm:
Closing this and branching @janchorowski's PR instead.

blocks/monitoring/__init__.py
janchorowski: I know, it has the evaluators. I wanted to have them in VC to have a full view of things for later. Feel free to branch off and remove them, or I can push something without them.

You are correct about the assumption about datasources. It worked well in GH and seems natural to me -- the dataset yields dicts, we match stuff by name. It also solves your ..todo in #90 - the evaluator walks the graph to find all inputs (not just those of the model) and assumes that they will be present in what the dataset yields.

bartvm: I think datasets should definitely have sources, it worked well in GH apparently, and it did in Pylearn2 as well. I was just pointing out that this hasn't been implemented yet. Working on that now!

janchorowski: > I think datasets should definitely have sources, it worked well in GH 
apparently, and it did in Pylearn2 as well. I was just pointing out that 
this hasn't been implemented yet. Working on that now!

Well GH is very limited - it just assumes that the data iterator yields 
dicts and that all theano input variables' names somehow match the keys 
in the dict. There is no mechanism to select what you want form a 
dataset, which is in PL2. On the other hand PL2 assumes that you will 
have many different sources with the same name, but different format, 
thus the batch cannot be a dict but has to be a tuple that you zip with 
the sources to get to the components. I find GH approach more intuitive, 
precisely because there is no source selection magic like in PL2 and 
there is no assumption that stuff with the same name can be different.

bartvm: I don't understand what you mean with having many different sources with the same name? If you mean across different datasets, then it is true that the source is always called `features`. However, I don't think that is much of a problem. We can't guarantee global uniqueness anyway, so if people want to combine multiple datasets they might as well prefix the dataset name in front of the source.

The ability to ask a dataset for specific sources only might be worth considering. It would be a shame if the user is forced to read large amounts of data from disk just to reject them afterwards (e.g. if a dataset consists of images, tags, and a category, and we only want the tags and category).

janchorowski: On 1/7/2015 6:27 PM, Bart van Merriënboer wrote:
>
> In blocks/monitoring/__init__.py 
> <https://github.com/bartvm/blocks/pull/90#discussion-diff-22600412>:
>
> > +
> > +   Monitored Theano variables might depend on inputs beyond those given to
> > +   the model. There should be a way of providing these to the monitoring
> > +   channels, collecting them, and compiling them as part of the Theano
> > +   function that performs monitoring.
> > +
> > +"""
> > +import logging
> > +from abc import ABCMeta, abstractmethod
> > +
> > +from blocks.utils import shared_expression, update_instance
> > +
> > +logger = logging.getLogger(__name__)
> > +
> > +
> > +class MonitorChannel(object):
>
> I don't understand what you mean with having many different sources 
> with the same name? If you mean across different datasets, then it is 
> true that the source is always called |features|. However, I don't 
> think that is much of a problem. We can't guarantee global uniqueness 
> anyway, so if people want to combine multiple datasets they might as 
> well prefix the dataset name in front of the source.
>
In PL2 you can ask the dataset to provide the same source in different 
spaces. Say you ask for tragets in 1-hot and integer representation (or 
however these are called in PL2). Both are called "features", though. 
Thus you cannot return the batch as a dict. I spoke once about it with 
Pascal Lamblin.
>
> The ability to ask a dataset for specific sources only might be worth 
> considering. It would be a shame if the user is forced to read large 
> amounts of data from disk just to reject them afterwards (e.g. if a 
> dataset consists of images, tags, and a category, and we only want the 
> tags and category).
>
Yes, this ability is useful. I don't know whether it should be automatic 
(as in PL2, or we walk the graph, find the inputs, and ask for their 
names), or manual.

PR #91, opened 2015-01-08
WIP: Datasets and monitoring

bartvm:
This is a continuation of #85 which now includes datasets, data streams, iteration schemes, data iterators, etc. I can't remember all of the exact details from our discussion at NIPS, but I believe this comes relatively close. It's still WIP, but I want to start running Travis and such (also because my computer or internet back in Montreal seems to have died, which makes it a bit harder to run my own tests). Hopefully this will finally get us a bit closer towards Bricks 0.1!


blocks/datasets/__init__.py
```diff
@@ -0,0 +1,223 @@
...
+        Returns
+        -------
+        tuple
+            A tuple of data.
+
+        """
+        raise NotImplementedError
+
+
+class DataStream(object):
```
janchorowski: We have a name clash with @rizar #92:
what he calls a data stream is an iterator over iterators,
what you call a data stream is a stateful iterable, each created iterator goes over an epoch.

We should settle on one, I prefer this one - a stream implements the `__iter__` protocol. Especially in extensions that do monitoring it feels weird to code:
```
iter = next(stream)
for batch in iter:
```
I prefer:
```
for batch in stream:
```

blocks/datasets/__init__.py
```diff
@@ -0,0 +1,223 @@
...
+        return DataIterator(self,
+                            iter(self.iteration_scheme)
+                            if self.iteration_scheme else None, self.sources)
+
+    @property
+    def sources(self):
+        if getattr(self, '_sources', None) is None:
+            return self.data.sources
+
+    @sources.setter
```
janchorowski: Do you intend that changing the sources property effectively changes what is returned by the stream?

I like it, I am just making sure I understand.

bartvm: Yes, although you normally wouldn't set it manually, you would pass it to the constructor. The reason I am using a property is because when `sources` is `None`, it nicely propagates up through all the data streams, to the point where it gets the `sources` attribute from the `Dataset`.

blocks/datasets/__init__.py
```diff
@@ -0,0 +1,223 @@
...
+    def sources(self, sources):
+        self._sources = sources
+
+    def open(self):
```
janchorowski: How do you intend to use it?
Shall I first create a dataset, then the streams, or shall I first create the stream, then open the dataset through them?

Since many streams will be attached to the same dataset (e.g. the training data will be typically accessed randomly in small batches to train and sequentially in larger batches to validate), maybe it is more natural to have the Dataset create the DataStreams, the streams are very lightweight then.

blocks/datasets/__init__.py
bartvm: I need to look more closely at #92, but I agree that `for batch in stream` looks more natural and intuitive, and is what I had in mind.

blocks/datasets/__init__.py
bartvm: My idea was to use it like this:
```python
mnist = MNIST('train')
data_stream = DataStream(mnist, SequentialScheme(mnist.num_examples, 10))
```
The `open` method gets called by the `DataStream` on construction. (I just noticed that currently `reset` gets called directly after `open`, need to change that.)

I'm not sure I understand why many streams will be attached to a dataset. The training and validation datasets are different instances of the same class, so I intended them separate in the way they are in e.g. Pylearn2:
```python
mnist_train = MNIST('train')
mnist_test = MNIST('test')
train_stream = DataStream(mnist_train, ShuffledScheme(mnist.num_examples, 10))
test_stream = DataStream(mnist_test, SequentialScheme(mnist.num_examples, 1000))
```
If you want to attach multiple streams to a dataset you could of course do that:
```python
mnist = MNIST('train')
stream_1 = DataStream(mnist, ShuffledScheme(mnist.num_examples, 10))
stream_2 = DataStream(mnist, SequentialScheme(mnist.num_examples, 10))
```

bartvm: Ah, I see you were referring to validating *on the training set*. In that case the last example I gave would be applicable.

janchorowski: I was thinking more along the line:

mnist = MNIST(....)
randomized_train_stream = mnist.get_stream(train, randomized, ...) #to 
be used for training
sequential_train_stream = mnist.get_stream(train, sequential, ...) #to 
be used for monitoring

It's a matter of taste though, so I will not convince you more.

On the other hand, I would like that we specify what the consumers of 
DataStreams will have access to. Can we assume the following:

class DataStream():
     @abstractmethod
     def __iter__(self):
         pass

     @property
     def sources(self):
         pass

     @sources.setter(self, sources):
         pass

I like it because:
1. It is very easy to write a custom datastream if my data ressemble 
other datasets
2. Technically, if I just hardcode ("features", "targets") for sources 
and disallow changing I don;t have to deal with them in 90% of cases - 
this again makes writing specific DataStreams easier.

rizar:
Now in the right pull request :)

So it all looks interesting, but I do not really understand the idea of your data stream, so I will start with an example. To train a phoneme recognizer on TIMIT I need three stages of data processing:

1. Read examples in random order.

2. Group them in batches with utterances roughly equal in terms of length (here some sorting is involved).

3. Apply some acoustic preprocessing, e.g. switch to spectrograms.

My plan to do that was the following: first I create the dataset object. Then I create the first data stream that fetches from the dataset using a random fetching schemes. Then I wrap the first data stream by a second one, which the examples yielded by the first one in batches. The third data stream applies a transformation to the data returned by the second one.

So the scheme in my plan belonged to the first data stream of the chain, the two others did not have one.

Now I see your code and it feels as if the scheme should to the top-most data stream. Could you explain what's the right way to implement a pipeline described above using your data streams?

janchorowski:
So for me it will be nice to separate two concepts:

1. The interface of the datasets to other parts of the library. Namely, 
what do we gitve to the trainer, what do we give to monitors, etc.
2. The convenience routines to write data iterators.

With respect to 1., the interface should be very minimal. Ideally, just 
the __iter__ protocol. Maybe a listing of source names. In my dream 
toolkit, the __iter__ function would assemble the pipeline.

With respect to 2., a library of stream processing routines is nice, but 
nearly impossible to snapshot. I have an exemplary implementation in 
https://github.com/janchorowski/data_iterators). Here there are classes 
that match exactly your steps: 
https://github.com/janchorowski/data_iterators/blob/master/example_kaldi_data.py#L147 
does 1., 
https://github.com/janchorowski/data_iterators/blob/master/example_kaldi_data.py#L187 
does 2., and 
https://github.com/janchorowski/data_iterators/blob/master/dataset/iterators.py#L185 
does 3.

bartvm:
@rizar I'll add some examples, but I believe we have exactly the same idea. The scheme in this case would belong to the first data stream. The other two won't necessarily have one, because schemes are optional.

I'll try and write an example of what you need and add it to the documentation or the tests. Right now, to give you an idea, caching works like this:

```python
from blocks.datasets.mnist import MNIST
from blocks.datasets import CachedDataStream, DataStream
from blocks.datasets.schemes import SequentialScheme

mnist = MNIST('train')
data_stream = DataStream(mnist, SequentialScheme(mnist.num_examples, 100))
cached_data_stream = CachedDataStream(SequentialScheme(mnist.num_examples, 3), data_stream)
for batch in cached_data_stream:
    pass  # Do stuff
```

In your case you would wrap data streams as well, but they wouldn't have an iteration scheme.

blocks/datasets/__init__.py
bartvm: I'll stick with this one for now, else it gets a bit more complicated. In your code you would basically have to postpone all the loading of the dataset (e.g. loading MNIST into memory) until the construction of the data stream, but you don't want to load MNIST into memory separately for each data stream, so you would probably want to do something clever there about checking if it is already loaded... This seemed simpler.

The way I envisioned it is that the data set constructor determiens *what* data to load (e.g. `train` or `test`, a particular subset, the location of the files to load, etc.) and the data stream determines the iteration (order, caching, etc.).

janchorowski: In my scheme the data would be loaded when you instantiate the MNIST 
object. Maybe its not so good. I start to see how your solution is more 
easy to serialize (the dataset is easy to serialize - it holds no state. 
Then the DataStream knows how to open whatever it needs once deserialized).

Anyway, can we make it such the open() method is internal to the 
DataStream - its users (trianers and such) only depend on the __iter__ 
protocol? I just don't want to force on someone who has an iterator to 
go over the data  a complicated structure.

bartvm: Oh, but `open` is entirely internal to `DataStream`? It's only used by the `DataStream` to get e.g. file handles and then pass them back to the `Dataset` when needed, allowing the `Dataset` to remain stateless. Any external users only depend on the `__iter__` protocol i.e. `for batch in data_stream`. No other part of the library ever needs to call `open` or `reset`.

janchorowski: That's perfect! Maybe then to make it more visible, we could make a base 
DataStream class, and then drive yours from it? The base on would 
document what interface its users can depend on. The derived one will 
become the basis of the common dataset routines.

My problem was that seeing the base class with so many methods, I 
assumed that some of them may be required.

bartvm: But `open` is required even on the base class because it's used when the data stream is wrapped by another data stream. The parent data stream will call `open` on the wrapped data stream, which will propagate down all the data streams till `open` is called on the dataset.

janchorowski: Yes. But it will be more visible to the (human) users that the rest of 
the library doesn't depend on open, reset and other methods. If you make 
them part of the base interface, then you don't know whether you need to 
implement them, or not.

bartvm: Documentation and more implementations will make that clear.

rizar: I agree with Jan that what the interface data streams provides to the 
rest of the library should be isolated in a base class. I had the same 
confusions as he had.

B.t.w., the interface you propose looks nice when `for batch in stream` 
iterates over the same data every time called, but what if an epoch is 
just a part of the data, not the whole data? Can `__iter__` method of a 
data stream return iterators over different data?

blocks/datasets/__init__.py
```diff
@@ -0,0 +1,233 @@
...
+    def sources(self, sources):
+        self._sources = sources
+
+    def open(self, state=None):
+        return self.data.open(state)
+
+    def close(self, satte):
+        self.data.close(self.data_state)
+
+    def get_data(self, state=None, request=None, sources=None):
```
rizar: Why do we need a state parameter if it is always `self.data_state`?

blocks/datasets/__init__.py
```diff
@@ -0,0 +1,233 @@
...
+        if batch_size >= len(self.cache[0]):
+            self._cache(state, sources)
+        data = []
+        for i, cache in enumerate(self.cache):
+            data.append(numpy.asarray(cache[:batch_size]))
+            self.cache[i] = cache[batch_size:]
+        return tuple(data)
+
+    def _cache(self, state, sources):
+        for cache, data in zip(self.cache, next(self.child_iterator)):
```
rizar: This `child_iterator` should at least be documented as an attribute. It is unclear from the `CachedDataStream` code what it is and when it was created.

blocks/datasets/__init__.py
```diff
@@ -0,0 +1,233 @@
...
+    def __init__(self, data, iteration_scheme=None, sources=None):
+        update_instance(self, locals())
+        self.data_state = None
+
+    def __iter__(self):
+        self.data_state = self.data.open(self.data_state)
+        # TODO Allow for multiple child datasets/data streams, so that they
+        # can be combined
+        if isinstance(self.data, DataStream):
+            self.child_iterator = iter(self.data)
```
rizar: Doesn't `iter` call `self.data.open` under the hood? It is not nice that it is called twice.

blocks/datasets/__init__.py
```diff
@@ -0,0 +1,233 @@
...
+    Notes
+    -----
+    Datasets should only implement the interface; they are not expected to
+    perform the iteration over the actual data. As such, they are
+    stateless, and can be shared by different parts of the library
+    simultaneously.
+
+    """
+    __metaclass__ = ABCMeta
+    sources = ('features', 'targets')
```
rizar: I do not think we should specify default source names here. Let users do their job of naming their data properly. In addition when I open a dataset class I want to see sources names and it is rather surprising for me that they are inherited.

blocks/datasets/__init__.py
bartvm: Yes, I wrote it last night, haven't documented it yet and now traveling,
but will do it soon.

Bart

blocks/datasets/__init__.py
bartvm: Don't have the code in front of me but I believe it was to badge l make the
interface the same as Dataset, so that date streams can be agnostic as to
whether they are wrapping a data set or another data stream.

blocks/datasets/__init__.py
bartvm: It should only be called once right now? I'll double check the code when I
can, but it should only be called by the __iter__ of the data stream, once
before each epoch.

blocks/datasets/__init__.py
bartvm: Okay.

blocks/datasets/__init__.py
```diff
@@ -0,0 +1,233 @@
...
+        return tuple(data)
+
+    def _cache(self, state, sources):
+        for cache, data in zip(self.cache, next(self.child_iterator)):
+            cache.extend(data)
+
+
+class DataIterator(six.Iterator):
+    """An iterator over data, representing a single epoch."""
+    def __init__(self, data_stream, iterator=None, sources=None):
```
bartvm: Sure.

blocks/datasets/__init__.py
```diff
@@ -0,0 +1,233 @@
...
+    sources : tuple of strings, optional
+        The sources of data to return. By default, all sources of the
+        dataset are requested.
+
+    """
+    def __init__(self, data, iteration_scheme=None, sources=None):
+        update_instance(self, locals())
+        self.data_state = None
+
+    def __iter__(self):
```
bartvm: Ehm, once again, don't have the code in front of me, but I believe that for
MNIST that should just make 60,000 epochs. I'll have a look.

bartvm:
Yeah, sounds like we are. I suggest you guys give me a list of crazy data
pipelines you think should be supported, and I'll try to implement a bunch
of them. Best way of seeing whether this works in practice!

blocks/datasets/__init__.py
rizar: Have a safe trip!

I can not think of any usecase for this agnosticity. Looks like data stream can be of two types: those that initiate data flow from a dataset and those that transform already existing streams. E.g. your `CachedDataStream` in practice can only wrap another data stream as it uses a `child_iterator`.

Based on these thoughts I propose the following changes:

* make `get_data` use `self.data_state` and fetch the next request from the request iterator. The only parameter left would then be `sources`.

*  the `DataIterator.__next__` then becomes that simple:
```python
def __next__(self):
     return self.data_stream.get_data(self.sources)
```

* this new `get_data` method can be used instead of `next(self.child_iterator)` by transforming data streams like your `CachedDataStream`

rizar:
Cool! Examples from me:

1) TIMIT pipeline described in this pull request. The challenge here is to decide how to serialize the data stream that reorders data. I would suggest to add a `is_picklable` routine to the data stream interface, make all data streams support it. It is reasonable to allow a data stream be unserializable once in a while.Then we can somehow handle it in the main loop or in an extension.

2) Data-driven epochs, be it different wikipedia documents or samples from a random distribution. Here I kind of dislike that

```python
a = list(stream)
b = list(stream)
assert a == b
```

fails. This is counter-intuitive. I understand that most frequently it will work well making it sensible to support the `__iter__` interface as you currently do, so perhaps we need to add something to the interface to account for such cases. E.g. in addition to `__iter__` we can have a more generic `next_epoch` interface.

```python
a = list(stream.next_epoch())
b = list(stream.next_epoch())
# it is OK that a != b
```

But this can wait: for version 0.1 it is OK to support only casual epochs as long as we keep in mind that we should take care of more complex cases.

janchorowski:
On 1/9/2015 3:45 PM, Dmitry Bogdanov wrote:
>
> 2) Data-driven epochs, be it different wikipedia documents or samples 
> from a random distribution. Here I kind of dislike that
>
> a=  list(stream)
> b=  list(stream)
> assert  a==  b
This will also fail with the common datasets and randomization. By 
iterating twice over the stream you should get examples in different order.

rizar:
Good point Jan! Perhaps we should not support `__iter__` at all to prevent such surprises.

By the way, iterating multiple times over MNIST in different random order goes to the list of crazy pipelines requested by @bartvm. For this an iteration scheme should span across multiple epochs, which as far as I see is currently not supported.

rizar:
Nice Bart, I kind of like what we ended up with! I do not mind merging this pull request - it is already quite big. Just a tiny piece of nit-picking: I would prefer a name `AbstractDataStream` to `_DataStream`.

B.t.w., I should be able to add training algorithms to my main loop pull request tomorrow. We are approaching something the very first complete version!

janchorowski:
Cool,

I am quite busy until Thursday, but I will prepare a configurator for 
then - a class that reads selectors and configures bricks. It's not 
super needed right now, but I guess that any reasonably sized experiment 
will to have most of the knobs in one place.

Jan

rizar:
Jan, what's your plan regarding your existing Selector pull request? If you rebase it and make tests pass I could try to merge it.

janchorowski:
If you and Bart like how the new selectors work I can rebase it and 
adapt your examples to use it. I was waiting for some feedback, and 
thinking how to make the selector language intuitive - it is already 
quite powerfull.

rizar:
OK, to be honest I have not really read it, perhaps I will do it tomorrow in the airplane and tell you my humble opinion.

bartvm:
Great! There is one last thing I'm struggling with: When to reset the state. If we want datasets to be plug-and-play with a single main loop, then they must reset their state automatically when needed. However, if we allow the case where one epoch doesn't go over the entire dataset, then we can't just reset the state after each epoch.

I see two ways around this:

1. We always reset the state after each epoch. This doesn't sound too extreme, considering the definition of an epoch is kind of exactly that: Having completed and starting anew i.e. resetting. We will have support for monitoring in between epochs, so that shouldn't be a problem.
2. The data set should implement a kind of `is_exhausted(state)` method, which takes the state of a data set and returns `True` if it needs to be reset (i.e. the next call to `get_data` won't return data anymore), `False` if we can go on (and maybe `None` if we can't tell, for data sets which can only be used in custom main loops).

(1) seems simpler, (2) could be more powerful.

janchorowski:
I would settle for option #1 - an epoch always resets. If one wants to 
do something more often, be it per sentence, per paragraph, etc he can 
use the per-batch callback and look at the data to know when to run.

bartvm:
Okay, let's stick with (1) for now. There are some weird scenarios which I can think of where something more complex could be needed e.g. with curriculum learning the data source might be different for later epochs. However, I think we can stick with (1) and implement something else later on e.g.

3. A midway solution which is easier but more flexible than (1): The data stream keeps track of which epoch we are in, and also passes this number on to the `get_data` method (and maybe also the `reset` method?). Crazy data sets can then use this information to open different files, read different data, etc.

janchorowski:
I guess that the solution is to have a per- epoch training extension 
which modifies the data stream, e.g. theels it to access more difficult 
examples.

It would be similar to e.g. learning rate scheduler.

bartvm:
A training extension might not be enough if you actually want to open different files to read the data from. But we can tackle this if/when the need ever arises.

bartvm:
To make sure that this all fits nicely with the main loop that @rizar is working on I'll merge this PR and defer extensive documentation and unit tests to a later PR (as usual... :p)

PR #92, opened 2015-01-08
WIP: Skinny main loop

rizar:
I would be especially eager to get some feedback about:

* the log interface, in particular whether we need the default values thingy altogether

* the `SimpleExtension` interface


blocks/extensions.py
```diff
@@ -0,0 +1,141 @@
+class TrainingExtension(object):
```
janchorowski: What about adding something like "on_timer", that would be fired every number of seconds or examples. GH uses something like that to save every 5 minutes for instance.

janchorowski: OK, sorry, I see that you have something more general below

blocks/extensions.py
```diff
@@ -0,0 +1,141 @@
...
+        if not key.startswith('_'):
+            def create_callback_overrider(key):
+                def simple_callback_overrider(self):
+                    self.execute(key)
+
+                def batch_callback_overrider(self, batch):
+                    self.execute(key, batch)
+                # TODO: use proper reflection here
+                return (simple_callback_overrider
+                        if key.find('batch') == -1
```
bartvm: Slightly more "Pythonic": `if 'batch' not in key`

rizar:
So it all looks interesting, but I do not really understand the idea of your data stream, so I will start with an example. To train a phoneme recognizer on TIMIT I need three stages of data processing:

1. Read examples in random order.

2. Group them in batches with utterances roughly equal in terms of length (here some sorting is involved).

3. Apply some acoustic preprocessing, e.g. switch to spectrograms.

My plan to do that was the following: first I create the dataset object. Then I create the first data stream that fetches from the dataset using a random fetching schemes. Then I wrap the first data stream by a second one, which the examples yielded by the first one in batches. The third data stream applies a transformation to the data returned by the second one.

So the scheme in my plan belonged to the first data stream of the chain, the two others did not have one.

Now I see your code and it feels as if the scheme should to the top-most data stream. Could you explain what's the right way to implement a pipeline described above using your data streams?

blocks/extensions.py
rizar: My plan is the following: we can have a `Timer` extension that a writes message to the main log when a period of time has passed since the last timer activation. Other extensions can handle this message. One should be careful with the order in which extensions are called in such a scheme of course.

rizar:
Here comes a draft of training algorithms. After giving it a thought I decided that it is better to keep the good old name `TrainingAlgorithm` from PyLearn2. The old part of the pull request is yet to be revised and I plan to add a test that uses `GradientDescent` in the `MainLoop`.

blocks/algorithms.py
```diff
@@ -0,0 +1,227 @@
...
+    -----
+        Changing `updates` attribute or calling `add_updates` after
+        the `initialize` method is called will have no effect.
+
+    .. todo::
+
+       Ideally the parameters should be extracted from the `cost`
+       computation graph. But in order for this to work our graph
+       traversal routine should recurse into ScanOp's, which is currently
+       not the case.
```
bartvm: Can't we walk the graph and collect all the bricks, and then query their `params` attributes? This should also give us all the parameters, no?

janchorowski: Even better: walk the graph and find all shared vars. A tag can be used to
indixate that a shared should not be trained. Lets not make the trainin
algo dependent on the brick concept. About the graph it has to know anyway.

15 sty 2015 03:13, "Bart van Merriënboer" <notifications@github.com>
napisał(a):
>
> In blocks/algorithms.py:
>
> > +    cost : Theano variable
> > +        The objective to be minimized.
> > +    params : list of Theano shared variables
> > +        The parameters to be tuned.
> > +
> > +    Notes
> > +    -----
> > +        Changing `updates` attribute or calling `add_updates` after
> > +        the `initialize` method is called will have no effect.
> > +
> > +    .. todo::
> > +
> > +       Ideally the parameters should be extracted from the `cost`
> > +       computation graph. But in order for this to work our graph
> > +       traversal routine should recurse into ScanOp's, which is
currently
> > +       not the case.
>
> Can't we walk the graph and collect all the bricks, and then query their
params attributes? This should also give us all the parameters, no?

rizar: The methods proposed by both of you will work as soon as we start going into the inner graphs of ScanOP's during the traversal done by the `ComputationGraph`. But it is not implemented right away, and I don't create to introduce dependencies between different tasks now. As soon as we have it, I am fine with getting rid of `params` at this point.

rizar: And as a side note: I like the idea that training algorithms should know nothing about bricks.

bartvm: On second thought. I think the best way to go is to keep it this way: Keep the parameters as an argument! This makes the training algorithm truly modular i.e. you can use it without the need to annotate the graph in one way or the other. I think it'd be great if people can actually build their own model entirely independently, and then just use a training algorithm class to get the Theano updates they need to perform. This is very much in line with the star-based dependency graph that we talked about.

janchorowski: +1

rizar: Agreed. If params are `None`, they can be retrieved from the computation graph, and this will be done in most cases.

bartvm: I would actually prefer it if the main loop is in charge of collecting the
parameters in all cases. I don't like the idea of having "soft
dependencies" everywhere; they tend to obfuscate functionality, and turn
into hard dependencies unnoticed.

I understand the desire for a minimal main loop, but I think we should
achieve this by pushing functionality into training extensions, not into
other parts of the library. Collecting parameters should never be the job
of a training algorithm.

rizar: I am not sure if is a good idea. The main loop knows nothing about the cost, in fact it is fully Theano agnostic. Besides, collecting parameters from a computation graph does not introduce too many dependencies: a training algorithm uses `ComputationGraph` which is a sort of very low level tool on top of Theano.

bartvm: Okay, this leads nicely on to another discussion... Do we need/want to be Theano agnostic? The first line of our documentation reads "Blocks is a framework that helps you build neural network models on top of Theano." I personally have no issue with assuming models are Theano based. I think we might needlessly complicate code by being Theano-agnostic if no one ever uses Blocks to build non-Theano models.

Also, I'd rather have the main loop making assumptions about the model (i.e. depending on an interface that returns parameters, interacting with `ComputationGraph`) than the training algorithm making these assumptions. The ideal IMO is still that each part of the library is completely independent and can be used separately from the rest, except for the main loop; the main loop is just a convenience utility that strings all parts from the library together based on certain assumptions.

rizar: We can not be and we are not fully Theano agnostic. But neither should we assume that everything is a Theano variable. That being said I would not like to see the assumption that the cost is a Theano variable in the main loop. For instance, that would forbid forever non-differentiable costs based on sampling.

I also like the idea that every part of the library should be useful on its own. But this is exactly why I want the training algorithms that take a differentiable Theano expression for cost as an input to have a default way to extract the parameters from the computation graph, and not rely on the main loop for this simple (assuming that using the `ComputationGraph` is allowed) job.

And just a side note: I did not mean `ComputationGraph` to be a model (and moreover _the_) model. I perceive it as a tool to work with annotated Theano graphs.

janchorowski: Hey,

I came up with another solution:

Make a Model base class, that would be subclassed by the user for each 
experiment.
Then the class would have methods as Model.get_params() and 
Model.get_grads() that by default scan the graph and call theano to get 
gradients.

Then: the common case is easy - the user just subclasses Model(). When a 
weird model comes in, the user has to override the functions that get 
paremeters and gradients. Then the main loop stays theano agnostic, and 
the trainer doesn't deal with parameter extraction and gradient evaluation.

bartvm: I'd prefer @janchorowski's solution over the training algorithm becoming responsible for collecting parameters using Blocks-specific interfaces, no matter how thin a layer the `ComputationGraph` wrapper is. I think that on the long term the benefit of this strict modularity will become only more clear.

blocks/main_loop.py
```diff
@@ -0,0 +1,90 @@
...
+                    self._run_extensions('after_batch', batch)
+                    self._check_finish_training()
+                self.log.status.epochs_done += 1
+                self._run_extensions('after_epoch')
+                self._check_finish_training()
+                self.log.status.last_epoch_end = (
+                    self.log.status.iterations_done)
+        except KeyboardInterrupt:
+            self._run_extensions('on_interrupt')
+        except StopIteration:
```
bartvm: This doesn't work. `StopIteration` is automatically caught and handled by the `for`-loop.

blocks/algorithms.py
rizar: I am not a big fan of the solution proposed.

First of all the name `Model` does not mean anything. It was one the most obscure concepts in both PyLearn2 and Groundhog. I did use the word model in the main loop class, but there it has very limited semantics of a handle to everything we want save for this experiment, with no assumption that a model has gradients or a Theano expression for cost or whatever else. So I guess that I can further discuss the proposal above using the name `Trainable` instead of the name `Model`.

Now we come to my main point: I believe there is no such thing as _the_ trainable. There are just no traits shared by all possible things that can be trained. I am looking forward to repeating the experiments by Google DeepMind people with a non-differentiable cost optimized by reinforcement learning: there are no gradients in this case. It is harder to point at concrete examples when some of my parameters are not Theano variables, but here comes another consideration: it is quite weird when the cost and the gradients are given by the user and parameters are provided by the main loop. This is one of PyLearn2's shortcoming: different components ask each other for different things and one struggles to understand the information flow.

So what I had in mind writing this code is: the user is fully responsible for configuring the algorithm. No obscure information passing through the main loop. And when it comes to a particular algorithm or to a family of them, this is where making assumptions is possible. And a base class of the family of algorithms, call it `DifferentiableCostAlgorithm` working with a differentiable Theano cost can provide convenience routines of collecting the parameters, because at this point it is quite reasonable to assume that those are shared Theano variables.

And this actually makes the `DifferentiableCostAlgorithm` usable without the main loop. I can just do
```python
algo = GradientDescent(cost)
for i in range(10):
   algo.process_batch(some_data)
```
without creating a `Model` or a `Trainable` instance. And I see no problem whatsoever that the `ComputationGraph` is needed for that, because there should be a lowest common denominator between all library components, and I do not care whether it is just Theano or Theano with our extension, such as the `ComputationGraph` is.

blocks/main_loop.py
rizar: I do not how, but it works for me:
```
> /home/rizar/LISA/pyhog6/blocks/tests/test_main_loop.py(35)test_main_loop()
     34     import ipdb; ipdb.set_trace()
---> 35     assert len(list(main_loop.log)) == 7
     36     for i in range(5):

ipdb> pp list(main_loop.log)
[(0, 'batch', {'data': 1}),
 (1, 'batch', {'data': 2}),
 (2, 'batch', {'data': 3}),
 (3, 'batch', {'data': 4}),
 (4, 'batch', {'data': 5}),
 (5, 'training_finished', True),
 (5, 'training_finish_requested', True)]
```

blocks/algorithms.py
janchorowski: Hey Dima,

I don't think I understood you correctly. We are considering an 
extremely popular use case here - how to train a model that is 
differentiable wrt its parameters. Even if there are use cases that fall 
outside of it we shouldn't stop supporting the common use case. Unless 
we will provide a good default implementation, we will end up with is a 
magical model setup code  that everyone will end up copy-pasting.

I think it is better to have a generic main loop and trainer and a 
Model/Trainable/Experiment class that is interfaced with them. No one is 
forced to use it. It is not a problem if someone uses just part of 
blocks. But all parts should be there for the common case.

blocks/main_loop.py
bartvm: Okay, I would need to look into why that works, because this is how a simple example works:

```python2
>>> class A(object):
...     def __init__(self):
...             self.n = 0
...     def __iter__(self):
...             return self
...     def next(self):
...             self.n += 1
...             if self.n == 5:
...                     raise StopIteration
...             return self.n
...
>>> try:
...     for i in A():
...             print i
... except StopIteration:
...     print "Ended"
...
1
2
3
4
>>>
```

rizar: I guess `for` loop only catches exception raises by the `__next__` method of the iterator.

bartvm: Do you mean this? It doesn't change anything. Mysterious.
```python2
>>> class A(object):
...     def __init__(self):
...             self.n = 0
...     def __iter__(self):
...             return self
...     def raiser(self):
...             raise StopIteration
...     def next(self):
...             self.n += 1
...             if self.n == 5:
...                     self.raiser()
...             return self.n
...
>>> try:
...     for i in A():
...             print i
... except StopIteration:
...     print "Ended"
...
1
2
3
4
>>>
```

rizar: No, I mean that this is gonna work:
```python
try:
   for i in A():
       raise StopIteration()
except StopIteration:
   print "Caught!"
```
for the reason that the exception is raised not in `A.__next__` but  in the loop.

bartvm: Ah, yeah, that would work. But it doesn't sound like a very good idea to rely on this behaviour at all... `StopIteration`'s definition is literally
> Raised by an iterator‘s `next()` method to signal that there are no further values.
So raising it from somewhere else seems a bit strange.

rizar: In fact this is what they do in PyLearn2. But okay, I will replace it with a custom exception.

blocks/algorithms.py
rizar: I fully support that we need a good default implementation for the case 
of differentiable cost. The question discussed is: how to achieve this 
goal. My point is: let the training algorithm decide what it wants to do 
with computation graph of the cost given to it, with no middlemen, be it 
the main loop or a "model" involved.

I have already added a stupid automatic fetching of parameters to the 
`TheanoTrainingAlgorithm` class.

rizar:
News:

* `TheanoTrainingAlgorithm` became `DifferentiableCostMinimizer` and illustrates quite well my vision of what is the responsibility of a training algorithm

* major refactoring of extensions: now all the logic is in the `do` method and different behavior can be achieved by specifying additional parameters

* a runnable MNIST training demo, unfortunate with no monitoring whatsoever as Bart removed evaluators from master

I guess the size of this pull request is coming to the point where decisions should be taken. I am waiting for your opinion on what and how should be merged.

After it is merged, extensions for monitoring and saving/loading will become the only major missing part.

blocks/algorithms/__init__.py
```diff
@@ -0,0 +1,239 @@
...
+    """A base class for all gradient descent algorithms.
+
+    By "gradient descent" we mean a training algorithm of the following
+    form:
+
+    .. code-block::  python
+
+        for batch in data:
+            for param in params:
+                param += step.compute(grad_wr_param_on_batch)
```
bartvm: Probably `-=`

blocks/algorithms/__init__.py
```diff
@@ -0,0 +1,239 @@
...
+
+    Parameters
+    ----------
+    step_rule : instance of :class:`StepRule`
+        An object incapsulating most of the algorithm's logic. Its
+        `compute_step` method is called to get a Theano expression
+        for the actual step to take for each parameter. Note, that
+        the step rule might have a state, e.g. to remember a weighed
+        sum of gradients from previous steps like it is done in
+        gradient descent with momentum. If ``None``, an instance of
```
bartvm: Nitpicking, but instead of saying "If `None`" (which suggests that the user is expected to pass that) you could say
```
step_rule : instance of :class:`StepRule`, optional
    Does this and that. By default, ...
```

blocks/algorithms/__init__.py
rizar: Currenly I have the minus in the `compute_step`. Do you think it should go here?

bartvm: Ah, no, code-wise I think it makes sense (since the step is in the negative gradient direction). I just read this docstring as saying "Gradient descent is when you do `param += grad_wr_param_on_batch`", thinking that `compute` meant "compute the gradient". It makes sense if you know the code, otherwise it might be a bit confusing, but it's minor.

blocks/examples/mnist/train.py
```diff
@@ -0,0 +1,34 @@
...
+
+def main():
+    mlp = MLP([Tanh(), Identity()], [784, 100, 10])
+    x = tensor.matrix('features')
+    y = tensor.lmatrix('targets')
+    cost = BinaryCrossEntropy().apply_for_indices(y - 1, mlp.apply(x))
+
+    mnist = MNIST("train")
+
+    main_loop = MainLoop(
```
bartvm: Might I just say that it's all coming together nicely! I very much like the way `main_loop` looks in the end.

blocks/examples/mnist/train.py
```diff
@@ -0,0 +1,34 @@
...
+from blocks.bricks.cost import BinaryCrossEntropy
+from blocks.main_loop import MainLoop
+from blocks.extensions import FinishAfter, Printing
+
+
+def main():
+    mlp = MLP([Tanh(), Identity()], [784, 100, 10])
+    x = tensor.matrix('features')
+    y = tensor.lmatrix('targets')
+    cost = BinaryCrossEntropy().apply_for_indices(y - 1, mlp.apply(x))
```
bartvm: MNIST doesn't use `BinaryCrossEntropy` (which suggests that there are two labels), but `CategoricalCrossEntropy` (which needs to be written still as a wrapper of `theano.tensor.nnet.categorical_crossentropy` instead of being a method of `BinaryCrossEntropy`).

blocks/algorithms/__init__.py
```diff
@@ -0,0 +1,239 @@
...
+        """Return inputs of the cost computation graph.
+
+        Returns
+        -------
+            list of Theano variables
+
+        """
+        return self._cost_computation_graph.inputs
+
+    @property
```
bartvm: Maybe I'm missing it, but this property doesn't seem to add anything to the behaviour of a normal attribute?

blocks/algorithms/__init__.py
```diff
@@ -0,0 +1,239 @@
...
+
+        Returns
+        -------
+            list of (Theano shared variable, Theano) expression tuples.
+
+        """
+        return []
+
+
+class DefaultStepRule(StepRule):
```
bartvm: Just a suggestion: `SteepestDescent` is a bit more descriptive.

blocks/extensions.py
```diff
@@ -0,0 +1,224 @@
...
+                self.do(callback_invoked, *(from_main_loop + tuple(arguments)))
+
+
+class FinishAfter(SimpleExtension):
+    """Finishes the training process when triggered.
+
+    Parameters
+    ----------
+    num_epochs : int
+        If not ``None``, training finish is requested after `num_epochs`
```
bartvm: Why is it optional? Without `num_epochs` the extension is useless, no?

blocks/main_loop.py
```diff
@@ -0,0 +1,96 @@
...
+
+    def __init__(self, model, data_stream, algorithm,
+                 log=None, extensions=None):
+        if not log:
+            log = RAMTrainingLog()
+        if not extensions:
+            extensions = []
+        update_instance(self, locals())
+
+    def _run_extensions(self, method_name, *args):
```
bartvm: Can't this functino (and `check_finish_training`) just be public? I don't see any particular reason to keep them private.

blocks/log.py
```diff
@@ -0,0 +1,241 @@
...
+
+class RAMStatus(AbstractTrainingStatus):
+    """A simple training status."""
+    def __iter__(self):
+        for attr, value in self.__dict__.items():
+            if not attr.startswith("__"):
+                yield attr, value
+
+
+class RAMTrainingLog(AbstractTrainingLog):
```
bartvm: Maybe we could just rename these to `TrainingStatus` and `TrainingLog`, basically signifying that the in-memory version is the "standard" implementation, and then we can come up with particular names for other implementations.

Personally just not a big fan of `RAM`; can't really explain why... Probably something to do with it invoking the feel of a very low-level implementation, while theoretically we're dealing with virtual memory anyway, not RAM.

bartvm:
Just read through it and made a few comments here and there, but it's all just small stuff. Overall it looks good to me. I have to say that I found the `SimpleExtension` thing a bit confusing though:

* The `MainLoop` calls `self._run_extensions('before_training')`.
* `_run_extensions` calls `SimpleExtension.before_training()`.
* `before_training()` has been overridden by `@replace_callbacks` to actually read `self.execute('before_training')`, which is called.
* `execute` checks if `'before_training'` is in `self._conditions`, and whether the optional accompanying `predicate` is `True`; if both are `True`, `self.do('before_training')` is called
* `do` now actually performs the action

I like to think there is a more straightforward way of doing this, but perhaps there just isn't.

I'm happy to merge once the merge conflict has been solved (and maybe some of my comments addressed, I can make a PR for that if you want).

rizar:
I read your description of it works and thought that this is indeed a real nightmare. Please check out a new scheme:

* `MainLoop` calls `TrainingExtension.dispatch`

* The overriden `dispatch` calls `do`.

I will address most of your notes myself. If you like, it would be nice if you write the `CategoricalCrossEntropy` class, as you did it in PyLearn2.

blocks/examples/mnist/train.py
rizar: Great! We will see how it looks with serialization and monitoring.

blocks/algorithms/__init__.py
rizar: Right, it does not. It is here mostly because I wanted to have a computation graph object as a private attribute and to have shortcuts to access it.  Maybe I will refactor this place later.

blocks/algorithms/__init__.py
rizar: I like `SteepestDescent`!

blocks/extensions.py
rizar: Technically no, just that it will have to be configured by calling `add_condition`. 

Anyway I think I will move `num_epochs` to the `SimpleExtension` constructor. It might be helpful for other extensions.

blocks/main_loop.py
rizar: Well, if it was Java or C++ they would be `protected`. I did not intend to give the user of the main loop any interface except for `run`. We can make them public if needed.

blocks/log.py
rizar: I am also not a big fan of RAM. Let's call them `TrainingStatus` and `TrainingLog`, although I think that this position should be in future occupied by a more mature implementation.

blocks/algorithms/__init__.py
rizar: I put optional in the description. As for the default values, I think it is enough to have them specified in the code.

rizar:
Okay, looks like I am done working in your suggestions. I will make friends with Travis and push the button.

bartvm:
Travis approves! Merging so that I can finish up #108

PR #93, opened 2015-01-10
WIP: revised Bart's datasets

rizar:
Here comes my revision of Bart's dataset proposal. The main changes are:

* I distinguish between two types of data streams, those that originate directly from a dataset and those that wrap existing ones. The former inherit from `InitialDataStream`, while the latter are descendants of `WrapperDataStream`.

* I change the interface to make a data stream behave just like Python file objects with an additional `next_epoch` method. As a consequence, the `__iter__` method does not reset iteration anymore.

Main open questions:

* do we actually need sources specification for the wrapping data streams? Perhaps it would be always enough to choose data sources when constructing the first data stream of the pipeline? 

* ~~some dataset states might turn out to be not picklable, e.g. a Python file can not be picked. Thus we need to add state serialization and deserialization routines to the `Dataset`, that by default will use pickling but can be overriden. E.g. for a file it is sufficient to remember the position in it. In addition we need to add smth like `is_pickable` to the `DataStream` interface.~~ By overloading `__getstate__` and `__setstate__` of the dataset's state we can ensure that the state is always serializable.

* iteration schemes should be extended to be able to span across several epochs. Adding smth like `next_epoch` method that will be called by the `next_epoch` method of a data stream should do the job.

* ~~I am not sure that a tuple is the best format for the data produced by a data stream. I would prefer something named, e.g. a named tuple or even a dict.~~ Perhaps a tuple is a good inner representation, a dict may be returned on request.

### DIFF
--- a/blocks/datasets/__init__.py
+++ b/blocks/datasets/__init__.py
@@ -9,13 +9,14 @@ from blocks.utils import update_instance
 class Dataset(object):
     """A dataset.
 
-    Dataset classes implement the interface to a particular dataset.
+    Dataset classes implement the interface to a particular dataset. The
+    interface consists of a number of routines to manipulate so called
+    "state" objects, e.g. open, reset and close them.
 
     Attributes
     ----------
     sources : tuple of strings
-        The sources this dataset can provide. By default, these are
-        ``features`` and ``targets``.
+        The sources this dataset can provide.
 
     Notes
     -----
@@ -26,51 +27,51 @@ class Dataset(object):
 
     """
     __metaclass__ = ABCMeta
-    sources = ('features', 'targets')
 
-    def __iter__(self):
-        """Use the default iteration scheme to construct a data stream.
-
-        .. warning::
-
-           A dataset only produces a single data stream using the default
-           iteration scheme, so multiple iterations in parallel are not
-           supported.
-
-        """
-        if not hasattr(self, 'default_scheme'):
-            raise NotImplementedError("Does not provide a default iterator")
-        elif not hasattr(self, 'default_stream'):
-            self.default_stream = DataStream(self, self.default_scheme)
-        return iter(self.default_stream)
-
-    def open(self, state=None):
+    @abstractmethod
+    def open(self):
         """Return the state if the dataset requires one.
 
         Datasets which e.g. read files from disks require open file
         handlers, and this sort of stateful information should be handled
         by the data stream.
 
-        Parameters
-        ----------
-        state : object
-            The state of the dataset after the previous epoch.
-
         Returns
         -------
         state : object
             An object representing the state of a dataset.
 
-
-        Notes
-        -----
-        The first time the dataset is opened this function always receives
-        `None`. Subsequent calls might pass the state at the end of the
-        previous operation. This allows for quicker opening e.g. by seeking
-        to the beginning of the file instead of re-opening it.
-
         """
         pass
+
+    def reset(self, state):
+        """Resets the state.
+
+        The default implementation closes the state and opens a new one.
+        A more efficient implementation can override the default one
+        in descendant classes.
+
+        Returns
+        -------
+        state : object
+            A reseted state.
+
+        """
+        self.close(state)
+        return self.open()
+
+    def next_epoch(self, state):
+        """Switches the state to the next epoch.
+
+        The default implementation simply resets the state.
+
+        Returns
+        -------
+        state : object
+            The state corresponding to the start of next epoch.
+
+        """
+        return self.reset(state)
 
     def close(self, state):
         """Cleanly close the dataset e.g. close file handles."""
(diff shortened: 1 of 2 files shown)


bartvm:
I'm at the airport now and the internet is terrible, so it might need to wait until tomorrow till I can have a good look at this. Without having looked at the code; I'm not a fan of the `next_epoch` method. The `__iter__` method is supposed to reset the iterator. I'm not sure if a `next_epoch` method is that easy e.g. in a situation where your epoch is shorter than the dataset, and you are reading your data directly from disk. In that case you can't re-use the same epoch without writing some code that would seek back in the file, and just re-using that iterator twice would result in strange things. Maybe for cases where you want to use the same epoch iteration twice we could just support the `deepcopy` method? So instead
```
for batch in stream:
    train(batch)
```
You could do
```
data_iterator = iter(stream)
second_iterator = data_iterator.copy()
```
if you wanted to use the same epoch twice. This would internally also copy the state, creating two file handlers, so that the second iteration would automatically result in the exact same epoch.

Other points:

* Returning a `dict` is fine by me. You can always `zip` with the sources, so it's a matter of preference.
* If we would allow iteration schemes to be stateful, it would be easy to return multiple iterators that cover different parts of the epoch.
* I think you can overwrite `__setstate__` and `__getstate__` to use any serialization method you want (just serialize in that method and return `None`) so we might not need custom methods.

blocks/datasets/__init__.py
```diff
@@ -26,51 +27,51 @@ class Dataset(object):
...
+        in descendant classes.
 
         Returns
         -------
         state : object
-            An object representing the state of a dataset.
+            A reseted state.
 
+        """
+        self.close(state)
```
bartvm: I consciously didn't do this because there might be faster ways than closing and opening to reset e.g. seek to the beginning of a file. Or if the dataset is an interface to a web API or a database, you don't want to redo the authentication process (but you might need to send some sort of reset signal).

rizar:
Having very limited time right now I will answer only one sentence of yours:

> The __iter__ method is supposed to reset the iterator. 

That's not pythonic. In Python you have containers (list, dict) that return a fresh iterator every time `__iter__` is called, but those are immutable. You also have iterators, like a file object. Those are stateful and `__iter__` just returns `self`. My idea is that a data stream, since it is stateful, should imitate behaviour of Python files. For reseting those have the `seek` method, that's why I added a `reset`  method to the data stream interface.

Will comment on the rest a bit later today.

rizar:
Another piece of response: I do not understand why you relate the `next_epoch` method to reusing of the epoch. I meant it to be called when an epoch is finished to start the next one. I did not mean to support reusing an epoch in any way because it is technically hard, as you say, and because I do not know who needs it. The `reset` method should reset the stream entirely to the beginning of the first epoch.

blocks/datasets/__init__.py
rizar: There definitely might be, but I think it makes sense to have this default implementation that works well in many cases.

rizar:
Here is a slightly revised version in which sources are set only for the first data stream in the chain. In fact I am not sure that selecting the necessary sources should be done by the dataset and can not be done by the user by manually selecting necessary entries from the returned dict. Can somebody argue why we need this?

bartvm:
I mentioned this to @janchorowski: There might be cases where we can save a lot of time by not retrieving certain data e.g. for a dataset with images, tags and categories where the user only needs tags and categories, it seems pointless to load gigabytes of images from disk to just discard them afterwards.

janchorowski:
> I mentioned this to @janchorowski: There might be cases where we can save a lot of time by not retrieving certain data e.g. for a dataset with images, tags and categories where the user only needs tags and categories, it seems pointless to load gigabytes of images from disk to just discard them afterwards.

Sources are needed, because a dataset can encompass many things (e.g. I may have filterbank and mfcc features for a speech experiment and I will want to use just one of them). OTOH, the user may be responsible for selecting them on the creation of the first stream. We do not need to provide control over sources to the consumer of the data.

bartvm:
However, there might be easier options than the current one. Instead of passing the sources to the data stream, we can make them an (optional) argument to the constructor of the data set. That way the sources can also be used to e.g. determine which files to open.

rizar:
OK guys, I see your point, perhaps sometimes it is too costly to load all data and just discard it after loading. But then I would agree with Bart that the simplest way to achieve that is to make them a parameter of the dataset. Moreover I do not see how it can be implemented now, since we only get to know the sources we need at the stage when data is retrieved, and at that time all files should be opened.

As datasets are very light-weight objects, I would probably make them immutable and specify the necessary sources at creation time.

bartvm:
I'll merge this PR into my PR and continue working on it. I agree that my original implementation was broken because `DataStream` is not a container, so `__iter__` [needs to return `self`](https://docs.python.org/2/library/stdtypes.html#iterator.__iter__). However, implementing a `next_epoch` method is also incorrect according to the same section in the documentation:

> The intention of the protocol is that once an iterator’s next() method raises StopIteration, it will continue to do so on subsequent calls. Implementations that do not obey this property are deemed broken.

I believe the correct way to do it according to the Python documentation is the way it is done in Pylearn2 i.e. by implementing a `get_iterator` method which returns a new `DataIterator` object similar to what I had in my original implementation (but there it was returned by `__iter__`, which was wrong).

rizar:
I agree that `next_epoch` contradicts Python documentation and think that something like `get_iterator` might be a good solution. Unlike PyLearn2 in our case `get_iterator` will not take dozen parameters since they are part of the iteration scheme. The name `get_iterator` might be not the best though because it does not illustrate semantics of changing the epoch.

bartvm:
`get_epoch_iterator` maybe? Just because `get_next_epoch_iterator` might be a bit long, although that's fine for me too.

janchorowski:
The maybe stream.epoch_iterator()?

Btw, I agree on specifying the sources in the lightweight dataset 
constructor.

rizar:
Both `get_epoch_iterator` and `epoch_iterator` are fine with me.

bartvm:
I think that in general we should strive to follow an active naming scheme for methods and functions i.e. they should be verbs (or verb phrases) describing what action they perform, so I'll go for `get_epoch_iterator`, which I think is a bit less ambiguous as `epoch_iterator` (which could be initializing it, resetting it, etc. instead of getting it).

janchorowski:
Ok, it's a sensible naming rule.
12-01-2015 21:19 użytkownik "Bart van Merriënboer" <notifications@github.com>
napisał:

PR #94, opened 2015-01-14
Module-level configuration

bartvm:
A small class which allows for module-wide settings. It has three levels of configuration, from low to high priority:

1. Defaults
2. YAML configuration file
3. Environment variables

By default it tries to use `~/.blocksrc` although this can be overridden using the `BLOCKS_CONFIG` environment variable.

Configuration values can be accessed using e.g. `blocks.config.data_path`. If a configuration isn't set and doesn't have a default it raises an error when accessed.

### DIFF
--- a/blocks/__init__.py
+++ b/blocks/__init__.py
@@ -1,2 +1,4 @@
 """The blocks library for parameterized Theano ops."""
+from blocks.config_parser import config  # flake8: noqa
+
 __version__ = '0.1a1'  # PEP 440 compliant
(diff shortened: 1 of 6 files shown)


bartvm:
@rizar @janchorowski I'll just merge it as it does not concern any existing code, but let me know if you have any comments and I can easily change/reverse things. The main choice to make is whether to use Theano's approach or Pylearn2's. Theano stores the entire configuration in a single environment variable e.g. `THEANO_FLAGS=floatX=float32,cxx_flags="-L /lib64" python` while Pylearn2 uses separate environment variables e.g. `PYLEARN_DATA_PATH=/data PYLEARN_VIEWER=eog python`. Considering we will probably have very few configuration settings (like Pylearn2) I prefer the latter, because Theano's often gets a mess of quotes and commas.

PR #95, opened 2015-01-14
Reintroducing the `next_epoch` method

rizar:
Here I reintroduce the `next_epoch` method without conflicts with Python documentation.

### DIFF
--- a/blocks/datasets/__init__.py
+++ b/blocks/datasets/__init__.py
@@ -79,6 +79,19 @@ class Dataset(object):
         self.close(state)
         return self.open()
 
+    def next_epoch(self, state):
+        """Switches the dataset state to the next epoch.
+
+        The default implementation for this method is to reset the state.
+
+        Returns
+        -------
+        state : object
+            The state for the next epoch.
+
+        """
+        return self.reset(state)
+
     def close(self, state):
         """Cleanly close the dataset e.g. close file handles."""
         pass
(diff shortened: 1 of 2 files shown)


bartvm:
But now this stream requires the calling of `reset` manually in order to start from the beginning, while others don't. That means that these datasets will require a slightly different main loop if you want to go over these two epochs multiple times. Or are you making the assumption here that we only iterate over the epochs once and never again?

rizar:
I assume that any real-world dataset will implement `next_epoch` that resets the state when data is exhausted. I updated the test to work like that.

bartvm:
I guess this is where my comment about it being difficult to see whether an iterator is exhausted comes in. In your case, where a separate iterator needs to be advanced once each epoch, it works nicely, but what about the (more realistic?) case where we read through part of a file each epoch. In that case we can't do `f.read(1) == ''` or something, because it would change the position in the file. For some files you could use `f.tell() == os.fstat(f.fileno()).st_size`, but I'm not sure if that's ideal?

Anyway, considering that this doesn't change anything for the normal implementation of data sets, I guess we can just merge it (although we will need to add some good documentation about how all of this fits together at one point).

rizar:
It is hard to tell what is realistic case here. I think that in many cases the user will be able to recognize that the current epoch is the last one like I do in the test. Or sometimes there is no last epoch like in your curriculum learning example.

PR #98, opened 2015-01-15
Fixes MNIST and runs test on Travis

bartvm:
Although it might take a few seconds to download MNIST, I think it's important to have at least MNIST and maybe something like PennTreebank (maybe an open alternative?) on Travis which we can run tests with. Pylearn2 doesn't do this, but often ends up breaking the core examples and tutorials because they can't be tested/run without data.

In the future we can start looking at this: http://docs.travis-ci.com/user/caching/ and cache miniconda and the datasets to speed things up (but we would need a workaround for not being able to use `sudo apt-get` which we need for BLAS).

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -12,21 +12,33 @@ env:
 before_install:
   - |
       if [ $TESTS ]; then
+        # Setup Python environment with BLAS libraries
         sudo apt-get install -qq libatlas3gf-base libatlas-dev liblapack-dev gfortran
         wget -q http://repo.continuum.io/miniconda/Miniconda-latest-Linux-x86_64.sh -O miniconda.sh
         chmod +x miniconda.sh
         ./miniconda.sh -b
         export PATH=/home/travis/miniconda/bin:$PATH
         conda update -q --yes conda
+        # Download MNIST for tests
+        (mkdir mnist
+        cd mnist
+        curl -O http://yann.lecun.com/exdb/mnist/train-images-idx3-ubyte.gz \
+             -O http://yann.lecun.com/exdb/mnist/train-labels-idx1-ubyte.gz \
+             -O http://yann.lecun.com/exdb/mnist/t10k-images-idx3-ubyte.gz \
+             -O http://yann.lecun.com/exdb/mnist/t10k-labels-idx1-ubyte.gz
+        gunzip *-ubyte.gz)
+        export BLOCKS_DATA_PATH=$PWD
       fi
 install:
+  # Install all Python dependencies
   - |
       if [ $TESTS ]; then
         conda install -q --yes python=$TRAVIS_PYTHON_VERSION nose numpy pip coverage six scipy
         pip install -q --no-deps git+git://github.com/Theano/Theano.git
         pip install -q nose2[coverage-plugin] coveralls
         git clone -q git://github.com/lisa-lab/pylearn2.git
-        (cd pylearn2; python setup.py -q develop)
+        (cd pylearn2
+        python setup.py -q develop)
       fi
   - |
       if [ $FORMAT ]; then
@@ -34,7 +46,17 @@ install:
         pip install -q git+git://github.com/bartvm/pep257.git@numpy
       fi
 script:
-  - if [ $TESTS ]; then THEANO_FLAGS=floatX=$TESTS,blas.ldflags='-lblas -lgfortran' coverage run --source=blocks -m nose2.__main__ tests; fi
-  - if [ $FORMAT ]; then flake8 blocks tests; pep257 blocks --numpy --ignore=D100,D101,D102,D103; fi
+  - |
+      if [ $TESTS ]; then
+        THEANO_FLAGS=floatX=$TESTS,blas.ldflags='-lblas -lgfortran' \
+        # Running nose2 within coverage makes imports count towards coverage
+        coverage run --source=blocks -m nose2.__main__ tests
+      fi
+  - |
+      if [ $FORMAT ]; then
+        flake8 blocks tests
+        # Ignore D100-103 errors (non-existing docstrings)
+        pep257 blocks --numpy --ignore=D100,D101,D102,D103
+      fi
 after_script:
   - if [ $TESTS ]; then coveralls; fi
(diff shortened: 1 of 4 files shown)


rizar:
What do you think about moving the downloading code to the dataset class? In fact a `download` routine could make it into the `Dataset` interface.

bartvm:
In the spirit of "lean and mean" I would argue it's not a feature that should go in the framework. Starting a Python interpreter and giving the `download` command is just as much time as typing MNIST into Google and clicking the download link (and a lot of people will end up doing that anyway, because they won't realise the feature exists). Then there's all the issues you would run into with not having read or write rights, URLs changing (and dead links lingering around till someone finally files a bug report), needing to (g)unzip files, servers being offline, and needing to give informative errors for all those things... I wouldn't bother.

rizar:
> Starting a Python interpreter and giving the download command is just as much time as typing MNIST into 
> Google and clicking the download link (and a lot of people will end up doing that anyway, because they 
> won't realise the feature exists). 

I am not sure about that. One can find numbers of formats of a dataset on the internet and only one of them is supported by the dataset class from Blocks, and it is never easy to tell which. I think you underestimate how much it discourages people from playing with framework. E.g. I have not yet tried your MNIST dataset mainly because I felt lazy to start an investigation which data is needed!

>  Then there's all the issues you would run into with not having read or write rights, URLs changing (and 
> dead links lingering around till someone finally files a bug report), needing to (g)unzip files, servers being > offline, and needing to give informative errors for all those things... I wouldn't bother.

For truly well-known public datasets such as MNIST, breast cancer data, etc., it is not gonna be an issue. For others we should not care that much. Moreover, I think that having all dataset classes in the main repo like PyLearn2 does is not the right way to go. Only a few examples should be included, for others we should have a separate repo, which apart from dataset classes can contain models from publications and so on. This repo, call it "blocks-contributions", does not have to be supported as actively as the main one. But for the examples that make into the main repo we should provide all comfort possible, including downloading them.

bartvm:
I think that the first concern is a valid one, but I think it should simply be fixed with documentation. I can write a very simple step for step guide on where to download the MNIST files, how to unzip them, and in which folder to put the files so that Blocks can find them.

If you feel that dataset handling should be more automated, I suggest we consider integration with e.g. https://jaberg.github.io/skdata/ instead of writing our own download methods.

I very much like the idea of only having a handful of important datasets in the main repo, and putting the rest in a different repo (although perhaps as a submodule, so that it is still automatically downloaded?)

rizar:
> I think that the first concern is a valid one, but I think it should simply be fixed with documentation. I can 
> write a very simple step for step guide on where to download the MNIST files, how to unzip them, and in 
> which folder to put the files so that Blocks can find them.

And this documentation of yours will be a subject to exactly the same problems you mention above! Even worse: the downloading code will be covered by tests, whereas the documentation will go outdated unnoticed.

The `download` method can use available solutions like the one you point to above. But I think it is important to have an executable instruction for getting the right data in important dataset classes, regardless whether it is implemented by launching a bunch of bash commands or by using a third-party framework. It saves the  user from endless doubts whether he did something wrong. 

And I do not think we have to be very verbose explaining why downloading failed. I would just run bash commands until one of them fails, then print the error code and standard error stream contents.

bartvm:
It's true that the documentation issues would be the same, but at least we wouldn't have broken code lying around.

I guess I'm a purist, but I'd prefer not to have Blocks littered with helper functions performing non-essential things (i.e. not things related to building, training and monitoring a model) for people who are too lazy to read the docs and Google "download MNIST".

Bash commands are already an issue, because on Windows there are no bash commands, and what if the person only has `curl` (or only `wget`)?

If we really do want it to be automated I would suggest just having `skdata` as an optional dependency with a minimal interface. I just had a quick look and it seems straightforward enough: For each dataset we say which `skdata` dataset it corresponds to, we overwrite `$SKROOT` to `$BLOCKS_DATA_PATH` and call the `fetch` method.

rizar:
> It's true that the documentation issues would be the same, but at least we wouldn't have broken code lying 
> around.

I don't understand your point about broken code. Suppose we have 5 truly important datasets in the main branch. If they are truly important, there typically exists a quite stable public source from which they can be loaded automatically. If such a source ceases to exist, we can replace the `download` implementation with `NotImplemented` right away.

> I guess I'm a purist, but I'd prefer not to have Blocks littered with helper functions performing 
> non-essential things (i.e. not things related to building, training and monitoring a model) for people who 
> are too lazy to read the docs and Google "download MNIST".

I do not consider fetching the right data an inessential part of a machine learning experiment. And I personally hate the stage of trying to understand whether the data I googled is the one a certain piece of code needs. I would prefer not to care about the way the data is stored, and this is achieved by having a place in the dataset description where an exact instruction to get the data is provided. If an instruction is exact, it can be executed.

You made a good point about platforms, `curl` and `wget`. But my opinion is that not supporting something at all or supporting at certain platforms (or with certain additional dependencies like `curl`) is better than ignoring the problem.

I fully support out-sourcing the loading job to a library designed specifically for loading purposes whenever it is possible. That is the `download` method can call the `fetch` method of the respective dataset.

bartvm:
I guess my main issue with all of this comes from experience with Pylearn2: It has a lot of these kind of "handy" functionalities which are rarely used. Downloading MNIST only needs to be done once for example (but Pylearn2 has a lengthy script for it: https://github.com/lisa-lab/pylearn2/blob/master/pylearn2/scripts/datasets/download_mnist.py), no one ever uses line search (but Pylearn2 has it), and there are lots and lots of other small things. This results in a lot of code bloat. (Pylearn2 has 131633 lines of code vs. 13474 in GroundHog... That's 10 times as much!) But more than the code bloat, it means that as the project grows these kind of things become badly maintained, and eventually you'll end up like Pylearn2: Fixing bugs reported by new users on the mailing list, complaining that their MNIST downloading script crashed because they updated to Python 3, etc.

But anyway, if we can keep the functionality out of the library and simply provide a minimal bridge to `skdata`, I'm happy :)

rizar:
Okay, created #105.

PR #102, opened 2015-01-16
BinaryCrossEntropy to work with indices direcly

rizar:
As far as I understand the PyLearn2'ish way of computing cross-entropy was to transform indices into one-hot vectors. I am afraid that this is rather memory inefficient, especially for a large number of classes. This pull request introduces a method that bypasses one-hot vectors and works with indices directly.

As the whole `cost.py` currently misses documentation, I did not document what the method does.

In case you case support this initiative, I will cover it by a test before merging.

### DIFF
--- a/blocks/bricks/cost.py
+++ b/blocks/bricks/cost.py
@@ -25,6 +25,15 @@ class BinaryCrossEntropy(CostMatrix):
         cost = tensor.nnet.binary_crossentropy(y_hat, y)
         return cost
 
+    @application
+    def apply_for_indices(self, y_ones_indices, y_hat):
+        flat_ones_indices = y_ones_indices.flatten()
+        return -(tensor.log(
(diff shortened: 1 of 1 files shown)


bartvm:
Coline and I implemented proper categorical cross-entropy in Pylearn2 over the summer (just use the `binary_target` argument for the `Softmax` layer).

You've implemented it here as part of the binary cross-entropy though. I would just make a different class, `CategoricalCrossEntropy`, which could even wrap [the relevant Theano function](http://deeplearning.net/software/theano/library/tensor/nnet/nnet.html#tensor.nnet.categorical_crossentropy) instead of writing it yourself. The Theano implementation supports both the one-hot vector targets as well as integers.

rizar:
Okay, I will write a new cost brick.

PR #103, opened 2015-01-16
as_dict for epochs

rizar:
An interface fix. It is more convenient to request data as dictionaries for the main loop.

### DIFF
--- a/blocks/datasets/__init__.py
+++ b/blocks/datasets/__init__.py
@@ -238,8 +238,7 @@ class AbstractDataStream(object):
                             if self.iteration_scheme else None,
                             as_dict=as_dict)
 
-    @property
-    def epochs(self):
+    def epochs(self, as_dict=False):
         """Allow iteration through all epochs.
 
         Notes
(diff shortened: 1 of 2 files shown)


bartvm:
I'm not a fan of the `epochs()`. As I mentioned earlier, I think that it should be `get_epochs()` then i.e. stick with verbs and verb phrases for methods and functions to make clear what they do, reserve nouns and noun phrases for properties and attributes.

Keep in mind that I actually implemented this for the `get_epoch_iterator` with the `as_dict` keyword argument, so maybe we can just keep `stream.epochs` as a convenience property, but use `get_epoch_iterator(as_dict=True)` in the main loop?

rizar:
I think we come to the point where subjective opinion matters. I like to have a fully functional wrapper over `get_epoch_iterator` that does iteration for me. In fact it is an `__iter__` method.

If you want to rename it to a verb phrase, `iterate_epochs` would work for me.

rizar:
I would like to proceed quickly with this small change. @bartvm , are you fine with `iterate_epochs`?

bartvm:
Yeah.

PR #104, opened 2015-01-16
WIP: First version of lazy properties

### LINKED ISSUE #96: Serialization of datasets (and preprocessing)
Now that we have an initial version of datasets, the next challenge: Serialization and preprocessing. This is a pretty difficult problem to tackle, so it's worth some thought.

Serialization is difficult because we don't want to pickle the data each time we save the model (e.g. no point in writing the same MNIST to disk time and time again). Not saving the data and loading it from disk when deserializing is the obvious solution, but comes with pitfalls: If the data isn't where it is was originally, the deserialization will fail and our saved model is now useless.

DWF gave this some thought for Pylearn2, so let me quote from his outline of what would be a solution:

> Automatically loading from disk on unpickling causes unpickling failure if the data has moved or if the path doesn’t exist on that machine. Don’t do one as a consequence of the other automatically.

> **Data loading should be overridable.**
> The lowest common denominator for every kind of dataset is to provide a directory path for loading or saving compatible disk representations. The load path should be specifiable via constructor, and alterable by a method call.

> **Data loading should be lazy.**
> To that end, data should be fetched from data on an as-needed basis. [...] In particular, no additional disk access should be triggered by unpickling: this gives the user the opportunity to modify the data load path, if necessary. This prevents a lot of nasty situations where you cannot unpickle something and you don’t have permissions to create the data paths necessary in order to make it unpickle cleanly.

-----

A separate problem, but one quite closely integrated with serialization, is preprocessing. If we want to support automatic preprocessing such as whitening, ZCA, contrast normalization, etc. there are a few things that should be considered:

* Some preprocessing is global (i.e. for whitening we need the dataset mean and variance), but might need to be able to be applied as an out-of-core algorithm for data that doesn't fit into memory.
* For out-of-core datasets we then have two options, depending on the kind of preprocessing performed: We apply the processing batch-wise during iteration, or we save the modified data to disk and then stream from disk again.
* For in-memory datasets, ideally, we should remove the original data from memory and only keep the processed data.
* Ideally all these pre-processing steps should be serializable/reproducible i.e. with only the original data and a pickle file, we should be able to recreate the pre-processed data exactly (this is where the strong link with serialization as discussed above shows up).

To quote from DWF again:

> **Different facilities for in-memory and out-of-memory data; in-code awareness of which you’re dealing with.**
> Whole-dataset preprocessors should be able to know that they are operating on an in-memory dataset, and fail if they don’t know how to do something intelligent with a certain kind of out-of-memory dataset.
> **Keep records of transformations done on data.**
> It should be possible to regenerate preprocessed data from the default loading path and the log of preprocessing operations upon deserialization. Of course generic containers in the spirit of “DenseDesignMatrix” needn’t provide a default loading path, but given a deserialized object, the load -> (optionally) fit preprocessor -> apply preprocessor should be a simple method call away.

:exclamation: However, I think the best point made is at the end of his document:

> **Is it really worth having the in-{Python,YAML} preprocessor facilities?**
> It seems like this is only of any use for relatively small, in-memory datasets. For large datasets (or even small datasets where you value reproducibility, and different machines will give you different ZCA results, for example), we could support it with a script like
> ```
python preprocess.py \
        --config <description-of-preprocessing> \
        --data_output_path <directory> \
        --output_fitted_preprocessor <pickle file>
```
> You could also add arguments for preprocessing from an already fitted preprocessor object, etc. To use a preprocessed dataset you’d then use the corresponding dataset class and override the load path in the constructor. Viewer code could be passed the serialized pickle file to know how to undo transformations and things like that.

It sounds to me like the pre-processing serialization, etc. would be a real nightmare, and I would argue using this approach for all datasets, in-memory or not. So pre-processing basically happens as an entirely separate step through a script that takes particular datasets (NumPy arrays, HDF5, CSV, etc.) and a configuration file describing what preprocessing to apply as an input. As outputs it produces a new dataset as well as a (pickle?) file which describes what preprocessing exactly has been performed, allowing you to reproduce it and/or reverse it.

### PR DESCRIPTION
**Pull request by bartvm**
**Update:** Just to be clear what the goal is here: The easy approach to stopping and resuming training is by simply dumping the `MainLoop` object in a pickle file. However, this would result in pickling any data held in memory by `DataSet` instances. To prevent that, this introduces a way of making sure that this data is transparently removed from memory before serializing, and reloading it again after deserialization. However, this reloading is done lazily, so that the user has time to intervene in case files have moved around or disappeared.

-----

Because many datasets will have the same approach I endeavored to automate some parts of dataset serialization.

This PR introduces a `lazy_properties` decorator which allows you to name properties which are "lazy". This means that they will be loaded by a call to `load`, and `load` will only be called the first time one of the properties is actually needed. Moreover, when the dataset is serialized the `__getstate__` method will remove all of the lazy properties so that they don't get pickled. When de-serializing, `load` won't be called again until one of the lazy properties is requested.

One shortcoming of this implementation is if *one* lazy property is requested, *all* will be loaded. I don't think that this is actually an issue though. The only thing we need is that they don't get loaded during de-serialization so that a user can fix the state of the dataset instance in case files have moved around.

I don't allow lazy loading of iterators, because their state would be lost, except for file handles (which should be handled through `dill` or our own implementation; I'll implement that soon). WIP because all tests and documentation are missing as of yet.

Fixes #96.

### DIFF
--- a/blocks/datasets/__init__.py
+++ b/blocks/datasets/__init__.py
@@ -158,9 +158,9 @@ class InMemoryDataset(Dataset):
     >>> import pickle
     >>> from blocks.datasets.mnist import MNIST
     >>> mnist = MNIST('train')
-    >>> print("{} MB".format(
-    ...     mnist.data['features'].nbytes // 1024 // 1024)) # doctest: +SKIP
-    179 MB
+    >>> print("{:,d} KB".format(
+    ...     mnist.data['features'].nbytes / 1024)) # doctest: +SKIP
+    183,750 KB
     >>> with open('mnist.pkl', 'wb') as f:
     ...     pickle.dump(mnist, f, protocol=pickle.HIGHEST_PROTOCOL)
 
--- a/blocks/datasets/__init__.py
+++ b/blocks/datasets/__init__.py
@@ -169,7 +169,7 @@ class InMemoryDataset(Dataset):
     and if the data file has not been moved, it will be as if nothing
     happened.
 
-    >>> with open('mnist.pkl') as f:
+    >>> with open('mnist.pkl', 'rb') as f:
     ...     mnist = pickle.load(f)
     >>> print(mnist.data['features'].shape)
     (60000, 784)
@@ -180,7 +180,7 @@ class InMemoryDataset(Dataset):
     >>> from blocks import config
     >>> correct_path = config.data_path
     >>> config.data_path = '/non/existing/path'
-    >>> with open('mnist.pkl') as f:
+    >>> with open('mnist.pkl', 'rb') as f:
     ...     mnist = pickle.load(f)
     >>> print("{} KB".format(mnist.data['features'].nbytes // 1024))
     Traceback (most recent call last):
--- a/blocks/datasets/__init__.py
+++ b/blocks/datasets/__init__.py
@@ -182,10 +182,10 @@ class InMemoryDataset(Dataset):
     >>> config.data_path = '/non/existing/path'
     >>> with open('mnist.pkl', 'rb') as f:
     ...     mnist = pickle.load(f)
-    >>> print("{} KB".format(mnist.data['features'].nbytes // 1024))
+    >>> print(mnist.data['features'].shape) # doctest: +SKIP
     Traceback (most recent call last):
       ...
-    IOError: [Errno 2] No such file or directory: ...
+    FileNotFoundError: [Errno 2] No such file or directory: ...
 
     Because the loading happens lazily, we can still deserialize our
     dataset, correct the situation, and then continue.
--- a/blocks/datasets/__init__.py
+++ b/blocks/datasets/__init__.py
@@ -216,6 +216,36 @@ class InMemoryDataset(Dataset):
 
 
 def lazy_properties(*lazy_properties):
+    r"""Decorator to assign lazy properties.
+
+    Used to assign "lazy properties" on :class:`InMemoryDataset` classes.
+    Please see the documentation there for a discussion on what lazy
+    properties are and why they are needed.
+
+    Parameters
+    ----------
+    \*lazy_properties : strings
+        The names of the attributes that are lazy.
+
+    Notes
+    -----
+    The pickling behaviour of the dataset is only overridden if the dataset
+    does not have a ``__getstate__`` method implemented.
+
+    Examples
+    --------
+    In order to make sure that attributes are not serialized with the
+    dataset, and are lazily reloaded by the :meth:`~InMemoryDataset.load`
+    method after deserialization, use the decorator with the names of the
+    attributes as an argument.
+
+    >>> @lazy_properties('features', 'targets')
+    ... class TestDataset(InMemoryDataset):
+    ...     def load(self):
+    ...         self.features = range(10 ** 6)
+    ...         self.targets = range(10 ** 6)[::-1]
+
+    """
     def lazy_property_factory(lazy_property):
         """Create properties that perform lazy loading of attributes."""
         def lazy_property_getter(self):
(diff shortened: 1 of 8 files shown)

### CODE REVIEW

**Comment by rizar** [comment]
I have a whole mixture of emotions regarding this idea of yours: it looks very nice and convenient, but is quite non-standard (e.g. skdata does not have that, one must call `fetch`) and implicit: one must open the dataset's code to understand when loading actually happens. @janchorowski, what do you think?

I suggest to make a separate class, e.g. `InMemoryDataset`, that will introduce the `load` method. It does not make sense for large datasets and might confuse the user.

**Comment by rizar** [comment]
Perhaps instead of automatic loading we could raise a nice exception reminding the user that the `load` method should have been called.

**Comment by bartvm** [comment]
I'm not sure that's a good idea, because it would require `load` to be called after de-serializing, while we want the object to be the in the exact same state it was before serializing (at least, it's interface must behave as if it is). Hence, `load` must be called automatically after deserializing, but not until needed, so that the user can intervene, but only if he *must* because data has moved.

My idea is this: When resuming training, we deserialize everything and the dataset object should automatically reload the MNIST data from disk as if nothing ever happened, and training can continue. Howver, if the data has moved, it will crash when you try to continue training. In that case, you can simply deserialize but **not** continue training, change the location from which MNIST should be loaded, and then continue. If it wouldn't perform lazy loading, the crash would occur during deserialization already, and would make the pickle file useless.

I'm happy to make an `InMemoryDataset`; I agree that if the dataset doesn't hold any data in memory, it doesn't need the `load` method.

**Comment by bartvm** [comment]
Fixes #100 by the way (but so does #92).

**Comment by rizar** [comment]
Thanks for the explanation, now I see better why we need this way of lazy loading.

File: blocks/datasets/__init__.py [hunk 1]
```diff
@@ -130,6 +143,37 @@ def get_default_stream(self):
...
+        for lazy_property in lazy_properties:
+            setattr(dataset, lazy_property,
+                    property(*lazy_property_factory(lazy_property)))
+        if not hasattr(dataset, '__getstate__'):
+            def __getstate__(self):
+                for lazy_property in lazy_properties:
+                    attr = getattr(self, '_' + lazy_property)
+                    if isinstance(attr, collections.Iterator):
+                        raise ValueError("Iterators can't be lazy loaded")
+                    delattr(self, '_' + lazy_property)
```
rizar: Wouldn't it be better to create a shallow copy of `__dict__` and change it? I do not like that `__getstate__` changes the object.

bartvm: You're right, that would be better.

bartvm:
Added a lengthy docstring about what lazy properties do. In the process ended up adding a new feature to the configuration class: Manually overriding configuration i.e. you can now do:
```python
blocks.config.data_path = 'blah'
```
and all subsequent calls to `blocks.config.data_path` throughout the library should return the new `'blah'` value.

rizar:
What about merging this guy?

bartvm:
Your call. I was planning to add a few tests, pickling data streams and unpickling them and seeing whether they keep their state. But if you need this for your work on serialization, feel free to merge and I'll add the tests separately.

rizar:
Okay, then I merge it and you can cover it with tests in further pull request. I guess you will also add dill as a dependency. It turned out that the standard pickle module can not cope with lambdas, which is quite unfortunate, so we need dill for the serialization of extensions as well. And both of them can not pickle generators: those have to be entirely forbidden in Blocks if we want to end up with working "pickle-everything" serialization.

bartvm:
Okay, we'll have to refactor the iteration schemes a bit then so that they don't return generators, but that should be straightforward.

rizar:
I will take care of it tomorrow as a part of the effort of making the main loop serializable.

bartvm:
Okay, I have actually already done that :) Dill can pickle list iterators, but only on Python 3.2 and above, so I actually had to write new iterators, but that only took a second. I'll make a PR in a bit.

PR #107, opened 2015-01-17
Categorical crossentropy

bartvm:


rizar:
The Theano function returns 1D-tensor, so it can not be a subclass of `CostMatrix`. But I merged this class in my pull request as a simple ancestor of `Cost`.

bartvm:
Okay!

PR #108, opened 2015-01-17
Replaced the Python 2 __metaclass__ = ABCMeta with six's version

### LINKED ISSUE #106: Our Abstract Classes Are Not Abstract Classes In Python3!
It turns out the syntax to declare metaclass has been changed in Python3. That's why our abstract methods can be called in Python3: they are not abstract in fact!

For more detail see the link below:
http://stackoverflow.com/questions/18513821/python-metaclass-understanding-the-with-metaclass/18513858#18513858

bartvm:
Fixes #106.

### DIFF
--- a/blocks/algorithms/__init__.py
+++ b/blocks/algorithms/__init__.py
@@ -3,11 +3,13 @@ from abc import ABCMeta, abstractmethod
 from collections import OrderedDict
 
 import theano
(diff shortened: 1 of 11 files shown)


rizar:
Can we wait with it until #92 is merged?

bartvm:
Sure.

PR #112, opened 2015-01-18
Serializable datasets

bartvm:
Removed the generators used in schemes and replaced them with custom
iterators so that dill can serialize them. Also added a test that
serializes an MNIST data stream and deserializes it, checking whether
the state is maintained.

@rizar

### DIFF
--- a/blocks/datasets/schemes.py
+++ b/blocks/datasets/schemes.py
@@ -1,3 +1,4 @@
+import itertools
 from abc import ABCMeta, abstractmethod
 
 import six
@@ -69,26 +70,10 @@ class ConstantScheme(BatchSizeScheme):
         self.times = times
 
     def get_request_iterator(self):
-        return ConstantIterator(self.batch_size, self.times)
-
-
-class ConstantIterator(six.Iterator):
-    def __init__(self, batch_size, times=None):
-        self.batch_size = batch_size
-        self.times = times
-        if times is not None:
-            self.current = 0
-
-    def __iter__(self):
-        return self
-
-    def __next__(self):
-        if self.times is not None:
-            if self.current == self.times:
-                raise StopIteration
-            else:
-                self.current += 1
-        return self.batch_size
+        if self.times is None:
+            return itertools.repeat(self.batch_size)
+        else:
+            return itertools.repeat(self.batch_size, self.times)
 
 
 class SequentialScheme(BatchScheme):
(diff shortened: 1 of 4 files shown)


blocks/datasets/schemes.py
```diff
@@ -91,5 +92,22 @@ def __init__(self, num_examples, batch_size):
         self.batch_size = batch_size
 
     def get_request_iterator(self):
-        return (slice(x, min(self.num_examples, x + self.batch_size))
-                for x in range(0, self.num_examples, self.batch_size))
+        return SequentialIterator(self.num_examples, self.batch_size)
+
+
+class SequentialIterator(six.Iterator):
```
rizar: Having such an iterator is also better in terms of memory, the old approach allocated a potentially quite large list.

rizar:
Great, however I am confused that pickling `repeat` works for you, because it does not work for me:

```
In [12]: dill.dump(repeat(3), open("f.pkl", "w"))

In [13]: dill.load(open("f.pkl"))
<ipython-input-13-788f69d0c637> in <module>()
----> 1 dill.load(open("f.pkl"))

/usr/local/lib/python2.7/dist-packages/dill/dill.pyc in load(file)
    197     pik = Unpickler(file)
    198     pik._main_module = _main_module
--> 199     obj = pik.load()
    200     if type(obj).__module__ == _main_module.__name__: # point obj class to main
    201         try: obj.__class__ == getattr(pik._main_module, type(obj).__name__)

/usr/lib/python2.7/pickle.pyc in load(self)
    856             while 1:
    857                 key = read(1)
--> 858                 dispatch[key](self)
    859         except _Stop, stopinst:
    860             return stopinst.value

/usr/lib/python2.7/pickle.pyc in load_newobj(self)
   1081         args = self.stack.pop()
   1082         cls = self.stack[-1]
-> 1083         obj = cls.__new__(cls, *args)
   1084         self.stack[-1] = obj
   1085     dispatch[NEWOBJ] = load_newobj

TypeError: Required argument 'object' (pos 1) not found

In [14]: 
```

rizar:
Okay, I guess I see what happened. It worked for me in Python 3, and perhaps you tried locally in Python 3 and your MNIST test uses the `SequentialScheme` only. If that's the case, can you please reintroduce your `ConstantIterator`?

blocks/datasets/schemes.py
bartvm: No, it was never a list, the parentheses made it a generator, so it used just as much memory.

rizar: My bad, I thought it was a big tuple.

PR #114, opened 2015-01-19
WIP: Language dataset

bartvm:
MNIST is great as an in-memory image-dataset, but I'd like to start adding some NLP stuff to test/play with.

I started work on supporting the monolingual One Billion Word benchmark from Google; a suggestion for another dataset is welcome, but this one is large, well standardized and freely available (unlike e.g. Penn Treebank). Also, it's nicely split up in 99 training parts and 50 validation sets. We can make one of these sets available on Travis to do our NLP testing with.

Data streams are proving to be a great idea: I've used them to create n-grams on the fly from sentence datasets, factoring this entire process out of the datasets.

WIP since I want to (eventually):

- [ ] Add tests for the one billion word datasets
- [x] Add tests for serializing datasets with file handles
- [ ] Add tests for n-gram creation
- [ ] Add a "sorting stream" and combine that with the cache to create a test/example of a cache being sorted in order to provide batches that require minimal masking
- [ ] Maybe write an RNN language modelling example (does anyone have a model with good hyperparameters lying around that I could reproduce?)

Some miscellanea:
- There was a bug in MNIST, fixed it and added a test for it
- Fixed typo in docs
- Added tests for the cached data stream (uncovered a bug regarding switching between epochs)


blocks/datasets/__init__.py
```diff
@@ -529,11 +541,66 @@ def get_data(self, request):
             self.cache[i] = cache[request:]
         return tuple(data)
 
+    def get_epoch_iterator(self, **kwargs):
+        self.cache = [[] for source in self.sources]
+        return super(CachedDataStream, self).get_epoch_iterator(**kwargs)
+
     def _cache(self):
         for cache, data in zip(self.cache, next(self.child_epoch_iterator)):
             cache.extend(data)
```
rizar: I think we should have an attribute documentation for the `cache` attribute, since it can be used by other classes.

bartvm: Good point, I'll write something up.

blocks/datasets/__init__.py
```diff
@@ -529,11 +541,66 @@ def get_data(self, request):
...
 
 
+class NGramStream(CachedDataStream):
+    """Return n-grams from a stream.
+
+    This data stream wrapper takes as an input a data stream outputting
+    batches of sentences. From these sentences n-grams of a fixed order
+    (e.g. bigrams, trigrams, etc.) are extracted and returned.
+
```
rizar: I think you should mention here that you are adding an additional source, and what this source is.

rizar:
Actually I am going to use this for an attention mechanism demo: I will try to train a script that removes spaces, and if it works, even backwards, that is to add spaces. Will use  this branch while you are working on it.

rizar:
Oops, actually I can't use exactly your code because it only return word indices, whereas I wanted to play on character level. We should add characters to the sources provided by the dataset.

bartvm:
Right, I want to factor that out into a "TextFile" dataset which basically just takes text files and a word or character dictionary. Just not entirely sure about the semantics of characters. Should it return a list of lists of lists? That is `batch -> sentence -> word -> character`? I guess we can use streams to then split that up into different words, combine with word-level data, etc.

rizar:
I guess I would split functionality in a few levels:

* The basic level is tokenization and encoding using a dictionary. This can be done in the dataset. The output of this level is a list of indices of every token, i.e. character, word or even syllable!

* The second level is grouping: here we form list of lists. Can be done by a datastream wrapper.

* The third level is packing into numpy arrays. Can be done by another datastream wrapper.

There may be other ways to separate this functionality, but I believe that to prevent tons of boilerplate code we should have multiple levels of preprocessing, each doing its own, well-defined and simple job.

bartvm:
Sounds good. I hadn't thought of the second level, but I guess it's good for datasets which can only process one example at a time (e.g. a file that is read line by line). The data stream can then handle turning them into batches. The last data stream would add the masks. Eventually I would also love a "merge" data stream and a "sort" data stream, so that machine translation becomes:

- Two datasets reading source and target language text files and binarizing them
- Merge them into a single stream
- Turn them into batches of size N
- Sort batch by sentence length of source language
- Cache this and return minibatches of near-uniform size

rizar:
Sounds like a good plan! Another idea is to have a wrapper that chains a sequence of data streams. That would be helpful for the billion words dataset.

bartvm:
@rizar This PR is already turning quite massive again, so maybe we should merge it tomorrow before I implement this entire chain of data streams. The pasts should be passing, except that it's not downloading MNIST anymore (`(7) couldn't connect to host` errors)... Not sure why that is. Maybe their server blocked Travis because we're downloading it too much :confused: If that's the case, I'll try to spend a bit of time setting up Travis with caches tomorrow or this weekend.

blocks/datasets/__init__.py
```diff
@@ -535,11 +604,72 @@ def get_data(self, request=None):
...
+    Parameters
+    ----------
+    ngram_order : int
+        The order of the n-grams to output e.g. 3 for trigrams.
+    data_stream : :class:`DataStream` instance
+        The data stream providing sentences. Each example is assumed to be
+        a list of integers.
+    target_source : str, optional
+        This data stream adds a new source for the target words. By default
```
rizar: Can you write the the target words is the one next to the n-gram? (as far as I understood from the code)

tests/__init__.py
```diff
@@ -0,0 +1,17 @@
...
+def temporary_files(*files):
+    def wrap_test(test):
+        def wrapped_test(*args, **kwargs):
+            if any(os.path.exists(file_) for file_ in files):
+                raise IOError
+            try:
+                test(*args, **kwargs)
+            finally:
+                for file_ in files:
+                    os.remove(file_)
```
rizar: I think we should check if they exists, for if a test failed, the file might not be there.

tests/datasets/test_text.py
```diff
@@ -0,0 +1,32 @@
...
+                         dictionary=dictionary, bos_token=None,
+                         preprocess=str.lower)
+    stream = text_data.get_default_stream()
+    epoch = stream.get_epoch_iterator()
+    assert len(list(epoch)) == 4
+    epoch = stream.get_epoch_iterator()
+    for sentence in zip(range(3), epoch):
+        pass
+    with open('text_stream.pkl', 'wb') as f:
+        dill.dump(epoch, f, fmode=dill.CONTENTS_FMODE)
```
rizar: I can not find documentation for what the `fmode` parameter stands for. Do you have a link?

rizar:
I also think we should not wait until the whole chain is implemented. Go ahead and merge when you want.

tests/datasets/test_text.py
bartvm: Dill's only documentation seems to be the source code: https://github.com/uqfoundation/dill/blob/master/dill/dill.py#L144-L153

However, here the documentation actually seems wrong. Even though it says that it "pickles the file contents", I tried pickling a large text file myself, and the resulting pickle file was less than a kilobyte, suggesting it hadn't saved the content, but only the position.

PR #115, opened 2015-01-19
Pickling the Main Loop and Resuming from Unpickled

rizar:
This pull request contains bunch of fixes that make it possible to pickle the whole main loop object, unpickle it and call `run` as if nothing happened at all. More specifically

* A small fix of the `Application` object to prevent infinite recursion because of overriden `__getattr__`.

* Refactoring of the main loop: it is now significantly harder to read but can be resumed in the middle of an epoch.

* Fixing serialization of the `ContainerDataset`'s state.

* Test of training resumption for a toy dataset and for MNIST

Finally I added a simple extension `SerializeMainLoop` that dumps the main loop to the given destination. And as usual, some refactoring.


blocks/bricks/__init__.py
```diff
@@ -658,10 +658,17 @@ def add_property(f):
         return add_property
 
     def __getattr__(self, attr):
+        # Prevents infinite recursion when no attribute "f" is available
+        # (happens during serialization under unknown circumstances)
+        try:
+            self_f = self.__getattribute__("f")
```
bartvm: I don't think this makes much sense; `__getattr__` is only called if the attribute (`attr`) couldn't be found as a usual attribute, so calling `__getattribute__` again will always fail. If you just use
```python
if attr == 'f':
    return {}
```
and not change anything else, does that work?

blocks/utils.py
```diff
@@ -429,3 +429,47 @@ def ipdb_breakpoint(x):
...
+
+    Parameters
+    ----------
+    next_function : callable
+        A function to call every time the next element is requested.
+
+    """
+    def __init__(self, next_function):
+        self.next_function = next_function
+
```
bartvm: You'll need to add
```python
def __iter__(self):
    return self
```
(`six.Iterator` just maps `next` to `__next__`, nothing else)

blocks/utils.py
```diff
@@ -429,3 +429,47 @@ def ipdb_breakpoint(x):
     """
     import ipdb
     ipdb.set_trace()
+
+
+class LambdaIterator(six.Iterator):
```
bartvm: Can we call it `FunctionIterator`, or does the argument need to be a lambda function per se?

blocks/extensions/__init__.py
```diff
@@ -92,18 +107,22 @@ class SimpleExtension(TrainingExtension):
         If ``True``, :meth:`do` is invoked before the first epoch.
     after_every_epoch : bool
         If ``True``, :meth:`do` is invoked after every epoch.
-    after_every_iteration : bool
-        If ``True``, :meth:`do` is invoked after every iteration.
+    after_every_batch: bool
```
bartvm: It seems like this argument list could grow very long... Is there maybe an easier way e.g. just ask the user to pass the string directly for all cases but the `after_n_epochs`?

blocks/extensions/saveload.py
```diff
@@ -0,0 +1,61 @@
...
+        ----------
+        destination : str
+            The destination where the state of the training process was
+            saved.
+
+        """
+        self.main_loop.log.current_row.saving_done_to = destination
+
+
+class DumpMainLoop(SaveLoadBase):
```
bartvm: Instead of "dump" maybe we could use "serialize"? Dump suggests that we just dump the memory state, while pickling involves a lot more than that.

blocks/utils.py
```diff
@@ -429,3 +429,47 @@ def ipdb_breakpoint(x):
...
+    Parameters:
+    ----------
+    sequence : list or tuple
+        The sequence to iterate over.
+
+    """
+    def __init__(self, sequence):
+        assert isinstance(sequence, Sequence)
+        self.sequence = sequence
+        self._offset = 0
```
bartvm: Same here, `__iter__` needs to return `self` I think.

tests/test_examples.py
```diff
@@ -19,7 +20,13 @@ def setup():
 
 
 def test_mnist():
-    mnist_test()
+    filename = 'mnist.pkl'
```
bartvm: Mostly a note to myself: A lot of tests are creating files now. In some of the tests I wrote I tested whether these files existed (to make sure our tests don't overwrite people's files unknowingly) and cleaned up the files at the end. Since we're doing this all over the place now I'll try to write a decorator that makes this easier to do.

blocks/main_loop.py
```diff
@@ -86,6 +77,62 @@ def run(self):
         finally:
             self._run_extensions('after_training')
 
+    def find_extension(self, name):
+        """Find an extension with a given name.
+
+        Parameters:
```
bartvm: No colons here

blocks/main_loop.py
```diff
@@ -86,6 +77,62 @@ def run(self):
...
+        ----------
+        name : str
+            The name of the extension looked for.
+
+        Notes:
```
bartvm: and here

bartvm:
Great! The main loop is indeed a bit harder to read now. I'm in class now, and having a hard time multi-tasking and trying to understand what is going on exactly; I'll try to have a closer look later today. I did make some smaller comments here and there.

bartvm:
Did you figure out the `PicklingError` yet? It seems mysterious, but just a lucky guess: Move the `Activation` class outside of the `_activation_factory` function. The error suggests that Python 3 tries to import the class directly, but can't find it because it's hidden in `_activation_factory`'s scope.

blocks/bricks/__init__.py
rizar: Well, but sometimes I do need the `f` value. That's why I first check that this value is available, otherwise the call `self.f` leads to infinite recursion.

bartvm: But if the `f` value is available, `__getattr__` would never be called. `__getattr__` is only called if the attribute isn't available.

blocks/utils.py
rizar: `FunctionIterator` sounds a bit like an iterator over functions. Any other suggestions?

bartvm: `FunctionCallIterator`? Otherwise just stick with `LambdaIterator` I guess.

blocks/extensions/__init__.py
rizar: It will not grow too long, because here only rather general and frequent conditions will be added. We could in principle add a parameter `on_callbacks` which value is the list of names of callbacks for which `do` should be called unconditionally, but so far I am not convinced that this is gonna be a better interface. Now it is at least homogeneous: a bunch of parameters which are independent of each other.

blocks/extensions/saveload.py
rizar: Hmm, but the method is called `pickle.dump`, isn't it?

blocks/utils.py
rizar: No problem, will fix that.

tests/test_examples.py
rizar: Would be nice! In theory we should have done everything without using the file system altogether, but sometimes using a real file is just a way simpler.

blocks/main_loop.py
rizar: Oops, forgot the syntax :)

blocks/bricks/__init__.py
rizar: But when `__getattr__` is called for an attribute which is indeed not present, the map `self.f` is requested. If it is not present as well, infinite loop starts.

blocks/extensions/saveload.py
bartvm: Fair enough, but maybe that's partly because the interface is supposed to be `pickler = pickle.Pickler(f)` followed by `pickler.dump(object)` i.e. the `pickler` is an object that serializes and the `dump` method performs the writing. But given that `dump` "writes a pickled representation" and "pickling is alternatively known as 'serialization'" according to the docs, I think `SerializeMainLoop` works too, and is a bit clearer for me... I can live with "dump" though.

blocks/utils.py
rizar: The thing when one says `WhateverIterator` it just feels like an iterator over a sequence of whatevers. Let it be `LambdaIterator` until we come up with something significantly better.

blocks/extensions/saveload.py
rizar: Not a big deal, let it be `SerializeMainLoop`. 

I am planning one more extension that it does it the Grounghog way, and this is where naming will be tough.

blocks/bricks/__init__.py
bartvm: Ah, I see why, but that's because `super(Application, self).__getattribute__(attr)` should really just read `raise AttributeError`. It doesn't make sense to call the `__getattribute__` of the parent class (if `__getattr__` was called it's clear that Python doesn't know where to find the attribute, so we should just end the chaing). So just do `if attr == 'f': return {}` and replace the last line with `raise AttributeError`.

bartvm: See e.g. http://code.activestate.com/recipes/389916-example-setattr-getattr-overloading/

rizar: OK, it works! Thanks!

rizar:
Sad but true, and most of the complexity comes from the fact that `for` loops can not be used. Also I split the code into functions, perhaps it would be clearer in a monolith form.

Pushing `Activation` out of the factory is not that easy: there is some class modification involved there. I will see what I can do, or if you know the solution, you can create a pull request to my pull request :)

bartvm:
Okay, I'll have a look and make a PR if I manage to fix it. Seems like it's a known problem (https://stackoverflow.com/questions/4647566/pickle-a-dynamically-parameterized-sub-class) but the solutions are a bit obscure.

bartvm:
I think my fix from https://github.com/bartvm/blocks/commit/0e36c69ea320d6d5ac6b11dc521b32e9a04c72b5 works. Let's see [what Travis says](https://travis-ci.org/bartvm/blocks/builds/47580678)...

bartvm:
Spoke too soon! Attempt 2 in d49adbdbf9c5b5797bd685c5a0436d5912f7bb8a... What I got so far: You need to use [`__reduce__`](https://docs.python.org/2/library/pickle.html#object.__reduce__). The problem with my first try was that when storing the activation functions as class attributes, the lambda operators were automatically converted into methods... Anyway, let's see what Travis says [this time](https://travis-ci.org/bartvm/blocks/builds/47583849)

bartvm:
With the possibility of speaking too soon, again (Travis is slow today), I think now it really runs! Basically, we shouldn't use `HIGHEST_PROTOCOL`. In Python 2 this links to protocol 3, which is a lot faster and considered a good practice. However, in Python 3 it links to the experimental protocol level 4, which is apparently known to be buggy. Dill actually selects the best level automatically (level 3), so we don't need to change anything besides removing the `protocol` argument.

Pylearn2's serialization now crashes, which means that the Pylearn2 Markov chain example no longer runs. Somehow that has to do with the fact that my changes force `cPickle` to try and pickle the lambda operators explicitly now, which it doesn't like. Dill could do it, but since the serialization happens in Pylearn2 code, I can't do much about it. I'm inclined to simply get rid of the Pylearn2 test, or at least skip it on Python 3...

rizar:
Okay, we got tests passing. Tell me if you mind merging it.

bartvm:
Merge away! I still haven't had time to look at the new main loop very closely, but let's merge so that we have something functioning, and I'll go through it later sometime instead of holding things up now.

PR #117, opened 2015-01-20
Miscellaneous fixes reported by Landscape.io

bartvm:
In an attempt to further automate code reviews (considering the little manpower we have) I'm trying out [Landscape.io](https://landscape.io/github/bartvm/blocks). It runs a tool called [prospector](http://prospector.readthedocs.org/en/latest/index.html) which is basically a combination of a whole bunch of static code analysis tools.

It can run things like PEP8, docstring tests, pylint, pyflakes, but also slightly more advanced things like code complexity analysis. The main problem right now is that it doesn't understand the `update_instance` method, which is problematic, because all useful errors are lost in a long list of errors related to `update_instance`.

Maybe that's a good reason to get rid of `update_instance`, which is something that I've been thinking about anyway. Although a convenient shortcut, it can make it pretty unclear which attributes have actually been assigned, and many classes have attributes they probably don't need, which are basically bugs waiting to happen... And of course, explicit is better than implicit. It would be quite a bit of search-and-replace work though, so I'll sleep on it :weary:


blocks/log.py
```diff
@@ -19,7 +19,7 @@ class AbstractTrainingStatus(object):
         The number of iterations done.
     epochs_done : int
         The number of epochs done.
-    _epoch_ends : list
+    epoch_ends : list
```
rizar: Oops, we can not do that right now. Currently private content of the log and the status is distinguished from public by a leading underscore. The `Printing` extension ignores private attributes. This is how I keep the output reasonably compact. While I have some ideas of better solutions (adding a sort of priorities to the attributes), I suggest we keep it like this for the 0.1 version.

rizar:
I think this is quite cool to have such sort of automatic supervision, but getting rid of `update_instance` is quite a high price. It does increase code readability in many cases.

blocks/log.py
bartvm: Okay, I'll revert those.

bartvm:
I have to say that although I loved it in the beginning because it was so much shorter, I'm actually falling out of love with it because I think it often does the opposite of increasing code readability. Instead of seeing immediately which attributes have been assigned, I need to: (1) see that `update_instance` is used and know what it does exactly; (2) read all the lines above it to see which local variables were created; (3) check the constructor arguments.

Perhaps more imporatntly, if we get more developers, I'm worried that they won't realize they are setting or overriding particular attributes with `update_instance` e.g. like this:
```python
def foo(self, foo, bar=None):
    self.bar = 'bar' if bar is None else bar
    update_instance(self, locals())  # Sets self.foo, but also overrides self.bar to None again!
```

rizar:
Ok, I see your point. Would be nice to come up with something less error-prone, manageable by landscape but also clearer than 

```python
self.a = a
self.b = b
self.c = c
```

because the latter is not that pleasant to read.

bartvm:
I guess we could use
```python
self.a, self.b, self.c = a, b, c
```
whenever it fits on a single line. It's explicit but doesn't take up as many vertical space. I can't think of any other way.

rizar:
We could introduce a code-style obligation to 

* group all trivial assignments together

* separate this section by blank lines

That would save people from the right-hand sides identifies: they must be necessarily equal to the left-hand side ones.

bartvm:
Sounds good!

bartvm:
Okay, so that's half an hour of my life I will never get back, but I replaced all the `update_instance` uses with explicit attribute assignments. It was actually far less bad than I expected, I don't think there is any place where it resulted in more than 4-5 lines being added.

@rizar Can we merge this? I don't want to end up with merge conflicts; it's becoming a pretty big PR again.

rizar:
I am afraid you did not separate the trivial assignments by blank lines as we agreed (risking to cause legitimate anger...).

As for merging, go ahead, I will merge my validation PR after you.

bartvm:
But I did! The other attribute assignments are the ones where e.g. defaults are set. The only other exception is that sometimes the `super` call was made before the attribute assignment. I kept those first, because there's a chance that the attribute assignments were overriding those from the `super` call, so interchanging them would break things.

So it should like this right now everywhere:

```python
super(SomeClass, self).__init__(**kwargs)
self.a = a
self.b = b

if c is None:
    c = []
self.c = c
```

rizar:
Well, my idea was 

```python 
super(SomeClass, self).__init__(**kwargs)
if c is None:
    c = []

self.a = a
self.b = b
self.c = c
```

That is I just thought that if the `update_instance` call is replaced by a bunch of trivial assignments and  I clearly see which block of code it is, I would be fine.

bartvm:
Ah, I misunderstood you then, my bad. I'll be happy to change it, but for the sake of my sanity, let me merge this one first and muster the courage to change everything again tomorrow or so :p

rizar:
No problem!

rizar:
And when do you do it, would you nice if you mention this policy in the developer guidelines, e.g. "all assignments of form self.a = a should be grouped together and clearly separated from the rest of the code".

bartvm:
Already did, as part of this PR. I'll update it to fix the misunderstanding we had.

PR #118, opened 2015-01-20
Developer guidelines

bartvm:
Quickly wrote some of them up, because many of these quirks tend to be forgotten! I'll merge straight away because it's just docs, but comments welcome of course.

### DIFF
--- a/docs/developer_guidelines.rst
+++ b/docs/developer_guidelines.rst
@@ -6,9 +6,8 @@ ensure the codebase is of high quality, we ask all new developers to have a
 quick read through these rules to make sure that any code you contribute will be
 easy to merge!
 
-Code style
-----------
-
+Formatting guidelines
+---------------------
 Blocks follows the `PEP8 style guide`_ closely, so please make sure you are
 familiar with it. Our `Travis CI buildbot`_ runs flake8_ as part of every build,
 which checks for PEP8 compliance (using the pep8_ tool) and for some common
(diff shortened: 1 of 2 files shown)


docs/developer_guidelines.rst
```diff
@@ -41,9 +40,92 @@ Blocks:
...
 
+Code guidelines
+---------------
+Some guidelines to keep in mind when coding for Blocks. Some of these are simply
+preferences, others stem from particular requirements we have e.g. in order to
+serialize training progress, support Python 2 and 3 simultaneously, etc.
+
+Validating function arguments
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+In general, be Pythonic and use `duck typing`_. That is, don't check the type of
```
rizar: Well, I think sometimes checks make a lot of sense. They prevent highly non-trivial bugs. I am not sure that explicitly encouraging people not to do checks (what they anyways tend not to do) is the right way to go for Blocks.

docs/developer_guidelines.rst
```diff
@@ -41,9 +40,92 @@ Blocks:
...
+Serialization
+~~~~~~~~~~~~~
+To ensure the reproducibility of scientific experiments Blocks tries to make
+sure that stopping and resuming training doesn't affect the final results. In
+order to do so it takes a radical approach, serializing the entire training
+state using dill_ (an extension of Python's native pickle_). Some things cannot
+be pickled, so their use should be avoided:
+
+* Generators
+* Dynamically generated classes (*possible but complicated*)
```
rizar: Unfortunately we have to add list iterators to this list of shame. And I just guess all other standard container iterators.

docs/developer_guidelines.rst
bartvm: Okay, I'll nuance it a bit. It's mostly to discourage checks like `isinstance(var, six.integer_types)` when a float, NumPy scalar, etc. would work too. Same thing for checks like `isinstance(var, list)` when a tuple, dictionary view, or any other iterable would work, etc. Checks such as `isinstance(var, TheanoSharedVariable)` make sense of course.

PR #120, opened 2015-01-20
Dataset evaluator is back!

rizar:
I refactored it a bit and fixed tests. As I am running locally on Python 3, let's see how it works for Python 2 at Travis.


bartvm:
If you think this is good, feel free to merge. Is this the last of our core features then? I want to finish up #114 so that we can showcase some language stuff at the tutorial, but otherwise I'll try to go over the entire library to see if there's stuff worth refactoring/cleaning up, and writing some docs/tuturials, for the next few weeks.

rizar:
Almost.  I will also need per-batch monitoring, so it will also go to the core features.

We will also have to care about monitoring not Theano expressions, but I guess this can wait a bit.

bartvm:
Ah, I thought per-batch monitoring was already in here. I would definitely consider that a core feature. Non-Theano expressions I think we need to keep in mind when determining the interface, but I agree that it can wait till later if needed.

rizar:
We have functional MNIST training with validation here! Summary of this pull request: 

* `DatasetEvaluator` is back, refactored, and tested

* a small extension for using the evaluator in the main loop

* added validation to MNIST examples

* added a `MisclassificationRate` brick

* fixed a bug in MNIST: test set size was not declared

* sort in alphabetical order when printing

bartvm:
Awesome! Forgot to mention the MNIST bug, sorry, I fixed it myself as part of #114 as well... Heading for lunch/lab now, but feel free to merge once the merge conflicts are resolved so that we have a completely functioning MNIST example to work with

PR #124, opened 2015-01-20
Make Cost and CostMatrix abstract classses

bartvm:


blocks/bricks/cost.py
```diff
@@ -18,6 +26,11 @@ def apply(self, y, y_hat):
         return self.cost_matrix.application_method(
             self, y, y_hat).sum(axis=1).mean()
 
+    @abstractmethod
+    @application
```
rizar: Did you try that such nested decorators work?

rizar:
Nice, merge when you want. I should say that `cost.py` definitely needs more docs...

blocks/bricks/cost.py
bartvm: Should do, I've used it before. It's just he order that becomes important
usually, but in this case they're complementary, so it should be good.

bartvm:
Agreed!

PR #127, opened 2015-01-21
WIP: Separate application instances for each brick instance

### LINKED ISSUE #125: Look into using MethodType to turn applications into bound classes instead of attributes with descriptors
Right now application methods (instances of `Application` such as `brick.apply`) are actually just class attributes. This means that a single application instance is shared by all instances of a brick. The way it is handled right now is that the descriptor sets `self.brick` and feeds that as `self` to the relevant functions.

This causes really weird behaviour!

```python
from blocks.bricks import application, Brick
class TestBrick(Brick):
    @application
    def apply(self, x):
        print self
        print x
test_brick_1 = TestBrick('test_brick_1')
test_brick_2 = TestBrick('test_brick_2')
apply_1 = test_brick_1.apply
apply_2 = test_brick_2.apply
apply_1(tensor.vector())
```
```
<__main__.TestBrick object at 0x7ff409ee1c50: name=test_brick_2>
test_brick_2_apply_output_0
```
Because the last descriptor call (`apply_2 = test_brick_2.apply`) set `self.brick` to `test_brick_2`, all consequent calls to any `TestBrick` application method thinks that they are `test_brick_2`, even if they are actually `test_brick_1`. Very bad!

I need to do some more reading, but I think it should be possible to actually bind the `Application` instances to the `Brick` instances using `MethodType` (or rather, the `six` wrapper [`create_bound_method`](http://pythonhosted.org/six/#six.create_bound_method)) or something similar.

The idea would basically be that whenever a brick gets instantiated, each application method gets replaced by a new, *bounded* instance of `Application`. This prevents different instances of bricks interfering with each other.

The immediate problem is that `MethodType` doesn't allow attributes to be set, so we might need to do something slightly different. Maybe something along the lines of [this](http://www.ianbicking.org/blog/2008/10/24/decorators-and-descriptors/index.html#comment-54812).

bartvm:
Very much WIP; doesn't work yet for recurrent bricks, some functionality is missing (like checking whether the caller is a parent), and some cleanup/simplification might still be possible.

Fixes #125.

### DIFF
--- a/blocks/bricks/__init__.py
+++ b/blocks/bricks/__init__.py
@@ -22,32 +22,33 @@ class _Brick(ABCMeta):
     """Replace application methods with :class:`Application` instances."""
     def __call__(cls, *args, **kwargs):
         brick = super(_Brick, cls).__call__(*args, **kwargs)
-        applications = {}  # application function -> application
+        applications = {}
         properties = {}
         delegates = {}
+
         for key in dir(brick):
             value = getattr(brick, key)
-            # Replace the application method by an Application instance
+            # Replace the application methods by Application instances
             if hasattr(value, 'application_signature'):
                 application = Application(value)
                 setattr(brick, key, application)
                 applications[value.__func__] = application
 
-            # Set a fake property on the Application instance
+            # Collect the application properties and delegates
             if hasattr(value, 'application_property'):
                 application_function, name = getattr(value,
                                                      'application_property')
                 properties[application_function] = name, value.__func__
-
-            # Delegate requests
             if hasattr(value, 'application_delegate'):
                 application_function = getattr(value, 'application_delegate')
                 delegates[application_function] = partial(value.__func__,
                                                           brick)
+
+        # Attach the properties and delegates to the Application instance
         for application_function, (name, func) in properties.items():
             applications[application_function].properties[name] = func
         for application_function, func in delegates.items():
-            setattr(applications[application_function], 'delegate', func)
+            applications[application_function].delegate = func
         return brick
 
 
@@ -511,20 +512,27 @@ class ApplicationCall(object):
 class Application(object):
     """A particular application of a brick.
 
-    Used by the :meth:`application` decorator. This wraps the original
-    application method.
+    When a brick is constructed, its application methods are automatically
+    replaced by an instance of this class.
 
     Parameters
     ----------
-    application : object
-        The application method (member of a brick) to wrap
+    application_method : object
+        The application method (member of a brick) to replace.
 
-    Raises
-    ------
-    ValueError
-        If this class is used without being a member of a brick.
+    Attributes
+    ----------
+    properties : dict
+        Properties are methods of the brick which are called when a
+        particular attribute is requested of the application.
+    delegate : function
+        When an attribute can't be found on the application instance and
+        this attribute is configured, the attribute is requested from the
+        object returned by this function.
 
     """
+    call_stack = []
+
     def __init__(self, application_method):
         self.application_method = application_method
         self.brick = application_method.__self__
@@ -551,7 +559,7 @@ class Application(object):
         .. warning::
 
             Properly set tags are important for correct functioning of the
-            framework. Do not provide args to your apply method in a way
+            framework. Do not provide inputs to your apply method in a way
             different than passing them as positional or keyword arguments,
             e.g. as list or tuple elements.
 
@@ -586,12 +594,10 @@ class Application(object):
         def copy_and_tag(variable, role, name):
             """Helper method to copy a variable and annotate it."""
             copy = variable.copy()
-            # This is the Theano name
-            copy.name = "{}_{}_{}".format(
+            copy.name = "{}_{}_{}".format(  # Theano name
                 self.brick.name, self.application_method.__name__, name)
             copy.tag.application_call = call
-            # This is the annotated name
-            copy.tag.name = name
+            copy.tag.name = name  # Blocks name
             copy.tag.role = role
             return copy
 
@@ -608,8 +614,15 @@ class Application(object):
                 kwargs[name] = copy_and_tag(input_, VariableRole.INPUT, name)
 
         # Run the application method on the annotated variables
-        outputs = self.application_method(*args, **kwargs)
-        outputs = pack(outputs)
+        if self.call_stack and self.call_stack[-1] is not self.brick and \
+                self.brick not in self.call_stack[-1].children:
+            raise ValueError
+        self.call_stack.append(self.brick)
+        try:
+            outputs = self.application_method(*args, **kwargs)
+            outputs = pack(outputs)
+        finally:
+            self.call_stack.pop()
 
         # Rename and annotate output variables
         for i, output in enumerate(outputs):
@@ -653,6 +666,12 @@ def application(*args, **kwargs):
     This decorator can be used both with and without passing attributes
     that will become part of the signature.
 
+    Technically speaking this decorator only tags the method in question as
+    being an application method. When the brick is constructed, the
+    :class:`_Brick` metaclass is responsible for replacing this method with
+    an instance of :class:`Application` which will perform the actual
+    wrapping.
+
     """
     assert (args and not kwargs) or (not args and kwargs)
 
@@ -679,9 +698,9 @@ def application(*args, **kwargs):
         application_function, = args
         return tag_application_function(application_function)
     else:
-        def wrap_application(application_function):
+        def application(application_function):
             return tag_application_function(application_function)
-        return wrap_application
+        return application
 
 
 class Random(Brick):
--- a/blocks/bricks/__init__.py
+++ b/blocks/bricks/__init__.py
@@ -1,8 +1,9 @@
 """The interface of bricks and some simple implementations."""
 import inspect
-import functools
 import logging
+from abc import ABCMeta
 from collections import OrderedDict
+from functools import partial
 from itertools import chain
 
 import numpy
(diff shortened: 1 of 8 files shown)


bartvm:
Update: Making sure the caller is a parent is now done.

Current approach:
* The decorator labels the application method as such, and does the same for delegates and properties
* When the brick is instantiated, the `_Brick` metaclass finds the tagged application methods and replaces them with an instance of `Application`
* The metaclass also finds the properties and delegates and attaches them to the new `Application` instance

Potential pitfall (only one I can think of so far):
* If someone tries really hard, they can attach attributes to the application method which will be lost when they are replaced on instantiation e.g. `Brick.apply.__func__.x = 2` will give `Brick.apply.x -> 2` but `Brick().apply.x -> AttributeError`. This can be fixed by selectively copying the relevant attributes, but not sure that's necessary.

rizar:
Cool, this is definitely a better approach. I should say I start believing in metaclasses: whereever you use them, code becomes significantly clearer. When you are done with recurrent bricks, I will sit down and read all the code carefully, since it is the core, but so far I like it!

bartvm:
Just a progress report: It becomes a bit of a nightmare because of some use cases in the recurrent blocks (e.g. getting attributes from classes instead of objects, `GatedRecurrent.apply.inputs`). After a bit more tinkering, I've gone back to the drawing board and am trying out something else. Instead of replacing the methods with `Application` instances on `Brick` instantiation, I'm going to create two classes: `Application`, which is bound to a particular `Brick` class, and `BoundApplication`, which is bound to a particular instance. So data can be stored on both a class level and instance level, and should fall back nicely. I'll update the PR when I have something functioning!

rizar:
In fact in my original code it was something like that: `brick.apply` transformed in creation of a middleman object.

bartvm:
New version! Basically back to where I was yesterday; only the recurrent bricks left to fix, but this time the behaviour seems a lot more logical. Some test cases I used to check the behaviour (I'll turn this into a proper test before merging):

```python
from abc import ABCMeta                
from types import MethodType           
                                       
import six                             
from numpy.testing import assert_raises
from theano import tensor

@six.add_metaclass(_Brick)
class Brick(object):
    @application(foo='bar')
    def apply(self, x):
        """A docstring!"""
        return x + 1

    @apply.property('name')
    def apply_property(self):
        return 'foo'


class ChildBrick(Brick):
    def __init__(self, sibling):
        self.sibling = sibling

    @application(name='foobar')
    def apply(self, x):
        return x + 1

    @apply.delegate
    def apply_delegate(self):
        return self.sibling.other_apply


class SiblingBrick(Brick):
    @application(bar='bar')
    def other_apply(self, x):
        return x - 1

    @other_apply.property('baz')
    def other_apply_baz(self):
        return 'baz'

brick = Brick()
x = tensor.scalar()
assert brick.apply(x).eval({x: 0}) == 1
assert Brick.apply.foo == 'bar'
assert brick.apply.foo == 'bar'
Brick.apply.foo = 'baz'
assert Brick.apply.foo == 'baz'
assert brick.apply.foo == 'baz'
brick.apply.foo = 'bar'
assert Brick.apply.foo == 'baz'
assert brick.apply.foo == 'bar'
Brick.apply.foo = 'foo'
assert brick.apply.foo == 'bar'

# Now properties
assert isinstance(Brick.apply.name, property)
assert brick.apply.name == 'foo'
assert_raises(AttributeError, setattr, Brick.apply, 'name', 'bar')
brick.apply.name = 'baz'
assert brick.apply.name == 'baz'

# Finally, delegation
sibling_brick = SiblingBrick()
child_brick = ChildBrick(sibling_brick)
assert child_brick.apply.bar == 'bar'
assert child_brick.apply.baz == 'baz'
sibling_brick.other_apply.baz = 'foo'
assert child_brick.apply.baz == 'foo'
child_brick.apply.baz = 'bar'
assert child_brick.apply.baz == 'bar'
assert sibling_brick.other_apply.baz == 'foo'
```

bartvm:
Passing tests now, FINALLY, (at least on my computer), but still not entirely happy with it. The "super" behaviour that several bricks rely on (like Fork) is a bit awkward. Since applications are now simply unbounded `Application` instances on the class, and `BoundApplication` on the instances, a call to `super(Brick, self).apply` results in the original `Application` instance. Which means that e.g. `super(Brick, self).apply.inputs` doesn't work anymore if `inputs` is a property, because properties only work on instances, not on classes (just like in normal Python). In the current master this works because of the descriptor behaviour, but that's buggy for the reasons mentioned in #125.

The solution would be to implement `__get__` to `Application` to automatically bind it, but in that case attributes won't be writable anymore. Maybe I can keep track of all bound versions of `Application` and make `__get__` return a bound version automatically when necessary?

All in all, it's a bit of a nightmare because several bricks (especially in the sequence generation) rely on very particular behaviour (e.g. using `super`, but also calling `__get__` explicitly at one point). I'll continue work again later.

bartvm:
Okay, ready for review @rizar :) I think this solves #125 nicely, and it's actually 27 lines shorter in the end than what we had!

bartvm:
OMG. So that's what you get for starting at the same bit of code too long... I thought I was working on a new branch, but was actually on master and pushed everything :/ Really didn't mean to. Luckily it's a version that is actually functional... I could reverse it if you want me to, but I'll just leave it for now. The commit history might become a bigger mess than it already is if I need to figure out where to roll back to!

blocks/bricks/sequence_generators.py
```diff
@@ -714,8 +714,9 @@ def apply_delegate(self):
         # I can write self.apply because it can be overriden.
         # Thus I have to hack.
         # TODO: nice interface for this trick.
-        AttentionTransition.do_apply.__get__(self, None)
-        return AttentionTransition.do_apply
+        # Bart: This doesn't work anymore, and probably shouldn't; without a
+        # brick attached, it's impossible to call do_apply(brick)...
+        return self.do_apply
```
bartvm: @rizar Please let me know if you want to do something about this.

blocks/bricks/__init__.py
```diff
@@ -1,22 +1,324 @@
...
+        """
+        self.delegate_function = f
+        return f
+
+    def __get__(self, instance, owner):
+        """Instantiate :class:`BoundedApplication` for each :class:`Brick`."""
+        if instance is None:
+            return self
+        elif instance in self.instances:
+            return self.instances[instance]
```
rizar: `self.instances[instance]` is a bit confusing. The dictionary should be renamed, I think, `self.bound_applications` perhaps.

blocks/bricks/__init__.py
```diff
@@ -1,22 +1,324 @@
...
+        else:
+            bounded_application = BoundApplication(self, instance)
```
rizar: Linguistic nitpicking: I guess you wanted to use the third form of the verb bind, which is bound, not bounded.

blocks/bricks/recurrent.py
```diff
@@ -349,8 +348,7 @@ def state_to_reset(self):
     def get_dim(self, name):
         if name == 'mask':
             return 0
-        if name in (GatedRecurrent.apply.sequences
-                    + GatedRecurrent.apply.states):
+        if name in (self.apply.sequences + self.apply.states):
```
rizar: I am not sure it is a nice change: what is the `apply` method is overridden in a descendant class?

blocks/bricks/__init__.py
```diff
@@ -1,22 +1,324 @@
...
+            self.application)
+        if not all(input_ in args_names + [varargs_name] for input_ in inputs):
+            raise ValueError("Unexpected inputs")
+        self._inputs = inputs
+
+    @property_
+    def name(self):
+        return self.application.__name__
+
+    def __call__(self, application, *args, **kwargs):
```
rizar: I would rename the argument number two into `bound_application`.

blocks/bricks/__init__.py
```diff
@@ -1,22 +1,324 @@
...
+
+        brick = application.brick
+
+        # Find the names of the inputs to the application method
+        args_names, varargs_name, _, _ = inspect.getargspec(
+            self.application)
+        args_names = args_names[1:]
+
+        # Construct the ApplicationCall, used to store data in for this call
+        call = ApplicationCall(brick, application)
```
rizar: We should update the documentation of the `ApplicationCall` class to reflect the fact that its `application` attribute is now a `BoundApplication` instance.

blocks/bricks/sequence_generators.py
rizar: Ehm, I am not sure I understand why `AttentionTransition.do_apply.__get__(self)` would not do the job. Doesn't it create a `BoundApplication` object similar to `self.do_apply`, with the exception of bypassing polymorphic method choice?

And by the way, I've just realized that we could in addition support `Brick.apply(instance)` interface, thus mimicking even closer behavior of usual methods. In order to do that the current `Application.__call__` should be renamed, and in fact there is no reason to call this function `__call__` currently. The new `__call__` method can create a bound application and pass the job the old `__call__` method.

blocks/bricks/recurrent.py
rizar: But OK, having polymorphism here kind of make sense, unlike the case with the `AttentionTransition`. There a delegate application is specified and there should be nothing polymorphic about that.

rizar:
You did a really good job! No need to reverse it, we will fix the issues on our way. I have written some notes in the code. 

I also think it is important to have your tests for adequate behavior from above in the repository.

tests/test_examples.py
```diff
@@ -33,7 +33,11 @@ def test_mnist():
 
 def test_pylearn2():
     filename = 'unittest_markov_chain'
-    pylearn2_test('train', filename, 0, 3, False)
+    try:
+        pylearn2_test('train', filename, 0, 3, False)
+    except OSError:
```
rizar: By the way, what's wrong with this one?

blocks/bricks/__init__.py
bartvm: It's "bound" when referring to being confined to something. "Bounded" is
when you're talking about running. You'll see that Python methods are also
bound, not bounded.

blocks/bricks/__init__.py
bartvm: Will do.

blocks/bricks/__init__.py
bartvm: Okay

tests/test_examples.py
bartvm: It ended up in an infinite recursion at one point, and for some reason it kept crashing my tests. I later found out that it is because of [this terrible line](https://github.com/lisa-lab/pylearn2/blob/master/pylearn2/utils/serial.py#L149). I silence the Pylearn2 logs in the tests, so I never saw that it was waiting for 5 minutes... I'll have another look now that it's done.

blocks/bricks/__init__.py
rizar: That's why I think that `bound_application` would be more appropriate than `bounded_application`, which you have in your code.

bartvm: Ignore me, I just woke up! I thought I had `bound_application` everywhere, but clearly not.

tests/test_examples.py
rizar: Well, in fact now I also have this awful waiting locally, and it is annoying.

blocks/bricks/sequence_generators.py
bartvm: You're absolutely right. It was in the version of my code that I didn't go with in the end that this didn't work anymore. I reintroduced it in #127, and also made the interface work as you suggested.

tests/test_examples.py
rizar: If this test is broken for a good reason let's switch it off, at least temporarily because currently it is just a pain in the ass!

bartvm: Haha, we can do that. I can try fixing it later, but it's not really a priority for me. It's probably got to do with `create_unbounded_method` or `property_` because the error says that `cPickle` fails trying to pickle a function object.

rizar: Okay, will do that in my next pull request coming in a few minutes.

PR #131, opened 2015-01-21
WIP: Training Data Monitoring

rizar:
In this pull request I implement what we typically call per-batch monitoring.

In order to do that the following changes are done:

* I separate a class `AggregationBuffer` from `DatasetEvaluator`. The new class is in fact simply a collection of aggregators and is agnostic to the way data is iterated.

* The `AggregationBuffer` is used in the `TrainingDataMonitoring` to do the job.

What I find quite cool about the way it is currently done is the user can control the period over which data is averaged by configuring the extension.

I plan to do a bit more before merging:

- [x] Add some more documentation to the extension.

- [x] Test the extension.

### DIFF
--- a/blocks/algorithms/__init__.py
+++ b/blocks/algorithms/__init__.py
@@ -171,7 +171,8 @@ class GradientDescent(DifferentiableCostMinimizer):
         self.step_rule = step_rule if step_rule else SteepestDescent()
 
         self.total_gradient_norm = named_copy(
-            sum(g.norm(2) for g in self.gradients), "total_gradient_norm")
+            tensor.sqrt(sum((g ** 2).sum() for g in self.gradients)),
+            "total_gradient_norm")
 
     def initialize(self):
         all_updates = self.updates
--- a/blocks/algorithms/__init__.py
+++ b/blocks/algorithms/__init__.py
@@ -184,7 +184,9 @@ class GradientDescent(DifferentiableCostMinimizer):
 
     def process_batch(self, batch):
         if not set(batch.keys()) == set([v.name for v in self.inputs]):
-            raise ValueError
+            raise ValueError("The names of the input variables of your"
+                             " computation graph must correspond to the"
+                             " data sources.")
         ordered_batch = [batch[v.name] for v in self.inputs]
         self._function(*ordered_batch)
 
--- a/blocks/algorithms/__init__.py
+++ b/blocks/algorithms/__init__.py
@@ -7,6 +7,7 @@ from six import add_metaclass
 from theano import tensor
 
 from blocks.graph import ComputationGraph
+from blocks.utils import named_copy
 
 
 @add_metaclass(ABCMeta)
@@ -168,6 +169,9 @@ class GradientDescent(DifferentiableCostMinimizer):
             else dict(
                 zip(self.params, tensor.grad(self.cost, self.params))))
         self.step_rule = step_rule if step_rule else SteepestDescent()
+
+        self.total_gradient_norm = named_copy(
+            sum(g.norm(2) for g in self.gradients), "total_gradient_norm")
 
     def initialize(self):
         all_updates = self.updates
--- a/blocks/monitoring/evaluators.py
+++ b/blocks/monitoring/evaluators.py
@@ -10,14 +10,117 @@ from blocks.graph import ComputationGraph
 logger = logging.getLogger()
 
 
+class AggregationBuffer(object):
+    """Intermediate results of aggregating values of Theano expressions.
+
+    Incapsulates aggregators for a list of Theano expressions. Collects
+    the respective updates and provides initialization and readout
+    routines.
+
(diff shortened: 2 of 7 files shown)


rizar:
Stumbled upon an interesting situation: if I do per-batch monitoring, average results and commit results to the log only at the end of the epoch, what should be my policy with data independent properties? E.g. should I report the average gradient norm during the epoch or the gradient norm after the last iteration? What behaviour should be default?

rizar:
Okay, there first funny issues with monitoring / aggregation

1. Current approach for aggregation schemes does not really work for anything except scalars. The main problem is that the real shape of the monitored tensor is only available with the data. Therefore we need to somehow change the shape of the accumulator when data arrives. After a number of tries and failures I gave it up, though I am pretty sure a solution exists. 

2. Depending on whether an aggregation scheme is used, data independent expressions (such as sum of matrix elements) are read out at different stages of training (see test_monitoring.py for an illustration of the problem). I think I know how to fix this one, by perhaps I would merge first.

What do you guys think about this pull request? 

@bartvm, @janchorowski

janchorowski:
On 1/21/2015 9:20 PM, Dmitry Bogdanov wrote:
>
> Okay, there first funny issues with monitoring / aggregation
>
> 1.
>
>     Current approach for aggregation schemes does not really work for
>     anything except scalars. The main problem is that the real shape
>     of the monitored tensor is only available with the data. Therefore
>     we need to somehow change the shape of the accumulator when data
>     arrives. After a number of tries and failures I gave it up, though
>     I am pretty sure a solution exists.
>
Only the number of dimensions has to match, it is known at compile time. 
The exact shape of a shared variable can change during run. At least it 
works for trainers in GH.

rizar:
That's true, but it is a bit more involved in practice. You can try yourself. I have an idea how to fix, though, but it is not worth telling before trying.

rizar:
Okay, the second weird issue seems to be combated by introducing another aggregation scheme. Now values of all expressions are computed right before another parameters are changed by the training algorithm.

Fixing tensor aggregation is rather out of scope of this pull request, so I will just create an issue.

blocks/algorithms/__init__.py
```diff
@@ -169,6 +170,9 @@ def __init__(self, step_rule=None, gradients=None, **kwargs):
                 zip(self.params, tensor.grad(self.cost, self.params))))
         self.step_rule = step_rule if step_rule else SteepestDescent()
 
+        self.total_gradient_norm = named_copy(
+            sum(g.norm(2) for g in self.gradients), "total_gradient_norm")
```
bartvm: Should probably be `tensor.sqrt(sum(g ** 2 for g in self.gradients))`; the L2 norm isn't additive.

blocks/monitoring/evaluators.py
```diff
@@ -4,20 +4,130 @@
...
+from blocks.monitoring.aggregation import _DataIndependent, Mean, TakeLast
 from blocks.graph import ComputationGraph
 
 logger = logging.getLogger()
 
 
+class AggregationBuffer(object):
+    """Intermediate results of aggregating values of Theano expressions.
+
+    Incapsulates aggregators for a list of Theano expressions. Collects
```
bartvm: *encapsulates

blocks/algorithms/__init__.py
rizar: Good point!

rizar:
The pull request is ready to merge, but I have a weird problem with importing PyLearn2 in one out of 4 test runs. @bartvm , can you look?

bartvm:
`fatal: read error: Connection reset by peer`, looks like Pylearn2 installation failed because of a connection error. I restarted it.

blocks/algorithms/__init__.py
```diff
@@ -169,6 +170,10 @@ def __init__(self, step_rule=None, gradients=None, **kwargs):
                 zip(self.params, tensor.grad(self.cost, self.params))))
         self.step_rule = step_rule if step_rule else SteepestDescent()
 
+        self.total_gradient_norm = named_copy(
```
bartvm: Maybe on the long term it would be useful to store all these variables in an attribute `TrainingAlgorithm.monitoring_channels` (or whatever term we want to use). In that case a user can just go through the algorithms `monitoring_channels` attribute to see what the algorithm suggests in terms of monitoring values. That's easier than having to go through the documentation of each algorithm to see what monitors it makes public that you could monitor.

rizar: Yep, I also had it mind. I see no problem in adding it a bit later, and seems we can release 0.01 without it.

bartvm: Cool, can you make a ticket for it? (I'm on my phone)

rizar:
Any objections to merging this guy?

bartvm:
Nope!

PR #137, opened 2015-01-22
Markov Chain Example

rizar:
I ported the markov chain example to use Blocks only. It became so much simpler and convenient to play with! 

In the process of doing so I also made the following changes:

* I changed the way `KeyboardInterrupt` is handled. Now the signal is caught and a respective record to the log is made. The main loop finishes training when it comes across such a record. In fact I think that a better and a more flexible way to do this would be to push the respective logic in an extension and add this extension by default. But as we currently do not have an institute of extensions added by default, I put it the main loop so far.

* Added another callback slot to the extension interface called `on_resumption`.

* Wrote a small script `continue.py` to continue training. 

* Switched off broken PyLearn2 test.


bartvm:
Cool! Let me just make a PR to this PR to change the way we do scripts. We had to fix those in Pylearn2, so we might as well do it right from the beginning here.

I think there should be a top-level folder called `bin`. Files in there should be called e.g. `blocks-continue` and look something like:
```python
#!/usr/bin/env python
import sys
from argparse import ArgumentParser

import dill
from six import exec_

from blocks.scripts import continue

if __name__ == "__main__":
    parser = ArgumentParser("Continues your pickled main loop")
    parser.add_argument(
        "path", help="A path to a file with a pickled main loop")
    parser.add_argument(
        "--rec-limit", type=int, help="The recursion depth limit")
    args = parser.parse_args()

    continue(path=args.path, rec_limit=args.rec_limit)
```
The file `blocks/scripts/__init__.py` could then have:
```
import sys

import dill

def continue(path, rec_limit=None):
    if rec_limit:
        sys.setrecursionlimit(rec_limit)
    main_loop = dill.load(open(path, "rb"))
    main_loop.run()
```
Finally, we need to add the script to `setup.py` using `scripts=['bin/blocks-continue'`.

This (a) installs scripts automatically with `pip` or by running `setup.py`, so `blocks-continue` becomes available on your path; (b) allows the script to be imported from within the namespace e.g. other scripts can do `from blocks.scripts import continue`; and (c) keeps script-specific code outside of the package.

rizar:
Wow, that's a serious approach! Of course I do not mind you making these changes.

bartvm:
How come you need to change the recursion limit btw?

rizar:
`dill.dump` crashes in markov chain example if I don't do this. I know this is weird, but somehow it manages to reach recursion depth above 1000!

blocks/main_loop.py
```diff
@@ -73,13 +77,14 @@ def run(self):
...
+            else:
+                self._run_extensions('on_resumption')
             while self._run_epoch():
                 pass
-        except KeyboardInterrupt:
-            self._run_extensions('on_interrupt')
         except TrainingFinish:
             self.log.current_row.training_finished = True
         finally:
+            signal.signal(signal.SIGINT, default_action)
```
bartvm: Finally have a bit of time to read the main loop code thoroughly. Just wondering what's happening here: `_handle_keyboard_interrupt` runs the `on_interrupt` extensions and makes a record in the log. Training then finishes up the rest of the batch before `_check_finish_training` finds the log entry, and raises `TrainingFinish`. This causes another log entry to be added (`training_finished`)... But then `default_action` gets called, killing the main loop before `after_training` extensions get called.

Where would you save the state of the model? You don't want to do it in `on_interrupt`, because you could be in the middle of anything. But Python dies before the `after_training` extensions get run.

P.S. Instead of storing `default_action`, I think you can just use `signal.SIG_DFL` handler.

rizar: `default_action` does not get called, it is just set as a new handler. Therefore `after_training` always gets called, and this is where saving is actually done.

There was indeed certain inconsistency: changing back the default handling scheme should be done last, not before `after_training` is handled. I fixed that plus added a bit more docs, see #143.

PR #139, opened 2015-01-22
Split bricks/__init__.py

### LINKED ISSUE #133: Split blocks/bricks/__init__.py
Coming in over 1260 lines this file is becoming a bit monstrous. I think we should split it up into two parts: Base classes like Brick, Application, ApplicationCall, BoundApplication, etc. and a file with implementations such as Linear, MLP, Sequence, the activations, etc. (should split it about 800-400 lines)

bartvm:
Fixes #133.


bartvm:
Also made the default seed configurable.

Issue #55, opened 2014-12-03
Switch from `RandomStreams` to `sandbox.MRG_RandomStreams`


bartvm:
It's quite a bit faster on GPU

bartvm:
Stalled for now because of technical reasons detailed in https://github.com/bartvm/blocks/pull/75#issuecomment-67707419

bartvm:
Closed though #141.

Issue #140, opened 2015-01-22
Should recurrent bricks attach updates to application_call instead of scan apply node?


bartvm:
For two reasons:
* Right now updates are spread over two places, which is confusing
* It will circumvent the bug holding up #75, which can actually make a huge difference for some models

bartvm:
Just remembered @rizar already commented on this:

> An alternative would be to attach updates to the ApplicationCall, but this does not cover the case when a user uses scan in an intermezzo between two blocks-based parts.

Although I agree, I think that should be okay. If a user doesn't use a brick in a particular segment, they can also add the updates themselves to the algorith manually using the `add_updates` method; simple enough step.

rizar:
After a few minutes of thinking I came up with an example where you can not simply add updates: what if you have updates in a computation graph of a variable monitored on an auxiliary dataset? There's no `add_updates` for it so far. 

We could use "fake" application call objects for this purpose. That would also allow to add auxiliary variables to the graph without using bricks.

rizar:
I had a sort of revelation and now it seems I see the big picture: let's introduce a notion of an annotation of a computation graph. An annotation is a bunch of auxiliary variables plus updates. The `ApplicationCall` becomes a subclass of the `Annotation` class. 

The fascinating thing coming from doing it this way is that the computation graph module should know nothing about bricks. It simply works with an annotated graph. The user can just work in bare Theano, add annotations and still benefit from using the framework.

bartvm:
Wait, I'm not sure I follow entirely: The `ApplicationCall` is a subclass of `Annotation`, but what functionality does it have that `Annotation` doesn't? And what do we attach `Annotation` to? To the `ComputationGraph` object?

Does this mean that you want to apply selectors to the `ComputationGraph` instead of to the `Brick` hierarchy?

rizar:
In addition to carrying auxiliary variables and updates an `ApplicationCall` instance also has references to the brick and to the application. We can attach annotations to the variables exactly in the same way `ApplicationCall`'s are currently attached, just that some of these annotations will be also instances of the `ApplicatoinCall` class. 

No, I do not want to apply selectors to the CG.

bartvm:
Closed through #141.

PR #141, opened 2015-01-22
Regularization

### LINKED ISSUE #126: Additional cost?
Question for @janchorowski, but maybe @rizar knows as well: What are the variable roles "cost" and "additional cost" supposed to represent, and how are they different? Can we done with one of the two? If not, we probably need some documentation as to why.

bartvm:
Following discussion in #126 this PR removes the `add_monitor` method and the `VariableRole.MONITOR` role. It also makes auxiliary variables more easily accessible from the `ComputationGraph` instance.

The goal is to extend the MNIST example with regularization.

Fixes #126, #55, #140 and #146 (yeah!).


blocks/graph.py
```diff
@@ -37,6 +37,7 @@ def _get_variables(self):
         self.applies = set()
         self.application_calls = set()
         self.updates = []
+        self.auxiliary_variables = set()
```
rizar: Maybe we should better add a routine to the ComputationGraph that filters variables according to their roles?

bartvm: Ah, we can do that too, but auxiliary variables don't have a role right now; they're auxiliary variables simply because they were attached to an application call. But yeah, we need an easy way to e.g. return auxiliary variables which have a `VariableRole.COST` role.

blocks/graph.py
```diff
@@ -47,6 +48,7 @@ def recursion(current):
                 if application_call not in self.application_calls:
                     self.application_calls.add(application_call)
                     for av in application_call.auxiliary_variables:
+                        self.auxiliary_variables.add(av)
                         av.tag.application_call = current.tag.application_call
```
bartvm: @rizar @janchorowski Is there any reason this tagging happens here and not in the `add_auxiliary_variable` method? We could simply do `expression.tag.application_call = self` over there, no?

rizar: Looks really weird, I guess there is no good reason.

blocks/graph.py
bartvm: On this topic: I was just looking at using selectors to e.g. return the L2 weight norms of a specific set of bricks. I feel that if we're not careful, we'll end up having very similar functionality between the brick selectors and the computation graph. Any ideas on how work should exactly be distributed between these two? Which one of the two should be in charge of collecting particular regularization variables from the application calls?

bartvm: Sorry for this monologue, but back to the first topic: I'll tag all auxiliary variables with `VariableRole.AUXILIARY`, that way the same method can be used to filter out `VariableRole.COST` and `VariableRole.AUXILIARY`. Makes more sense.

rizar: Well, no reasonable brick will declare L2 penalty as an auxiliary variable, since it is too generic and to easy to evaluate in a way your describe above. The auxiliary variables is a mechanism for more involved cases, when there exists a highly specialized local variable that is hard to compute otherwise, like Jan's penalty.

rizar: I like the plan from above.

bartvm: Wait, why wouldn't it attach it as an auxiliary variable? My idea was this: All the linear bricks automatically assign their weight decays (L1, L2 norms) as auxiliary variables. When you want to apply weight decay you would simply query the computation graph (or use selectors?) to give you the relevant weight decays which you can then add to the cost you are training on. What I haven't really figured out is how we would filter out a particular weight decay if we only want to use it for 1 brick.

rizar: Okay, I see, I had a rather different plan for L2 and L1: have a routine `add_L2_regularization(cost, params)` where cost is the cost Theano variable, and `params` are the parameters that you want to include in the penalty calculation. This routine would return a new Theano variable corresponding to the cost with regularization. 

So when I want to apply this regularization only for parameters of certain bricks I first use `Selector` to choose those bricks and extract their parameters and then pass this parameters to `add_L2_regularization`.

I like this plan more than what you propose, because when separation in a few consecutive steps possible, it should be done. When it is not possible, that is when the regularization term only makes sense in the context of a particular brick, the brick can add this term to the graph as an auxiliary variable.

I can do a pull request for L2 if you want.

bartvm: Although the interface looks okay, my worry is about getting the correct the parameters out, because this means knowing for each brick which of its parameters are weights. Do we want to start tagging parameters with their roles e.g. `VariableRole.WEIGHTS` and `VariableRole.BIASES`? Otherwise how is a user expected to e.g. select the 3rd and 7th set of weights in a 20-layer deep autoencoder created through the `MLP` class.

Although I appreciate the argument about splitting things up into parts, I also don't like the idea of some regularizers being done this way, and others in yet another way. It would be nice to have a single unified interface for that, especially since the interface is pretty complex already; it will never be as easy as Pylearn2's `model.get_weight_decay()`. The great thing of attaching regularizers and other costs to application calls then, is that you can do some crazy things. Let's say it's now a thing to penalize the L2 difference between successive weight matrices in a deep autoencoder, but you want to exclude the middle layers from this. For an MLP, this would be easy by using the application call:
```python
for num, (i, j) in enumerate(zip(self.linear_transformations[:-1],
﻿                                 self.linear_transformations[1:])):
    application_call.add_auxiliary_variable(
        (i.W - j.W).norm(2), roles=[VariableRole.COST],
        name='L2_difference_{}'.format(num))
```
```python
costs = cg.get_variables(roles=[VariableRole.COST])
costs = [cost for cost in costs if cost.name != 'L2_difference_9']
```
The alternative would be to write a new function, `add_successive_L2_regularization(param)`*, but now we're stuck with the problem of getting the parameters out again.

I think it would be nice if we could come up with a very specific example of what a complicated case would look like exactly. Let me give this as a first attempt (attempts, really):

```python
mlp = MLP(activations=[Tanh() for _ in range(4)], dimensions=[np.random.randint(11)])
y_hat = mlp.apply(x)
cost = CategoricalCrossEntropy().apply(y, y_hat)

cg = ComputationGraph([y_hat])
regularizers = cg.get_variables(roles=[VariableRole.COST])
# Maybe it would be nicer if start using roles more heavily,
# so that we could say e.g. cg.get_variables(roles=[VariableRole.L2_WEIGHT_DECAY])
l2_weight_decay = [var for var in regularizers if 'L2' in var.name]
# This would require variables to be returned in topological order
l2_weight_decay = l2_weight_decay[:5] + l2_weight_decay[6:]
# Probably cleaner
l2_weight_decay = [var for var in l2_weight_decay
                   if var.tag.brick is not mlp.linear_transformations[6]]
```
I guess that this is which is where I started wondering whether maybe we should stop using selectors on bricks and select on the computation graph itself instead. This is in line with out idea of being more of a Theano framework, and would make best use of the annotation. Of course, we could have both: Selectors for bricks, and for the computation graph. Then we could do something like.
```python
# This or a more fancy kind of matching language like Jan's if you will
weight_norms = cg.get_variables(roles=[VariableRole.L2_WEIGHT_DECAY],
                                exclude_bricks=[mlp.linear_transformations[5]])
```

*Not sure why you wanted to return the cost directly, it seems easier to return just the regularization term so that the user can add that to the cost himself, potentially with a variable regularization coefficient.

rizar: Let me start from this point:

> Although I appreciate the argument about splitting things up into parts, I also don't like the idea of 
> some regularizers being done this way, and others in yet another way.

I think the situation here is very similar to the one with models: I believe there is no such thing as "the regularization" and there is nothing in common between different regularization. Examples:
dropout is a blunt force insertion in the computation graph, L2 is addition of a term to the cost, matrix norm constraint is a change in the training algorithm and so on and so forth. Thus we can not have a common interfaces for all regularizations possible. What we should do though, is to provide tools to build new methods and have a set of implementation of most common ones. 

Next, back to weight decay: there was a bit of my fault here as I forgot that biases are typically not included in the weight decay. But that in fact even strengthens my point: do you want every brick to put three auxiliary variables with roles `WEIGHT_DECAY`, `WEIGHT_DECAY_WITH_BIASES` and `BIASES_DECAY` into the computation graph? And code of every brick would have to know about all regularizations supported by the library.

Instead, I find your idea of having additional roles for parameters more attractive. 

So let's go from a simple cases to more complicated ones:
```python
mlp = MLP(activations=[Tanh() for _ in range(4)], dimensions=...)
y_hat = mlp.apply(x)
cost = CategoricalCrossEntropy().apply(y, y_hat)

# Case 1: I am a machine learning newbie. I want weight decay and do not want to know too much 
# about the way your library works.
cost = cost + weight_decay(cost=cost)
# The call above simply selects all parameters tagged as WEIGHTS and returns sum of their square norms.

# Case 2: My weights are OK, but my biases are too large. What shall I do?
cost = cost + weight_decay(cost=cost, roles=[BIASES])

# Case 3: I only care about the third layer of my MLP.
cost = cost + weight_decay(
      params=[param for param in Selector([
                   mlp.linear_transformations[2]]).get_params().values()
              if param.role == WEIGHTS])
# Note: Jan's new selectors should make it look better.
# Alternatively we can provide an interface
cost = cost + weight_decay(bricks=[mlp.linear_transformations[2]])
# which is a short-cut to the whole list comprehension above.
```

Also about difference between `Selector` and `ComputationGraph`: the former deals with the bricks hierarchy and pathes in it, whereas for the latter the world is flat. Computation graph might have routines like `get_variables` from your example, but it should never treat in recursive fashion.
E.g. if I have an MLP and every layer of it adds something to the graph, then in order to exclude variables belonging to all its ancestor I would write

```python
weight_norms = cg.get_variables(roles=[VariableRole.USEFUL_SOMETHING],
                                exclude_bricks=Selector(mlp1).select('//').get_params().values())
```

although this is currently not implemented.

bartvm: Okay, see #146 for my attempt at consolidating all of this!

bartvm:
Okay, first working version! Not much has changed actually, but this is the full list:

* I created an `Annotation` class; it was just a convenience to make `Brick` a subclass of it, let me know if you want me to attach the brick-level variables to the `Application` instead. Users can add annotations manually and they will be picked up by the `ComputationGraph`'s `_get_variables` method.
* To keep the `ComputationGraph` agnostic to bricks, it basically just has a `get_shared_variables(predicate)` function now. I introduced a `VariableFilter` class which allows one to easily create predicates that filter variables based on their roles and/or owning bricks.
* Instead of returning sets of variables, I am now returning a list of variables which is ordered according to the topological order of the apply nodes. For more complicated DAG models this won't help too much, but for a simple MLP this should ensure that all variables are returned in the same order as the layers, which is quite useful.
* Cleaned up the interface of `ComputationGraph` a bit
* Now attaching `updates` from `scan` to the application call as discussed in #140. This allowed me to switch to `MRG_RandomStreams` (which fixes #55).
* Introduced a bunch of new variables roles such as `WEIGHTS`, `BIASES` and `PARAMETER` and some convenience functions to add these roles to variables.
* Bit of documentation.

Main thing missing: Tests, lots of tests, considering this is a pretty crucial part.

bartvm:
@rizar All yours to review!

bartvm:
Forgot to add: Should we prefix the names of e.g. parameters and auxiliary variables with the names of the brick and/or application? Right now it's a bit irritating that you simply end up with `W_norm` and `W` 10 times for an MLP.

blocks/bricks/__init__.py
```diff
@@ -146,12 +147,15 @@ def __init__(self, input_dim, output_dim, **kwargs):
         self.output_dim = output_dim
 
     def _allocate(self):
-        self.params.append(shared_floatx_zeros((self.input_dim,
-                                                self.output_dim),
-                           name="W"))
+        W = shared_floatx_zeros((self.input_dim, self.output_dim), name='W')
+        VariableRole.add_role(W, VariableRole.WEIGHTS)
+        self.params.append(W)
+        self.add_auxiliary_variable(W.mean(), name='W_mean')
```
rizar: I am kind of sceptical that mean weights would be of interest for anybody, weight norms would make more sense.

blocks/bricks/base.py
```diff
@@ -21,6 +21,103 @@ def create_unbound_method(func, cls):
 property_ = property
 
 
+class Annotation(object):
+    """Annotations on Theano variables in a graph.
+
+    Blocks automatically adds annotations to variables created using
```
rizar: Linguistic nitpicking: "In Blocks annotations are automatically added"?

blocks/bricks/base.py
```diff
@@ -738,13 +838,53 @@ def init(self, *args, **kwargs):
...
+    OUTPUT = 'output'
+    #: Any parameter of the model
+    PARAMETER = 'parameter'
+    #: The weights of a particular linear transformation
+    WEIGHTS = 'weights'
+    #: The biases added after a linear transformation
+    BIASES = 'biases'
+
+    @classmethod
+    def add_role(cls, var, role):
```
rizar: In fact this could be achieved by a hierarchy of roles:

```
class Parameter(VariableRole):
     pass

class Weights(Parameter):
     pass
```

blocks/graph.py
```diff
@@ -32,45 +59,58 @@ def __init__(self, outputs):
...
+        variables = [var for var in list(chain(
+            *[apply_node.inputs for apply_node in sorted_apply_nodes]))
+            if not (var in seen or seen.add(var))] + self.outputs
+
+        # While preserving order add auxiliary variables, and collect updates
+        i = 0
+        seen = {'application_call': set(), 'annotation': set(), 'brick': set()}
+        while i < len(variables):
+            var = variables[i]
+            for tag in ('application_call', 'annotation', 'brick'):
```
rizar: This is not exactly what we agreed on in #146. In the comment https://github.com/bartvm/blocks/issues/146#issuecomment-71247985 I meant the following: every variable has a single annotation: `x.tag.annotation`. 
This annotation can be attached manually, bust most often it is attached automatically in `Application.apply`. In this case 
the annotation is an instance of `ApplicationCall`, which subclasses `Application`. The annotation attached in `Application.apply` consists of variables put there by the user and the auxiliary variables of the brick this application belongs to.

I just want to keep the computation graph code clean from brick-related concepts. If it is a thing on its own it will be easier to maintain and keep documented.

Perhaps a better way to implement what is described above: let a variable have several annotations just they currently can have several roles and just as in your PR, with the distinction that all of them belongs to `x.tag.annotations` list. Then in `Application.apply` simply two annotations can be added: one from the brick and another one just an `ApplicationCall` instance.

rizar:
Nice @bartvm , you did a good job of aggregating out yesterday's thought, but to be honest I am still not entirely happy with the results.

I will try to describe a sequence of changes that would make me happy:

1. The current `VariableRole` seems to have come from the pages of an OOP textbook as an example how one should not do. Let's make all this constants a class hierarchy: no need for `add_role` method, easy to add new roles, no need to put all roles in a single file. I would suggest that the root class moves to `graph.py`, together with a few basic roles which are not from the world of bricks, such as PARAMETER, WEIGHTS, BIASES.

2. Instead of having a set of different annotations attributes we are looking for, let's look for only one:  `x.tag.annotations`. Then I am even fine with having a brick in this list, as it just plays a role of yet another annotation and has no influence on the CG's code. It will make code of the predicates that check for variables created by a certain brick a bit more complicated, since they will have to first search for the right annotation, but since this is done in the `VariableFilter`, it does not touch the user, unless he needs extreme custom selection logic, in which case certain familiarity with the framework is a prerequisite.

3. A bit of refactoring: `Annotation` goes to `graph.py`, `VariableFilter` goes to a separate file, `graph.py` does not have `from blocks.bricks import ...` statements. 

That's it so far, I might read it a few more times and find something else.

blocks/graph.py
rizar: There is a revised version of this proposal on the "conversations" page.

blocks/graph.py
```diff
@@ -32,45 +59,58 @@ def __init__(self, outputs):
...
+                annotation = getattr(var.tag, tag, None)
+                if annotation and annotation not in seen[tag]:
+                    seen[tag].add(annotation)
+                    new_avs = [av for av in annotation.auxiliary_variables
+                               if av not in variables]
+                    variables = variables[:i + 1] + new_avs + variables[i + 1:]
```
rizar: Although speed is not our top priority, let's avoid quadratic complexity where it can be done at low cost.  I think one will almost never need to select both usual and auxiliary variables in one call, so they can simply be appended to the end of `variables` list. Also when you do a ` if av not in variables` check, please make sure that `variables` is a set.

examples/mnist.py
```diff
@@ -30,6 +32,11 @@ def main(save_to, num_epochs):
     cost = CategoricalCrossEntropy().apply(y.flatten(), probs)
     error_rate = MisclassficationRate().apply(y.flatten(), probs)
 
+    cg = ComputationGraph([cost])
+    W1, W2 = VariableFilter(roles=[VariableRole.WEIGHTS])(cg.variables)
```
rizar: I really like your idea about returning variables in topological order! This piece of code looks like magic.

examples/mnist.py
```diff
@@ -30,6 +32,11 @@ def main(save_to, num_epochs):
...
+    cost = cost + .00005 * W1.norm(2) + .00005 * W2.norm(2)
```
rizar: Actually L2 regularization means adding not norms but norms squared, so I guess you need `(W1 ** 2).sum()` instead.

bartvm:
Okay, I'll change things around to address your comments. I know it wasn't exactly agreed as #146 (see my first bullet point), but I had already coded something up before that, and wanted to finish the PR before refactoring it -- will get to the refactoring today.

Fully agreed with all the comments, except that I think what @janchorowski had in mind for `VariableRole` is a kind of enumerated type. Sadly, those exist only [in Python 3.4](https://docs.python.org/3.4/library/enum.html), so he settled for a poor man's version. I'm not sure whether an OOP textbook would propose just making a long list of dummy classes. There is a [semi-official backport of the `Enum` type](https://pypi.python.org/pypi/enum34) which we can use, which would be my preferred way of keeping Python 2 support.

bartvm:
Okay, I believe that addresses all your comments, @rizar.

There's on additional thing I added: I wanted to be able to filter parameters by the bricks that they belong to as well. In order to save people from having to add the brick manually each time I made it so that any parameter is automatically annotated with the brick.

rizar:
Thank you for making these changes for me!

I want to get back to the question of roles and try and to explain better why I think those should form a class hierarchy. First, one can not introduce a new role. More exactly, one can use an arbitrary constant but this will look slightly odd when the standard is to write `VariableRole.ROLE`. Apart from this aesthetic issue we have another, a more serious one: suppose I want to introduce a new role, let's say RECURRENT_WEIGHTS or CONVOLUTIONAL_WEIGHTS, and I do not want manually add the PARAMETER and WEIGHTS roles. The only way to do it for me is to change your `Variable.add_role` routine.

What I expect is that the `add_role` routine will essentially go upwards in the role hierarchy and add all ancestors. But we get this for free if we use classes, and those are a ready tool to work with a hierarchical structure. One just has to use `isinstance` for checking roles.

If we go this way, every submodule will be able to declare its legitimate roles that play well with all the rest:
```
# in recurrent.py
class RecurrentWeightsRole(WeightsRole):
     pass
```
And I see no problem that they are scattered around the code: similarly in Theano different types are scattered around the code, and care is taken to make addition of new types possible.

Btw, I do not know much about enumerations in Python, but from my experience with C++ enumerations, I do not anticipate that using them will solve the problem.

blocks/bricks/base.py
```diff
@@ -21,6 +23,51 @@ def create_unbound_method(func, cls):
...
+        ------
+        ValueError
+            If the parameter isn't a shared variable or ``None``.
+
+        """
+        if is_shared_variable(value):
+            VariableRole.add_role(value, VariableRole.PARAMETER)
+            annotations = getattr(value.tag, 'annotations', []) + [self.brick]
+            value.tag.annotations = annotations
+        elif value is not None:
```
rizar: Wait a minute, I remember you wanted parameter substitution to be possible, like `decoder.W  = encoder.W.T`, didn't you?

bartvm: Ah, that's it! When I was coding this I could swear there was a reason I didn't to it initially, but I couldn't come up with it anymore...

bartvm:
Okay, I'll change to use a class hierarchy, although I would like to keep the ability to add multiple roles. I think it's too strong an assumption to say that a single role will always cover everything.

bartvm:
Okay, that should be it. I don't like the idea of using classes as if they're constants, so I instantiated them and use the objects as the constants. I guess if we want to be really neat, we could add a singleton metaclass to the `VariableRole` to make sure there is only one instance of each role at any given time, but I'll defer that to later.

I like the current ability of filtering the CG a lot. I think it can be really powerful if combined with good brick selectors, while still letting users deal directly with the Theano graph.

We'll have to take closer to the tutorial about what you think we should highlight about Blocks, but after taking a step back, I think that I like the idea of branding it as a "Theano toolbox for **building** and **managing** computation graphs, plus features to support experiments".

Basically, bricks are used to **build** large complicated models. Applications allow bricks to interact, asking each other what their inputs, outputs, etc. are. Bricks also allow you to initialize parameters well.

After building your model with bricks, you end up with a highly annotated computation graph. `ComputationGraph` **manage** your graph by filtering out parameters, auxiliary variables, inputs, outputs, monitors, costs, biases, or whatever you care about. At this point, the original bricks can still be used for even more powerful selection and filtering.

Then there are the features such as logging, training algorithms, etc. which are really just a collection of convenience functions that allow you to train and analyze the Theano model. Bricks play no role here, at all.

bartvm:
Added documentation (so also some basic tests, because of the doctests)

blocks/bricks/base.py
```diff
@@ -21,6 +24,48 @@ def create_unbound_method(func, cls):
...
+        """Annotates the variable.
+
+        Raises
+        ------
+        ValueError
+            If the parameter isn't a shared variable or ``None``.
+
+        """
+        if isinstance(value, Variable):
+            add_role(value, PARAMETER)
```
rizar: Just to make sure we're on the same page: this line means that the substituted parameters, which typically are not shared variables, can also bear the PARAMETER role, which in turn means that the reliable way to extract all trainable parameters is to request all shared variables with PARAMETER role. Am I right?

blocks/filter.py
```diff
@@ -0,0 +1,115 @@
...
+    return get_annotation(var, ApplicationCall)
+
+
+class VariableFilter(object):
+    """Filters Theano variables based on a range of criteria.
+
+    Parameters
+    ----------
+    roles : list of :class:`VariableRole` instances, optional
+        Matches any attribute which has one of the roles given.
```
rizar: Did you want to write "variable" instead of "attribute"?

blocks/graph.py
```diff
@@ -94,6 +130,144 @@ def get_theano_function(self):
                                updates=self.updates)
 
 
+def add_role(var, role):
+    r"""Add a role to a given Theano variable.
+
+    Some roles will imply others, using this helper function will make
+    sure that these roles are also added.
```
rizar: This docstring should be updated.

rizar:
Nice Bart, I like what we have now! Also the pull request has become too big and rather inconvenient to work with, I would not mind if you merge it even before covering it with tests.

docs/cg.rst
```diff
@@ -0,0 +1,81 @@
...
+When you apply a brick to a Theano variable, it automatically *annotates* this
+Theano variable, in two ways:
+
+* It defines the *role* this variable plays in the computation graph e.g. it will
+  label weights matrics and biases as parameters, keep track of which variables
+  where the in- and outputs of your bricks, and more.
+* It constructs *auxiliary variables*. These are variables which are not an
+  output of your brick, but might still be of interest. For example, if you are
+  training a neural network, you might be interested to now the norm of your
+  weight matrices, so Blocks attaches these as auxiliary variables to the graph.
```
rizar: Regarding attaching the norms as auxiliary variables: as far as I understand what you want to achieve is automatic monitoring in PyLearn2 style. I agree that this feature will have its customers, but it has to be thought through.
I see it done in the following way: we introduce a new role, call it e.g. `DiagnosticRole`. I think we need to have this subclass of `AuxiliaryRole` because I can imagine a situation where an auxiliary variable is a very bad candidate for monitoring. 

Then the monitoring extensions can accept the cost variable as an optional argument, in which case they automatically extract all diagnostic variables from the computation graph. If we equip the cost bricks with a lot of diagnostic (e.g. categorical cross-entropy can automatically add misclassification error to the CG as a candidate to be observed), we might get PyLearn2 level of simplicity:

```python
cost = CategoricalCrossEntropy(y.flatten(), mlp.apply(x))
main_loop = MainLoop(...,
    extensions=[..., DataStreamMonitoring(cost=cost)])
```

I  am sharing this thought with you because I know you care a lot about making simple examples really simple. I guess for my experiments  I would prefer to carefully choose myself what is monitored.

rizar:
As for the high-level vision of Blocks you present above, I have very little to add. We wanted to have different components independent of each other and we have mostly achieved this goal. At the tutorial we should just go through the components and show how one can assemble a complicated experiment setup with them. I think it is good idea to start from the computation graph, as this is something everybody knows well and show how we expanded it. After the computation graph is introduced, there are two almost orthogonal topics left: bricks and the selector is the first one, all the experiment support is the second one. They can be presented in any order.

blocks/filter.py
bartvm: Yes, missed that one

blocks/bricks/base.py
bartvm: Yup

docs/cg.rst
bartvm: That sounds like a good idea, it would be a nice by-product, although it actually wasn't the main goal of this endeavor. I mostly wanted to get closer to the idea of a "managed Theano graph" without the need for using bricks.

bartvm:
Sounds good to me, I'll merge and add the `DiagnosticRole` when I have some time.

PR #144, opened 2015-01-23
Fetching character level data from text files

rizar:


blocks/datasets/text.py
```diff
@@ -109,13 +109,14 @@ def get_data(self, state=None, request=None):
...
             sentence = self.preprocess(sentence)
+        data = [self.dictionary[self.bos_token]] if self.bos_token else []
         if self.level == 'word':
-            data = [self.dictionary[self.bos_token]] if self.bos_token else []
             data += [self.dictionary.get(word, self.dictionary[self.unk_token])
                      for word in sentence.split()]
-            data += [self.dictionary[self.eos_token]] if self.eos_token else []
         else:
-            raise NotImplementedError
+            data += [self.dictionary.get(char, self.dictionary[self.unk_token])
```
bartvm: We need to update the docs if you want to do it this way (I wrote there that I would turn character level into a list of lists, each sublist being word, whereas you are replacing word boundaries by another number.

rizar: OK, I updated the PR.

Issue #148, opened 2015-01-23
More DataStreamWrappers!


rizar:
As discussed in #114, we need a whole lot of them. Good thing about writing such things is that it can be split up into parts. 

I would put priority on the following ones: 
a) one that splits a data stream into batches 
b) one to pack lists of lists of lists of ... of tuples of whatever into tuples of numpy arrays and masks.

Having those would make it possible to finally train a demo with an attention mechanism.

bartvm:
Both those have been introduced, so probably need more tickets with specific requests.

PR #149, opened 2015-01-24
Language streams

bartvm:
Needs some polishing and tests, but here are some crude versions of the batching and masking data streams.

The masking scheme should probably (a) return data with the proper (time, batch, features, ...) axes (or should that be another wrapper? or up to the user to transform before feeding into the model?) and (b) use the proper `dtype`. If given a Python list of integers, there is no `dtype` to copy, so it should probably check what the Python type of the elements are to determine whether to return floating point or integer numbers.

Addresses #148.

### DIFF
--- a/blocks/datasets/__init__.py
+++ b/blocks/datasets/__init__.py
@@ -613,65 +613,46 @@ class CachedDataStream(DataStreamWrapper):
             cache.extend(data)
 
 
-class NGramStream(CachedDataStream):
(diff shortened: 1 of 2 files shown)


bartvm:
Quick note for myself: I think that Dill might be failing to pickle *closed* file handles, so after closing a file handle it should be removed by the data stream. Need to confirm this later though.

Issue #65, opened 2014-12-13
MNIST Example with Regularization using PyLearn2 Wrappers


rizar:
It would be nice to have an example on the dataset everybody is familiar with showcasing they way we separate model constructor and regularization.

@sherjilozair, you wanted to contribute that?

PR #156, opened 2015-01-26
No more wrappers

### LINKED ISSUE #119: PyLearn2 Markov Chain Test Is Broken
#115 breaks it by changing the way activation classes are serialized.

Perhaps we should just not use class generation? Apart from serialization problems it is quite hard to read for people with limited Python knowledge.

bartvm:
No need to merge this now, and maybe not for a while, but I think that eventually we should get rid of the wrappers. I don't believing we're using or maintaining them anymore (at least I'm not), so they'll probably turn stale and become a liability at one point (e.g. people getting a bad impression because the first thing they try are the non-working wrappers, us spending time fixing Pylearn2's pickling issues, etc.).

Fixes #119 and #65.


rizar:
We can get rid of them as soon as all the examples are ported, and it should probably be done before your talk.

bartvm:
@rizar Let's merge? We've got a bunch of Blocks-only examples now, and these files are just taking up time when I'm cleaning up docstrings, running tests, etc.

rizar:
Okay, go ahead. The only thing we will miss is an example of carrying over the hidden state.

PR #157, opened 2015-01-26
Language streams

rizar:
@bartvm , I am finishing here the work on language streams you started a few days ago.

### DIFF
--- a/blocks/datasets/__init__.py
+++ b/blocks/datasets/__init__.py
@@ -656,13 +656,18 @@ class BatchDataStream(DataStreamWrapper):
         data = [[] for _ in self.sources]
         for i in range(request):
             try:
-                [source_data.append(example) for source_data, example
-                 in zip(data, next(self.child_epoch_iterator))]
+                for source_data, example in zip(
+                        data, next(self.child_epoch_iterator)):
+                    source_data.append(example)
             except StopIteration:
-                if self.strict:
-                    raise
-                else:
+                if self.strict and data[0]:
+                    raise ValueError("Not enough examples to form a batch of"
+                                     " requested size")
+                # If some data has been extracted and `strict` is not set,
+                # we should spit out this data before stopping iteration.
+                if data[0]:
                     break
+                raise
         return tuple(numpy.asarray(source_data) for source_data in data)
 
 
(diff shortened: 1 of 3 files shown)


rizar:
I made your wrappers work, wrote tests, generalized to the case of more than 1-D sequences, wrote a bit more docs and renamed one of them.

rizar:
As this does touch any core components, I merge it and close the original #149 PR.

bartvm:
Okay! Please have a look at my LSH branch; I ran into a bug with these
which I fixed there. This bug might only happen if you use chains of data
streams, so not sure if your tests will catch them (on my phone now so
can't really look).

Basically, you need to make sure that if you catch StopIteration you raise
it again the next time `get_data` gets called. Also, you can't pickle
closed file handles.

rizar:
My tests use chains of data streams and I remember fixing this bug:

```python
except StopIteration:
  if self.strict and data[0]:
  raise ValueError("Not enough examples to form a batch of"
      " requested size")
  # If some data has been extracted and `strict` is not set,
  # we should spit out this data before stopping iteration.
  if data[0]:
     break
  raise
```

bartvm:
I don't think that's enough. What basically happens is that when you hit `break`, the function returns data normally, and the stream's parent never receives a `StopIteration` signal. This means that it will try to call `get_data` again, because it doesn't know the epoch ended, which might result in things like trying to read closed files. See [these lines](https://github.com/bartvm/blocks/blob/lsh/blocks/datasets/text.py#L225-L227) in my branch.

rizar:
My code works well under assumption that `get_data` can be called after `StopIteration` and all such subsequent calls will also raise `StopIteration`. Do we have a dataset that fails to satisfy this condition?

bartvm:
Yes, for `TextFile` a subsequent call will crash because [this line](https://github.com/bartvm/blocks/blob/master/blocks/datasets/text.py#L103) will try to read from a closed file before it gets the chance to raise `StopIteration` again. I'll fix it, because as I mentioned in https://github.com/bartvm/blocks/pull/149#issuecomment-71394864 there is another bug there that stops the data stream from being pickled right now.

bartvm:
Fixed in 4c8f8ab995dda5ed0849bbc42283b03c3a52b630

Issue #151, opened 2015-01-25
Tiny quirks in monitoring


bartvm:
I haven't had the time to look into the details of these, but here are some usability comments and some strange quirks I ran into when using the extensions and monitoring with these settings:

```python
DataStreamMonitoring([real_cost], test_n_gram_stream,
﻿﻿                      prefix='test', every_n_batches=5000)
TrainingDataMonitoring([cost], prefix='train',
﻿﻿                        after_every_batch=True),
Printing(every_n_batches=50),
SerializeMainLoop('lsh.pkl', every_n_batches=5000)])
```

* There is no output while it's calculating the first validation set score, it would be nice if it printed something so that I know things have successfully started.
* There are no headers for the intermediate results, I get this each 50 batches
```
-------------------------------------------------------------------------------
-------------------------------------------------------------------------------
```
* After 5000 batches it says this, even though there are far more batches in a single training iteration.
```
TRAINING HAS BEEN FINISHED:
```
* It should probably also give a message after catching `KeyboardInterrupt`, right now it just seems like it's unresponsive, while it's actually just running `on_interrupt` extensions.

rizar:
Okay, I will take care of this tomorrow. What header would you like to see for intermediate results? I simply put none because that can be any event at all.

bartvm:
Maybe just something like `COMPLETED 5000 BATCHES`? Or maybe no lines at all, or just leave it like this. I just wasn't sure whether it was by design or not.

rizar:
> There is no output while it's calculating the first validation set score

I can only add `logger.info` statements for this information. There's no way to integrate it with the `Printing` extension.

>  Or maybe no lines at all, or just leave it like this
Let's leave it like this until we come up with a significantly better format.

> After 5000 batches it says this, even though there are far more batches in a single training iteration.
How's that?  I can not imagine it happening when reading the code below:
```python
elif which_callback == "after_training":
     print("TRAINING HAS BEEN FINISHED:")
```

> It should probably also give a message after catching  `KeyboardInterrupt`.
Yep.

See #158.

rizar:
Closed by #158 with discussion continued in #154

PR #158, opened 2015-01-26
Minor enhancements of printing and monitoring

rizar:
Addressing #151


bartvm:
Looks good to me! Although at one point we'll have to sort out the logging properly e.g. set the level to `INFO` for the entire library, and make sure that a logger hierarchy exists by always using `getLogger(__name__)`, and add lots of `logger.debug` statements everywhere so that logging really becomes useful for error reports and debugging.

rizar:
By the way, do you know any guideline for writing good log messages? If there's one, it could make it to our developer's guide. It would be nice to have all this printing looking more or less homogeneous.

PR #161, opened 2015-01-27
Scripts

bartvm:
Some scripts to create word/character vocabularies out of text files.

I like the UNIX way of doing these kind of things, so I tried to make the scripts fit in. This means that if you e.g. want to lowercase words before counting, limit your vocabulary to a 1000, but number the words alphabetically, you can now do
```bash
cat *.en.txt | tr [:upper:] [:lower:] | blocks-count-tokens | head -n 1000 | sort -t ' ' -k 2 | blocks-create-vocabulary vocab.pkl
```

### DIFF
--- a/bin/blocks-continue
+++ b/bin/blocks-continue
@@ -1,14 +1,15 @@
 #!/usr/bin/env python
-from argparse import ArgumentParser
+from argparse import ArgumentParser, FileType
 
 from blocks.scripts import continue_training
 
-if __name__ == "__main__":
+if __name__ == '__main__':
     parser = ArgumentParser("Continues your pickled main loop")
     parser.add_argument(
-        "path", help="A path to a file with a pickled main loop")
+        'infile', type=FileType('rb'),
+        help="File with a serialized main loop")
     parser.add_argument(
-        "--rec-limit", type=int, help="The recursion depth limit")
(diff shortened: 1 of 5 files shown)


rizar:
Just repeating your point: are you sure that we want to have such scripts in our library and not redirect the user to an NLP toolbox? Same concerns about support and so on.

bartvm:
I completely agree, but I never really found implementations. On the one hand it's something so trivial that no library seems to implement it, on the other hand it's something so standard that it seems silly to make everyone rewrite a word counting script every time.

I'll have another look to see if I can find a library that does this.

bartvm:
NLTK's `FreqDist` seems to be the answer, but I think people will still have to use the Python interpreter to run it, there's no scripting interface. On second though, I'm quite fine with that, so for now, let's not merge this. I'll have a closer look later to see if `FreqDist` and NLTK can replace these scripts easily enough.

rizar:
OK, let's merge it for now and then we will see.

bartvm:
Haha, I would actually suggest *not* merging it. No need to duplicate NLTK functionality, like you said. If we write documentation on e.g. language modelling examples, we can just refer users to NLTK (or spaCy) in there.

rizar:
Oh, I did not see the "not" word :) OK, I am fine in both cases, let's be truly lean and mean.

Issue #163, opened 2015-01-27
Naming scheme for auxiliary variable should be the same as for usual ones.


rizar:
Currently input and output variables have two names: `x.name` and `x.tag.name`. The first name is composed from three parts as follows: "brick_application_variable-name". This quite helpful for debugging and I suggest that `ApplicationCall.add_auxiliary_variable` should use the same naming scheme (currently it is just https://github.com/bartvm/blocks/blob/master/blocks/graph.py#L268)

rizar:
Closed by #164

Issue #123, opened 2015-01-20
Order of Arguments for IsotropicGaussian


rizar:
That's a really minor thing, but anyway: when I write `IsotropicGaussian(0.01)` I expect it to have zero mean and 0.01 standard deviation, which is currently the other way round. Should I fix that?

bartvm:
I don't have much of a preference, but I guess that it would make sense to stick with e.g. the NumPy and Theano interface, which both take the mean as the first argument (http://docs.scipy.org/doc/numpy/reference/generated/numpy.random.normal.html and http://deeplearning.net/software/theano/library/tensor/raw_random.html#raw_random.RandomStreamsBase)

rizar:
Okay, let's wait until we have users and can make polls :)

rizar:
Well, I stumbled upon this one more time!  I have a good argument why in our case the standard deviation should go first: because `IsotropicGaussian` is an initialization scheme, and in the context of weights initialization sigma is always zero.

bartvm:
I ran into it the other day as well actually, I'm happy to change it!

bartvm:
Fixed in #164.

PR #164, opened 2015-01-27
Various fixes

rizar:
While working on an attention mechanism demo I came across a few bugs and had to do a few impovements. An incomplete lists of changes:

* adds some logging

* fixes bug in gradient norm computation

* changes the naming scheme for auxiliary variables #163 

* changes the order of the `IsotropicGaussian` parameters #123 

* partial enables monitoring of not-scalar expression by using the `TakeLast` aggregation scheme

### DIFF
--- a/blocks/algorithms/__init__.py
+++ b/blocks/algorithms/__init__.py
@@ -1,6 +1,7 @@
 """Training algorithms."""
 from abc import ABCMeta, abstractmethod
 from collections import OrderedDict
+import logging
 
 import theano
 from six import add_metaclass
@@ -8,6 +9,8 @@ from theano import tensor
 
 from blocks.graph import ComputationGraph
 from blocks.utils import named_copy
+
+logger = logging.getLogger(__name__)
 
 
 @add_metaclass(ABCMeta)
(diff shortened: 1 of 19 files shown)


blocks/algorithms/__init__.py
```diff
@@ -164,17 +167,20 @@ class GradientDescent(DifferentiableCostMinimizer):
     """
     def __init__(self, step_rule=None, gradients=None, **kwargs):
         super(GradientDescent, self).__init__(**kwargs)
-        self.gradients = (
-            gradients if gradients
-            else dict(
-                zip(self.params, tensor.grad(self.cost, self.params))))
+        self.gradients = gradients
+        if not self.gradients:
+            logger.info("Taking the cost gradient")
```
bartvm: I would make this `logger.debug` and not `info`, it's pretty detailed information.

rizar: Actually taking the gradient can take more then half-a-minute when you are working with scan. So you want this logging to be always on, and you do no want to see those really detailed messages you get at the debug level.

bartvm: Fair point

bartvm:
LGTM

PR #168, opened 2015-01-28
WIP: Live plotting

bartvm:
I very much like the idea of live plotting, as recently [implemented in Pylearn2](https://github.com/lisa-lab/pylearn2/blob/master/pylearn2/train_extensions/live_monitoring.py) by Dustin. However, I don't want to write 400 lines of code and use a messaging framework to do it...

So I tried some other solutions. Yesterday I set up InfluxDB and Grafana, which turned out to be not that straightforward, complete overkill, and actually not very good to work with (because it assumes time stamped data). I looked into Bokeh this morning (it's a library I've been following for a while) which seems to work quite nicely.

As this PR stands, the user would have to type `bokeh-server` to start the plotting server, and can then start training with `Plot('Experiment title')` as an extension. You can then automatically watch live updating plots of all your monitors!

screen shot 2015-01-28 at 9 15 54 am

It works quite nicely as a proof of concept, but it's obviously a bit rough around the edges and hence WIP:

- [x] All lines have the same color right now
- [x] The user needs to be able to control which monitors get plotted
- [x] The user should be able to decide which plots go in the same figure, and which go in separate figures
- [ ] The x-axis should start at 0
- [x] Currently it's quite slow; it takes a second or two for the plot to update each time, which isn't a problem for longer epochs, but if your epochs are less than a second long (like in the MNIST example) it kind of defeats the purpose
- [x] Start Bokeh server automagically if not started already

### DIFF
--- a/blocks/extensions/__init__.py
+++ b/blocks/extensions/__init__.py
@@ -1,11 +1,6 @@
 from __future__ import print_function
 
 from abc import ABCMeta, abstractmethod
-try:
-    from bokeh.plotting import figure, output_server, show, cursession, push
-    bokeh_available = True
-except ImportError:
-    bokeh_available = False
 
 from six import add_metaclass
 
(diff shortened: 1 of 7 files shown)


rizar:
Wow, that's definitely very cool!

blocks/extensions/__init__.py
```diff
@@ -300,3 +305,40 @@ def do(self, which_callback, *args):
                 log.status.iterations_done))
             self._print_attributes(log.current_row)
         print()
+
+
+class Plot(SimpleExtension):
```
rizar: I think this extension deserves a separate file.

bartvm:
Lines now have different colors, and the user can choose which channels to plot in which figures. I've also made a setting to start the plotting server automatically. There's a few user-interface choices to make though: I think it might be a good thing for the extension to crash if the user doesn't have a running server and doesn't pass `start_server` explicitly. Mostly because I don't want either of these options:

* User uses plotting extension, doesn't read the documentation and plotting server starts
* The user stops training, and immediately loses all his plots

* User uses plotting extension, doesn't read the documentation and plotting server starts
* The user stops training, but the plotting server keeps running
* The user starts/resumes another model using the plotting extension, but all crash immediately because the try to start a server on the same port the old one is running on

* User uses plotting extension, doesn't read the documentation and plotting server starts
* The user stops training, but the plotting server keeps running
* The user stops the plotting server manually, not realizing that this wipes out all of his/her plots

bartvm:
I'll leave the axes minutiae for a later ticket. I think this PR is pretty much done. I'll make a test for it, but beyond checking whether things won't crash, it won't do much.

examples/mnist.py
```diff
@@ -63,6 +64,12 @@ def main(save_to, num_epochs):
                         prefix="train",
                         after_every_epoch=True),
                     SerializeMainLoop(save_to),
+                    Plot(
+                        'MNIST example',
+                        monitors=[
```
rizar: Maybe channels is a better word here?

bartvm: Done

bartvm:
Tests are passing so I'm merging; will try it out in the lab when I have a chance. I don't think anything else is needed to be able to just go to e.g. `http://eos17:5006` to keep track of your training now :)

rizar:
Btw, how robust is it with respect to training interruptions and resumptions?

bartvm:
It's so-so. I'm not sure what the best behaviour would be. Right now the only way it can go wrong is that *if* the user starts the server using Blocks, and then resumes training later on (or starts another model with the same setting) then it will give an error, because it will be trying to start a second server. That's why I didn't make this default behaviour and added a lengthy warning about it (see [here](https://github.com/bartvm/blocks/blob/master/blocks/extensions/plot.py#L43-L52)).

PR #169, opened 2015-01-28
Timing

rizar:
An extension that keeps track of time. It would be better to implement that with two extensions, but let's start with something easy. On the bright side, it should be pretty robust to restarts.


rizar:
OK, I will just merge it: it is feature full, tested and documented.

@bartvm , please tell me if you have suggestions.

PR #172, opened 2015-01-28
WIP: More Robust Serialization

rizar:
While it is very nice that simple pickling of the main loop object works in Blocks, issues with pickling and unpickling certain Theano objects AFAIK are still there. This pull request proposes serialization a-la Groundhog, which saves only certain components of the main loop. It can be used as an alternative in situations when the pickle-based solution does not work for technical reasons. I also plan to add an extension that automates using this serialization. 

Also:

* I add a new light-weight example `examples/sqrt.py` which is more suitable for using in tests.

* `silence_outputs` wrapper to hide verbose output of the `Printing` extension

Todo:

- [x] write an extension


bartvm:
The problems being that you still need a GPU to unpickle a GPU-trained model, or are you referring to anything else?

rizar:
This is the most important problem. But another thing is that unpickling my main loop with attention mechanism and stuff takes 5 minutes at my laptop (I guess some recompilation takes place). This is unacceptable. In addition for some reason it works only for Python 2, I should make a bug report from this.

I anticipate that will keep having issues with the pickled-based method for years, because we need complete support from Theano and every other library we use for that. Meanwhile I want Blocks to be ready for machine translation, speech recognition and other large-scale projects right now.

rizar:
Yet another example of a Pickle-caused trouble: I can not restart training because my training script had this:

```python
lambda log: math.isnan(log.current_row.total_gradient_norm)),
```

which quite naturally results into `NameError: global name 'math' is not defined` when running `blocks-continue`

bartvm:
Why is this unacceptable? 5 minutes is negligible if it allows you to pick up right were you left off after a server failure without having to re-initialize your momentum, adadelta parameters, etc. In general, if speed is a big issue, I would suggest a different approach: Pickle some parts of the main loop separately. That way, if you just want to view the log, you don't need to wait a few minutes for the model to load, and you don't need a GPU either.

However, I agree that in general it's a very tough one and that the problems are non-negligible, but I hate the idea of supporting two completely separate serialization methods, because making one work nicely is more than difficult enough. So I would argue that we should move to a new method entirely eventually, if that's possible without losing too much functionality (but is it?)

If we were to try that, I would propose that we should try to use `dill` wherever we can i.e. for all non-Theano stuff. It's simply the easiest way of making sure that things (dataset iterators, the log) really will be exactly the way they were on saving.

I'n assuming that saving the Theano parameters separately means that we will need to have a minimal "model" concept that recreates the Theano graph on loading? Can we limit this model to a single function which takes Theano variables in, and spits Theano variables out? I don't like the idea of bricks becoming a necessity for the main loop again, so I would like it if we could pass `mlp.apply` instead of `mlp`, so that I can theoretically also just pass `lambda x: (x ** 2).mean()` if I want.

But, for now my main worries are: What about extensions and algorithms that create Theano variables? For example, I am about to run some experiments with a second-order method which will keep track of negative curvature directions, and resuming my training will be quite important (because they're very long jobs). However, I don't feel like having to implement an entire custom serialization scheme to make sure that my algorithm plays nice with your serialization scheme. I also don't like the idea of there being a list of extensions which are and aren't suitable for a particular way of serialization.

I have to say that I'm very much on the fence about this issue. I understand that the problems are real, but maybe the limitations there are right now can be dealt with in a less radical way.

bartvm:
How about this as an alternative solution:

* We stick with the current serialization method *in principle*, because ensuring reproducibility otherwise is just a nightmare when you start looking at extensions and training algorithms that have running averages and such all over the place
* To make sure that users don't need to wait for 5 minutes if they just want to look at the log, we can look into splitting the pickling up in pieces, so we pickle the log, the algorithm, extensions, model and data stream all separately. (I'm not sure whether/how this works, because the algorithm and extensions could both have links to the full Theano graph, so maybe we should just pickle the log and data stream separately.)
* We also dump the parameters to `.npz` files and allow for a way of reloading them, but make *no attempt whatsoever* to ensure training reproducibility. This is basically a fail-safe: If the pickle files break or whatever, you always have the most important thing, the parameters, from which you can continue training. The log was also saved separately, so worst case scenario you can always start your data stream kind of where you left it.

bartvm:
Oh, and as far as lambda functions go, I was just talking to DWF about them: We should just not use them. Not just because they're hard to pickle, but because they make debugging a library a nightmare in general (your traceback and debugger get really cryptic).

rizar:
@bartvm, the alternative solution you propose above is very similar to the code from this pull request with two differences:

* I think that we should not make piece-wise saving the default behaviour. This is convenient to have a file that just has all you need to continue training. I would like to have a separate extension that save pieces separately, which one can use when he is not satisfied with unpickling speed or has to continue training at a different architecture, with different version of libraries or just feels scary. 

* I care not only about dumping model parameter in `.npz`, but also I want to load those automatically. That can be done "before_training" by an extension.

Altogether this gives the user Grounhog-ish experience. It is impossible to guarantee complete reproducibility without pickling: I agree with that. However, I planned this poor man's saving and loading since long, and you can see that the main loop and the extensions I wrote are mostly stateless: the store everything in the `log.status`. Of course aggregation buffers of `TrainingDataMonitoring` will be lost, but one can live with that. The biggest shortcoming is that there is currently no way to save the algorithm's state without pickling. While it is quite bad, we are at least as good as Groundhog, because there it was not supported as well.

Just a motivational side-note: I do not try to argue with the pickling's role of the primary serialization method in Blocks. But we have to buy Grounhog users, and for this we have to support CPU/GPU stuff right now, out of the box.

rizar:
And regarding lambdas: as for the library's internals I don't know, but forbidding the users to use them: that's definitely too much. I am a huge lambda aficionado and I do want my scripts to be readable. 

It boils down to the following statement: to have `blocks-continue` working one just has to put all the logic in a file other then the executable scripts, keeping the size of the latter as small as possible, ideally having only stupid command line parsing there. Then lambda are not an issue. But if I made such a mistake, it is nice to be able to break the pickle into pieces and feed this pieces with to the new training process: that's what this PR is for.

bartvm:
Okay, then we're on the same page! I also think reloading should be possible and automatic. I wouldn't worry about pickling the state of extensions then. Just to summarize to make sure I understand:

* Default method of serialization is pickling everything
* As an alternative and/or backup, you can also save the parameters to a `.npz` file, and pickle the log and data stream separately. This (a) can be used to make sure that if anything goes wrong with pickling, you are unlikely to lose your parameters and (b) is a faster and more cross-platform alternative to the first method in cases where you don't care much about things being reproducible and/or extensions keeping their state.

Pickling things separately might still be worth a look on the long term. If you want to make it transparent, you could basically pickle 3 separate files (log, data stream, everything else) and then tar them so that you still have a single file. I just like the idea of a user not needing to wait for 5 minutes and have a GPU to plot the log. Of course, we could make this optional.

I wouldn't dare to take anonymous functions away from you! Just suggesting that we keep them to a minimal in the library itself. I think they're great as throw-away functions for things like `sorted`, but let's try not to pass them around, just to make debugging a bit easier.

bartvm:
Sorry for the Travis error. Ran the tests locally, thinking it tried the install arguments, but obviously it didn't so missed the strange `yaml`/`pyyaml` duality.

rizar:
OK, now that it is functional we should think about terminology. I would probably call the folder with three files a dump, and replaced the word state with the word dump everywhere in code and comments. That would kind of suggest that loading from dump might not return you to exactly the same state where you were.

And I thought a bit more about pickling: in fact it is very very binding. Any single change in API of any single library used would make unpickling of the whole thing impossible. This is at the same time good and bad. It is good because it is reliable, if you managed to unpickle the main loop, you might really hope that you are continuing the experiment the right way. It is bad because one often wants to continue an experiment started in an environment A in another environment B, even though it is risky.

Summing up, we do need two serialization methods, one as restrictive possible and another as loose as possible. I am fine with prioritizing the restrictive one, but I guess we will have to develop the loose one as well. So I think we should add a developer guideline, that extensions, if possible, should be stateless, in order for the loose method to work in as many cases as possible. 

I like your idea with tar, by the way.

bartvm:
Yes, I'm very aware of that limitation (of not being able to unpickle if your library changes), because it's something that Pylearn2 tries to fix by endless backward compatibility hacks in `__setstate__`.

In practice, I think it's a limitation I am happy to live with. I think I mentioned this somewhere else before, but my long term idea is this: Use something along the lines of [this Pylearn2 metaclass](https://github.com/lisa-lab/pylearn2/blob/master/pylearn2/utils/track_version.py) which automatically tracks the Blocks, Theano, NumPy, etc. versions, if possible down to the git commit. When unpickling it should warn if any of the versions are different, even if the rest of the unpickling fails. Something along the lines of:

```
WARNING: This training loop was serialized using a different version of Blocks (Blocks commit 2bd32f, this is 93fa32). Correct resumption of training is not guaranteed. If any problems occur, please see the documentation on how to (temporarily) revert to an old version of Blocks.

# Pickling crashes here
```

And it would give a similar warning for Theano and/or NumPy version mismatches. Theano and Blocks are the tricky ones, because people often don't use official releases and install straight from GitHub, but all the others are easy enough.

I guess that you're right though, and that this is another reason for supporting two (complementary) ways of serializing: One is for short-term training resumption on an identical setup. This is great in cases where you are running on a cluster with time limits or want to resume training after a network/disk error. The method of dumping raw parameters is better for long-term model saving and switching between systems.

I like the dump vs. serialization terminology.

rizar:
Okay, that seems to be. @bartvm, would you look through it and say if I can merge it without creating too much confusion and chaos?

bartvm:
Yeah, in class now, but let me have a look this afternoon!

blocks/dump.py
```diff
@@ -0,0 +1,220 @@
...
+    Returns
+    -------
+    A dictionary of (parameter name, numpy array) pairs.
+
+    """
+    if isinstance(bricks, Brick):
+        bricks = Selector([bricks])
+    if not isinstance(bricks, Selector):
+        raise ValueError
+    return OrderedDict([(name, variable.get_value(borrow=True))
```
bartvm: Could we switch this (eventually, not necessarily now) to the `VariableFilter(roles=[PARAMETER])` approach. That would make this kind of serialization also work for models which aren't necessarily entirely created using bricks.

I've done this a few times in my experiments already e.g. I created an MLP with bricks, then added a strange extra thing using Theano, and then fed the new cost to the training loop. If we use `VariableFilter(roles=[PARAMETER])` I could just do `add_role(crazy_thingy, PARAMETER)` and still serialize happily.

rizar: `ComputationGraph` has to go into scan for that. And also now we get the names of the parameters from their location in the brick hierarchy, how would you get names in your proposal? Note, `param.name` attribute are very often the same for different variables.

bartvm: I guess it would become tricky to find the parameters upon loading though. Maybe that can be fixed eventually by changing our naming scheme for variables to something that is unique (which wouldn't be a bad idea anyway) e.g. the name would be `/mlp/linear_1.apply[W]` instead of the current `linear_1_apply_W`.

blocks/dump.py
```diff
@@ -0,0 +1,220 @@
...
+
+    Notes
+    -----
+    Requires the model to be a Brick or a list of Bricks.
+
+    """
+    def __init__(self, folder):
+        self.folder = folder
+
+    def path_to_parameters(self):
```
bartvm: These 3 methods would probably be good candidates for properties.

blocks/extensions/saveload.py
```diff
@@ -56,6 +45,61 @@ def __init__(self, path, **kwargs):
...
-        with open(self.path, "wb") as destination:
-            dill.dump(self.main_loop, destination, fmode=dill.CONTENTS_FMODE)
-        self.log_saving_done(self.path)
+        try:
+            self.main_loop.log.current_row[SAVING_DONE_TO] = self.path
+            with open(self.path, "wb") as destination:
+                dill.dump(self.main_loop, destination,
+                          fmode=dill.CONTENTS_FMODE)
+        except:
+            self.main_loop.log.current_row[SAVING_DONE_TO] = None
```
bartvm: Will `None` represent failure? Maybe we should document that somewhere.

blocks/extensions/saveload.py
```diff
@@ -56,6 +45,61 @@ def __init__(self, path, **kwargs):
...
+        if not os.path.exists(self.manager.folder):
+            logger.info("No dump found")
+            return
+        logger.info("Loading the state from {} into the main loop"
+                    .format(self.manager.folder))
+        try:
+            self.manager.load_to(self.main_loop)
+            self.main_loop.log.current_row[LOADED_FROM] = self.manager.folder
+        except:
+            logger.error("Failed to load the state:\n{}"
```
bartvm: Better to use `reraise_as("Failed to load state")` from `blocks.utils` here, that way you can still use PDB to do debugging.

blocks/extensions/saveload.py
```diff
@@ -1,30 +1,19 @@
...
-from blocks.extensions import SimpleExtension
+from blocks.extensions import SimpleExtension, TrainingExtension
+from blocks.dump import MainLoopDumpManager
 
+logger = logging.getLogger(__name__)
 
-class SaveLoadBase(SimpleExtension):
-    """The base class for save-load extensions.
+LOADED_FROM = "loaded_from"
+SAVING_DONE_TO = "saving_done_to"
```
bartvm: `SAVED_TO`?

blocks/dump.py
rizar: I think that same names are not a bug but a feature! It would so annoying to try and make them all unique (and I am gonna have 50 parameters in my models). Thanks to the brick hierarchy we do not have to do that.

bartvm: *That was a cross-post

Why would it need to go into scan? Parameters themselves won't be hidden inside the scan apply node or anything.
```python
x = theano.shared(np.random.rand(10, 10))
y = theano.shared(np.random.rand(10, 10))
z, _ = theano.scan(fn=lambda row: row + y, sequences=x)
ComputationGraph(z).shared_variables
```
```
[<TensorType(float64, matrix)>, <TensorType(float64, matrix)>]
```
I'm happy to merge the current version if you just want to get to using it, but on the long term I think that we should remove the requirement for bricks. The only thing we need for this to work is that each parameter needs some sort of unique identifier. If we can make that unique identifier a simple string or tag, than I would much prefer that over imposing an entire bricks hierarchy.

bartvm: I keep taking too long to actually press "Comment", because we keep cross-posting. Agreed that we don't necessarily need to change all the names, but we can just add a tag to each variable `var.tag.id` which must be set to a unique string. I would like avoiding the reintroduction of the dependency on bricks into the main loop for an essential extension like this one.

How exactly this ID would be set, I am not sure. The naive way would require each brick to be aware of its parents, which it currently isn't (and it doesn't even need to be unique), but there has to be a easy solution...

rizar: Regarding scan: damn, I did not know that! Indeed, the shared variables are just inputs to Scan apply nodes.

As for ids, an additional requirement for them is that they should be human readable, like paths in the hierarchy are. In fact the paths are the best ids, just that I do not see a simple way to reconstruct them.

bartvm: One way of doing it would be by a similar data structure as I use for the parameters. Basically, this would give brick a `parents` attribute (because you can have multiple) that is kept entirely in sync with the `children` attributes (so when you modify the children of a brick, they automatically get parents added/removed). That would then allow us to reconstruct the path from any brick.

However, this path is not unique as long as we allow one brick to be the child of multiple parents. How is this currently handled? I guess that you can just ignore it; you will simply save the identical parameters twice, and then reload them twice. But if want to create a unique ID, you need to make a decision as to what how you deal with multiple parents (e.g. choose the first parent, use both).

rizar: We can define the left-most path as the one of the paths from the root to the leave that always chooses the child with the least possible index. This is what actually happens now.

After sleeping with this idea I agree that in general it is not nice that the dumping mechanism is tied to bricks. But I would not be looking for an automatic solution, because it would be very implicit and hard to understand. Instead, we can require from the user to rename the parameters in the graph himself if he wants to use dumping. It would look like that:

```
cost = CategoricalCrossEntropy(y, mlp.apply(x))
cost = rename_parameters(cost, mlp)
...
main_loop = MainLoop(..., extensions=[..., Dump(cost)])
```

That would make the extension however only usable in the case of a differentiable cost. So I think the time has come to give the `model` parameter of the main loop some semantics: let it encapsulate the logic of retrieving parameters from whatever we try to train. 

But I would like to merge this one first, to fix the fact that we are not worse then Groundhog.

PR #175, opened 2015-01-30
WIP: Reintroduce parameter collection

bartvm:
- [x] Write tests
- [ ] Move graph modifications like `collect_parameters` and `apply_noise` to a separate file?

### DIFF
--- a/blocks/graph.py
+++ b/blocks/graph.py
@@ -496,6 +496,8 @@ def collect_parameters(computation_graph, params):
         param_shapes.append(param_values[-1].shape)
 
     new_params = shared_floatx_zeros(sum(param_sizes))
+    new_params.set_value(numpy.concatenate([value.flatten()
+                                            for value in param_values]))
     new_params.name = 'collected_params'
 
     replacements = {}
--- a/blocks/graph.py
+++ b/blocks/graph.py
@@ -13,7 +13,8 @@ from theano.scan_module.scan_op import Scan
 from toolz import unique
 
 from blocks.config import config
-from blocks.roles import add_role, has_roles, AUXILIARY, PARAMETER, DROPOUT
+from blocks.roles import (add_role, has_roles, AUXILIARY, PARAMETER, DROPOUT,
+                          COLLECTED)
 from blocks.utils import (is_graph_input, is_shared_variable, dict_union,
                           shared_floatx_zeros, shared_like)
 
@@ -499,6 +500,8 @@ def collect_parameters(computation_graph, params):
     new_params.set_value(numpy.concatenate([value.flatten()
                                             for value in param_values]))
     new_params.name = 'collected_params'
+    add_role(new_params, COLLECTED)
+    new_params.replacement_of = params
 
     replacements = {}
     for param, shape, i, j in zip(params, param_shapes,
--- a/blocks/graph.py
+++ b/blocks/graph.py
@@ -466,11 +466,12 @@ def collect_parameters(computation_graph, params):
 
     Examples
     --------
-    >>> from blocks.bricks import MLP, Sigmoid
+    >>> from blocks.bricks import MLP, Logistic
     >>> from blocks.bricks.cost import SquaredError
     >>> from theano import tensor
     >>> x = tensor.matrix()
-    >>> mlp = MLP(activations=[Sigmoid(), Sigmoid()], dims=[784, 100, 784])
+    >>> mlp = MLP(activations=[Logistic(), Logistic()],
+    ...           dims=[784, 100, 784])
     >>> cost = SquaredError().apply(x, mlp.apply(x))
     >>> cg = ComputationGraph(cost)
     >>> new_cg = collect_parameters(cg, cg.shared_variables)
--- a/blocks/graph.py
+++ b/blocks/graph.py
@@ -14,7 +14,7 @@ from toolz import unique
 
 from blocks.config import config
 from blocks.roles import (add_role, has_roles, AUXILIARY, PARAMETER, DROPOUT,
-                          COLLECTED)
+                          COLLECTED, COLLECTOR)
 from blocks.utils import (is_graph_input, is_shared_variable, dict_union,
                           shared_floatx_zeros, shared_like)
 
@@ -475,18 +475,20 @@ def collect_parameters(computation_graph, params):
     >>> cg = ComputationGraph(cost)
     >>> new_cg = collect_parameters(cg, cg.shared_variables)
 
-    The new graph only has a single shared variable.
+    The new graph only has a single shared variable. This variable receives
+    the :const:`COLLECTOR` role.
 
     >>> new_cg.shared_variables
     [collected_params]
 
     The bricks' variables have been replaced with reshaped segments of this
-    single shared variable.
+    single shared variable. These replacements are given the
+    :const:`.COLLECTED` role.
 
     >>> from blocks.filter import VariableFilter
     >>> from blocks.roles import PARAMETER
-    >>> var_filter = VariableFilter(roles=[PARAMETER])
-    >>> var_filter(new_cg.variables) # doctest: +SKIP
+    >>> var_filter = VariableFilter(roles=[COLLECTED])
+    >>> var_filter(new_cg.variables)  # doctest: +SKIP
     [Reshape{1}.0, Reshape{1}.0, Reshape{2}.0, Reshape{2}.0]
 
     """
@@ -500,15 +502,15 @@ def collect_parameters(computation_graph, params):
     new_params.set_value(numpy.concatenate([value.flatten()
                                             for value in param_values]))
     new_params.name = 'collected_params'
-    add_role(new_params, COLLECTED)
-    new_params.replacement_of = params
+    add_role(new_params, COLLECTOR)
 
     replacements = {}
     for param, shape, i, j in zip(params, param_shapes,
                                   numpy.cumsum([0] + param_sizes[:-1]),
                                   numpy.cumsum(param_sizes)):
         new_param = new_params[i:j].reshape(shape)
-        new_param.tag = param.tag
+        new_param.replacement_of = param
+        add_role(new_param, COLLECTED)
         replacements[param] = new_param
     return computation_graph.replace(replacements)
 
--- a/blocks/graph.py
+++ b/blocks/graph.py
@@ -3,8 +3,9 @@ import logging
 from collections import OrderedDict
 from itertools import chain
 
-from picklable_itertools.extras import equizip
+import numpy
 import theano
+from picklable_itertools.extras import equizip
 from theano import Variable
 from theano.gof import graph
 from theano.sandbox.rng_mrg import MRG_RandomStreams
@@ -14,7 +15,7 @@ from toolz import unique
 from blocks.config import config
 from blocks.roles import add_role, has_roles, AUXILIARY, PARAMETER, DROPOUT
 from blocks.utils import (is_graph_input, is_shared_variable, dict_union,
-                          shared_like)
+                          shared_floatx_zeros, shared_like)
 
 logger = logging.getLogger(__name__)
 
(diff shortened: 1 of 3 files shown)


bartvm:
@rizar Let me know if you have any issues with merging this

blocks/graph.py
```diff
@@ -319,3 +320,84 @@ def apply_noise(graph, variables, level, seed=None):
...
+    param_values, param_sizes, param_shapes = [], [], []
+    for param in params:
+        param_values.append(param.get_value(borrow=True))
+        param_sizes.append(param_values[-1].size)
+        param_shapes.append(param_values[-1].shape)
+
+    new_params = shared_floatx_zeros(sum(param_sizes))
+    new_params.set_value(numpy.concatenate([value.flatten()
+                                            for value in param_values]))
+    new_params.name = 'collected_params'
```
rizar: I guess we are missing a `PARAMETER` role for the new parameter.

bartvm: Right, that's where it gets tricky, and it really all depends on the semantics we want to use. Instead I give the `PARAMETER` roles to the reshaped slices of this shared variable (see line 401). I don't want to do both, because otherwise filtering by `PARAMETER` would give you each parameter twice.

I think the current behavior probably makes most sense, because you can still filter by `WEIGHTS` and `BIASES` and retrieve the correct tensor variables, even if they aren't shared anymore. If you want to e.g. noise the weight matrices afterwards, you can still use `apply_noise(new_cg, VariableFilter(role=[WEIGHTS])(new_cg.variables))`.

This does mean that you can't rely on a `PARAMETER` variable being shared, but we already said that this was the case (for my hypothetical case at the time where the weight matrix is the transpose of another).

rizar: Okay, but I am just scared that without adding this role the automatic parameter extraction will not work, because so far we planned that the shared _and_ equipped with the `PARAMETER` role variables will be extracted by the algorithm and trained. So nothing will be fetched from the graph built by your function.

bartvm: In that case maybe it would be better if the algorithm extracts all shared variables, but warns if it finds some that don't have the `PARAMETER` role? So it could say something like:
```
WARNING: Training algorithm found shared variable(s) without PARAMETER role: collected_parameters.
Are you sure you want to train this? If not, pass the parameters to algorithm manually.
```
With all the amazing graph filtering we have going on, it should be easy enough for a user to pick out which parameters he does or doesn't want to train manually if he needs to.

rizar: In case of dropout and noise addition we have shared variables of random streams and we have to rule them out. This is a casual case and I would not like to have warnings in it.

bartvm: Yeah, but you can filter out random streams anyway, you're not going to train those.

rizar: But the case of an MLP with dropout on MNIST to be handled automatically, and the plan was to select only those shared variables marked by a `PARAMETER` role. The computation graphs produced by your routine break this strategy. I am not sure I understand what's your plan instead.

bartvm: Just select all the non-random shared variables. This works in 99% of the cases, including on usual MLPs with dropout. However, if a shared variable shows up without a `PARAMETER` role, the algorithm will warn you, and you have to decide whether to intervene and exclude those shared variables or not.

I can't really think of many cases where you would have shared variables as part of the computation graph that you do not want to train. If that does happen, all you need to do is heed the warning and use `params=[var for var in cg.shared_variables if var not in [excluded_var, another_excluded_var]]`.

If ever the situation arises that this becomes an actual issue, you could always solve it by adding a `NON_PARAMETER` role to the shared variables that should be ignored by the algorithm.

rizar: > Just select all the non-random shared variables.

Do you know a reliable way to do that? I thought we want to stay away from technical details of Theano as much as possible.

> I can't really think of many cases where you would have shared variables as part of the computation graph that you do > not want to train

That happens quite often when people want to freeze weights. But I think the way to support it is to assign zero learning rates. Apart from this example and the variables related to random generation, I do not know other examples.

bartvm: If you want to freeze variables, you could use the same trick of just passing `params=[var for var in cg.shared_variables if var not in frozen_vars]`, but either way is fine.

Selecting the random shared variables is trickier than I thought. For the normal shared random streams it is easy, because you can just do `isinstance(var, RandomStateSharedVariable)`, but I just realized that `MRG_RandomStreams` shared variables don't inherit from this for some reason, and are just usual `TensorSharedVariable` objects...

They are created [here](https://github.com/Theano/Theano/blob/master/theano/sandbox/rng_mrg.py#L1207-L1219), so it depends on how much you feel these are "technical details", but we could make a two-line PR that sets `node_rstate.tag.is_mrg_stream = True`. It seems like something that should be possible in Theano anyway, identifying which shared variables are RNG streams.

rizar: Okay, for identifying which shared variables are there to model randomness we can probably ask for support from Theano developers. But I have just remembered that the learning rate will very soon become a shared variable as well. One can mark it with a role `NOT_PARAMETER`, but I feel bad about it.

And I fact I should say I do not like the idea that a non-shared variable can be a parameter. This conflicts with the standard semantic of the word: parameter is something external to the formula, something independent. We can introduce another role `PARAMETER_SUBSTITUTE` for the cases when a parameter is replaced by a function of other parameters. Or we can use a combination of two roles `PARAMETER` and `SUBSTITUTE` for these cases. 

The one-liner `apply_noise(new_cg, VariableFilter(role=[WEIGHTS])(new_cg.variables))` becomes by the way extremely dangerous when you work with an autoencoder, because noise will be added twice, both to the encoder weights and to the decoder weights obtained by transposition of the former. And to be honest, it is not going to be easy to add different noise to these two matrices...

bartvm: Keep in mind that the learning rate is only part of the computation graph for the *updates*, not of the one for the cost, which is what the algorithm would be dealing with, so I don't think that will be a problem.

I agree with your argument though, but it's more complicated than that, because the alternative can be very annoying and/or dangerous as well: With an autoencoder I will only retrieve one weight matrix, which might be the behavior that I expect. However, if I added noise to weight matrices in an MLP, I will retrieve the matrices without noise, and will be left without a way of retrieving the ones with noise. That is problematic. Moreover, if I used the `collect_parameters` function, then my MLP will no longer have *any* weight matrices or biases. That is very unintuitive as well. Something as simple as getting the norm of the weight matrices is now impossible, because I have no way of retrieving them from the graph.

The advantage of separating the "shared variables" from the `PARAMETER` role, instead of making the latter a subset of the other, is that it gives us more granularity (per definition). Your autoencoder example can then be solved by taking "shared variable AND `PARAMETER`", getting the noised weight matrices of an MLP is done by doing "`PARAMETER`", while the unnoised ones can be retrieved by just taking the "shared variables" filtered by their owning bricks. Same thing for collected variables: Getting the weight matrices is done with `WEIGHTS`, finding the collected ones is done by getting the shared variables again (there will be only one).

Although more flexible, I agree that the semantics are awkward, and I am not a fan of `SUBSTITUTE` either. Ideally we want a distinction between *trainable* parameters i.e. the actual Theano shared variables with respect to which we will take the gradient, and the *semantic/bricks* parameters, which will always be the weight matrix that a linear brick uses to multiply, whether it be shared, the slice of another variable, with or without noise, etc. There is a linguistic confusion here that we can do little about, but maybe we could name the former `TRAINABLE` and the latter `PARAMETER`. In most cases, a variable (e.g. a weight matrix) will be both a `PARAMETER `and will be `TRAINABLE`, but e.g. after adding noise the original variable will only be `TRAINABLE` while the noised version will be the brick's `PARAMETER`/`WEIGHTS`.

rizar: > Keep in mind that the learning rate is only part of the computation graph for the updates, not of the one for the cost, which > is what the algorithm would be dealing with, so I don't think that will be a problem.

Good point, but the shared variable that carries over the state of my language model to the next batch is a part of the cost graph and should not be trained. That being said, I do not think that choosing all shared variables and filtering those that do not belong to random streams is a good default way of choosing what to train.

Answers to the rest of your post is coming, I just wanted to say this quick point first.

bartvm: Right, but my suggestion was to give a warning in that case, it would say something like: `Shared variable (transfer_state) found without PARAMETER role. Are you sure you want to train this? ...` Not ideal, but better than nothing. Looking forward to the rest of your reply! :p

rizar:
> However, if I added noise to weight matrices in an MLP, I will retrieve the matrices without noise, and will 
> be left without a way of retrieving the ones with noise

Whatever we decide here on roles, I think it would be nice to return the new variables in addition to returning the CG when doing things like `add_noise` and `collect_parameters`. 

> Your autoencoder example can then be solved by taking "shared variable AND PARAMETER

I am not sure. The thing is that the variable `W` and `W.T` are consequent in the graph, and I am afraid `theano.clone` can not handle this case:

```python
x = tensor.matrix('x')
y = x + 1
z = y + 1
debugprint(z)
print
debugprint(theano.clone([z], {x: x + 7}, {y: y + 7}))
```
Run it and you get:
```
Elemwise{add,no_inplace} [@A] ''   
 |Elemwise{add,no_inplace} [@B] ''   
 | |x [@C]
 | |DimShuffle{x,x} [@D] ''   
 |   |TensorConstant{1} [@E]
 |DimShuffle{x,x} [@F] ''   
   |TensorConstant{1} [@E]

# There should be two times 7 in the debugprint below
Elemwise{add,no_inplace} [@A] ''   
 |Elemwise{add,no_inplace} [@B] ''   
 | |Elemwise{add,no_inplace} [@C] ''   
 | | |x [@D]
 | | |DimShuffle{x,x} [@E] ''   
 | |   |TensorConstant{7} [@F]
 | |DimShuffle{x,x} [@G] ''   
 |   |TensorConstant{1} [@H]
 |DimShuffle{x,x} [@I] ''   
   |TensorConstant{1} [@H]
```

> getting the noised weight matrices of an MLP is done by doing "PARAMETER", while the unnoised ones > can be retrieved by just taking the "shared variables" filtered by their owning bricks

Do not understand that. Do you suggest to withdraw the `PARAMETER` role from shared variables after noise addition? If you do, how would they be extracted for training? As the example with the carried over hidden state shows, selecting _all_ shared variables for training is a dangerous option.

The proposal you make in the last paragraph seems reasonable to me. As far as I understood, every shared variable created in a brick and assigned to the parameter array's cell receives at this point two roles: `TRAINABLE` and `PARAMETER`. A non-shared variable under the same circumstances receives only the `PARAMETER` role. When a `PARAMETER` variable is replaced in a graph transformation, the replacement takes its `PARAMETER` role over, but the `TRAINABLE` role stays with the replaced variable. This seems to work.

But when you think about it from a more general perspective, it becomes quite confusing: some roles are "inheritable" and some are not. For instance what should we do with the `INPUT` and `OUTPUT` roles? What if we introduce a new role `REPLACEMENT` for the variables that were added to the CG by transformations like `apply_noise`? Such a role, I believe, should not be taken away by new replacements because the variable keeps being a replacement.

To be honest, thinking about variables losing their roles makes me dizzy, and it would so nice to forbid that,

bartvm:
I agree that it would be nice to forbid it and that it's not as simple as I would like, but so far it seems like the best solution to me. Mostly for myself, here a summary of all the reasonable options as I understand them:

**(1) Introduce ``TRAINABLE``**
```python
>>> VariableFilter(roles=[PARAMETER])(cg.variables) 
[W]
>>> new_cg = add_noise(cg, W)
>>> VariableFilter(roles=[PARAMETER])(new_cg.variables) 
[noised_W]
>>> VariableFilter(roles=[TRAINABLE])(new_cg.variables)
[W]
>>> new_cg.shared_variables
[W, carried_over_state]
>>> GradientDescent(cost=cost)  # Only selects shared which are TRAINABLE
```

**(2) Replaced variables get PARAMETER role as well**
```python
>>> new_cg = add_noise(cg, W)
>>> VariableFilter(roles=[PARAMETER])(new_cg.variables) 
[W, noised_W]
>>> new_cg.shared_variables
[W, carried_over_state]
>>> VariableFilter(roles=[PARAMETER])(new_cg.shared_variables)
[W]
>>> GradientDescent(cost=cost)  # Takes only shared variables with PARAMETER role
```

**(3) Introduce a `REPLACEMENT` role**
```python
>>> new_cg = add_noise(cg, W)
>>> new_cg = collect_parameters(new_cg)
>>> VariableFilter(roles=[PARAMETER])(new_cg.variables) # Or only W keeps PARAMETER?
[W, noised_W, collected_parameters]
>>> VariableFilter(roles=[PARAMETER])(new_cg.shared_variables) 
[collected_parameters]
>>> VariableFilter(roles=[REPLACEMENT])(new_cg.variables) 
[noised_W, collected_parameters]
>>> GradientDescent(cost=cost)  # Takes shared variables with PARAMETER tag?
```

**(4) NEW: Introduce a `REPLACED` role**
```python
>>> VariableFilter(roles=[PARAMETER])(new_cg.variables) 
[W, noised_W, collected_parameters]
>>> [var for var in VariableFilter(roles=[PARAMETER])(new_cg.variables)
...  if var not in VariableFilter(roles=[REPLACED])(new_cg.variables)]
[collected_parameters]
```

**Issues**

I actually like the (2) example here, but the part I find problematic is still there: There is no easy way to select the *actual* weight matrices being used by the linear transformations.
```python
>>> new_cg = add_noise(ae_cg, W)
>>> VariableFilter(roles=[PARAMETER])(new_cg.variables)
[noised_W, W, W_T]
>>> VariableFilter(roles=[PARAMETER])(new_cg.shared_variables)
[W]
```
But I want `[noised_W, W_T]`!

The `REPLACEMENT` suggestion doesn't help here either. Take the more involved case where I have an autoencoder followed by a linear transformation, where I want to retrieve the actual linear transformations used (so `[noised_W1, W1_T, W2]`):
```python
>>> VariableFilter(roles=[PARAMETER])(new_cg.shared_variables)
[W1, W2]
>>> VariableFilter(roles=[PARAMETER])(new_cg.variables)
[noised_W1, W1, W1_T, W2]
>>> VariableFilter(roles=[REPLACEMENT])(new_cg.variables)
[noised_W1]
```
(1), the split between `TRAINABLE` and `PARAMETER`, *does* work here:
```python
>>> VariableFilter(roles=[TRAINABLE])(new_cg.shared_variables)
[W1, W2]
>>> VariableFilter(roles=[PARAMETER])(new_cg.variables)
[noised_W1, W1_T, W2]
```
And so does (4):
```python
>>> [var for var in VariableFilter(roles=[PARAMETER])(new_cg.variables)
...  if var not in VariableFilter(roles=[REPLACED])(new_cg.variables)]
[noised_W1, W1_T, W2]
```

I think (1) is nicer when it comes to selecting variables, because you can just do `VariableFilter(roles=[TRAINABLE])`, but (4) is nicer because variables don't magically lose their roles. However, selecting the right ones now becomes a bit more complicated. Perhaps that's not too bad though, considering we're probably considering pretty exceptional corner-cases here. So for now, let's say I'm leaning towards (4).

rizar:
Sorry @bartvm, I did not really have time today to think your last post through and form an opinion. I will surely do that tomorrow morning!

rizar:
I am not sure if you had it mind, but I  have just realized that when one does `new_cg = theano.clone(old_cg, {old_variable: new_variable})` and `new_variable` is a function of `old_variable`, then the original `old_variable` becomes a part of `new_cg` as well:

```python
>>> x = tensor.matrix('x')
>>> y = x + 1
>>> z = theano.clone(y, replace={x: x + 1})
>>> z.owner.inputs[0] == y
False
>>> z.owner.inputs[0].owner.inputs[0] == x
True
```

I might be wrong, but looks like it means we can _never_ change tags of shared variables, because we do not want to affect the pure graph on the one hand, and we do want them to share parameter on the other hand. This conflicts with the options (4) and (1).

rizar:
Another consideration: blindly copying all the tags is also dangerous. Their semantic is defined by the user, and doing that might break certain important invariants. For instance, if an `add_dropout` routine is applied to an input of a brick, copying the tag implies that there are now two inputs with same name, which is a contradiction.

I am trying to think of an austere counter-proposal that would involve no role copying and no adding roles to existing variables, but still support all selections you need.

bartvm:
@rizar This was updated to follow the new role replacements pattern. The single shared variable that gathers everything gets the role `COLLECTOR`, the slices of this variable (i.e. the actual replacements) get the role `COLLECTED` and `tag.replacement_of = original_parameter`.

Ready to merge now?

blocks/graph.py
```diff
@@ -421,6 +423,91 @@ def apply_noise(computation_graph, variables, level, seed=None):
...
+    add_role(new_params, COLLECTOR)
```
rizar: Do you plan to give this variable also the `PARAMETER` roles? Without this it will not be picked by `cg.parameters`.

bartvm: Mmm... This role-tagging is still tricky. So I figured that we weren't copying roles, so in that case I would argue that the correct way of discovering this parameter should strictly be:

* Follow the `replacement_of` tag of the `COLLECTED` variables, which will show you that they are parameters.
* Noticing that they aren't shared variables, walk the graph to see what shared variables they depend on (`COLLECTOR` in this case). This tells you what Theano shared variable you should train.

rizar: Sorry, forgot about this!

I think it would be perfectly okay to give the collector a `PARAMETER` role. I do not consider it as a role copying, you are introducing something new into the graph and you should be allowed to give it a role.

The algorithm of parameter discovery you propose seems quite custom to me.  I am not sure that we should have something like that as our default logic. For all other graph substitutions we currently have the original shared variables keep being a part of the graph and will be found by `ComputaionGraph.parameters` call.

bartvm: I'm not sure how custom the logic is, but I need to think about it a bit more. The way we retrieve parameters in Blocks is starting to bug me a bit (since we have `Selector.get_params`, `Model. get_params`, and `ComputationGraph.params`, which just seems overkill). For now, I can just give it a `PARAMETER` role if you want, fine by me.

rizar: A `PARAMETER` role would do the job for now, agreed.

PR #181, opened 2015-01-31
No more global seed

### LINKED ISSUE #180: Change use of global seed
@dwf pointed out correctly that in the current implementation, bricks created separately from each other will use the same seed, and so will have exactly the same weights (assuming they use the same initialization scheme).

It could be an idea to create a single global RNG to attach to the `Brick` class. That way we can still set a global seed that this RNG will be initialized with, but it won't be re-used

bartvm:
Instead, draw a seed from a global RNG for each brick. This prevents
bricks from having the exact same weight initialization.

Fixes #180.

### DIFF
--- a/blocks/bricks/__init__.py
+++ b/blocks/bricks/__init__.py
@@ -36,9 +36,24 @@ class Random(Brick):
         A ``MRG_RandomStreams`` instance.
 
     """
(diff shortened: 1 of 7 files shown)


bartvm:
I should probably given the Theano RNG the same treatment, will do that later.

PR #183, opened 2015-01-31
Move variable roles to separate file

bartvm:
Roles all over the place started annoying me, and I actually like this much better. This way there is confusion about where to find them.

### DIFF
--- a/blocks/bricks/__init__.py
+++ b/blocks/bricks/__init__.py
@@ -9,22 +9,10 @@ from theano.sandbox.rng_mrg import MRG_RandomStreams
 
 from blocks import config
 from blocks.bricks.base import application, _Brick, Brick, lazy
-from blocks.graph import add_role, ParameterRole
+from blocks.roles import add_role, WEIGHTS, BIASES
 from blocks.utils import pack, shared_floatx_zeros
 
 logger = logging.getLogger(__name__)
-
-
-class WeightsRole(ParameterRole):
-    pass
-
-WEIGHTS = WeightsRole()
-
-
-class BiasesRole(ParameterRole):
-    pass
-
-BIASES = BiasesRole()
 
 
 class Random(Brick):
--- a/blocks/bricks/base.py
+++ b/blocks/bricks/base.py
@@ -8,8 +8,8 @@ from six import add_metaclass
 from theano import tensor
 from theano.gof import Variable
 
-from blocks.graph import (add_annotation, add_role, Annotation, INPUT, OUTPUT,
-                          PARAMETER)
+from blocks.graph import add_annotation, Annotation
+from blocks.roles import add_role, PARAMETER, INPUT, OUTPUT
 from blocks.utils import pack, repr_attrs, reraise_as, unpack
 
 
--- a/blocks/graph.py
+++ b/blocks/graph.py
@@ -10,6 +10,7 @@ from theano.gof.sched import make_dependence_cmp, sort_apply_nodes
 from theano.sandbox.rng_mrg import MRG_RandomStreams
 
 from blocks import config
+from blocks.roles import add_role, AUXILIARY
 from blocks.utils import (is_graph_input, is_shared_variable, dict_union,
                           shared_like)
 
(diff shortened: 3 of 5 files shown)


rizar:
I would prefer  to have the roles declared as needed, that is INPUT and OUTPUT roles should be declared in the bricks submodule. One reason is that their documentation should refer to other components.

The annoying inconvenience you experienced is the result of us not having something like `blocks.all` which would import important stuff from all other modules. In this concrete pull request you could just import all the roles in `blocks/roles.py` instead of moving the declarations. Then documentations would be located in the same files where the components they refer to are.

bartvm:
So you're suggesting moving things out of `graph.py` and moving everything into `bricks/*.py` files?

Although I agree that we should eventually start importing objects from other files judiciously so that they are more easily accessible to users, it wasn't the importing that bugged me here. It just seems that the locations of parameter roles are completely arbitrary right now. Why would `PARAMETER` and  `AUXILIARY` be in `graph.py` but `WEIGHTS` in `blocks/__init__.py`? In essence, parameter roles are the one thing that are common to bricks and the computation graph: The bricks assign them, the graph reads them and allows them to be filtered. For me that means that both of them should import the roles from a central place which is neither bricks nor graph-related.

rizar:
I thought that the roles AUXILIARY and PARAMETER should be declared in the `graph.py`, the rest in `blocks/bricks/__init__.py`. If I need at some point RECURRENT_WEIGHTS, I would declare them in `recurrent.py`. That is I think the roles should be declared in the context in which they make sense.

bartvm:
I don't see why `PARAMETER` makes sense in `graph.py` but not in `bricks/__init__.py`, which I think shows that these are pretty arbitrary distinctions. Only bricks have parameters, so if `WEIGHTS` is going to go to the bricks, then I think `PARAMETER` should go there too...

I would prefer them to (1) either go all in one file, so that there will never be any confusion as to where they are, or (2) if you insist on having them spread over different files, than at least they should all go in one part of the library (the bricks part) instead of being arbitrarily spread out over two parts of the library which we've been trying to keep completely separate.

rizar:
I think that `PARAMETER` belongs to `graph.py` because a person might want not to use bricks at all but have his parameters automatically extracted from the graph. That is, it is an abstraction of a level lower then bricks. Same as `AUXILIARY`.

bartvm:
I can give exactly the same argument for `WEIGHTS` being part of `graph.py`, if I don't want to use bricks at all but just want to extract my weight matrices... You could also argue that `PARAMETER` is to `Brick` as `WEIGHTS` is to `Linear`.

On a meta-level: If the two people who came up with these things can't decide on which variable role goes where, then obviously the distinction isn't easy to make, so I would go for the simpler approach of at least putting all of them in one part of the library. This also leads to a much cleaner separation; at least if I need to find the docstring of a variable role somewhere, I know that I need to find it in the relevant file in `blocks/*`.

rizar:
Okay, I see. I do not mind too much putting these roles we currently have together, but if this specialization goes on and the list of roles grows longer, we might need to reconsider this decision later.

bartvm:
I'm okay with spreading them throughout the `bricks/*` files instead of putting them all together. Right now that means that some would go in `base.py` (e.g. `AUXILIARY`, `PARAMETER`), some in `__init__.py` (e.g. `WEIGHTS`, `BIASES`) and eventually other files as well (e.g. `RECURRENT_WEIGHTS` in `recurrent.py`). I guess you prefer that over having a `roles.py`, right?

rizar:
But I do not see any reason for `AUXILIARY` to be declared in `bricks/*`, that's the problem.

bartvm:
Okay then, let's merge this for now and if the list grows longer then we can re-evaluate.

rizar:
A late suggestion: maybe we at least move `INPUT` and `OUTPUT` to `bricks/*`? They make zero sense outside of this folder.

bartvm:
Sure, but should I import them into `roles.py` then so that you can still `from blocks.roles import INPUT`?

rizar:
First I thought "yes", but then I thought this can lead to awful import loops. We should better import all roles somewhere at an upper level.

Issue #177, opened 2015-01-30
The Theano toposort we use is n^2


rizar:
Which makes it run 1 minute on my gradient computation graph.

I found lots of toposort's in Theano and can not figure out what is the simplest way to get my apply nodes sorted in linear time. @nouiz, @lamblin, can you give me a hint?

nouiz:
toposort isn't always the same implementation. It depend when this happend.
At the start of Theano optimization, when there is no op with destroy_map,
we use a faster version that do not check for  cycle in the graph. That is
faster. But as soon as we have op with destroy_map, there can be cycle in
the graph and you must add this extra check that is costly.

From what you told, the toposort is done in the grad. So it should use the
"fast" version. Where do you find the n^2 number? I never computed it. But
I never saw a problem with the "fast" version.

Can you show the toposort that cause problem? When I want to make Theano
optimization faster, most of the time if there is problem of slowness due
to toposort, I try to call it less often. And up to now, this always worked.

rizar:
I am sorry, I did not make myself clear.

In Blocks we offer the user some additional tools to work with Theano graphs. In order to do so, the first thing we do with the Theano graph is extracting all the variables in the topological order. Here is a link to the code that does it: https://github.com/bartvm/blocks/blob/master/blocks/graph.py#L86

The problem is that the `sort_apply_nodes` routine we use is quadratic (due to https://github.com/Theano/Theano/blob/master/theano/gof/sched.py#L163). So my question is: which toposort function from Theano should we use. There are a quite a few of them with very little documentation, and it is not clear which one will not be, for instance, removed in the next release.

I am not sure I understand how can we have cycles in Theano graphs.

nouiz:
I use theano.gof.graph.io_toposort(inputs, outputs) when I need a
deterministic order but, do not care if that order is a valid execution
order. To my knowledge it is the fastest implementation in Theano.

I do not think we will remove some of the implementation. Try it. If it is
too slow, tell me.

rizar:
I've tried and it seems to work correctly and fast. But I am not sure I understand what you mean by "deterministic but not valid execution order". Doesn't it ensure that if a node A has an output of a node B as its input, then B will appear earlier in the returned list? From the code it seems it does, and why isn't it a valid execution order?

nouiz:
The case you describe is done by this method. The problem is related to
destroy_map. Do you know what it is?

Quickly, if a Theano op have an output that is a view of an input like
[...]

Issue #52, opened 2014-12-03
 Base Feedforward Brick


rizar:
Seems like we need a new abstraction: a brick with a single input and a single output that supports setting the input and output dimensions, i.e. has some writeable properties `input_dim` and `output_dim`. Or has `set_input_dim`, `set_output_dim` methods. I suggest an to create an abstract brick `Feedforward` for it.

Motivation: before I used MLPs everywhere for intermediate transformations and configured them with  `mlp.dims[0] = smth`. Now we also have `LinearMaxout` and later we will have `DaisyChain`. There should be a shared configuration interface for all these guys.

Opinions?

bartvm:
Fixed in #184.

PR #184, opened 2015-01-31
Attention Mechanism Demo (raw) and Related Changes

rizar:
I have been working for a week on an example of using sequence generators. On my way I had to do quite a lot of bug fixing, optimization and refactoring. This night the cost finally went down to zero, although it blew up quickly again due to the fact that I still do not have gradient clipping. But I would like to merge my partial results.

Overview of changes:

1. A small script that loads a main loop pickle and dumps it using `blocks/dump.py`. 

2. **Changes in bricks**:
 
  * Fixed a bug in `AttentionTransition` that led to crazy nested scans and huge slowdowns. Put a warning to prevent people from doing it again. Briefly: ones underspecifies the `contexts` attribute of his recurrent method, everything can still work but the optimizer is driven crazy.
  
  * Introduced the `Feedforward` abstraction as proposed in #52. Using it added a missing `tanh` to the `SequenceContentAttention`. That's a crucial fix, because otherwise the whole alignment computation mechanism is linear and does not have chance to train at all.

3. **Changes in the computation graph**: use a faster toposort see #177. Added a test for `get_snapshot`. Unfortunately `get_snapshot` crashes on my example with sequence generators, quite deep in Theano so I could not debug it.

4. The `AggregationBuffer` builds `ComputationGraph` only once and then uses its `has_input` routine for its purposes: got rid of quadratic complexity at this point.

5. A filtering wrapper `DataStreamFilter`.

6. The example itself!

@bartvm, can you please take a look at the changes in the core? (i.e. the `Feedforward` interface)


blocks/bricks/__init__.py
```diff
@@ -158,7 +159,56 @@ def _push_initialization_config(self):
...
+
+    Many bricks have just one input and just one output (activations,
+    :class:`Linear`, :class:`MLP`). To make such bricks interchangable
+    in most contexts they should share an interface for configuring
+    their input and output dimensions. This brick provides such an
+    interface.
+
+    """
+    @abstractproperty
+    def input_dim():
```
bartvm: These should still take `self` as an argument.

blocks/bricks/__init__.py
```diff
@@ -158,7 +159,56 @@ def _push_initialization_config(self):
...
+        """The input dimension of the brick."""
+        pass
+
+    @abstractproperty
+    def output_dim():
+        """The output dimension of the brick."""
+        pass
+
+
+def override_with_attribute(property_, attribute=None):
```
bartvm: Where does the need for this abstraction come from exactly? These properties seem to add no new behaviour whatsoever, so why can't you just do `self.input_dim = input_dim` in the `__init__` of `FeedForward`?

blocks/bricks/recurrent.py
```diff
@@ -130,6 +130,18 @@ def only_given(arg_names):
...
                 batch_size = kwargs.pop('batch_size')
 
+            # Handle the rest kwargs
+            rest_kwargs = {key: value for key, value in kwargs.items()
+                           if key not in scan_arguments}
+            for value in rest_kwargs.values():
+                if (isinstance(value, Variable) and not
+                        is_shared_variable(value)):
+                    warnings.warn(
+                        'Your function uses a non-shared variable other then'
```
bartvm: *Other *than*

blocks/dump.py
```diff
@@ -89,7 +89,7 @@ def extract_parameter_values(bricks):
     if isinstance(bricks, Brick):
         bricks = Selector([bricks])
     if not isinstance(bricks, Selector):
-        raise ValueError
+        bricks = Selector(bricks)
```
bartvm: I guess `bricks` can now be a list of bricks? Docstring needs to be updated then.

bin/blocks-dump
```diff
@@ -0,0 +1,16 @@
+#!/usr/bin/env python
+from argparse import ArgumentParser
+
+from blocks.scripts import dump
+
+if __name__ == "__main__":
+    parser = ArgumentParser("Dumps your pickled main loop")
```
bartvm: Could eventually do with a bit more documentation e.g. telling the user that this extracts the parameters in NumPy files and allows the model to be shared with people on other platforms.

examples/reverse_words/main.py
```diff
@@ -0,0 +1,27 @@
+#!/usr/bin/env python
+"""Learm to reverse words of natural text."""
```
bartvm: \**learn*

bartvm: Also, I think we talk about *natural language*, but *natural text* isn't really a thing, so I would just say *Learn to reverse the words in a text.*

bartvm:
Looks pretty good, the only thing I don't understand is the `override_with_attribute` function, which I'm not a fan of. Otherwise, just minor things.

blocks/bricks/__init__.py
rizar: Because `input_dim` is deliberately made an abstract property so that every attempt to access it would cause a failure. I did not include a default interpretation in the `Feedforward` to prevent mistakes when people forget to override the required properties. `override_with_attribute` overrides an abstract property with a very simple ones, behaving exactly like an attribute.

bartvm: If you just want it to given an error on access, then I don't see why the usual way doesn't work:
```python
@property
def input_dim(self):
    if not hasattr(self, '_input_dim'):
        raise ValueError("Your feed-forward brick forgot to set the input_dim")
    return self._input_dim

@input_dim.setter
def input_dim(self, value):
    self._input_dim = value
```
Either way, the user is allowed to change `input_dim` or `output_dim` up to allocation, right? So I think that this is definitely the best option. An `abstractproperty` doesn't make sense if you are allowing people to set attributes instead of properties.

rizar: I want to give an error if the author of the class forgot to override the properties. E.g. the `MLP` bricks currently redirects the `input_dim` and `output_dim` requests to its `dims` array. If you remove these implementations, you will get an error from accessing an abstract property. Not only when reading, but also at writing, unlike your implementation above.

bartvm: An `abstractproperty` does not give an error on access! It gives an error when a class does not implement the property at all. But now you're forcing the user to explicitly implement a property (that mimics an attribute) if they just want to use an attribute (like in `Linear` where you just want to do `self.input_dim = input_dim`)... That's not very nice at all. A user should be able to just use an attribute if he wants to.

The only requirement I see is that at the time of allocation, there is an `input_dim` and `output_dim` that makes sense. This is the same requirement we have for every other attribute on every other brick. I think that in this case we just need to forgo the complexity of forcing a user to use a non-obvious way of overriding an abstract property with a property that acts like an attribute in order to set the attribute. Just reading that sentence makes me even more convinced that this is really adding needless complexity; just add to the docstring: "FeedForward bricks are expected to have a `input_dim` and `output_dim` attribute accessible on allocation."

rizar: > It gives an error when a class does not implement the property at all.

Right, but this is just what I wanted in fact. 

There two ways to declare  interfaces: in the code or in the docs. The disadvantage of the second one is that nothing is checked. I am trying to use the first one here to gently remind the user that he forgot to implement something which is required. Indeed it comes with a price: one can not just use an attribute with same name, one has to override the property.
To avoid boiler-plate code that does it, I implemented `override_with_attribute`.

So this is a typical dilemma of security vs. readability. I can switch to using java-like setters and getters, `get_input_dim`, etc.

bartvm: Yes, but we're not writing in Java, and Python gives a lot more freedom to the user. I mean, private attributes don't even exist in Python, and to define an abstract class you need to import metaclasses... That kind of shows what Python's intended design is.

In this case I think the readability side definitely weights heavier. For a few reasons:

* This is not about a user passing wrong values or anything, where we want to raise informative errors, but about someone building their own bricks. Someone who is going to build their own bricks can be expected to have read the documentation of the bricks he is inheriting from. We can't babysit them every step along the way. We can even put this `Feedforward` brick in an upcoming "How to write your own brick"-tutorial.
* You are trying to help the user by reminding them to set these attributes. What will happen in reality is that a user who doesn't read the documentation will simply type:
```python
class MyNewBrick(Feedforward):
    def __init__(self, input_dim, output_dim):
        self.input_dim = input_dim
        self.output_dim = output_dim
```
This will crash with `TypeError: Can't instantiate abstract class MyNewBrick with abstract methods input_dim`. Considering we are dealing with someone who was in a rush, and maybe not very good at Python, he won't spend any time on figuring out this error, and simply write:
```python
class MyNewBrick(Brick):
    def __init__(self, input_dim, output_dim):
        self.input_dim = input_dim
        self.output_dim = output_dim
```
The user doesn't care about the fact that he's not using a base class we wrote for him. He just wanted the `input_dim` attribute, and if that's actually getting *more* difficult to set than easier, then he's just going to use something else.
* In general, considering we are dealing with a framework aimed at computer scientists who want to do rapid prototyping and development, I care a lot more about readability.
* Python is based around EAFP and duck-typing. These kind of security checks which are central to Java, C++, etc. are very non-Pythonic.

rizar: Okay, let's keep it simple. I am missing a few features from those stricter languages in Python, in fact, and declaring an interface just by describing the set of class attributes in plain text seems rather weak for me. Because the error message "<something> does not have an attribute input_dim" does not really convey the message that the interface is not respected.

I've just thought that I can override `__getattr__` of `Feedforward` to give a better error message.

rizar:
I am merging it, and if you @bartvm  have any more comments, I will be happy to make respective fixes.

blocks/bricks/__init__.py
bartvm: You call them features, but I guess "Pythonistas" would argue that it's a design philosophy, and that these kind of things have very consciously been excluded from Python. Improving the readability of the error messages sounds like a very good way to go.

bartvm:
LGTM.

Sidenote: I spent a bit of time getting the docs into shape yesterday (still a long way to go, but it's getting better), and realized that if you want to a link to an object or class outside of the current file you need to do one of two things:
```
:class:`~blocks.bricks.base.Brick`
```
```
:class:`.Brick`
```
If you instead use
```
:class:`Brick`
```
Sphinx won't be able to resolve the object you are referring to.

rizar:
Okay. Guess we should eventually have a sphinx cheatsheet with the relevant constructions, because its documentation is rather bulky I can not make myself to parse it...

bartvm:
Yeah, its documentation is a nightmare. Too bad, because I think it's actually pretty powerful. I'll make a ticket for that cheat sheet.

Issue #154, opened 2015-01-26
Error handling of main loop


bartvm:
More small things regarding the main loop: Right now the `after_training` extensions are run whenever an error occurs during training (because it's in the `finally` block). I'm just wondering whether this is the correct thing to do; maybe another event should exist for this e.g. `on_error`?

For example, I was having errors with pickling again (recursion limit), which means that the error got caught, the `Printing` extension got called and suddenly said "Training finished!", after which the `SaveLoad` extension tried to save the model, crashing again. It took me a while to figure out what was happening here. I guess that in general we should:

* Give a warning, saying that an error occurred, but that we are trying to run some extensions before exiting
* Call the extensions with the `on_error` event, so that they can behave differently
* Save the model to a different file; we don't want to be overriding a good model with a broken one if the error was e.g. a filesystem error.

rizar:
That sounds reasonable, it is especially bad that we corrupt a good model now when an error happens.

bartvm:
Partially addressed in #158 (still needs saving to a different model on interrupt).

bartvm:
I think all the of these issues have been resolved.

PR #191, opened 2015-02-01
Also handle SIGTERM signals

### LINKED ISSUE #171: Respond to SIGTERM as to SIGINT
Clusters often send a `SIGTERM` to warn the process that it's going to end. We should respond to this as we do to a user interrupt (`SIGINT`).

bartvm:
@rizar This should be straightforward, but let me know if I overlooked anything. This should make Blocks play nice with all the clusters.

Fixes #171.

### DIFF
--- a/blocks/main_loop.py
+++ b/blocks/main_loop.py
@@ -76,8 +76,10 @@ class MainLoop(object):
         a `training_finish_requested` record in the log.
 
         """
-        self.original_handler = signal.signal(
-            signal.SIGINT, self._handle_keyboard_interrupt)
+        self.sigint_handler = signal.signal(
+            signal.SIGINT, self._handle_interrupt)
+        self.sigterm_handler = signal.signal(
+            signal.SIGTERM, self._handle_interrupt)
         try:
             logger.info("Entered the main loop")
             if not self.status._training_started:
@@ -98,7 +100,8 @@ class MainLoop(object):
             self.log.current_row.training_finished = True
         finally:
             self._run_extensions('after_training')
-            signal.signal(signal.SIGINT, self.original_handler)
+            signal.signal(signal.SIGINT, self.sigint_handler)
+            signal.signal(signal.SIGTERM, self.sigterm_handler)
 
     def find_extension(self, name):
         """Find an extension with a given name.
@@ -156,16 +159,17 @@ class MainLoop(object):
         # the iteration the corresponding log record can be found only in
         # the previous row.
         if (self.log.current_row.training_finish_requested or
-                self.log.current_row.keyboard_interrupt_received or
-                self.log.previous_row.keyboard_interrupt_received):
+                self.log.current_row.interrupt_received or
+                self.log.previous_row.interrupt_received):
             raise TrainingFinish
 
-    def _handle_keyboard_interrupt(self, signal_number, frame):
+    def _handle_interrupt(self, signal_number, frame):
         # After receiving a first keyboard interrupt signal,
         # ignore all following ones.
-        signal.signal(signal.SIGINT, self.original_handler)
+        signal.signal(signal.SIGINT, self.sigint_handler)
+        signal.signal(signal.SIGTERM, self.sigterm_handler)
         self._run_extensions('on_interrupt')
-        self.log.current_row.keyboard_interrupt_received = True
+        self.log.current_row.interrupt_received = True
 
 
 class TrainingFinish(Exception):


blocks/main_loop.py
```diff
@@ -156,16 +159,17 @@ def _check_finish_training(self):
...
-                self.log.current_row.keyboard_interrupt_received or
-                self.log.previous_row.keyboard_interrupt_received):
+                self.log.current_row.interrupt_received or
+                self.log.previous_row.interrupt_received):
             raise TrainingFinish
 
-    def _handle_keyboard_interrupt(self, signal_number, frame):
+    def _handle_interrupt(self, signal_number, frame):
         # After receiving a first keyboard interrupt signal,
         # ignore all following ones.
```
rizar: That's not true any more and in fact I do not know what the correct policy is. I guess we should catch the first interrupt and try to handle it gracefully, and with the second and all the following we should just finish immediately. What do you think?

rizar:
That's a good thing to do, but I am not really satisfied with the current policy of handling SIGINT, so we should reconsider both.

blocks/main_loop.py
bartvm: Yeah, but that's the current behavior, right? We restore the handler to the original one in the next few lines, and the original handler makes Python quite immediately, so pressing Ctrl + C twice should make things crash with a `KeyboardInterrupt` straight away (which I think is the correct behavior). I've done this a few times (just press Ctrl + C a few times, sometimes it doesn't catch the second one) and it works just fine.

rizar: Not exactly, because the `finally` clause is still run (https://github.com/bartvm/blocks/blob/interrupt/blocks/main_loop.py#L101), and I think after the second interrupt it shouldn't. And in case of the cluster we might need different behaviour I guess: try hard to finish gracefully until killed blatantly by SIGKILL.

bartvm: Right, that explains why you it didn't catch it on the second time, because you need to press it 3 times: The first time `_handle_interrupt` gets called and the signals get restored. The second time, you cause a `KeyboardInterrupt` which brings you straight to the `finally` block that calls `self._run_extensions('after_training')`. The third time actually kills the program.

Maybe this would be good behavior:
* On SIGINT, try to finish the epoch. A second SIGINT will try to finish the batch. The third one will kill your program like usual.
* On SIGTERM, we try to finish the batch and then exit.

rizar: I am fine with your proposed SIGTERM behavior. As for waiting until the epoch's end: basically it means that a lot of users will have to always press it twice, because for many application epochs are very long. If we print something like "press another Ctrl + C to finish faster", this is also OK for me. 

And perhaps we should move printing related to handling interrupts from the `Printing` extension, because it is tightly related to interrupt handling logic which is currently simply hard-coded in the main loop. 

But well, this does not have be a part of this pull request, so if you want to have SIGTERM handled right away, you can merge it. Later I think we should consolidate all discussions on handling all sorts of interrupts and errors in one issue and decide what our policy is.

bartvm: Yeah, a message like that is what I had in mind.
```
Finishing epoch... Press CTRL + C again to terminate after this batch instead.
Finishing batch... Press CTRL + C again to kill training now (WARNING: Extensions won't be called!)
```
This is also quite closely related to #154. We don't want an impatient user to press CTRL + C three times and end up with a corrupted model,

rizar: That's true, but I do not like the idea of always saving to another file when interrupted casually (not by an exception), because I do it all the time and want to resume training simply. An easy way to prevent corrupted writes is to pickle to a temporary file and then move the result in a critical section. I agree with you though that when training is ended as a result of an exception, we should save to a different path.

bartvm: That's what I meant, actually. I think we should just pickle to `.model.pkl.tmp` or something, and once pickling has completed successfully, we move `.model.pkl.tmp` to `model.pkl`. There will always be a split second in the middle of copying in which an interruption might happen, but at least we will never corrupt an good copy because of recursion errors, pickling errors, etc.

bartvm: Actually, you can just remove the old one and rename the current one, that gives you zero chance of ruining a good copy with a bad copy.

rizar: And you can also ignore all the signals in this world while doing this.

bartvm:
Closing this, see continued discussion at #287

Issue #182, opened 2015-01-31
SteepestDescent conceivably bakes learning rate constant into graph


dwf:
In looking how to implement other learning rules, I noticed SteepestDescent is baking the learning rate into the graph as a constant. This is bad because it doesn't allow for learning rates to be updated on a schedule, or changed during training for other reasons.

I can see a few ways of addressing this:

* Have the StepRule own a shared variable that other objects can modify.
* Simply change the documentation to encourage users to pass in a shared variable or symbolic expression; then some other sort of object can own the shared variable or symbolic expression. 

I suppose that an object that inherits from both `Brick` and `Extension` can be used to provide a gradually annealed learning rate. If this sounds like a good direction I'll send a pull request when I get a minute.

bartvm:
I guess that we can just copy Pylearn2's approach here, and give the `StepRule` a shared variable. We can then implement all sort of learning rate schedules through extensions.

I wouldn't mix `Brick` and `Extension`, the philosophy is very much that bricks help you construct a Theano graph, but that by the time that you start the main loop, the dependencies on that part of the library are minimal/non-existent.

rizar:
Making the learning rate accessible as a shared variable attribute to the `StepRule` seems like a good solution to me.

rizar:
Solved by #192

PR #192, opened 2015-02-02
Gradient clipping, shared learning rate, composite rule

### LINKED ISSUE #185: Gradient Clipping
When designing the `StepRule`'s interface I made a mistake: I forgot that for gradient clipping one needs access to gradients w.r. all parameters at the same time, not only to each one separately. Thus `StepRule` should be redesigned a bit to make gradient clipping possible.

rizar:
Fixes #185, #182.


rizar:
Update: added a `CompositeRule` for chaining purposes.

rizar:
Tests pass, waiting for an LGTM :)

blocks/algorithms/__init__.py
```diff
@@ -222,6 +224,29 @@ def compute_step(self, param, grad_wr_param):
...
 
+    def compute_steps(self, gradients):
+        """Build a Theano expression for steps for all parameters.
+
+        Override this method if you want to process the gradient
+        with respect to all parameters as a whole, not parameter-wise.
+
+        Parameters
+        ----------
+        gradients : dict of (:class:`~tensor.TensorSharedVariable`
```
dwf: You really want this to be an `OrderedDict` for determinism reasons.

blocks/snippets.py
```diff
@@ -0,0 +1,14 @@
...
+def L2_norm(tensors):
+    """Computes the total L2 norm of a set of tensors.
+
+    Parameters
+    ----------
+    tensors : iterable of :class:`~theano.Variable`
+        The tensors.
+
+    """
+    return tensor.sqrt(sum((t ** 2).sum() for t in tensors))
```
bartvm: Do we really need to filter this out? It seems that `tensor.sqr(tensors).norm(2)` would work as well, and that's short enough to not warrant a separate function.

If we do want to start a collection of Theano convenience expressions, I would call them `blocks.expressions` and not `blocks.snippets`.

blocks/algorithms/__init__.py
```diff
@@ -244,12 +269,66 @@ class SteepestDescent(StepRule):
...
+
+    """
+    def __init__(self, threshold=None):
+        if threshold:
+            self.threshold = theano.shared(threshold)
+
+    def compute_steps(self, gradients):
+        if not hasattr(self, 'threshold'):
+            return gradients
+        norm = L2_norm(gradients.values())
```
bartvm: There's no need to actually take the L2-norm here. You could just do `tensor.sqr(gradients.values()).sum()` and square the threshold. Saves some computation.

blocks/algorithms/__init__.py
```diff
@@ -244,12 +269,66 @@ class SteepestDescent(StepRule):
...
+    """Clips the total gradient to make it not exceed a threshold.
 
-    def compute_step(self, param, grad_wr_param):
-        return -self.learning_rate * grad_wr_param
+    Parameters
+    ----------
+    threshold : float, optional
+        The maximum permitted L2 norm for the gradient. The gradient
+        will be rescaled to be not higher than this quanity.
+        If ``None``, no rescaling will be applied.
```
bartvm: In what case would we use a `GradientClipping` step rule which doesn't perform clipping? If in some situation step rules are automatically chained through another function, I would say it's that functions responsibility to leave out the clipping, instead of being able to create a non-clipping gradient-clipper :p

dwf: Given that this is sort of a "meta-steprule" that is sort of designed to
wrap/be wrapped by other step rules, I can sort of see the point. This
might be handy for automated hyperparameter search, where special-casing a
lot of the initialization logic quickly turns into a pain.

bartvm: Fair enough. Regarding hyperparameter optimization, @dwf do you know if hyperopt is any good, and worth looking into integrating eventually? I have to say that I haven't given hyperparameter searches a lot of thought, but we'll need to make sure it's easy enough to do.

dwf: Let's discuss it after the ICML deadline :) It's a complicated topic.

bartvm: Sure. That does sound ominous...

blocks/algorithms/__init__.py
rizar: Oh come on, one sqrt of a scalar per an iteration, I believe it's negligible.

blocks/algorithms/__init__.py
rizar: In order to fetch configuration from a text file, e.g.:

```python
step_rule = CompositeRule([Rule1(), GradientClipping(config["clipping"])])
```

blocks/algorithms/__init__.py
rizar: But in `initialize` (lines 189 and 190) the updates are added in the deterministic order, isn't that enough?

dwf: Hmm, yes, you're right. However, subsequent refactorings (or different clients of this code) might miss this subtlety! As such I would recommend either a) a comment around L189 stressing the importance of the deterministic iteration order or b) future-proofing by just using `OrderedDict`.

rizar: I will surely do (a), but as for (b), I think it is an unnecessary complication of the step rule concept. Step rules live at a high level of abstraction where the order of computations does not matter and simplify provide the user with the expressions he needs, and it is up to him how exactly he wants arrange these computations.

blocks/snippets.py
rizar: I think we should filter it out: I hate finding errors in hastily entered snippets like this. The way you do it above is shorter, but to be honest I do not understand how it works and `tensor.sqr?` gives me quite a piece of documentation to read, so I would leave the code I currently have which I believe everyone can read.

I do not like `blocks.expressions` because it is too general. `blocks.theano_snippets` would work best for me.

bartvm: `theano_expressions` then? The thing is that a "snippet" is almost always used to refer to a piece of code that you copy and paste, which is the opposite of what you're doing here.

I don't think that using `tensor.sqr` is in any way confusing though. It's a simple element-wise operator (the name of which describes exactly what it does), which takes multiple inputs, like all element-wise operators in Theano. To me that line is actually much easier to read and shorter than using a generator within a function. I'd also be surprised if the documentation of `tensor.sqr` is longer than 10 words...

And you'll probably say this is negligible, which might be true, but it also makes for a much smaller computational graph than summing in two sweeps instead of one, which might help save on compilation time (assuming that Theano optimizes your expression correctly).

```
Elemwise{sqrt,no_inplace} [@A] ''
 |Sum{acc_dtype=float64} [@B] ''
   |Elemwise{sqr,no_inplace} [@C] ''
     |Join [@D] ''
       |TensorConstant{0} [@E]
       |DimShuffle{x,0} [@F] ''
       | |<TensorType(float32, vector)> [@G]
       |DimShuffle{x,0} [@H] ''
       | |<TensorType(float32, vector)> [@I]
       |DimShuffle{x,0} [@J] ''
       | |<TensorType(float32, vector)> [@K]
       |DimShuffle{x,0} [@L] ''
       | |<TensorType(float32, vector)> [@M]
       |DimShuffle{x,0} [@N] ''
         |<TensorType(float32, vector)> [@O]
```
While the current one gives
```
Elemwise{sqrt,no_inplace} [@A] ''
 |Elemwise{add,no_inplace} [@B] ''
   |Elemwise{add,no_inplace} [@C] ''
   | |Elemwise{add,no_inplace} [@D] ''
   | | |Elemwise{add,no_inplace} [@E] ''
   | | | |Elemwise{add,no_inplace} [@F] ''
   | | | | |TensorConstant{0} [@G]
   | | | | |Sum{acc_dtype=float64} [@H] ''
   | | | |   |Elemwise{pow,no_inplace} [@I] ''
   | | | |     |<TensorType(float32, vector)> [@J]
   | | | |     |DimShuffle{x} [@K] ''
   | | | |       |TensorConstant{2} [@L]
   | | | |Sum{acc_dtype=float64} [@M] ''
   | | |   |Elemwise{pow,no_inplace} [@N] ''
   | | |     |<TensorType(float32, vector)> [@O]
   | | |     |DimShuffle{x} [@P] ''
   | | |       |TensorConstant{2} [@L]
   | | |Sum{acc_dtype=float64} [@Q] ''
   | |   |Elemwise{pow,no_inplace} [@R] ''
   | |     |<TensorType(float32, vector)> [@S]
   | |     |DimShuffle{x} [@T] ''
   | |       |TensorConstant{2} [@L]
   | |Sum{acc_dtype=float64} [@U] ''
   |   |Elemwise{pow,no_inplace} [@V] ''
   |     |<TensorType(float32, vector)> [@W]
   |     |DimShuffle{x} [@X] ''
   |       |TensorConstant{2} [@L]
   |Sum{acc_dtype=float64} [@Y] ''
     |Elemwise{pow,no_inplace} [@Z] ''
       |<TensorType(float32, vector)> [@BA]
       |DimShuffle{x} [@BB] ''
         |TensorConstant{2} [@L]
```

blocks/algorithms/__init__.py
bartvm: I'm a bit confused as to why this would be an "unnecessary complication". I believe that @dwf is simply suggesting you replace this line by `OrderedDict((param, self.compute_step(param, gradients[param]) for param in gradients)`, a very minor change that will make sure if any future refactor that potentially uses this return value directly (instead of looping over it using keys from an ordered list) will still be deterministic.

rizar: And this is exactly what I call an unnecessary complication. Just compare the two constructions:

```python
{param: self.compute_step(param, gradients[param]) for param in gradients}
```

```python
OrderedDict((param, self.compute_step(param, gradients[param]) for param in gradients)
```

The first one uses Python nice built-in syntax for dictionaries and is more readable.

rizar: But well, let's not waste our time for that, I will make it an `OrderedDict`.

blocks/snippets.py
rizar: > theano_expressions then? The thing is that a "snippet" is almost always used to refer to a piece of code that you copy 
> and paste, which is the opposite of what you're doing here.

Literately it is, but this file is exactly to avoid snippets, so this is how that name is carried over to the file. 

>  which takes multiple inputs, like all element-wise operators in Theano.
Okay, but why does it have a single output then, which it apparently has according to the snippet you propose.

> I'd also be surprised if the documentation of tensor.sqr is longer than 10 words

Try `tensor.sqr?`, make some tea of coffee, and enjoy reading. I am not saying that theano is badly documented though.

Smaller CG is the decisive argument for me, will change this one.

bartvm: Oh, but that's the (internal) documentation of the Op for element-wise operations for some reason! The [actual docstring](http://deeplearning.net/software/theano/library/tensor/basic.html#tensor.sqr) is this: "Returns a variable representing the square of a, ie a^2." :)

That reasoning does not make sense to me. Calling a file "snippets" because you don't want to use snippets is like calling a file "old_functions" and filling it with new functions that replace the old ones... Either way, a "snippet" is used to refer to a piece of code that you *don't* factor out, but that you copy and paste all the time. We don't copy and paste and *do* factor out, so they're not snippets, per definition.

rizar: Well, all software engineering is about using old notion in new senses. Our bricks are also not used for real-world building.

I call this file snippets for the reason that its elements are almost like parametrized snippets I can do with my vim plugin. The name `expressions` seems inappropriate for me, because we use this term for Theano variables all across the code, and not for functions returning them. Unfortunately I can not invent a better name, the only other thing I have in mind is `templates`, but I find the confusion even worse then with `snippets`.

bartvm: But they're not like snippets, at all! The only thing that distinguishes snippets from helper-functions/macros is that the former are per definition copy-pasted into your code (most of the time by your IDE). Using that argument, a snippet and a helper-function are the same thing, which they're not.

I think the best alternative is `expressions`, which describes exactly what they are: factored out Theano expressions Yes, strictly speaking they are functions that *return* expressions, but we name functions and files all the time after things that they return/provide; I've never seen someone call a file "fibonacci_returnes" instead of "fibonacci". I've recently already replaced a number of "Theano expressions" by "Theano variables", so I'd rather finish up doing that if you really think the dual use of the word "expressions" is going to confuse people.

Alternative names would be "theano_macros", "theano_utils", where the latter has my preference (because macros, like snippets, have a specific use that is not the case here).

rizar: Bricks are also not like bricks at all, they have zero clay in them!

For me the difference between calling my plugin to substitute inputs in a small piece of code or calling a function to do that is minor. I like to call things according to their purpose. There is a need for a file with small pieces of Theano code: I call it snippets, never caring whether the code is actually copy-pasted or not.

I would be opposed to replacing `expressions` with `variables` in many cases in the code, so let's keep it slow.

`theano_utils` has a disadvantage that for many functions from `utils.py` it is very hard to say where they should belong, to `theano_utils.py` or to `utils.py`. I am fine with "theano_macros¨, though your arguments apply fully to this name as well.

bartvm: It's not about literalism vs. figurativeness here, it's about not reappropriating commonplace understandings of terms to suit your own interpretation. Snippets are--quoting Wikipedia here--"literal text, or written in a simple template language", which "are a small-scale form of copy and paste programming". Hence, these helper functions are not snippets, regardless of what you name them. For many people, including myself, it will not be obvious what `snippets.py` contains. When seeing the name `expression.py` (in the context of a Theano framework) or `theano_expressions.py` I think everyone will have an immediate understanding of the kind of functions he will find inside that file.

rizar: I am not sure I would immediately guess what's in `theano_expressions.py`.  It sounds for me like yet another abstraction level Blocks have on top of Theano. It is not clear that I should look there for `L2_norm` or other one-liners.

So I am fine with `theano_macros.py`, `theano_oneliners.py`, `oneliners.py`, but not with `theano_expressions.py`, because for me expression is not a callable object, and this name gives me much less clue on the content of the file compared to all aforementioned ones.

rizar: @bartvm , I am citing the `Elemwise` documetation: "All the inputs must have the same number of dimensions". I would like to hardcode such a constraint in the `L2_norm` code.

bartvm: Okay, I guess we'll just have to agree to disagree :) If you look through the Theano documentation, you'll find the word "expressions" used throughout to refer to functions applied to variables. Actually, on the first page of the Theano documentation you'll find it says that in Theano you "build expressions for how to put [those] variables together". See also [this part](http://deeplearning.net/software/theano/library/tensor/basic.html#converting-from-python-objects) for example, where it talks about the result of "arithmetic expressions" being a tensor variable and the "inputs to all expressions" needing to be tensor variables.

I actually doubt anyone will ever look for L2-norm anywhere and just type it, considering how short an expression it is, which is why I question the necessity of factoring it out at all, but we already had that discussion.

If you really want to pick from that list, I would go for `theano_macros.py`, else you'll need to rename it the moment you ever want to write something with two lines.

bartvm: Ah, that's true, that would be reason to factor it out to `tensor.sqrt(tensor.sqr([var.flatten() for var in vars]).sum())` then.

rizar: Okay, if it is the idea that an expression is something that is callable and builds variables is a part of Theano slang, I can live with that. It does not seems naturally for me and maybe did not for Jan, as it was he who started using the term expression differently in Blocks.

rizar: Another thing: `var.flatten()` seems to ignore scalars.

Computing L2 norm is not as easy as it seems!

bartvm: Apparently so! Quite absurd really, it seems like such a simple operation. I tried a few variants, but your original one seems to do as well as any other I can think of that handles scalars, so let's stick with it. My bad!

blocks/theano_expressions.py
```diff
@@ -0,0 +1,18 @@
...
+    Parameters
+    ----------
+    tensors : iterable of :class:`~theano.Variable`
+        The tensors.
+
+    """
+    flattened = [tensor.as_tensor_variable(t).flatten() for t in tensors]
+    flattened = [(t if t.ndim > 0 else tensor.alloc(t, 1))
+                 for t in flattened]
+    joined = tensor.join(0, *flattened)
```
bartvm: It worked on my examples, but you're right. I thought element-wise operations performed a `join`, but they actually perform a `as_tensor_variable` on the list, which does a `join` as well, but only after prepending an empty dimension. By the way, `t.dimshuffle('x')` should add a dimension in place instead of allocating new memory for it.

You might want to update the dosctring by the way; right now it suggests that the inputs need to be tensors, but you allow for numeric inputs as well (and use that in the tests). If you only want to support tensor variables, then remove the `as_tensor_variable` part and just change the tests to look like `L2_norm([x, y]).eval({x: 1, y: [1, 2]})`.

Otherwise, enough nitpicking and LGTM!

PR #194, opened 2015-02-03
Rename expressions

bartvm:
@rizar To solve all the confusion

Rename expressions to variables in order to distinguish clearly between a Theano expression (e.g. x + y) and its result (which is a variable). Hence all arguments and return values of functions are variables (not expressions). Functions and methods can represent Theano expressions.


rizar:
Okay, nice, I like it more than I thought I would. Makes me wondering though, what is the true meaning of the word expression, but this is a whole another story.

PR #195, opened 2015-02-04
Scrutinizer config

bartvm:


rizar:
Will we use scrutinizer instead of landscape or in addition to it?

bartvm:
They basically do the same thing (run Pylint and flake8) so I was thinking of replacing Landscape.io with it. The main reason being that it gives immediate feedback on the PR, and that we can move the docstring/PEP8 checks to there instead of running those on Travis. It also does code coverage, so that means we can get rid of coveralls (assuming it works well on Scrutinizer), which means one less service to worry about.

PR #196, opened 2015-02-04
Finalize the Attention Mechanism Example.

rizar:
It works quite well, though the current way to train it requires 20000 iterations, which on my CPU takes some 9 hours. 

This pull request adds a `save_separately` option to the `SerializeMainLoop` extension, which proved to be extremely important for smooth workflow. Unpickling the model is instant, whereas unpickling the log and the algorithm can take quite some time.


rizar:
A question arises: how should this example be tested? With all optimizations it takes about 80 seconds to compile (in addition to 12 seconds to take a gradient).  In the `fast_compile` mode it takes 6 + 12 seconds, which it still a lot. In addition it uses data we currently do not download for testing (billion words dataset).

bartvm:
I was thinking about that the other day... At one point you mentioned splitting datasets into a second repo, and maybe that's a good idea here as well. Basically, we will have two repos: Blocks, which contains the core, with a few very lightweight tests based on MNIST and a text file (for the recurrent stuff).

Then we have a second repo with all the dataset wrappers and examples, models from papers, etc. The advantage is that we don't end up with this situation like Pylearn2 where every paper that is written/reproduced with Pylearn2 ends up in the main codebase, which can really slow down development if you have to fix 5-year old models that no one really cares about anymore. We should still try to keep things working in this separate repo of course, but there will be less pressure, and the development cycle there can be a bit slower.

I'll try to add the billion words dataset to Travis at one point.

rizar:
Right, we should have this additional repo. I guess we should create it before the tutorial session, in order to give people a clear instruction: push your stuff there.

But I think we should have a functional demo for all major components, and this one covers the attention bricks. So eventually it should be tested automatically, just that 20 seconds is quite a lot for a test, but I can not help it.

blocks/extensions/saveload.py
```diff
@@ -26,6 +26,12 @@ class SerializeMainLoop(SimpleExtension):
     ----------
     path : str
         The destination path for pickling.
+    save_separately : list of str, optional
+        The list of the main loop's attributes to be pickled separately
+        to their own files. The pathes will be formed by adding
```
bartvm: *paths

Would it be a good idea to set sensible defaults for this i.e. didn't we say that the log and the dataset iterator should probably be saved separately>

blocks/extensions/saveload.py
```diff
@@ -26,6 +26,12 @@ class SerializeMainLoop(SimpleExtension):
...
+        the attribute name preceeded by a underscore before the
```
bartvm: *preceded

bartvm: *an underscore

bartvm:
I agree that we should have a functional demo, but maybe we can come up with something more light weight than this eventually, and then move this to the other repo.

On the topic of attention mechanism/sequence generators and the demo next week: #200

PR #198, opened 2015-02-04
Check data sources in DatasetEvaluator

### LINKED ISSUE #176: Informative error on missing input
The current one goes:
```
  File "deep.py", line 69, in <module>
    main_loop.run()
  File "/home/vanmerb/lisa/blocks/blocks/main_loop.py", line 88, in run
    while self._run_epoch():
  File "/home/vanmerb/lisa/blocks/blocks/main_loop.py", line 120, in _run_epoch
    self._run_extensions('before_epoch')
  File "/home/vanmerb/lisa/blocks/blocks/main_loop.py", line 145, in _run_extensions
    extension.dispatch(method_name, *args)
  File "/home/vanmerb/lisa/blocks/blocks/extensions/__init__.py", line 252, in dispatch
    self.do(callback_invoked, *(from_main_loop + tuple(arguments)))
  File "/home/vanmerb/lisa/blocks/blocks/extensions/monitoring.py", line 50, in do
    value_dict = self._evaluator.evaluate(self.data_stream)
  File "/home/vanmerb/lisa/blocks/blocks/monitoring/evaluators.py", line 206, in evaluate
    self.process_batch(batch)
  File "/home/vanmerb/lisa/blocks/blocks/monitoring/evaluators.py", line 181, in process_batch
    batch = dict_subset(batch, self.buffer_.input_names)
  File "/home/vanmerb/lisa/blocks/blocks/utils.py", line 300, in dict_subset
    result = [(key, extract(key)) for key in keys]
  File "/home/vanmerb/lisa/blocks/blocks/utils.py", line 297, in extract
    return dict_[k]
KeyError: None
```

rizar:
Fixes #176

### DIFF
--- a/blocks/monitoring/evaluators.py
+++ b/blocks/monitoring/evaluators.py
@@ -189,8 +189,6 @@ class DatasetEvaluator(object):
                 "Not all data sources required for monitoring were"
                 " provided. The list of required data sources:"
                 " {}.".format(self.buffer_.input_names)))
-
-
         if self._accumulate_fun is not None:
             self._accumulate_fun(**batch)
 
--- a/blocks/monitoring/evaluators.py
+++ b/blocks/monitoring/evaluators.py
@@ -185,10 +185,10 @@ class DatasetEvaluator(object):
         try:
             batch = dict_subset(batch, self.buffer_.input_names)
         except KeyError:
-            reraise_as(Exception(
+            reraise_as(
                 "Not all data sources required for monitoring were"
                 " provided. The list of required data sources:"
-                " {}.".format(self.buffer_.input_names)))
+                " {}.".format(self.buffer_.input_names))
         if self._accumulate_fun is not None:
             self._accumulate_fun(**batch)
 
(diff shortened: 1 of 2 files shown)


rizar:
Waiting for an LGTM.

blocks/monitoring/evaluators.py
```diff
@@ -181,7 +182,13 @@ def initialize_aggregators(self):
         self.buffer_.initialize_aggregators()
 
     def process_batch(self, batch):
-        batch = dict_subset(batch, self.buffer_.input_names)
+        try:
+            batch = dict_subset(batch, self.buffer_.input_names)
+        except KeyError:
+            reraise_as(Exception(
```
bartvm: You should be able to just provide a string to `reraise_as`, there's no reason to change a different error type here, and theoretically it also means a `KeyError` can be caught again somewhere higher in the call stack (unlikely in this case, but anyway).

tests/monitoring/test_evaluators.py
```diff
@@ -1,6 +1,7 @@
 import numpy
 
 import theano
+from nose.tools import assert_raises
```
bartvm: Maybe just import from `numpy.testing` instead... Bit silly maybe, because `numpy.testing` is mostly just a redirect to `nose.tools`, but we might as well keep things simple by always using `numpy.testing` for our test suites, instead of mixing `numpy.testing`, `nose.tools`, `nose2`, `unittest2`... Thinking about it, testing is a mess!

PR #199, opened 2015-02-04
Secure dill dump

rizar:
Addresses #154

### DIFF
--- a/blocks/main_loop.py
+++ b/blocks/main_loop.py
@@ -1,6 +1,7 @@
 """The event-based main loop of Blocks."""
 import signal
 import logging
+import traceback
 
 from blocks.log import TrainingLog
 from blocks.utils import unpack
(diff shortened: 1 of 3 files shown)


rizar:
Update: I added some logging for the case of error, still deferring writing to a different file.

rizar:
Waiting for an LGTM. That does not fully close #154, but significantly improves the situation.

blocks/main_loop.py
```diff
@@ -105,6 +106,10 @@ def run(self):
                 pass
         except TrainingFinish:
             self.log.current_row.training_finished = True
+        except Exception as e:
+            logger.error(traceback.format_exc(e))
+            logger.info("Still trying to finish gracefully")
```
bartvm: Maybe something slightly more informative, like: "An error occurred during the training. Attempting to run extensions before exiting..."

bartvm:
LGTM (might want to look at the slightly vague log entry)

blocks/main_loop.py
rizar: My idea was that they should have seen the ERROR right above. But okay, I will change that.

bartvm: Oh, yeah, but it was more about the "running extensions" part. It's not very clear what "finish gracefully" means.

Issue #200, opened 2015-02-04
Docs sequence generators, attention mechanism, and related bricks


bartvm:
I've spoken to at least 3-4 different people who tried to get started with that part of the library but got completely lost (which I can understand, it took me a long time to figure out what is going on, and I still need to sit down and figure out some of the details I think). However, it's a major draw of the library, considering that it's something for which e.g. Pylearn2 has zero support.

For example, the parallel bricks need to be updated with clear documentation and examples in the docstrings (i.e. doctests), avoiding the use of terminology such as "channels" and "mixing". We also need good documentation and an example of a very simple plug-and-play sequence generator. In most cases people just want something that takes an initial hidden state and spits out a series of softmaxes, so we should give them that with clear documentation and examples as well. It might also be a good idea to reoganise the docs to highlight the bricks that people will actually need, because many bricks like `AbstractFeedback` are very cryptic.

I think the key to documenting these kind of things is to avoid the use of as much jargon as possible, and give examples in the docstrings which are as small and self-contained as possible. That's a lot easier to digest than the sparsely commented and lengthy scripts we have in `examples/*`.

Lastly, on the long term it might be worth a look to see if there is anything in that code that we can refactor and/or simplify.

@rizar BTW: With the ICML deadline coming up in two days and an internship interview next week I'm incredibly swamped, so I'm afraid I won't be of much help for anything until later next week...

rizar:
While it definitely should be done, that's not going to be an easy one. Terminology is lacking for all this sequential (and parallel) stuff, because it has become so popular only very recently. I will try to clean it up and document gradually, but that will take some time and most probably will need another person's participation, somebody who is willing to understand.

rizar:
Sequence generators are now documented. Examples are still missing, but for those separate can be created when needed.

PR #204, opened 2015-02-05
Refactor the Parallel brick

rizar:
This is a first little step in resolving #200. Please look at this refactored `Parallel` brick and tell me if it is crystal clear now what it does or not.


bartvm:
Great, much better!

The only thing I would consider changing is making the `weights_init` a `Constant(1)` and adding `new_x.eval({x: [1 1]})` and `new_y.eval({y: [2 2 2]})` to the docstring, the output of which should then be `[2 2 2 2]` and `[6 6 6 6 6]`. Always nice to see numerical results with actual shapes.

bartvm:
Shall we merge this?

rizar:
I would like to finish the two other bricks first but got distracted trying to make my phoneme recognition work. But, okay, I will merge it today after checking that it does not break the attention example.

bartvm:
Sure, no rush, was just wondering whether you forgot about this PR or whether there was more to do.

rizar:
Turns out having numerical values in doctests is rather hard because we test with both float32 and float64. We could run doctests only with one of these two values or avoid numerical checks in them.

bartvm:
Ah, just add `# doctest: +ELLIPSIS` to the end of the last docstring command, and give as expected output `array([[ 6.,  6.,  6.,  6.,  6.]]...`

rizar:
Okay, I will finish this refactoring when I have time, but at least the root brick of the hierarchy is now documented.

PR #207, opened 2015-02-05
Refactor SimpleExtension, add `every_n_epochs`.

### LINKED ISSUE #203: every_n_epochs
Seems like `every_n_epochs` is missing. You could of course implement this with a counter in your object state but you could make the same argument for `every_n_batches`.

dwf:
Fixes #203 and also addresses @rizar's comments there.

### DIFF
--- a/blocks/extensions/__init__.py
+++ b/blocks/extensions/__init__.py
@@ -206,6 +206,7 @@ class SimpleExtension(TrainingExtension):
                                    predicate=predicate)
             else:
                 raise KeyError("Invalid condition: {}".format(key))
+        return self  # For chaining calls.
 
     def add_condition(self, callback_name, predicate=None, arguments=None):
         """Adds a condition under which a :meth:`do` is called.
--- a/blocks/extensions/__init__.py
+++ b/blocks/extensions/__init__.py
@@ -232,8 +232,11 @@ class SimpleExtension(TrainingExtension):
         if not arguments:
             arguments = []
         if not predicate:
-            predicate = lambda log: True
-        self._conditions.append((callback_name, predicate, arguments))
+            self._conditions.append((callback_name, lambda log: True,
+                                     arguments))
+        else:
+            self._conditions.append((callback_name, predicate,
+                                     arguments))
         return self
 
     @abstractmethod
--- a/blocks/extensions/__init__.py
+++ b/blocks/extensions/__init__.py
@@ -155,10 +155,13 @@ class SimpleExtension(TrainingExtension):
         self._conditions = []
         super_kwargs = {}
         trigger_keywords = self.BOOLEAN_TRIGGERS | self.INTEGER_TRIGGERS
+        conditions = {}
         for key, value in kwargs.items():
-            if key not in trigger_keywords:
+            if key in trigger_keywords:
+                conditions[key] = value
+            else:
                 super_kwargs[key] = value
-        self.set_conditions(**kwargs)
+        self.set_conditions(**conditions)
         super(SimpleExtension, self).__init__(**super_kwargs)
 
     def set_conditions(self, **kwargs):
@@ -201,6 +204,8 @@ class SimpleExtension(TrainingExtension):
                 predicate = predicate_factories.get(key, lambda: None)(value)
                 self.add_condition(conditions.get(key, key),
                                    predicate=predicate)
+            else:
+                raise KeyError("Invalid condition: {}".format(key))
 
     def add_condition(self, callback_name, predicate=None, arguments=None):
         """Adds a condition under which a :meth:`do` is called.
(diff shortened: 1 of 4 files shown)


blocks/extensions/__init__.py
```diff
@@ -141,36 +143,39 @@ class SimpleExtension(TrainingExtension):
...
+        # Freeze the keys as a list so that we can safely modify kwargs.
+        for key in list(kwargs.keys()):
+            if key in self.BOOLEAN_TRIGGERS and kwargs[key]:
+                self.add_condition(key, predicate=predicates.get(key, None))
+                del kwargs[key]
+            elif key in self.INTEGER_TRIGGERS:
+                if kwargs[key] is not None:
+                    factory = predicate_factories[key]
+                    predicate = factory(kwargs[key])
+                    self.add_condition(key, predicate=predicate)
```
bartvm: I don't know why it took me so long to realize: Here you're adding conditions like `after_n_epochs` and `every_n_batches`, but the condition should be `after_batch` (see e.g. line 212 of the removed code).

blocks/extensions/__init__.py
```diff
@@ -141,36 +143,39 @@ class SimpleExtension(TrainingExtension):
...
+                lambda log: log.status.iterations_done == n_batches,
+            'every_n_epochs': lambda n_epochs:
+                lambda log: log.status.epochs_done % n_epochs == 0,
+            'after_n_epochs': lambda n_epochs:
+                lambda log: log.status.epochs_done == n_epochs,
+        }
+        # Freeze the keys as a list so that we can safely modify kwargs.
+        for key in list(kwargs.keys()):
+            if key in self.BOOLEAN_TRIGGERS and kwargs[key]:
+                self.add_condition(key, predicate=predicates.get(key, None))
```
bartvm: Same here, there is no condition `after_every_batch`, the condition here is `after_batch`. In general, the keyword arguments and the conditions don't always match.

bartvm:
Woooh, where did that bot come from.

dwf:
So, the failing test uses a public API which is a bit of an outlier. It's only used in that one place and it's hard for me to think of a use-case for adding more trigger conditions post-creation when there isn't a way of removing or editing the conditions already in place. @rizar, do you think there's a need for these public methods?

blocks/extensions/__init__.py
rizar: By the way, do we want to keep it this way. Consider for instance the `after_every_batch` keyword argument and `after_batch` condition. On the one hand the first reads better, on the other hand using `after_batch` allows very easy matching with the main loop code.

I am inclined to switch to using the callback names verbatim, then all the callbacks can be automatically added to the conditions list.

rizar:
What do you think about moving all logic from `__init__` into a separate `set_conditions` method, that would first clear the conditions list and then process kwargs in the way you do here? That seems to solve all the problems: it gives us public API, it allows to remove the existing conditions. It could look like this:

```python 
extensions=[...,
    DataStreamMonitoring(data_stream, [cost]). # by default runs every epoch
       .set_conditions(before_first_epoch=True, after_every_epoch=True, every_n_batches=1000),
       # now runs also every 1000 batches
    ...]
```

As for editing the existing conditions, that would be rather more difficult to implement because the predicates are currently just lambds. But I think we can live without that, one will just have to specify all he needs in the `set_conditions` call.

dwf:
@rizar I think that sounds good. Should the constructor pass these kwargs to set_conditions also?

rizar:
Yes, I would make the constructor consist of essentially a single call of `set_conditions`. The only thing to be taken care of are the `kwargs` that should go to the base class. One can can have a private `_set_conditions` method that filters the unknown `kwargs` without crashing and returns them and a public counterpart, that would crash on every unknown keyword argument. The first one would be used in the constructor and the second one would serve general public. But those are implementations details, it is up to you.

blocks/extensions/__init__.py
```diff
@@ -141,36 +143,64 @@ class SimpleExtension(TrainingExtension):
...
-        if after_n_batches:
-            self.invoke_after_n_batches(after_n_batches)
-        if every_n_batches:
-            self.invoke_every_n_batches(every_n_batches)
+        super_kwargs = {}
+        trigger_keywords = self.BOOLEAN_TRIGGERS | self.INTEGER_TRIGGERS
+        for key, value in kwargs.items():
+            if key not in trigger_keywords:
+                super_kwargs[key] = value
+        self.set_conditions(**kwargs)
```
bartvm: Just one tiny thing before we finally merge this (taking awfully long!): I might be overlooking something, but right now all `kwargs` get passed on to `set_conditions`, regardless of whether they are in `trigger_keywords` or not, right? And if it's not in `trigger_keywords`, it get's silently ignored there. Wondering whether we should raise an error somewhere in those cases. Otherwise, if you mistype an argument your extension won't get triggered and you probably won't even realize, because you didn't get an error.

dwf: I was actually thinking the same thing. I can imagine this being horribly
annoying source of bugs in user code.

rizar:
It looks so much better now, thanks!

By the way, I do not get notifications from Github when a pull request is updated without any comment. If you want to me to push the button, write anything so that I get an email.

Issue #205, opened 2015-02-05
SGD with Momentum and Required Refactoring


rizar:
I wanted to implement SGD with momentum and realized that current `StepRule` interface is still very far from perfection. In order to implement it one has to add some updates that have gradients on their right hand side, but the `additional_updates` method from [here](https://github.com/bartvm/blocks/blob/master/blocks/algorithms/__init__.py#L255) has no parameters. There are a few possible options I see:

1. Add gradients a parameter to `additional_updates`. That would solve the problem, but potentially can lead to a lot of code duplication between this method and `compute_step`. 

2. Change the `compute_steps` return value, make it return both the step and the updates. Eliminates the problem with code duplication at the cost of a slightly more complex interface.

3. `compute_steps` determines the updates and stores them as the object's attributes, these attributes are returned by `additional_updates`. Does not require any change of interface, no code duplication, but adds a life-cycle to an otherwise stateless object.

4. A radical solution - follow the PyLearn2 way of doing things and have `StepRule` return updates only. I mention it for completeness, but it seems to be inappropriate because there is no way to chain different rules.

I am inclined to follow the path number (2), because it allows to make the `StepRule` stateless, which I think would be nice. All the shared variables can be created in the `compute_steps` call. The returned value of `compute_steps` would then deserve a name, e.g. `GradientStep`, and can also have additional attributes for monitoring purposes.

dwf:
Option 1 needn't lead to a lot of duplication with the right auxiliary methods, which could be provided by `StepRule` once it's clear what the useful primitives are. But I agree with you that 2 seems most attractive.

bartvm:
@rizar Was this fixed in #210?

rizar:
Sure!

PR #210, opened 2015-02-06
Momentum step rule and required changes

rizar:
Sorry for the mess: I pushed my branch to the wrong repository first.

Addresses #205

I did not make StepRule stateless - although that sounds nice, the code would become significantly more complex. Also is hard to think about anybody reusing a StepRule.


blocks/algorithms/__init__.py
```diff
@@ -245,22 +248,18 @@ def compute_steps(self, gradients):
 
         Returns
         -------
-        An ordered dictionary of the same form as `gradient`, with the
-        proposed steps as values.
+        A tuple (steps, updates). `steps` is ordered dictionary of the
```
dwf: should this tuple be in backquotes @bartvm?

bartvm: Backquotes by themselves actually don't do much in Sphinx I think (they just make the text italic). If you want something to be monospace, you need to use double backquotes. In this case proper format should make Sphinx figure things out by itself, but this isn't proper format actually (we've been sloppy about that in large parts of the library). It should be:

```
Returns
-------
steps : OrderedDict
    Ordered dictionary of the same form as :attr:`gradients`
updates : list
    A list of tuples, representing the updates to perform.
```
or alternatively
```
Returns
-------
OrderedDict
    Ordered dictionary of the same form as :attr:`gradients`
list
    A list of tuples, representing the updates to perform.
```
If you want a link to `OrderedDict` to show up you could use ```:class:`~collections.OrderedDict```` in the first case. In the second version (with types only, no names) you can't use this because of a bug in Sphinx (https://bitbucket.org/birkenfeld/sphinx-contrib/issue/82/napoleon-return-type-containing-colons-is)

bartvm: Just to correct/clarify a little bit: @dwf is right that functions, objects, etc. should be in backquotes according to the NumPy docs, but judging from the documentation, only when referring to in-library objects (I couldn't find any ``` `int` ``` or ``` `tuple ` ``` for example). In our case, we use ```:class:`.Brick` ``` in many of those cases though, so that it automatically links things. Backquotes should be used for referring to parameters though (which is something you can't reference in Sphinx).

PR #211, opened 2015-02-06
Convert log to pandas.DataFrame

rizar:
Our current log provides absolute no means for data analysis. And it shouldn't: there is pandas for it. Currently there is only one log and it is  very easy to convert to a `DataFrame` (see the code), but we should have more, and for some of them, e.g. for a one based on sqlite, there will be no easy strategy for conversion. That's why I reserve a place for a generic implementation that will use the log's public interface.

### DIFF
--- a/blocks/log.py
+++ b/blocks/log.py
@@ -3,6 +3,11 @@ from abc import ABCMeta, abstractmethod
 from collections import defaultdict
 
 from six import add_metaclass
+try:
+    from pandas import DataFrame
+    pandas_available = True
+except:
+    pandas_available = False
 
 
 @add_metaclass(ABCMeta)
@@ -203,6 +208,10 @@ class AbstractTrainingLog(object):
         if not isinstance(time, int) or time < 0:
             raise ValueError("time must be a positive integer")
 
+    def to_dataframe(log):
+        """Convert to :class:`.DataFrame`."""
+        raise NotImplementedError
+
 
 class TrainingStatus(AbstractTrainingStatus):
     """A simple training status."""
@@ -245,3 +254,18 @@ class TrainingLog(AbstractTrainingLog):
 
     def get_status(self):
         return self._status
+
+    def to_dataframe(self):
+        return DataFrame.from_dict(self._storage, orient='index')
+
+
+def to_dataframe(log):
+    """Convert a log into a :class:`.DataFrame`."""
+    if not pandas_available:
+        raise Exception("The pandas library is not found. You can"
+                        " install it with pip.")
+    try:
+        return log.to_dataframe()
+    except NotImplementedError:
+        # TODO: some default method of converting a log into a dataframe
+        raise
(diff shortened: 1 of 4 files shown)


rizar:
@bartvm , why do tests pass without adding `pandas` to the list of dependencies? That's rather surprising for me..

bartvm:
Looks like it's installed as a dependency of bokeh.

rizar:
Should I add `pandas` to `extras_require` or should I just merge?

bartvm:
You should add it either to `install_requires` or add it to `extra_requires` under a `log` heading or something. I have no problem with adding it to `install_requires`, considering how common a library it is, but else you can just add a new `log` key in the dictionary I guess.

blocks/log.py
```diff
@@ -245,3 +254,18 @@ def __iter__(self):
...
         return self._status
+
+    def to_dataframe(self):
+        return DataFrame.from_dict(self._storage, orient='index')
+
+
+def to_dataframe(log):
+    """Convert a log into a :class:`.DataFrame`."""
+    if not pandas_available:
+        raise Exception("The pandas library is not found. You can"
```
bartvm: Just raise an `ImportError` here. (In general, I don't think there are many cases where you want to raise `Exception`, one of its more specific subclasses is almost always better.)

blocks/log.py
```diff
@@ -245,3 +254,18 @@ def __iter__(self):
 
     def get_status(self):
         return self._status
+
+    def to_dataframe(self):
+        return DataFrame.from_dict(self._storage, orient='index')
+
+
+def to_dataframe(log):
```
bartvm: Why not just make this a method of the `Log` class? It acts directly on the class instance, so seems like a perfect candidate.

blocks/log.py
```diff
@@ -245,3 +254,18 @@ def __iter__(self):
...
+    """Convert a log into a :class:`.DataFrame`."""
```
bartvm: You'll need to add `pandas` to InterSphinx by adding `'pandas': 'http://pandas.pydata.org/pandas-docs/stable/'` to the `intersphinx_mapping` dictionary in `docs/conf.py`. After that, you can link to the pandas documentation by using 
```
:class:`~pandas.DataFrame`
```

blocks/log.py
rizar: No really good reason for that, but how would keep the header that checks pandas availability? Would have to implement that with a publlc-private method pair, like `to_dataframe` and `_to_dataframe`. But I guess it is not such a big deal, because if the person is brave enough to write a new log class, he should be able to work with such patterns.

bartvm: Yeah, agreed.

rizar:
Anything else I should take care of?

blocks/log.py
```diff
@@ -203,6 +208,16 @@ def _check_time(self, time):
         if not isinstance(time, int) or time < 0:
             raise ValueError("time must be a positive integer")
 
+    def to_dataframe(log):
```
bartvm: As Scrutinizer points out, this method and the next one probably need `self` as their first argument.

rizar:
Oops, that was an accident!

PR #213, opened 2015-02-07
Tests

bartvm:
Bits and pieces:

* Use `ContainerDataset` for `test_cache` so that it can run without MNIST
* Introduce a scheme for shuffled batches
* Add functionality to `ConstantScheme`, allowing it to return batch sizes until they sum to a given `num_examples`
* Use `BytesIO` instead of files where possible
* ~~Skip tests instead of failing if files already exist~~
* Use `tempfile` to create temporary files instead of creating files in the working directory
* Make the tests on Travis more verbose, so that we can see at which test it stalls

### DIFF
--- a/blocks/datasets/schemes.py
+++ b/blocks/datasets/schemes.py
@@ -74,9 +74,9 @@ class ConstantScheme(BatchSizeScheme):
         The size of the batch to return.
     num_examples : int, optional
         If given, the request iterator will return `batch_size` until the
-        sum reaches `num_exampes`. Note that this means that the last batch
-        size returned could be smaller than `batch_size`. If you want to
-        ensure all batches are of equal size, then pass `times` equal to
+        sum reaches `num_exam;pes`. Note that this means that the last
+        batch size returned could be smaller than `batch_size`. If you want
+        to ensure all batches are of equal size, then pass `times` equal to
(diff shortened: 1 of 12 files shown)


bartvm:
@rizar I also fixed the joining of paths in `dump.py`. In general it's not a good idea to use `{}/{}.format(dir, file)` because on different operation systems the directory separator might be different (e.g. it's `\` on Windows). It's best to use `os.path.join` for this.

blocks/datasets/schemes.py
```diff
@@ -102,27 +135,48 @@ class SequentialScheme(BatchScheme):
...
+        return BatchIterator(self.num_examples, self.batch_size,
+                             shuffled=True, rng=self.rng)
 
 
-class SequentialIterator(six.Iterator):
-    def __init__(self, num_examples, batch_size):
-        self.num_examples = num_examples
+class BatchIterator(six.Iterator):
+    def __init__(self, num_examples, batch_size, shuffled=False, rng=None):
+        indices = numpy.arange(num_examples)
```
rizar: This will hurt if this iterator is ever used with a large out-of-memory dataset. It is rather unlikely that such datasets will support indexing at all, but when I said that at NIPS you had a counter-example in mind.

PR #215, opened 2015-02-07
Explicitly raise error on empty epoch iterator.

### LINKED ISSUE #201: Check at Least One Batch in an Epoch
Subject.

dwf:
Fixes #201.

Not sure if this needs a test, or how to test for it exactly with the main loop's error-catching in the way. I'd just call this method but it is soft-private.

### DIFF
--- a/blocks/main_loop.py
+++ b/blocks/main_loop.py
@@ -4,7 +4,7 @@ import logging
 import traceback
 
 from blocks.log import TrainingLog
-from blocks.utils import unpack
+from blocks.utils import reraise_as, unpack
 
 logger = logging.getLogger(__name__)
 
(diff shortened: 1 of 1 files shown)


blocks/main_loop.py
```diff
@@ -135,6 +135,7 @@ def find_extension(self, name):
     def _run_epoch(self):
         if not self.status._epoch_started:
             try:
+                self._received_first_batch = False
```
rizar: I would like to move this attribute to `self.log.status`. In Blocks we have yet another serialization mechanism called "dumping" (see `blocks/dump.py`) for the situations when training was started in one environment, but for some reason has to be continued in other one. The most common example is switching from GPU to CPU. The idea is to save only the model parameters and the log and load them later in a constructed from scratch main loop object. In order for the main loop to catch up it has to store everything in the log, that's why `_epoch_started` for instance is there. If `_received_first_batch` is an attribute of the main loop, resumption from dump is no longer possible.

So if you guys agree with this reasoning, let's keep `MainLoop` clean from attributes other than `model`, `log`, `algorithm`, `data_stream` and `epoch_iterator` and only add new ones when they can not be hosted by the `log`.

PR #221, opened 2015-02-07
WIP: Convolutional and max pooling bricks

bartvm:
Parts of this have been based of @dmitriy-serdyuk's work for cats and dogs: https://github.com/dmitriy-serdyuk/cats_vs_dogs/blob/master/cats_vs_dogs/bricks.py

First time working with convolutional networks, so figuring things out as I go along. This introduces the basic convolution brick and a max pooling brick.

- [x] Add tests
- [x] Combine convolution, pooling and activation into a single brick
- [ ] Combine these bricks into a sequence with an easy interface
- [x] Make Convolution accept input shape as argument
- [x] Add biases
- [ ] Move `Flattener` and document it


dwf:
This opens a can of worms with respect to the convolution implementation to use, and how to structure the API surrounding that. Your options in this respect are:

* Theano's `conv2d`: spatial domain convolution, bc01 ordering, dog slow on GPU
* Theano's `conv2d_fft`: faster, bc01 ordering, but memory intensive
* cuda-convnet Op wrappers in pylearn2. Fast on Fermi era cards, but not really that fast of Kepler. c01b ordering. GPU only.
* `theano.sandbox.cudnn`: I don't know what ordering (the docs don't specify), supposedly quite fast but requiring compute capability 3.0 or higher, convolution gradient non-deterministic (! arrgghhh), GPU only.

There's also GPU max pooling in (at least) vanilla theano, cudnn, and cuda_convnet. 

There's also cuda-convnet2 and various kernels from Caffe and Torch that would need wrappers written, but the above are the options as of today. Features like strided convolution, rectangular vs. square input, and various border modes are supported or not across different backends.

It would be nice to have an API that can support the use of each simply. My thinking would be that all but those that use a non-bc01 data ordering could be handled by Brick constructor arguments.

bartvm:
But as far as I understand, the standard `conv2d` op can be replaced by the CuDNN, CorrMM (Caffe implementation), and FFT implementations if you enable the right optimizations, correct? That's why I figured I would use this one, and basically let people chose using the optimization flags. The CuDNN and FFT implementations also seem quite a bit faster than the cuda-convnet ones (going of https://github.com/soumith/convnet-benchmarks).

bartvm:
Theano reference about optimization flags here btw: http://deeplearning.net/software/theano/library/tensor/nnet/conv.html#module-conv

blocks/bricks/conv.py
```diff
@@ -0,0 +1,163 @@
...
+        output : :class:`~tensor.TensorVariable`
+            A tensor with the same number of dimensions as `input_`, but
+            with the last two dimensions downsampled.
+
+        """
+        output = max_pool_2d(input_, self.pooling_size, st=self.step)
+        return output
+
+
+class ConvolutionalLayer(Sequence, Initializable):
```
bartvm: @dwf Any naming suggestions for this? I really have no idea what to call the convolution + nonlinearity + pooling group.

dmitriy-serdyuk: @bartvm When I was implementing this I thought that it would be a good idea that `ConvolutionalLayer` could compute the output dimension if the input dimension is provided. Do you know an easy way to do it?

dwf: The wrappers of cuda-convnet in pylearn2 have the necessary logic for that
in the C code. I can dig it out and send a pull request to this branch if
you like.

bartvm: I don't think there's an easy way besides just doing all the messy calculations that take into account all the possible settings (the border mode, the pool stride, the filter stride, etc.)

But actually the main reason I didn't implement this is because it's not clear to me what the "dimension" of a 2D input or output is. I guess there are basically two options: `height * width` or `(height, width)`. Then the question is whether they should share the interface with `FeedForward` (which so far has always assumed that `dim` is an integer). Curious to hear what @dwf and @rizar have to say about that.

bartvm: *Cross-post; I always take too long to write my replies.

dmitriy-serdyuk: I found: theano ops have `getOutputShape` and `out_shape` for convolution and pooling respectively.

rizar: Sorry, I am terrible busy right now preparing for a talk, but I should 
find some time tomorrow to comment on this.

dwf: From our cuda-convnet wrappers (for the forward pass):

```python
        div_ms_y = "((imgSizeY - 2*paddingStart - filter_rows) / moduleStride)"
        div_ms_x = "((imgSizeX - 2*paddingStart - filter_cols) / moduleStride)"
        mod_ms_y = "((imgSizeY - 2*paddingStart - filter_rows) % moduleStride)"
        mod_ms_x = "((imgSizeX - 2*paddingStart - filter_cols) % moduleStride)"
        target_rows = "%s + ((%s > 0) ? 1 : 0) + 1" % (div_ms_y, mod_ms_y)
        target_cols = "%s + ((%s > 0) ? 1 : 0) + 1" % (div_ms_x, mod_ms_x)
```
paddingStart is a negative number or 0. moduleStride is the step (cuda-convnet requires that the x stride and y stride be equal). 

To get the input buffer's image dimensions from the output buffer and filters dimensions: 
```
        #target_rows = "(hidActsSizeY + filter_rows + 2 * paddingStart) * moduleStride - 1"
        #target_cols = "(hidActsSizeX + filter_cols + 2 * paddingStart) * moduleStride - 1"
```

harmdevries89:
I quickly reviewed your code, and it looks good to me, except for the bias terms that seems to be missing in the convolutional layer. What's the best way to add this to the code? (I'm sorry but I'm still quite unexperienced with github.)

bartvm:
Fork this repository (top right) and then checkout this branch (git checkout conv). Then make the changes you want and push to your own fork (git add/commit/push). After that you should be able to go your fork (or this repo) and click on the green button on the index page (create a pull request). You then have two dropdown menus at the top; make sure that as a base you choose this branch (bartvm/conv) while comparing with your own (harmdevries89/conv), and then create a pull request. I can merge that then so that it becomes part of this PR. It's really just like making a normal PR, except that you probably need to change the base (which is master by default) to this branch.

harmdevries89:
Ok, thanks! I have created a PR. I saw you've created a new role FILTERS for the weights of a conv layer. For now, I have just BIASES for the biases of the conv layer.

bartvm:
Merging this so that we can start using it, but I'll make an issue for the outstanding tasks.

PR #222, opened 2015-02-07
Don't creat in-memory indices if not necessary

bartvm:


blocks/datasets/schemes.py
```diff
@@ -161,10 +161,11 @@ def get_request_iterator(self):
 
 class BatchIterator(six.Iterator):
     def __init__(self, num_examples, batch_size, shuffled=False, rng=None):
-        indices = numpy.arange(num_examples)
+        indices = range(num_examples)
```
dwf: Use `six.range` which will get you `range` on Py3 and `xrange` on Py2.

bartvm: The problem is that you can't slice in `xrange`, so using `six.moves.range` will require a few more changes. I can have a look though, because creating the list in memory in Python 2 still isn't great.

PR #225, opened 2015-02-08
WIP: Refactor parallel2

rizar:
I continue refactoring of `parallel.py`. The most hated `Mixer` became `Merge`, which was a linguistic challenge for me, and I am still not quite satisfied with the current naming scheme. Please tell me if you have better suggestions.

- [x] Doctest for `Merge`
- [x] Change names of the `Parallel`'s children. 
- [ ] Somehow check that this commit does not break the sequence generator

The last item of TODO is quite non-trivial by the way. Unit tests are a nice way to check that simple bricks work, but for proper testing big components like the sequence generator another mechanism is needed. At the company I used to work at people checked outputs of large scripts for this purpose, and those gold-standard output were part of the repo. Guess we also have to go this way. @bartvm, would you mind loading a moderate-sized model to Travis for testing purposes?

### DIFF
--- a/blocks/bricks/parallel.py
+++ b/blocks/bricks/parallel.py
@@ -183,11 +183,11 @@ class Distribute(Fork):
     >>> x = tensor.matrix('x')
     >>> y = tensor.matrix('y')
     >>> z = tensor.matrix('z')
-    >>> merge = Distribute(target_names=['x', 'y'], source_name='z',
-    ...                    target_dims=dict(x=2, y=3), source_dim=3,
-    ...                    weights_init=Constant(2))
-    >>> merge.initialize()
-    >>> new_x, new_y = merge.apply(x=x, y=y, z=z)
+    >>> distribute = Distribute(target_names=['x', 'y'], source_name='z',
+    ...                         target_dims=dict(x=2, y=3), source_dim=3,
+    ...                         weights_init=Constant(2))
+    >>> distribute.initialize()
+    >>> new_x, new_y = distribute.apply(x=x, y=y, z=z)
     >>> new_x.eval({x: [[2, 2]], z: [[1, 1, 1]]}) # doctest: +ELLIPSIS
     array([[ 8.,  8.]]...
     >>> new_y.eval({y: [[1, 1, 1]], z: [[1, 1, 1]]}) # doctest: +ELLIPSIS
--- a/blocks/bricks/parallel.py
+++ b/blocks/bricks/parallel.py
@@ -105,7 +105,7 @@ class Fork(Parallel):
     the copies are applied to the input to produce different outputs.
 
     A typical usecase for this brick is to produce inputs for gates
-    of a gated recurrent bricks, such as
+    of gated recurrent bricks, such as
     :class:`~blocks.bricks.GatedRecurrent`.
 
     >>> from theano import tensor
@@ -127,9 +127,6 @@ class Fork(Parallel):
         Names of the outputs to produce.
     input_dim : int
         The input dimension.
-    prototype : instance of :class:`.Brick`
-        A prototype for the input-to-fork transformations. A copy will be
-        created for every output channel.
 
     Attributes
     ----------
@@ -210,6 +207,10 @@ class Merge(Parallel):
         values are dimensions.
     merged_dim : dict
         The dimension of the merged input.
+
+    Notes
+    -----
+    See :class:`.Initializable` for initialization parameters.
 
     """
     @lazy
--- a/blocks/bricks/parallel.py
+++ b/blocks/bricks/parallel.py
@@ -19,8 +19,7 @@ class Parallel(Initializable):
 
     >>> from theano import tensor
     >>> from blocks.initialization import Constant
-    >>> x = tensor.matrix('x')
-    >>> y = tensor.matrix('y')
+    >>> x, y = tensor.matrix('x'), tensor.matrix('y')
     >>> parallel = Parallel(
     ...     input_names=['x', 'y'],
     ...     input_dims=dict(x=2, y=3), output_dims=dict(x=4, y=5),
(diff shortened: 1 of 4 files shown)


rizar:
Okay, I am sort of done here. Somebody please check that it is now possible to understand what bricks from `parallel.py` actually do. And @bartvm, would you mind using a trained model for a test in Travis?

bartvm:
Mmm... So I completely agree of course that unit testing (white-box testing), although essential, doesn't replace the need for system testing/black-box testing. I'm just wondering how to best structure these tests. A couple of reasons not to run them on Travis in the main repo:

* If we're running entire experiments, it won't be long before Travis starts taking hours to complete its runs. This can make for frustratingly slow development. Keep in mind that there is a bit of an avalanche effect here: Long builds cause subsequent builds to start later, in which time people's commits will have triggered more builds, and before you know it you're waiting 3 hours to hear from Travis.
* It will end up blurring the line between the core repo and the "contrib" repo we talked about, where we would host datasets and examples of models.
* Travis doesn't run GPU tests, which we'll need soon (e.g. to test GPU-only convolutional layers).

Pylearn2 technically speaking has a buildbot (and so does Theano) which runs all the tests that either don't fit on Travis, need a GPU, require datasets, etc. on a computer at LISA. The downside is that no one ever cares to look at its logs, so things break without anyone ever noticing. However, I think the idea of having a separate build that does more extensive testing could be a good idea.

So although I don't have all the details figured out, at all, I think it's worth considering a setup where the core repo only hosts unit tests and a second, more black-box set of tests, is run as part of this "contrib" repo. These tests would make sure that our example scripts run from start to end, and that our results are reproducible. It would be great if this test suite would be easy to run outside of Travis on a computer with a GPU and a time limit high enough to actually train some models to convergence.

The only question then becomes how we use/relay the statuses from this Travis build of the contrib repo and the buildbot. We could either do this manually, or write some sort of script that triggers builds and reports the results.

rizar:
I agree that we should keep running time of Travis tests reasonably small. But I think the main factor for deciding where to host a test, at Travis or somewhere else, should be not its type (system test or unit test) but the time it takes to run it. Of course, there is no way to run training tests at Travis, but its quite possible to test the costs or the samples produced by the model. This is what I would like to do now to ensure that we are on the safe side.

As for core-repo and contrib: a couple of examples covering the main components should go to the core repo, and there can be no simpler example of sequence generation with an attention mechanism than the one reversing words, except for the data: it could be synthesized.

I fully support the initiative of testing training procedures on a separate computer with GPU. I do not think such checks can be performed for every pull request though, but we could at least run them on daily basis. Would be nice to have the results as an additional badge on Github repo page. I anticipate that finding computational resources for such checks might be challenging.

dwf:
As a relative Bricks neophyte I will give my first-read opinion as soon as
I'm at a computer.

bartvm:
Although the tests will be harder to write, I don't see why in principle the `SequenceGenerator` can't be unit tested with a series of artificial inputs. That should be more than enough to know if a particular internal change will break the current interface or change the outputs.

I'm actually starting to wonder whether we really should have examples in the core repo at all, for a variety of reasons:

* We clearly separate the framework from its usage: The core repo contains the actual framework, and unit tests to ensure that everything behaves as expected. The contrib contains examples of what can be done with the actual framework. In our case the contrib repo doubles as a place where we do our black-box testing. This is a completely natural split: You won't find any lengthy examples of what can be done with Theano, NumPy, etc. either in their repositories either.
* It fixes the installation process. Keep in mind that if you install Blocks, the examples scripts won't be copied (current behaviour) or they will be hidden somewhere deep in the file system e.g. `/usr/lib/python2.7/site-packages/blocks/examples`. By splitting the examples from the core, the installation becomes clear: The core is something you install and don't think about anymore, it's a Python package like any other. This will allow us to eventually put Blocks on PyPI as well. Now, if you want to use particular models as a starting point to play with, use the dataset wrappers provided by people, etc. you clone (not install!) the contrib repo and you can hack away at the code from there.

We could run our daily builds on the same computer as Pylearn2 and Theano I guess, I can talk with Fred about that in due time.

rizar:
Any volunteers to skim over new code of `parallel.py`?

@pbrakel, I remember you found code of this file very hard to understand, would you like check that it became better?

pbrakel:
The doc strings are much clearer now. I'm still not sure if the name merge really capture what that brick does though. Perhaps it's more accurate to say that it 'distributes' the output of a single transformation and adds it to the 'input' or 'changed' variables. The thing is that the name 'merge' kind of suggests that it's the opposite of fork while it is more like a fork of which the output is added to a list of output variables (with of course the convenience that the dimensions are correct). This makes me wonder if it wouldn't be more natural to just have a fork and do the pairwise summation over the two lists of variables whenever this is required (or make a brick for that). It's also not directly clear from the description that addition is carried out by default. Finally, I was wondering why Parallel needs both input names and a dict with the names and the dimensions. Isn't the latter making the former redundant?

Anyway, the examples are very clear and shows what the bricks do even if the naming is a bit tricky in my opinion.

rizar:
>  Finally, I was wondering why Parallel needs both input names and a dict with the names and the 
> dimensions. Isn't the latter making the former redundant?

This is required for lazy initialization. The names are required to create children bricks, that is in the constructor, and the dimension are only required at the allocation stage. 
 
>This makes me wonder if it wouldn't be more natural to just have a fork and do the pairwise summation 
> over the two lists of variables whenever this is required (or make a brick for that). 

This is a good question, and I will answer it in detail, also to make it clear for myself. I want to have a clear interface for the cases when information from one source should be merged into others. The example situation is when a weighted average of a context (i.e. glimpse) is merged into inputs of a recurrent transition.
I would not like to hard-code that summation is the last step of doing this. Who knows, maybe later we will later realize that multiplying is better (hmm, I should try that!).  So this poorly named `Merge` brick serves in fact two purposes: 1) it defines an interface of a brick of this kind 2) proposes a merging procedure, the one with summation as the last operation.

I will try a `Distribute` now and update the PR.

rizar:
@pbrakel , is it better now?

pbrakel:
@rizar, to me it is but I might be a bit biased now :). I guess the general issue is that it's just very difficult to come up with the words for these kind of things that are the most intuitive to people. Hopefully other people share their opinions as well.

rizar:
Okay, at we two are happy now. I will refactor the `AttentionTransition` next and ask you to read it as well.

PR #226, opened 2015-02-08
Using toolz here and there

bartvm:


rizar:
It definitely improves readability, but as I said in the issue, it looks a bit weird after you argued againts `map` and `filter`.

PR #229, opened 2015-02-09
Added dimension computation

dmitriy-serdyuk:
It should be useful for lazy initialization.

### DIFF
--- a/blocks/bricks/conv.py
+++ b/blocks/bricks/conv.py
@@ -157,19 +157,22 @@ class ConvolutionalLayer(Sequence, Initializable):
                                  self.pooling.apply],
             **kwargs)
 
-    @property
-    def output_dim(self):
-        if hasattr(self, '_output_dim'):
+    def get_dim(self, name):
+        if name == 'input_':
+            return self.input_dim
+        if name == 'output':
+            if hasattr(self, '_output_dim'):
+                return self._output_dim
+            conv_out_dim = ConvOp.getOutputShape(self.input_dim[1:],
+                                                 self.convolution.filter_size,
+                                                 self.convolution.step,
+                                                 self.convolution.border_mode)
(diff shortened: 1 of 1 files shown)


blocks/bricks/conv.py
```diff
@@ -148,13 +148,26 @@ class ConvolutionalLayer(Sequence, Initializable):
...
+                                         num_channels, conv_step, border_mode)
+        self.pooling = MaxPooling(pooling_size, pooling_step)
         super(ConvolutionalLayer, self).__init__(
-            application_methods=[convolution.apply, activation, pooling.apply],
+            application_methods=[self.convolution.apply, activation,
+                                 self.pooling.apply],
             **kwargs)
 
+    @property
+    def output_dim(self):
```
rizar: In fact this is typically done by `get_dim` function in the rest of the library.

dmitriy-serdyuk: I was confused by `MLP` class. Is it done differently there?

rizar: In the MLP class `output_dim` is there to set dimension, but the main interface for asking them is `get_dim` (see `Brick` documentation). There is certain philosophy behind all this and I swear to explain it tomorrow! :)

blocks/bricks/conv.py
```diff
@@ -147,14 +147,33 @@ class ConvolutionalLayer(Sequence, Initializable):
...
-            application_methods=[convolution.apply, activation, pooling.apply],
+            application_methods=[self.convolution.apply, activation,
+                                 self.pooling.apply],
             **kwargs)
 
+    def get_dim(self, name):
+        if name == 'input_':
+            return self.input_dim
+        if name == 'output':
+            if hasattr(self, '_output_dim'):
```
bartvm: I think it's okay to repeat this computation, no need to store it in a private variable.

PR #232, opened 2015-02-09
More verbose exception from GradientDescent when cost inputs and data sources don't match

jbornschein:
With this patch a mismatch between the cost functions input and the datasets sources will result in a error message like this:

```
ValueError: The names of the input variables of your computation graph (features) must correspond to the data sources (features, targets). Maybe your Dataset supports the 'sources' keyword to explicitly select which sources you want?
```

I guess this touches a few policy question: 

1) Do you want more verbose exceptions at the cost of more complex error handling code?
2) Writing unit tests for these branches seems overkill to me, but not writing them will decrease test coverage.

### DIFF
--- a/blocks/algorithms/__init__.py
+++ b/blocks/algorithms/__init__.py
@@ -203,10 +203,10 @@ class GradientDescent(DifferentiableCostMinimizer):
             raise ValueError("The names of the input variables of your"
                              " computation graph (%s) must correspond to the"
                              " data sources (%s). Maybe your Dataset supports"
-                             " the 'sources' keyword to explicitly select which"
-                             " data sources you want?" %
-                                (", ".join([v.name for v in self.inputs]),
-                                 ", ".join(batch.keys())))
+                             " the 'sources' keyword to explicitly select"
+                             " which data sources you want?" %
+                             (", ".join([v.name for v in self.inputs]),
+                              ", ".join(batch.keys())))
         ordered_batch = [batch[v.name] for v in self.inputs]
         self._function(*ordered_batch)
 
--- a/blocks/algorithms/__init__.py
+++ b/blocks/algorithms/__init__.py
@@ -201,8 +201,12 @@ class GradientDescent(DifferentiableCostMinimizer):
     def process_batch(self, batch):
         if not set(batch.keys()) == set([v.name for v in self.inputs]):
             raise ValueError("The names of the input variables of your"
-                             " computation graph must correspond to the"
-                             " data sources.")
+                             " computation graph (%s) must correspond to the"
+                             " data sources (%s). Maybe your Dataset supports"
+                             " the 'sources' keyword to explicitly select which"
+                             " data sources you want?" %
+                                (", ".join([v.name for v in self.inputs]),
+                                 ", ".join(batch.keys())))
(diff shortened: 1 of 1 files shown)


bartvm:
All datasets should support the `sources` keyword argument, so you can change the phrasing to e.g. ```Maybe you can use the `sources` keyword argument to....```

Regarding (2), I agree there's no need for unit tests. The code coverage doesn't need to be a 100%, it's more there to be an incentive for if we start neglecting tests over time.

I think policy question (1) is a very interesting one, which I've thought about and am very conflicted about. On the one hand, I feel that error messages should be terse and not try to make assumptions about what went wrong. Error messages can also quickly end up taking more space than the actual source code, which makes code harder to read. On the other hand, informative and useful error messages are the perfect way to explain to a user what likely went wrong, and where they can start with fixing it. Nothing more frustrating than having to reverse engineer code to figure out a mistake that the error message could have easily told you about.

My preferred solution so far is the idea of using short error messages in the code, but concatenating it with longer error messages which are defined at the end of a file (or maybe just above/below the class) e.g.

```python
raise ValueError("mismatch variable names and data sources" +
                  mismatch_error.format(sources=sources, variables=variables,
                                        dataset=dataset))
```
```python
mismatch_error = """

Blocks tried to match the sources ({sources}) of the dataset ({dataset.class.__name__})
to the names of the Theano variables ({variables}), but failed to do so. If you want to
train on a subset of the sources that your dataset provides, pass the `sources` keyword
argument to its constructor."""
```

This gives (a) the user a nice and informative error message; (b) keeps the code clean while still giving people who are reading code a hint about what the error is; and (c) makes it a lot easier to type error messages because you can use the full 79 characters instead of awkwardly concatenating strings over multiple lines. @rizar @dwf Either of you have objections to putting this in the developer guidelines?

bartvm:
I'll merge this, but still waiting to hear from the others how/if we should make this the standard way of giving error messages. One problem with this approach that I just realized is that triple quotes will maintain the newlines, preventing the terminal from wrapping/reflowing the text properly, so maybe this would be better:
```python
mismatch_error = """

Blocks tried to match the sources ({sources}) of the dataset ({dataset.class.__name__}) \
to the names of the Theano variables ({variables}), but failed to do so. If you want to \
train on a subset of the sources that your dataset provides, pass the `sources` keyword \
argument to its constructor."""
```
or alternatively
```python
mismatch_error = (""
""
"Blocks tried to match the sources ({sources}) of the dataset ({dataset.class.__name__}) "
"to the names of the Theano variables ({variables}), but failed to do so. If you want to "
"train on a subset of the sources that your dataset provides, pass the `sources` keyword "
"argument to its constructor.")
```

jbornschein:
Yes, especially because we don't know the length of e.g. {sources} or {variables}.

Te me the version with backslash-escaped newlines looks quite a bit cleaner than the version with individually quoted and concatenated strings.

rizar:
I like the proposed way of writing error messages: with a big separate message cut into lines by backslahes. Nothing to add so far, we will see if this policy causes any trouble.

bartvm:
Cool.

I agree, and think these kind of lengthy expressions should probably be reserved for situations like these anyway, where a mistake is likely to be common and has a clear solution that wouldn't be obvious from a short message.

PR #233, opened 2015-02-09
WIP: blocks-contrib

bartvm:
This PR would separate blocks into two separate repositories

### blocks

The core of the framework
* The actual framework
* Does not rely on external data.

---

* Unit tests
* Except for development, possible to install directly through `pip` (e.g. no examples or other code that is expected to be read by users)
* Does not rely on external data when testing

### [blocks-contrib](https://github.com/bartvm/blocks-contrib)

To keep the installation process simple I'm using namespace packages. This means that things in the `blocks-contrib` repository can be made available under `blocks.contrib` when installed.

* Dataset implementations e.g. MNIST, BillionWords, anything else
* Simple toy models that users can play with to get familiar with the library
* Reference implementations of people who publish models with Blocks
* Models from papers reproduced

---

* Unit tests
* Black-box tests
    * Example scripts tested from start to end
    * Ensure that reference implementations of models perform as expected
* Preferred method of installation is by cloning/downloading raw files so that user has direct access to examples script
    * Maintains the possibility of installing directly and making components accessible in the `blocks.contrib` namespace if user wants easy access to user-contributed code

Clearly, both repositories will still be closely connected. We should aim to run a full test suite on `blocks-contrib` at least daily with the latest master from `blocks`. This should let us know if we broke anything that our unit tests didn't catch, whether numerical results have changed, etc.

On the other hand, if we need to or want to, we can make changes to the core repo that break tests in `blocks-contrib` without halting development on the core framework. This will be especially beneficial once we start accumulating large numbers of models in `blocks-contrib` over the years. Although we will still want to test these models and endeavour to keep them working, we don't want to stall every PR to the framework by the need to fix every single unused/archived model written in the past X years.

### Todo

- [ ] Run Travis and Scrutinizer on `blocks-contrib`
- [ ] Make `test_serialization.py` no longer dependent on MNIST
- [ ] Update documentation
- [ ] Move BillionWords dataset


dwf:
Hmm. Namespace packages are bad mojo in my experience. Main reason that sklearn is no longer scikits.learn and skimage is no longer scikits.image...

bartvm:
What can go wrong with them? Doesn't e.g. Django still use them? I just
tested things on my system and it worked perfectly. Our use case is pretty
simple, because the dependency only goes one way, maybe that makes things
easier?

dwf:
http://mail.scipy.org/pipermail/scipy-dev/2011-July/016357.html lists all
of sklearn's reasons for it

dwf:
Some of those obviously don't apply, but arguably some of them do.

bartvm:
Mm, interesting. I'm not entirely convinced these cons outweigh the benefits in our case though. Mostly because in the case of `scikit` they were dealing with separately coordinated projects, which could conceivably be installed in any order. In our case we can coordinate perfectly between the two packages, and they will always be installed in the same order.

(a) The `__init__.py` of `blocks-contrib` can stay empty; that should make merging it trivial.
(b) This would be an issue I guess if `blocks` is installed by the admin and you want to install `blocks-contrib` without admin rights. Also, we can reconsider the namespace package if we ever become part of EPD or Debian package manager :p 
(c) Don't really see this to be much of an issue, is it?
(d) Can't imagine this being such a slowdown that it becomes problematic for the size of our framework.
(e) What would we shorten it to? `blocks_contrib` is just as long, and I can't really think of an abbreviation that makes sense...

What would be the alternative? Installing a separate module called `blocks_contrib` is an option. Are there others?

rizar:
Before entering the discussion about the proper name: shouldn't we first launch blocks-contrib, establish regular testing there, and only then remove all this things from `blocks`? That would be a safer way: the code would be all the time covered by system tests, which are very important.

bartvm:
I was just planning on making the switch in one go, so that the two repos don't end up diverging. But yeah, testing needs to be set up there first.

Main issue I ran into so far is the documentation: I would like to be able to refer to `blocks-contrib` and use its datasets and models in e.g. tutorials, but that means that we still need to install `blocks-contrib` on Travis for the main repo if we want to keep running doctests.

rizar:
Maybe we keep a few scripts in `blocks` as we planned before?  That would solve your problem with docs: if a tutorial for something is in `blocks`, then we should have all required code, data and models, if necessary, in  `blocks`.


I am not sure we should do everything exactly is renowned packages like `matplotlib` do, because those are quite different from ours. They do not have this problem that the model is a black box and its inside can not be checked by `assert`'s. That's why in my view system tests have higher importance for us than for typical packages.

bartvm:
Closing following discussion at https://github.com/bartvm/blocks/issues/517#issuecomment-84522386

PR #234, opened 2015-02-10
Added LSTM brick

pbrakel:
I wrote a simple LSTM brick.The code is based on the optimized version Mohammad wrote for the speech project. I guess it's a bit odd that the input sequence dimensionality needs to be four times the state dimensionality but I couldn't directly think of a cleaner way to do it.

### DIFF
--- a/blocks/bricks/recurrent.py
+++ b/blocks/bricks/recurrent.py
@@ -305,9 +305,9 @@ class LSTM(BaseRecurrent, Initializable):
     u"""Long Short Term Memory.
 
     Every unit of an LSTM is equipped with input, forget and output gates.
-    This implementation aims to do as many computations in parallel as possible
-    and expects the last dimension of the input to be four times the output
-    dimension.
+    This implementation aims to do as many computations in parallel as
+    possible and expects the last dimension of the input to be four times
+    the output dimension.
 
     Parameters
     ----------
--- a/blocks/bricks/recurrent.py
+++ b/blocks/bricks/recurrent.py
@@ -379,15 +379,8 @@ class LSTM(BaseRecurrent, Initializable):
             Next cell activations of the network.
 
         """
-        # Unfortunately, I'm not aware of a general way to slice the last
-        # dimension of a theano tensor.
         def slice_last(x, no):
-            if x.ndim == 3:
-                return x[:, :, no*self.dim: (no+1)*self.dim]
-            elif x.ndim == 2:
-                return x[:, no*self.dim: (no+1)*self.dim]
-            elif x.ndim == 1:
-                return x[no*self.dim: (no+1)*self.dim]
+            return x.T[no*self.dim: (no+1)*self.dim].T
 
         activation = tensor.dot(states, self.W_state) + inputs + self.biases
         cellW_cell = tensor.dot(cells, self.W_cell)
--- a/blocks/bricks/recurrent.py
+++ b/blocks/bricks/recurrent.py
@@ -11,6 +11,7 @@ from theano import tensor, Variable
 from blocks.bricks import Initializable, Identity, Sigmoid
 from blocks.bricks.base import Application, application, Brick, lazy
 from blocks.initialization import NdarrayInitialization
+from blocks.roles import add_role, WEIGHTS, BIASES
 from blocks.utils import (pack, shared_floatx_zeros, dict_union,
                           is_shared_variable)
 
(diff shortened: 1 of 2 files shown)


blocks/bricks/recurrent.py
```diff
@@ -300,6 +301,116 @@ def apply(self, input_=None, state=None, mask=None):
...
+        Returns
+        -------
+        states : :class:`~tensor.TensorVariable`
+            Next states of the network.
+        cells : :class:`~tensor.TensorVariable`
+            Next cell activations of the network.
+
+        """
+        # Unfortunately, I'm not aware of a general way to slice the last
+        # dimension of a theano tensor.
```
bartvm: Sadly Theano still doesn't support the ellipsis syntax from NumPy, but `x.T[no * self.dim:(no + 1) * self.dim].T` should work, no?

dwf: Alternatively, you can construct a tuple of `slice(None)` objects.

```
def fake_leading_ellipsis(ndim, slice_obj):
    return tuple([slice(None)] * (ndim - 1) + [slice_obj])
x[fake_leading_ellipsis(x.ndim, slice(no * self.dim, (no + 1) * self.dim)]

Comment (dwf) [inline blocks/bricks/recurrent.py]: Or perhaps nicer:

```python
def slice_last(x, slice):
    leading = [slice(None)] * (x.ndim - 1)
    return x[tuple(leading + [slice])]

slice_last(x, slice(no * self.dim, (no + 1) * self.dim))
```

dwf: Incidentally, I've opened an issue about the ellipsis: https://github.com/Theano/Theano/issues/2487

pbrakel: I fixed it with the transpose for now. Not sure why I didn't think of that myself :)

blocks/bricks/recurrent.py
```diff
@@ -300,6 +301,109 @@ def apply(self, input_=None, state=None, mask=None):
...
+    """
+    @lazy
+    def __init__(self, dim, **kwargs):
+        super(LSTM, self).__init__(**kwargs)
+        self.dim = dim
+
+    def get_dim(self, name):
+        if name == 'inputs':
+            return self.dim * 4
+        else:
```
bartvm: This is a bit dangerous, because it means that if a person requests a non-existing `name` he will still get a return value. It's best to check for specific names here (and I think @rizar has always given `mask` a dimension of ) e.g. 
```python
if name == 'mask':
    return 0
if name in ['states', 'cells']:
    return self.dim
return super(LSTM, self).get_dim(name)  # This raises an informative error
```

blocks/bricks/recurrent.py
```diff
@@ -300,6 +301,109 @@ def apply(self, input_=None, state=None, mask=None):
...
+    the output dimension.
+
+    Parameters
+    ----------
+    dim : int
+        The dimension of the hidden state.
+
+    Notes
+    -----
+    See :class:`.Initializable` for initialization parameters.
```
rizar: Could you please put a reference?

blocks/bricks/recurrent.py
```diff
@@ -300,6 +301,109 @@ def apply(self, input_=None, state=None, mask=None):
...
+        def slice_last(x, no):
+            return x.T[no*self.dim: (no+1)*self.dim].T
+
+        activation = tensor.dot(states, self.W_state) + inputs + self.biases
+        cellW_cell = tensor.dot(cells, self.W_cell)
+        in_gate = tensor.nnet.sigmoid(slice_last(activation, 0) +
+                                      slice_last(cellW_cell, 0))
+        forget_gate = tensor.nnet.sigmoid(slice_last(activation, 1) +
+                                          slice_last(cellW_cell, 1))
+        next_cells = forget_gate * cells + \
```
rizar: What do you guys think about not  using line breaks in Blocks?

```
The preferred way of wrapping long lines is by using Python's implied line continuation inside 
parentheses, brackets and braces. Long lines can be broken over multiple lines by wrapping
expressions in parentheses. These should be used in preference to using a backslash 
for line continuation.
```
from http://legacy.python.org/dev/peps/pep-0008/

pbrakel: I agree with you and normally try not to use this unless I feel I'm forcing it too much with the parentheses. I guess this case is a bit on the border for me but I'll change it.

rizar:
I did not check the formulas, but apart from the small remarks I put, I would be happy to merge it.

blocks/bricks/recurrent.py
bartvm: That's just part of PEP8, so it's already part of the guidelines (I just missed this one). Note however that PEP8 makes exceptions for a few cases (such as the `with` statement), and that it's my personal preference for `if` statements (on which PEP8 explicitly does not take a stance, but `pep8` complains if you indent the second line too much). So this should be okay:

```python
with open('file1.txt') as f, \
     open('file2.txt') as g:
    pass

if (a and b) or \
        (c and d):
    pass
```

blocks/bricks/recurrent.py
pbrakel: I'm currently asking @mohammadpz where he got the formulas and will also mention that I used his code for this.

mohammadpz: Hi there,
Here's the references:
[1] Gers, Felix A., Nicol N. Schraudolph, and Jürgen Schmidhuber. "Learning precise timing with LSTM recurrent networks." The Journal of Machine Learning Research 3 (2003): 115-143.
[2] Graves, Alex. "Generating sequences with recurrent neural networks." arXiv preprint arXiv:1308.0850 (2013).

pbrakel: I changed it but I'm not sure how the references will appear in the docs. What is the easiest way to check that?

bartvm: There's no easy way that I know of. You can build the documentation
yourself if you have sphinx and the RTD theme installed. It's quite
straightforward but I haven't added the instructions to the docs yet. For
now just assume it's okay, we'll fix it if they don't turn out right.

dwf: If it's a generic LSTM implementation then it should cite Hochreiter et al.
Bricks that implement more specific variants could cite other papers,
though.

pbrakel: I just double checked the computations to see how specific this variant is and apparently it uses peephole connections from the cells to the gates. Somehow the output gates get the current cell values while the other gates receive the cell values from the previous timestep. This is how the LSTM is defined in the paper by Graves but I'm not sure how to name this variant. It's the same architecture that he used for speech recognition as well. Anyway, I saw there is a mistake in the implementation because some of the matrices involved should be diagonal (probably best implemented using element-wise vector multiplications) so I should fix that first before this gets merged. I'm considering to cite Hochreiter et al. as well because that paper coined the name of the architecture.

rizar: I think most importantly we should cite the paper where exactly this version is used, which is the Graves one. But in addition we can also cite the source where the term comes from, Hochreiter et al. in this case.

blocks/bricks/recurrent.py
rizar: I see why an exception should be made for the first case, but why don't you like

```python
if ((a and b) or
        (c and d)):
    pass
```
?

bartvm: Because I could swear that PyLint complained about, but I just tried and it actually doesn't seem to. Maybe it used to? Or maybe I'm just crazy. Anyway, that was my only reason for not using it.

rizar: I bet you are not :)

Okay, then we are just strictly adhering PEP8 at this point.

rizar:
Okay, I see the references are there. Are you done with fixing bugs, can I press the button?

pbrakel:
You can press the button :)

rizar:
Cool, thanks, very nice to have a good LSTM implementation. I wish somebody sped up the `GatedRecurrent` in a similar way...

mohammadpz:
Hey guys,

I have implemented GRU in a same way as LSTM (if you used my
implementation). I can add it after getting familiar with Blocks.

rizar:
That would be nice!

7hil:
I am wondering why the dim of the input should be 4 times of the output's. Is it because this code assumes that we've done the (Wx+b) part of LSTM outside of this brick? So I guess I should apply a Linear brick to x before applying this LSTM brick?

rizar:
Exactly! In Blocks only really recurrent part of the computation is in the recurrent brick. Everything that can be moved out, including biases and linear transformations, is moved out.

7hil:
Thanks for the timely reply!

I think this feature is different with most people's expection of LSTM layer. This brick is more like a Op but not a Layer. And there is no document mentions this, which makes it harder to use this brick.

rizar:
http://blocks.readthedocs.org/en/latest/rnn.html explains the usual RNN, and it mentions that the linear transformation has to be done separately.

Having only the core functionality in the recurrent bricks makes their code concise and ensures modularity. We could however add all-batteries-included recurrent bricks, renaming those that we have now into `CoreSimpleRecurrent`, `CoreGatedRecurrent` and `CoreLSTM`. You can open a discussion by creating an issue that proposes such change.

7hil:
Now I finally notice and understand what it means by saying in http://blocks.readthedocs.org/en/latest/rnn.html: 
> Note that in order to double the input we had to apply a bricks.Linear brick to x, even though ht=f(Vht−1+Wxt+b) is what is usually thought of as the RNN equation.

I agree this provides more flexibility and the naming is just fine. It will be better to have more documentation about this design.

Issue #228, opened 2015-02-09
Add data set loading/configuration to tutorial


bartvm:
Right now the tutorial uses MNIST but doesn't explain which files to download, where to put them, and how to set the data path.

arasmus:
I just set up the thing, so I'll share here what were the steps required. Maybe it's helpful. I don't know whether you have plans for automatic data download scripts (which travis could use too) or simply instructing users to do something like this:

```
echo "data_path: /Users/rasmusa/code/datasets" > ~/.blocksrc
mkdir -p /Users/rasmusa/code/datasets/mnist
cd /Users/rasmusa/code/datasets/mnist
curl -O http://yann.lecun.com/exdb/mnist/train-images-idx3-ubyte.gz \
             -O http://yann.lecun.com/exdb/mnist/train-labels-idx1-ubyte.gz \
             -O http://yann.lecun.com/exdb/mnist/t10k-images-idx3-ubyte.gz \
             -O http://yann.lecun.com/exdb/mnist/t10k-labels-idx1-ubyte.gz
gunzip *
```

bartvm:
We discussed automatic data downloading scripts, which resulted in #105. I feel that we shouldn't make them part of the core library, to make sure we stay small, but if we could do it with a minimal dependency on `skdata` then it's worth considering.

rizar:
Yep, that's the current policy. But instruction for getting dataa should be part of the tutorials and/or documentation.

bartvm:
Yeah, I consider that part of this issue. I should probably change the title, but it's in the description :)

bartvm:
Added in 04914ba306c09d7eff200490d67b39ddd2ec50ce

PR #238, opened 2015-02-10
Fix typos in tutorial.rst

vdumoulin:
In addition to the proposed fixes, I think the paragraph containing the sentence `MNIST defines two data sources: 'features' and 'targets'` could be expanded a little to make it clearer. The phrasing makes it unclear whether MNIST refers to the actual data or to the dataset class.

A quick explanation of what's a data source and why the variable names must match the data source of a given dataset would help.

I can come up with something and make it part of this pull request if you'd like, or I can leave that for a future pull request.


bartvm:
Great, thanks!

Yeah, the section about data sets needs to be expanded. It also needs to deal with setting the data path (see #228). A future pull request would be very welcome of course :)

PR #239, opened 2015-02-11
Implement ADADELTA as a StepRule

vdumoulin:


vdumoulin:
One thing which I'm not clear on is if we'd like to enforce some sort of sanity check on the value of the decay parameter (_i.e._ being in [0, 1]).

bartvm:
I would say yes, never hurts to check. Perhaps I should make that clearer in the [developer guidelines](http://blocks.readthedocs.org/en/latest/development/index.html#validating-function-arguments) as well. I'd say we should encourage these kind of checks (whether values are sane), but discourage type checking.

One tiny comment regarding the citation: To make it consistent and render nicely in the documentation you could use Sphinx references as [done here](https://github.com/bartvm/blocks/blob/master/blocks/bricks/__init__.py#L262-L267). Otherwise it looks good to me!

vdumoulin:
@bartvm Good to know about citation formatting. I did the requested changes, plus I put the `nose` import at the right place after reading the developer guidelines more carefully.

Regarding `nose`, I see you're using `nose2` for Travis-CI; is it still okay to mix in `nose` imports in the code?

bartvm:
`numpy.testing` is actually partly a wrapper of `nose` (and hence `nose` is an indirect dependency), but you could also use `numpy.testing.assert_raises`, which is what I've done so far.

vdumoulin:
For the sake of uniformity, I'll use that as well.

bartvm:
Great, I'll merge when Travis gives you its blessing!

Issue #241, opened 2015-02-11
Developer installation command wrong for required_extras


arasmus:
http://blocks.readthedocs.org/en/latest/setup.html#development says that one should use
```
$ pip install --upgrade -e git+git://github.com/user/blocks.git#egg=blocks[tests] --src=$HOME
```
but the extras is actually called ```test``` not ```tests```, see

https://github.com/bartvm/blocks/blob/master/setup.py#L40

arasmus:
Actually, for developers, you might want to use [test,plot,docs].

PR #242, opened 2015-02-11
Fix installation instructions for developers

arasmus:
There are two fixes. Can take both or only the first one
- Fix required_extra that fixes the command so that it doesn't fail anymore
- Another, that changes the default protocol to ssh (requiring user to have setup their ssh key on github) and allowing pushing. Also performs minor style change to highlight which part of the command is supposed to be changed.

I opened issue #241 for the former.


rizar:
Thanks!

Issue #240, opened 2015-02-11
Momentum step rule implementation is counter-intuitive


vdumoulin:
Following a conversation with @bartvm I now understand why `Momentum` is implemented the way it is (it's meant to be chained with `SteepestDescent`, so that `Momentum`'s `self.momentum * velocity + gradient` translates to something like `momentum * velocity - learning_rate * gradient` when used in conjunction with `SteepestDescent`).

It has the unfortunate effect of obfuscating the implementation (which can be solved by proper documentation), and it also requires the user to be very careful about how things are chained:

* `Momentum` cannot be used alone, or it would lead to counter-intuitive results.
* For some step rules like ADADELTA, there is no such thing as a learning rate and using it after `SteepestDescent` doesn't make much sense.

On the other hand, there may be some advantages to having `Momentum` be composable with other step rules.

My proposition would be to rename `Momentum` to make it clear that it's not intended to be used alone and to wrap `SteepestDescent + Momentum` in a `Momentum` step rule which chains them together.

rizar:
I see your point, but am not sure if I agree with it. In fact the full name of the algorithm is "gradient descent _with_ momentum", which means that having momentum as an additional component makes sense. In PyLearn2 that was a stand-alone learning rule and maybe that may cause confusions, but so far I am inclined to think that Blocks way is better.

I agree that it is weird that `Momentum` can not be used alone. I think the reason is that multiplication by minus one should be moved out to the `GradientDescent` class, and #243 does it. After #243 using `Momentum` alone would imply learning rate equal to one.

ADADELTA is sometimes used with gradient rescaling, @janchorowski did it when working with TIMIT. Maybe using `Adadelta` and `SteepestDescent` in a chain looks weird, then we can add a clone of `SteepestDescent` called `Rescale` that reads better.

vdumoulin:
I wouldn't revert back to having monolithic step rules either, but I think that having a set of pre-baked combinations also makes sense.

Here's an example. In order to get the correct behaviour for `SteepestDescent + Momentum`, `SteepestDescent` needs to come first and `Momentum` comes after. In implementing RMSProp (see #246 ) the way `Momentum` is implemented, I realized that the correct chaining order in that case is to use RMSProp first, and _then_ to use `SteepestDescent`.

[...]

PR #243, opened 2015-02-11
Move minus to the GradientDescent

rizar:
Inspired by the concerns of @vdumoulin from #240, I propose the following modification: in step rules instead of the something to be added to the parameters return something to be subtracted. The reason that to conform with semantic of "the step to be added" every step rule should have a multiplication by -1, but this would make chaining impossible. 

I suggest to keep calling the subtracted entity a "step", because it is just  very convenient. If we call it something like `antistep`, that would be more to the point but less readable.

### DIFF
--- a/blocks/algorithms/__init__.py
+++ b/blocks/algorithms/__init__.py
@@ -154,16 +154,18 @@ class GradientDescent(DifferentiableCostMinimizer):
             for param in params:
                 param -= steps[param]
 
+    Note, that the step is _subtracted, not added_! This is done in order
+    to make step rule chaining possible.
+
     Parameters
     ----------
     step_rule : instance of :class:`StepRule`, optional
         An object encapsulating most of the algorithm's logic. Its
         `compute_steps` method is called to get Theano expression for
-        steps. Note, that the step rule
-        might have a state, e.g. to remember a weighted sum of gradients
-        from previous steps like it is done in gradient descent with
-        momentum. If ``None``, an instance of :class:`SteepestDescent` is
-        created.
+        steps.  Note, that the step rule might have a state, e.g. to
+        remember a weighted sum of gradients from previous steps like it is
+        done in gradient descent with momentum. If ``None``, an instance of
+        :class:`SteepestDescent` is created.
     gradients : dict, optional
         A dictionary mapping a parameter to an expression for the cost's
         gradient with respect to the parameter. If ``None``, the gradient
@@ -210,8 +212,8 @@ class GradientDescent(DifferentiableCostMinimizer):
         if not set(batch.keys()) == set([v.name for v in self.inputs]):
             raise ValueError("mismatch of variable names and data sources" +
                              variable_mismatch_error.format(
-                                sources=batch.keys(),
-                                variables=[v.name for v in self.inputs]))
+                                 sources=batch.keys(),
+                                 variables=[v.name for v in self.inputs]))
         ordered_batch = [batch[v.name] for v in self.inputs]
         self._function(*ordered_batch)
 
@@ -298,7 +300,7 @@ class SteepestDescent(StepRule):
 
 
 class Momentum(StepRule):
-    """Accumulates the step with exponential discount.
+    """Accumulates gradients with exponential discount.
 
     Parameters
     ----------
--- a/blocks/algorithms/__init__.py
+++ b/blocks/algorithms/__init__.py
@@ -152,7 +152,7 @@ class GradientDescent(DifferentiableCostMinimizer):
         for batch in data:
             steps = step_rule.compute_steps(params, gradients_wr_params)
             for param in params:
-                param += steps[param]
+                param -= steps[param]
 
     Parameters
     ----------
@@ -201,7 +201,7 @@ class GradientDescent(DifferentiableCostMinimizer):
         # the parameters were given. Keep it like that to ensure
         # reproducibility.
         for param in self.params:
-            all_updates.append((param, param + self.steps[param]))
+            all_updates.append((param, param - self.steps[param]))
         all_updates += self.step_rule_updates
         self._function = theano.function(self.inputs, [], updates=all_updates)
         logger.info("The training algorithm is initialized")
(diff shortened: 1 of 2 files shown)


rizar:
This is quite an important change (and fast to review) change, and I want your opinions guys.

@vdumoulin , @bartvm , @dwf

vdumoulin:
@rizar Could you give me an example of why the current situation makes chaining impossible and why changing the sign solves that? It's not clear to me yet.

bartvm:
Haven't had time to look at the code yet (sorry, @rizar, crazy day) but at first sight I'm a bit confused as well. I don't understand why adding a `-` in `GradientDescent` would be any different from adding a `-` to each step individually (which maintains the semantics of the word "step").

rizar:
If I don't miss anything `SteepestDescent` and `Adadelta` currently can not be chained. Though it seems useless, Jan Chorowski found it useful to prescale the gradient before passing it to Adadelta. The reason I think they can not be chained now is that both of them expect the gradient and produce a step in a direction sort-of-opposite to it. When we combine them, the step will be done in a direction along which the function increases, not decreases as we would definitely prefer.

@bartvm , because this would ensure that minus is added only once, and not in each learning rule, which would work for odd number of learning rules and crash for even!

bartvm:
I think we have two separate objectives when it comes to these design decisions:

1. A user-friendly interface for quick prototyping and 90% of use cases.
2. Flexibility to extend and construct more complicated algorithms.

We shouldn't confuse the two! That is, to expect every possible algorithm you can think of to work without the need to write a single line of code. We should however make it easy to write this code. I've always considered it part of the philosophy of the library that it doesn't shield you from writing bits and pieces of code where needed. What that means IMHO concerning this discussion:

1. We want a `Momentum` class that works out of the box, and likewise for `AdaDelta`. I believe this also means that a `Step` should be an actual step, instead of changing all the semantics to deal with a few corner cases. Common use cases come first.
2. We provide a `CompositeRule` and an `AddMomentum` (or whatever we end up calling it) that allows us to add momentum to steps other than the common `SteepestDescent` option. However, for the uncommon case where people want to combine `AdaDelta` with another step they have a multiple options that don't take more than a few lines of code, for example:

```python
class FlipRule(StepRule):
    def compute_step(self, param, gradient):
        return - gradient, []

rule=CompositeRule([AdaDelta(...), FlipRule(), SteepestDescent(...)])
```
```python
class FlipCompositeRule(StepRule):
    def compute_steps(self, gradients):
        result = gradients
        updates = []
        for rule in self.components:
            result, more_updates = rule.compute_steps(result)
            result = - result
            updates += more_updates
        return - result, updates

rule=FlipCompositeRule([AdaDelta(...), SteepestDescent(...)])
```

I prefer this approach: For the vast majority of users the semantics make sense (step is a step, momentum trains with momentum) and that's where it ends, while for those that want to try out new things, it takes 3 lines of code (`FlipRule`) to combine `AdaDelta` with `SteepestDescent`.

rizar:
> We want a `Momentum` class that works out of the box, and likewise for `AdaDelta`. I believe this also 
> means that a step should be an actual step, instead of changing all the semantics to deal with a few corner > cases.

Why? As I said before, I do not mind having a standalone `Momentum` rule that combines both scaling and accumulation of the gradient. This has nothing to do with the design decision on what exactly is the return value of step rule.

The change I propose will only affect those people who dare to write their own step rules. They will have to read documentation in `blocks/algorithms/__init__.py` and remember once and forever that step is something to be subtracted in Blocks. And tell me that those people who will ever have to do that are not cornercases!

On the other hand, people who just play with making random combination of rules are the majority. They don't even care how it works, but as of now we have weird principles of which rules can be chained with which and which can not. `GradientClipping` does not add minus, `Adadelta` does. If we accept this change of mine, it will be possible to use any rule in combination with any other rule, without even reading documentation. Isn't that user friendly? Save them from thinking that `Momentum` and `GradientClipping` are compatible, but `AddMomentum` and `GradientClipping` are not, because gradient is not reversed.

blocks/algorithms/__init__.py
```diff
@@ -276,7 +278,7 @@ def compute_steps(self, gradients):
 
 
 class SteepestDescent(StepRule):
-    """A step in the direction opposite to the gradient.
+    """A step in the direction proportional to the gradient.
```
bartvm: This gets a bit confusing now...

If we want this terminology to make sense than I would argue that all `StepRule`s no longer take `gradient` as an input (because it's no longer receives a gradient, it's either the negative gradient, or something completely different in the case of chaining). The `SteepestDescent` rule would then be `ScaleRule` or something, because that is all it does (scaling the step it receives, the first step being hardcoded to be the negative gradient by the `GradientDescent` class).

rizar: The whole purpose of this pull request is to ensure that the input is a gradient (clipped, accumulated), not negative gradient.

Renaming `SteepestDescent`  into `Scale` is a reasonable thing to do.

bartvm: Right, sorry, that's what you get for trying to read code while also following class.

Anyway, the argument still holds: `gradient` is confusing, because it's no longer the gradient if it's part of a chain. So I would rename the argument to `step` everywhere, the first step being the gradient as passed by the `GradientDescent` class.

bartvm:
Okay, just had a look at the code and change seems reasonable to me, but we do need to change terminology here and there to have it all make sense I think.

rizar:
Well, that's true that the `gradient` argument is not always a gradient  any more. But should we change its name? I do not know: the rules are typically formulated in terms of what they do with the gradient, and if we replace the `gradient` argument  with `step` argument we will lose in terms of readability. 

And also I would like to merge these changes first and think about refactoring after that, because this change influences other PRs (#246).

bartvm:
I really think we gain in terms of readability. I find it very confusing that `gradient` actually isn't the gradient, but some step that could be the output of a complex step like AdaDelta. If you look at `StepRule`s as things that modify a given step, and output a new one, the argument `step` makes good sense to me.

bartvm:
That said, if you don't have much time, feel free to merge so that we can finish up #246

vdumoulin:
I'd be in favour of merging now (so I can adapt #246 right away) and then refactor the names in a subsequent pull request.

rizar:
I created #266 for accumulation of concerns regarding step rule terminology.

PR #246, opened 2015-02-11
Implement RMSProp as a StepRule

vdumoulin:
There are two new classes introduced: `RMSPropChainable` and `RMSProp`.

The former is implemented in a way that allows it to chain step rules, much like `Momentum` is implemented right now, while the latter wraps `RMSPropChainable` and `SteepestDescent` to produce a step rule that corresponds to what users would expect out of the box.

The same principle could also be applied to `Momentum` in order to make things a bit more natural while retaining modularity.

### DIFF
--- a/blocks/algorithms/__init__.py
+++ b/blocks/algorithms/__init__.py
@@ -406,6 +406,11 @@ class RMSPropChainable(StepRule):
     step rule, _e.g._ :class:`SteepestDescent`. For an
     all-batteries-included experience, look at :class:`RMSProp`.
 
+    In general, this step rule should be used _before_ other step rules,
+    because it has normalization properties that may undo their work.
+    For instance, it should be applied first when used in conjunction
+    with :class:`SteepestDescent`.
+
     For more information, see [RMSProp]_.
 
     .. [RMSProp] Geoff Hinton, *Neural Networks for Machine Learning*,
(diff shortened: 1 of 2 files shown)


bartvm:
I think @rizar might disagree, but I'm inclined to agree with the distinction between `Momentum` and `MomentumChainable`.

In common usage you would often say you trained something "with momentum", in which case it's simply assumed that you trained using the steepest descent direction. There are few cases where you would use a learning rate fixed at 1, or a different descent direction.

Moreover, there's a pretty strong case to be made for removing the redundancy of *always* having to add `SteepestDescent` followed by `Momentum`. This would be so common, it's worth making it the more accessible `Momentum` the default, while keeping the more powerful and modular chained version, `MomentumChainable`, as an additional option for people who want to use more complex algorithms.

dwf:
+1

bartvm:
So this looks good to me, but we should probably finish up with #243 and possible rebase/adapt this PR before merging.

rizar:
My preference would be to call the basic rule something like `PureRMSProp` instead of `RMSPropChainable`, because `RMSProp`  is also absolutely chainable now.

vdumoulin:
I agree with you, but I'm not sure I like the term "pure" (I'm being picky, I know). How about `BasicRMSProp`? Or my original name `RMSPropBase`?

rizar:
`BasicRMSProp` is just perfect!

rizar:
May I merge it or do you want to do something else with it?

vdumoulin:
I feel like I'm done for that pull request.

Issue #245, opened 2015-02-11
Serialization problem with Adadelta


arasmus:
I get
```RuntimeError: maximum recursion depth exceeded```

I have no idea what's going on. I tried to change mnist example to use Adadelta, but serialization starts failing. I tracked the problem a little bit, but couldn't figure it out easily. See this commit for repro instructions:

https://github.com/arasmus/blocks/commit/f7c2bd97a9c870d13389d769b5d563f3743a1914

dwf:
CC @vdumoulin

rizar:
There is even issue about that: #152.

For some reason, dill, the package we use for serialization, uses crazy deep recursion. So try simply increasing the recursion limit until we find a more general solution for that.

vdumoulin:
@rizar If I understand correctly, `AdaDelta` exhibits the problem with recursion limit but is not directly causing it? I'd be curious to know if there's something I did while writing `AdaDelta` which should be done in another way in order to prevent these things from happening.

bartvm:
There's nothing I'm aware of that you could easily change to prevent this from happening. The problem is basically that a large, annotated Theano graph creates a deep object hierarchy with many circular references. When `cPickle` or `dill` tries to serialize this, it quickly runs into recursion errors. (Note that `mlp.py` in Pylearn2 also increases the recursion limit to deal with this.) It's annoying, but it's not actually a problem; the recursion limit in Python is set pretty low, to make sure that infinite recursions don't take too long to break out of. Simply increasing it solves all cases I've seen so far.

Issue #152, opened 2015-01-25
Increase recursion limit by default


bartvm:
For my model, I quickly ran into the recursion limit issue (just as with the markov chain model). It might be a good idea to increase the recursion limit by default.

rizar:
Sound like a reasonable thing to do. The worst thing is that `blocks-continue` called without changing the maximum recursion depth can screw your model up. Btw, I think we should pickle to a temporary file and then perform replacement, because pickling errors are so frequent and their consequences are so disastrous.

bartvm:
How can it screw your model up? If the file is opened in read-only mode, I really don't see how it could screw up a correctly pickled model?

bartvm:
Oh wait, I guess that you mean it tries to overwrite the old, correct model and then fails when doing so, got it! Never had that, because I always move my models before resuming training.

bartvm:
Fixed in #259.

PR #247, opened 2015-02-11
Change recursion limit in main loop

rizar:
Let's just change it in the main loop so that a person first likes Blocks and only then stumbles upon this shit.

#245, #152.


bartvm:
+1 @dwf's comment. I guess we can do `sys.setrecursionlimit` in scripts, but we'll need to be more careful in library imports.

rizar:
For technical reason moving to #259.

PR #250, opened 2015-02-11
Added extension for parameter adaptation during the training process

dmitriy-serdyuk:


dwf:
This really needs a better name, I think.

dmitriy-serdyuk:
Any suggestions? `ModifyParameter`? As I understand, one should use a verb as an extension name.

blocks/extensions/training.py
```diff
@@ -0,0 +1,28 @@
...
+
+    Applies function to compute the new value of a shared parameter each
+    iteration.
+
+    This class can be used to adapt over the training process parameters like
+    learning rate, momentum, etc.
+
+    Parameters
+    ----------
+    parameter : shared variable to be adjusted
```
bartvm: Just a note about the docstring formatting, it's:
```
parameter_name : type
    A lengthy explanation of what this
    variable does
```
So in this case:
```
parameter : :class:`~tensor.TensorSharedVariable`
    The shared variable to be adjusted.
function : callable
    Function that takes the total number of examples seen (``int``)
    as an input, and outputs a numeric value to which the
    given shared variable will be set.
```

blocks/extensions/training.py
```diff
@@ -0,0 +1,28 @@
...
+    """
+    def __init__(self, parameter, function, **kwargs):
+        super(AdjustParameter, self).__init__(**kwargs)
+        self.learning_rate = parameter
+        self.function = function
+        self.num_examples = 0
+
+    def after_batch(self, batch):
+        self.num_examples += batch.values()[0].shape[0]
+        self.learning_rate.set_value(self.function(self.num_examples))
```
bartvm: Perhaps this function could also take the current value of `self.learning_rate` as an input? That we `function` can just be `def decay(current_value, num_examples_seen): current_value * 0.9` if you want to do something batch-wise instead of example-wise.

bartvm:
Methods and functions should have verbs or verb phrases as names, but classes don't have to. So something like `SharedVariableAdjustor`, `SharedVariableUpdater`, `SharedVariableModifier`, etc. would also work.

blocks/extensions/training.py
```diff
@@ -0,0 +1,28 @@
...
+    Parameters
+    ----------
+    parameter : shared variable to be adjusted
+    function : function which input number of aggregated examples and outputs
+               the parameter value
+
+    """
+    def __init__(self, parameter, function, **kwargs):
+        super(AdjustParameter, self).__init__(**kwargs)
+        self.learning_rate = parameter
```
bartvm: If you want this class to be generic enough for updating any shared variable (which I think is a good idea), then this attribute should probably be `self.parameter` and not `self.learning_rate`.

blocks/extensions/training.py
```diff
@@ -0,0 +1,28 @@
...
+        self.function = function
+        self.num_examples = 0
```
bartvm: There is no need to keep track of this yourself! The total number of examples seen so far is stored in the log. I'm not entirely sure about the precise attribute (@rizar?) but I think it should be something like `self.main_loop.log.current_row.iterations_done`.

blocks/extensions/training.py
```diff
@@ -0,0 +1,28 @@
...
+
+    Parameters
+    ----------
+    parameter : shared variable to be adjusted
+    function : function which input number of aggregated examples and outputs
+               the parameter value
+
+    """
+    def __init__(self, parameter, function, **kwargs):
+        super(AdjustParameter, self).__init__(**kwargs)
```
bartvm: You should probably set a default callback here i.e. `kwargs.setdefault("after_every_batch", True)` will make sure this extension is run after each batch by default.

blocks/extensions/training.py
```diff
@@ -0,0 +1,45 @@
+import inspect
+from blocks.extensions import TrainingExtension
+
+
+class SharedVariableModifier(TrainingExtension):
```
bartvm: You should probably inherit from `SimpleExtension` instead and move `after_batch` to a `do` method. The advantage is then that the user could choose to e.g. run this only once an epoch, or every `n` batches, etc. making it a bit more flexible (and fixing the current Travis error).

bartvm:
This looks like a good step towards learning rate scheduling! I'll let @rizar have a quick look in case he had any other ideas on how to do this, but as far as I'm concerned this looks good.

blocks/extensions/training.py
```diff
@@ -0,0 +1,45 @@
+import inspect
+from blocks.extensions import SimpleExtension
+
+
+class SharedVariableModifier(SimpleExtension):
+    """Adjusts shared variable parameter using some function.
+
+    Applies function to compute the new value of a shared parameter each
```
rizar: a function

blocks/extensions/training.py
```diff
@@ -0,0 +1,45 @@
...
+    iteration.
+
+    This class can be used to adapt over the training process parameters
+    like learning rate, momentum, etc.
+
+    Parameters
+    ----------
+    parameter : :class:`~tensor.TensorSharedVariable`
+        shared variable to be adjusted
```
rizar: Sorry for being picky, but we everywhere else in Blocks the parameter docstrings are started with a capital letter.

rizar:
LGTM after fixing docs. If you want to do it ideally, use the `unittest` package to join two tests you added  and reduce code duplication. But to be honest, we have copy-pasted tests everywhere in Blocks and this much better than not to have them at all.

Issue #101, opened 2015-01-16
The Softmax Brick Works Only For Matrices


rizar:
See subject. It is using `tensor.nnet.softmax` which raises the following error: 

```
ValueError: ('x must be 1-d or 2-d tensor of floats. Got ', TensorType(float64, 3D))
```

When this is fixed it could be used in `sequence_generators.py`.

rizar:
After giving it another thought I am inclined to just have a wrapper that adapts 2D-only blocks to work with more-D data, like the one @bartvm wrote ages ago. Otherwise we can end up with lots of boilerplate code dealing with dimensions.

vdumoulin:
Just throwing in an idea here, but how about _sandwiching_ the softmax operation in reshapes?

I'm thinking of something like this:

```python
from theano import tensor

def nd_softmax(a, axis=-1):
    swapped_a = a.swapaxes(axis, -1)
    matrix_a = swapped_a.reshape((swapped_a.shape[:-1].prod(), a.shape[-1]))
    softmax_swapped_a = tensor.nnet.softmax(matrix_a).reshape(swapped_a.shape)
    return softmax_swapped_a.swapaxes(axis, -1)
```

The downside of that approach is that the reshape operations may complicate the graph optimization and prevent some optimizations being applied (I've run into that issue in the past).

bartvm:
You could make the reshaping conditional though, no?
```python
shape, ndim = a.shape, a.ndim
if ndim > 2:
    a = a.reshape((a.shape[:-1].prod(), a.shape[-1]))
softmax_a = tensor.nnet.softmax(a)
if ndim  > 2:
    softmax_a = softmax_a.reshape(shape)
```

This seems like code that would be generally useful though, so I personally had something in mind like:

```python
class As2D(Brick):
    def __init__(self, application_method):
        self.application_method = application_method

    def apply(self, x):
        reshaped_x = x.reshape(...)  # As above
        y = self.application_method(reshaped_x)
        reshaped_y = y.reshape(...) 
```
That way you can do
```python
y = As2D(Softmax().apply).apply(x)
```

vdumoulin:
I like that idea. Do you think there's a use case for applying softmax on another axis than the last one? If so, something like

```python
class WithAxesSwapped(Brick):
    def __init__(self, application_method, axis_0, axis_1):
        self.application_method = application_method
        self.axis_0 = axis_0
        self.axis_1 = axis_1

    def apply(self, x):
[...]

PR #251, opened 2015-02-11
Documented flattener

dmitriy-serdyuk:
I wrote how I supposed it to be used. Should I add shomething?


bartvm:
I'll merge for now, but ideally we should also document the `apply` function, describing the type and exact format of the inputs and outputs. I'm also planning to move this brick to `bricks/__init__.py` because we actually need the exact same thing for #101

PR #256, opened 2015-02-12
Picklable itertools

### LINKED ISSUE #254: Switch to picklable_itertools
Where possible, simplify code by using [picklable_itertools](https://github.com/dwf/picklable_itertools).  The main targets will be data streams and request iterators.

bartvm:
Picklable itertools are awesome.

Fixes #254.

### DIFF
--- a/blocks/datasets/__init__.py
+++ b/blocks/datasets/__init__.py
@@ -2,9 +2,10 @@ import collections
 from abc import ABCMeta, abstractmethod
 
 from six import add_metaclass
+from picklable_itertools import izip
 
 from blocks.datasets.streams import DataStream
-from blocks.utils import LambdaIterator, SequenceIterator
+from blocks.utils import SequenceIterator
 
 
 @add_metaclass(ABCMeta)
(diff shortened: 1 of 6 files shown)


bartvm:
There's probably more places where they would come in handy, but this is a nice beginning (removed 90 lines).

@rizar There's one significant change I made: I don't think we need to force wrappers to implement a `get_data` method. For simple wrappers we can just return an iterator from `picklable_itertools` directly instead of returning a `DatasetIterator`, which then calls `get_data`, which would then finally call the same iterator from `picklable_iteratools`...

rizar:
Nice, code simplifications are pretty significant!

I have a concern about the change you made: it seems to me that the datastream wrapper that overrides `get_epoch_iterator` like you did can not be wrapped further, for the reason that its `child_iterator` is in fact `child_iterator` over data produced by its parent. Am I right?

bartvm:
Ehm, I don't think so, although I need to have a closer look to be sure.

A data stream that wraps one of the data streams that now uses `picklable_itertools` would call `get_epoch_iterator` and receive an iterator from the `picklable_itertools` toolbox (instead of the usual `DatasetIterator`). Calling `next` on this will result in the wrapped data, as expected. It's okay that it's a `picklable_itertools` iterator instead of a `DatasetIterator`, because the only interface we rely on is that we can call `next`. Internally, the `picklable_itertools` iterator will be calling `next` on the `DatasetIterator` that calls `get_data` of the underlying data stream and modify it.

rizar:
Okay, I see, the `child_iterator` is created by the wrapper, not by the wrapped.

Then LGTM.

PR #259, opened 2015-02-12
And recursion limit to the config and change it temporarily in the code

rizar:
Let's just change it in the main loop so that a person first likes Blocks and only then stumbles upon this shit.

Better version of #247.

#245, #152.


bartvm:
Would be good to add this new configuration setting to the docs, otherwise LGTM

PR #263, opened 2015-02-12
Fix MNIST start and stop when value is 0

arasmus:
For a very minor bug. If someone wants to set start or stop = 0, it doesn't work.


rizar:
Thanks!

PR #269, opened 2015-02-12
Fixed convolutional

dmitriy-serdyuk:
Fixed several issues

### DIFF
--- a/blocks/bricks/conv.py
+++ b/blocks/bricks/conv.py
@@ -98,7 +98,8 @@ class Convolutional(Initializable):
             W = self.params
 
         output = conv2d(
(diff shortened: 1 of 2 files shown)


blocks/bricks/conv.py
```diff
@@ -98,7 +98,8 @@ def apply(self, input_):
             W = self.params
 
         output = conv2d(
-            input_, W, image_shape=self.input_dim, subsample=self.step,
+            input_, W, image_shape=(None,) + self.input_dim,
```
bartvm: `(None,) + self.input_dim if self.input_dim else None`

PR #270, opened 2015-02-13
When supplying GradientDescent with gradients, they implicitly define which parameters to update

jbornschein:
Currently the 'params' attribute of DifferentiableCostMinimizer is always initialized to contain all shared variables of the cost-graph; even when the user supplies an explicit gradients-dict to GradientDescent. 

1) I guess we can trust the user and automatically set it to gradients.keys()?
2) I wonder whether we should guard GradientDescent.gradients with a property/setter?

### DIFF
--- a/blocks/algorithms/__init__.py
+++ b/blocks/algorithms/__init__.py
@@ -182,7 +182,9 @@ class GradientDescent(DifferentiableCostMinimizer):
     def __init__(self, step_rule=None, gradients=None, **kwargs):
         super(GradientDescent, self).__init__(**kwargs)
         self.gradients = gradients
-        if not self.gradients:
+        if self.gradients:
+            self.params = gradients.keys()
+        else:
             logger.info("Taking the cost gradient")
             self.gradients = dict(
                 zip(self.params, tensor.grad(self.cost, self.params)))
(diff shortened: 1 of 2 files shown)


blocks/algorithms/__init__.py
```diff
@@ -180,7 +180,11 @@ class GradientDescent(DifferentiableCostMinimizer):
 
     """
     def __init__(self, step_rule=None, gradients=None, **kwargs):
-        super(GradientDescent, self).__init__(**kwargs)
+        if gradients and 'params' not in kwargs:
```
bartvm: I guess this should be `'gradients'`?

bartvm: Oh, but wait, `gradients` is an argument, so it will never be in `kwargs`

jbornschein: This indeed looks confusing -- maybe it should read:

`if gradients is not Null and 'params' not in kwargs:`

I was also wondering: Maybe DifferentiableCostMinimizer shouldn't populate `params` at all? Identifying parameter variables that can be optimized is maybe something a more specialized subclass should do? E.g. GradientDescent. That would simplify/streamline the code here a bit.

rizar: Would this do the job?

```python
kwargs.setdefault("params", gradients.keys())
```

jbornschein: if gradients:
        kwargs.setdefault("params", gradients.keys())

would work and looks better than the current proposal.

rizar: Cool, let's do that and merge then!

PR #271, opened 2015-02-13
Naming of Linear and MLP parameters according to the Brick name

arasmus:
If I create an MLP, it will name all the weights "W" and all the biases "b". Unique names would be nice based on the brick name. 

For instance, consider the following autoencoder:

```python
    x = tensor.matrix('features')
    enc = MLP([Rectifier(), Rectifier()], [784, 1000, 2000], name='enc')
    dec = MLP([Rectifier(), None], [2000, 1000, 784], name='dec')
    cg = ComputationGraph(dec.apply(enc.apply(x)))
    print VariableFilter(roles=[PARAMETER])(cg.variables)
```
```[b, b, b, b, W, W, W, W]```

One solution is to use ```ComputationGraph``` and its tools to rename them according to the owning ```Brick``` at query time if user wants. There's also ```var.tag.name``` which role is unclear and not documented, but if the library is intended as light-weight as possible, I think most users would benefit from properly named ```var.name```.

So, to the implementation of ```var.name```. A problem is that ```MLP``` names all its ```Linear``` bricks in ```__init__``` before it passes ```kwargs[name]``` to its ancestor ```Brick``` which sets the ```self.name```. So MLP doesn't know its name when it creates ```Linear```s.

These commits would name them

```[dec_l1_b, dec_l0_b, enc_l1_b, enc_l0_b, enc_l0_W, enc_l1_W, dec_l0_W, dec_l1_W]````

Do you want such behavior and what's the most elegant way to achieve this? Two solution candidates in the latter commit.

### DIFF
--- a/blocks/bricks/__init__.py
+++ b/blocks/bricks/__init__.py
@@ -503,7 +503,11 @@ class MLP(Sequence, Initializable, Feedforward):
     def __init__(self, activations, dims, **kwargs):
         self.activations = activations
 
-        self.linear_transformations = [Linear(name='linear_{}'.format(i))
+        # Another solution duplicates code from Brick and initializes Linears
+        # straight away with right name
+        # self.name = kwargs.get('name', self.__class__.__name__.lower())
+
+        self.linear_transformations = [Linear()
                                        for i in range(len(activations))]
         # Interleave the transformations and activations
         application_methods = [brick.apply for brick in interleave(
@@ -512,6 +516,8 @@ class MLP(Sequence, Initializable, Feedforward):
             dims = [None] * (len(activations) + 1)
         self.dims = dims
         super(MLP, self).__init__(application_methods, **kwargs)
+        for i, brick in enumerate(self.linear_transformations):
+            brick.name = '{}_l{}'.format(self.name, i)
 
     @property
     def input_dim(self):
--- a/blocks/bricks/__init__.py
+++ b/blocks/bricks/__init__.py
@@ -206,15 +206,16 @@ class Linear(Initializable, Feedforward):
         self.output_dim = output_dim
 
     def _allocate(self):
-        W = shared_floatx_zeros((self.input_dim, self.output_dim), name='W')
+        W = shared_floatx_zeros((self.input_dim, self.output_dim),
+                                name=self.name + '_W')
         add_role(W, WEIGHTS)
         self.params.append(W)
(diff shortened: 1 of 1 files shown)


rizar:
Hold on, we are approaching certain fundamental (though unfortunately undocumented) principles of the bricks hierarchy.

One of our starting decision was that names of bricks and variables created by them do not have to be unique. Because if you go this way, you end up with names like "W3_enc_rec_att_proj". In Blocks the name only identifies the bricks among his siblings, that is the path from a root brick serves as a full identifier. You propose to make an exception from this principle for the MLP brick, but that effectively raises a question why not do it for all other bricks.

I would recommend you to look at the [selector](https://github.com/bartvm/blocks/blob/master/blocks/select.py) instead, which is designed to deal with such hierarchies. The current selector is extremely primitive, but a much better is waiting for two months in #73: someone should rebase, clean it up and merge. You can see how it can be used in the word reversing demo, `examples/reverse_words`.

On the other hand, prepending the brick's name to a parameter's name attribute seems reasonable and would help in certain situations. We already do something like that for input and output variables [here](https://github.com/bartvm/blocks/blob/master/blocks/bricks/base.py#L266). We keep two names, `x.name` and `x.tag.name`, because the first serves mainly for debugging purposes (it is more verbose), and the second is the true Blocks name of a variable.

arasmus:
Yes, that selector idea is much more elegant! That's something similar I was referring with my comment but did not know it exists. I'll close this.

> One solution is to use ComputationGraph and its tools to rename them according to the owning Brick at query time if user wants. I'll close this.

bartvm:
I think this is still a point of discussion though. I agree that variable names and block names shouldn't be unique, but on the other hand I would like to see more informative names than `[W, W, W, W]` when I retrieve the weights from an MLP. I think it would be nice to create a `parents` attribute for each brick at one point, which is automatically updated to be in sync with the user-writable `children`'s attribute (kind of like the `parameters` attribute is right now).

We could then automatically set the name attribute to something like `mlp/linear_0/W`. This would for debugging reasons only; it would be really nice to be able to identify variables like this when you `debugprint` a Theano graph for example. You won't be able to use it for selection though, especially considering that a brick can have multiple parents, so the path is not unique. In a separate tag, `tag.name`, we can store the actual name of the variable (`W`) which can be filtered by.

PR #272, opened 2015-02-13
A new data stream wrapper to convert numpy arrays to floatX

rizar:


blocks/datasets/streams.py
```diff
@@ -225,6 +227,26 @@ def get_data(self, request=None):
...
+        super(ForceFloatX, self).__init__(data_stream)
+
+    def get_data(self, request=None):
+        if request is not None:
+            raise ValueError
+        data = next(self.child_epoch_iterator)
+        result = []
+        for piece in data:
+            if (isinstance(piece, numpy.ndarray)
+                    and str(piece.dtype).startswith('float')
```
bartvm: `piece.dtype.kind == 'f'`

bartvm:
Although this is nice to have, datasets should definitely support loading and creating data with the correct `dtype` from the beginning, else you might run out of memory long before you ever get to this step. In cases memory is scarce, this step can actually be problematic, because I think it temporarily creates both the `float64` and `float32` versions of the array in memory. It would be better to use `allow_input_downcast=True` to be set.

rizar:
If data is stored as `float64` the conversion is inevitable. It is better to do it in a wrapper rather than have `astype` in all such datasets.

`allow_input_downcast` is yet to be somehow propagated to the `DatasetEvaluator` now. So I do not see now a better solution than this wrapper. For cases when a batch of data takes more than half of your RAM something else has be designed, but well, those should be pretty rare.

bartvm:
It should often be up to the dataset to load with the correct `dtype`. For example, in MNIST the data is originally loaded as `uint8`, and the dataset then needs to decide to decide whether to convert it to `float32` or `float64`. There could be many other cases where the decision can be made more efficiently when loading the data instead of casting it later on (e.g. when generating random data).

Cases where a batch of data takes more than half of your RAM are far from rare! When doing optimization you might want to use full-batch or very-large batches to make your gradient less noisy and collect better second-order information.

Not opposing this PR, just saying that it's not a full solution; datasets will still need to be able to control the `dtype` they load, and `allow_input_downcast` might still be needed in cases where you want to do the casting only when transferring data to GPU in order to save CPU memory.

rizar:
> Not opposing this PR, just saying that it's not a full solution

Agreed. But it is nice to have for quick and dirty experimentation.

PR #273, opened 2015-02-13
Refactor recurrent bricks

rizar:
There are a few issues I would like to solve with this commit. First, I do not like that a very generic name `Recurrent` belongs to a particular, the simplest one in fact, recurrent transition. Second, it is nice to have uniform argument names, such as "inputs" and "states" for all recurrent transition. Finally, there were some out-dated todo's in the code.

### DIFF
--- a/blocks/bricks/recurrent.py
+++ b/blocks/bricks/recurrent.py
@@ -269,9 +269,9 @@ class SimpleRecurrent(Recurrent, Initializable):
         Parameters
         ----------
         inputs : :class:`~tensor.TensorVariable`
-            The inputs, in the shape (batch, features).
+            The 2D inputs, in the shape (batch, features).
         states : :class:`~tensor.TensorVariable`
-            The states, in the shape (batch, features).
+            The 2D states, in the shape (batch, features).
         mask : :class:`~tensor.TensorVariable`
             A 1D binary array in the shape (batch,) which is 1 if
             there is data available, 0 if not. Assumed to be 1-s
--- a/blocks/bricks/recurrent.py
+++ b/blocks/bricks/recurrent.py
@@ -16,7 +16,7 @@ from blocks.utils import (pack, shared_floatx_zeros, dict_union,
                           is_shared_variable)
 
 
-class Recurrent(Brick):
+class BaseRecurrent(Brick):
     """Base class for brick with recurrent application method."""
     has_bias = False
 
@@ -215,7 +215,7 @@ def recurrent(*args, **kwargs):
         return wrap_application
 
 
-class SimpleRecurrent(Recurrent, Initializable):
+class SimpleRecurrent(BaseRecurrent, Initializable):
     """The traditional recurrent transition.
 
     The most well-known recurrent transition: a matrix multiplication,
@@ -286,7 +286,7 @@ class SimpleRecurrent(Recurrent, Initializable):
         return next_states
 
 
-class LSTM(Recurrent, Initializable):
+class LSTM(BaseRecurrent, Initializable):
     u"""Long Short Term Memory.
 
     Every unit of an LSTM is equipped with input, forget and output gates.
@@ -409,7 +409,7 @@ class LSTM(Recurrent, Initializable):
         return next_states, next_cells
 
 
-class GatedRecurrent(Recurrent, Initializable):
+class GatedRecurrent(BaseRecurrent, Initializable):
     u"""Gated recurrent neural network.
 
     Gated recurrent neural network (GRNN) as introduced in [CvMG14]_. Every
@@ -571,7 +571,7 @@ class Bidirectional(Initializable):
 
     Parameters
     ----------
-    prototype : instance of :class:`Recurrent`
+    prototype : instance of :class:`BaseRecurrent`
         A prototype brick from which the forward and backward bricks are
         cloned.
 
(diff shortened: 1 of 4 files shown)


bartvm:
Not a big fan of the name change I must say. People will expect to be able to just use a `Recurrent` layer, and will be unpleasantly surprised when they find out it our `Recurrent` brick doesn't implement an application method (it's not an abstract class either, so people will be able to instantiate it unknowingly). The name `SimpleRecurrent` is just confusing; simple compared to what? 

I think that in this case user-friendliness outweighs the technical correctness of giving a particular implementation the generic name for recurrent. Correct me if I'm wrong, but I think that most people would refer to a "recurrent layer" as exactly this anyway, unless they further qualify it as a recurrent layer "with LSTM", "with GRU", "with a deep transition function", etc.

I would even argue that the `Recurrent` brick should really get `Tanh()` as a default activation, so that people won't be unpleasantly surprised if they want to build a traditional recurrent neural net and are unable to train it because they unknowingly don't use an activation function. I don't really see much of a direct use case for using `Recurrent` with the `Identity` activation?

@pbrakel Will be working with the recurrent toolbox, so maybe he has more to say.

rizar:
Well, the main reason for me to do this name change is that I want to rename `AttentionTransition` to `AttentionRecurrent`, and I do not want this to be associated with a particular type of recurrent brick.

Also the association between the notion "recurrent network" and "tanh(Wx + b)" is getting weaker and weaker every day. Similarly, "Wx + b" is not the neural network, but a neural network.

My suggestions to address your concerns: `TraditionalRecurrent` (`ClassicRecurrent`) instead of `SimpleRecurrent`, `BaseRecurrent` instead of `Recurrent` and abstract class.

bartvm:
If someone says "a neural network layer" without qualification, I would still think `nonlinearity(Wx + b)`, and the same goes for a recurrent neural network. Without qualification, I assume a "recurrent layer" is still `nonlinearity(Wx + Us + b)`. For user-friendliness, I don't want to be hiding the most simple and traditional recurrent layer under something as obscure a name as `TraditionalRecurrent` or the sorts.

That said, I think there's an argument to be made for not giving a default activation function, to be consistent, since we don't do it for `MLP` or `Linear` either.

I think this is really a usability issue; what do users expect these layers to be called? So let's just hear what end-users of the toolkit such as @pbrakel, @laurent-dinh, @mohammadpz, @jych feel is the best naming scheme.

rizar:
Interesting. When someone says "a neural network" I would rather think about the one with a single hidden layer, and most probably not think anything. 

Adding @kelvinxu to the best naming scheme pool.

bartvm:
To clarify the question to all these people: Currently the `Recurrent` brick refers to the traditional `activation(Wx + Us + b)` transition. There is a base class for all recurrent bricks called `BaseRecurrent`. @rizar is proposing renaming the former to `SimpleRecurrent` or `TraditionalRecurrent`, and naming the base class `Recurrent` instead (so `Recurrent` will no longer be usable layer, just one you inherit from).

pbrakel:
I don't have a direct solution to the naming problem here because I degree a bit with both of you and think it all depends a bit on how much experience a person has with blocks. That said, I think that some of the confusion lies in the fact that the name 'Recurrent' seems to refer to both the type of application a brick supports and a specific neural network architecture, while potentially the recurrent decorator could also be used for HMMs, MCMC algorithms and many other things. Moreover, most of the bricks are more like definitions or transitions than RNNs in the traditional sense that contain input and output weights. Perhaps there should be a distinction between the abstraction of a recurrent/transition brick and the specific case of a recurrent neural network to start with. I would still expect a RecurrentNeuralNetwork or RNN class to implement nonlinearity(Wx Us + b) though. Perhaps 'BasicRNN' is another option. Long story short: I don't have a direct solution but think it's good to put some thought in the whole naming scheme for 'things that have recurrent transformations' in general.

rizar:
Clarification to what @bartvm says: I do not mind the base class to be called  `BaseRecurrent`, I just want not to give the name `Recurrent` to the simplest transition - it is too general for it.

mohammadpz:
I think it's good to replace ```Recurrent``` with ```SimpleRecurrent```as been proposed in Elamn 1988 paper as ```SRN```. ```BaseRecurrent``` is just okay.

pbrakel:
I guess for now I agree with the last two posts of @rizar and @mohammadpz and that would mean Recurrent would simply disappear entirely as a class name. That way, trying to instantiate it cannot cause any confusion. SimpleRecurrent is not very catchy but is kind of common enough terminology that looking through the list of classes would probably let people find what they're looking for.

jych:
I also agree with @rizar. We can call non-linearity(Wx + Uh + b) as 'SimpleRecurrent' or 'VanillaRecurrent' (I think 'Simple' or 'Vanilla' seem to be clear than 'TraditionalRecurrent'). We can use 'Recurrent' or 'BaseRecurrent' as the abstract class for all the Recurrent layers. Further more we can make 'GRU' and 'LSTM' to inherit from 'Recurrent' or 'BaseRecurrent' (I'm currently letting GRU or LSTM inherit from 'VanillaRecurrent' in my code). We could also have a class 'StackedRNN' or 'StackedRecurrent' that corresponds to MLP.

bartvm:
Okay then, let's have `BaseRecurrent` and `SimpleRecurrent` :)

PR #274, opened 2015-02-13
And on_error callback

rizar:
Solves #154.

Such things are not really convenient to test, so I will carefully describe here in plain English what the new error handling strategy is. When a error occurs, we make a record in the log and notify user that the extensions are called. Then the `on_error` callbacks (new!) are invoked. If any of those fail, we just go ahead, calling `logger.error` though. Finally, we reraise the exception.  A care is taken not to invoke "after_training" callbacks when finishing because of an error.

The advantages of the new scheme:

* one can use the debugger to figure out what the problem was

* it is possible to do something with the errored main loop, for instance to save it under another name

The saving for the last item is not implemented yet, but can be easily coded when this PR is merged.

### DIFF
--- a/blocks/main_loop.py
+++ b/blocks/main_loop.py
@@ -9,6 +9,14 @@ from blocks.utils import reraise_as, unpack, change_recursion_limit
 
 logger = logging.getLogger(__name__)
 
+error_message = (
+    """An error occurred during the training.
+
+Attempting to run `on_error` extensions before exiting...""")
+
+error_in_error_handling_message = (
+    """While calling `on_error` extension another error occured.""")
+
 
 class MainLoop(object):
     """The standard main loop of Blocks.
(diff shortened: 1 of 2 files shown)


blocks/main_loop.py
```diff
@@ -109,13 +117,17 @@ def run(self):
...
-                    "Attempting to run extensions before exiting...")
-                # TODO: change the serialization destination here
+                self.log.current_row.got_exception = traceback.format_exc(e)
+                logger.info(error_message)
+                try:
+                    self._run_extensions('on_error')
+                except Exception as inner_e:
+                    logger.error(traceback.format_exc(inner_e))
+                    logger.info(error_in_error_handling_message)
+                reraise_as(e)
```
bartvm: `raise` should work as well I think

bartvm:
Major improvement simply because it means that we can use PDB again; been a very frustrating issue which I just haven't had time to fix this week.

blocks/main_loop.py
rizar: I tried and the stack trace was lost.

bartvm: Strange... Maybe because of the second, nested `except` statement? Either way, leave it then

rizar:
Merging despite the decreased coverage: testing the main loop would be hell.

PR #275, opened 2015-02-13
Refactor monitoring

rizar:
The main thing I do here is moving the prefix logic to a mixin. Apart from prettying the code a bit, this allows to provide a `record_name` interface, which is very helpful when working with the `Plot` extension.


bartvm:
LGTM

PR #277, opened 2015-02-13
Added feedback dimension to source dimensions

dmitriy-serdyuk:
@rizar we discussed this bug at NIPS.


rizar:
True, this is needed for the shortcut from the previous word feedback to the generation of the next word, right?

dmitriy-serdyuk:
Yes, right.

PR #279, opened 2015-02-13
Beam search

dmitriy-serdyuk:
Work is still in progress.

- [x] Theano functions for beam search
- [x] Extract necessary inputs
- [x] Search itself
- [x] Multinomial probabilities in Theano for ndim > 2
- [ ] Test beam search


rizar:
Are we closing the first PR?

blocks/bricks/sequence_generators.py
```diff
@@ -439,7 +462,8 @@ class SoftmaxEmitter(AbstractEmitter, Initializable, Random):
     Interprets readout elements as energies corresponding to their indices.
 
     """
-    def _probs(self, readouts):
+    @application
+    def emit_probs(self, readouts):
```
rizar: Can we call it just `probs`? I know I have not really documented or even explain terminology, but only outputs are emitted, because there stochasticity may be involved. All the rest is computed deterministically.

dmitriy-serdyuk:
Yes

rizar:
I would suggest to start fixing tests, you can click on `Details` it see how exactly they failed, e.g. [like this](https://travis-ci.org/bartvm/blocks/jobs/50659411).

blocks/bricks/sequence_generators.py
```diff
@@ -226,7 +238,7 @@ def generate_states(self):
     @generate.property('outputs')
     def generate_outputs(self):
         return (self.state_names + ['outputs']
-                + self.glimpse_names + ['costs'])
+                + self.glimpse_names + ['costs', 'probs'])
```
dmitriy-serdyuk: @rizar Is it going to ruin something? What should I return for emiters which have no support for `probs`?

rizar: Well, there is no sense to do beam search when the emitter is deterministic.

In fact I would like to suggest a different way of retrieving probabilities for beam search from the sequence generator. It seems to me that one can use the generation computation graph and retrieve the probabilities from there using the `VariableFilter`. Knowledge that probabilities are an output of a certain application method of a certain brick is sufficient for that. That would allow to keep the interface a bit cleaner. Check the reverse words demo for an example of using `VariableFilter` for getting intermediate results from the computation graph.

blocks/bricks/sequence_generators.py
```diff
@@ -439,19 +463,25 @@ class SoftmaxEmitter(AbstractEmitter, Initializable, Random):
...
         return tensor.nnet.softmax(readouts.reshape(
             (tensor.prod(shape[:-1]), shape[-1]))).reshape(shape)
 
     @application
     def emit(self, readouts):
-        probs = self._probs(readouts)
-        return self.theano_rng.multinomial(pvals=probs).argmax(axis=-1)
+        probs = self.probs(readouts)
+        # Implemented in Theano only for pvals.ndim == 2
+        # TODO: Refactor reshapes when implemented in Theano
```
rizar: Do you think we might need that for more than 2D readouts?

blocks/bricks/sequence_generators.py
dmitriy-serdyuk: Okey, I'll try to do it.

blocks/bricks/sequence_generators.py
dmitriy-serdyuk: The readouts are 2D, so the probabilities are 3D (time, batch, index).

dmitriy-serdyuk:
@rizar Can you take a look at `compile` function? As I understand it should look like this.

dmitriy-serdyuk:
Can you say, where can I get masks for outputs?

dmitriy-serdyuk:
It's a silly question about masks, I got it.

blocks/search.py
```diff
@@ -0,0 +1,261 @@
...
+
+        """
+        generator = self.sequence_generator
+        attended = self.attended
+        attended_mask = self.attended_mask
+        init_generated = generator.generate(attended=attended,
+                                            attended_mask=attended_mask,
+                                            iterate=False, n_steps=1,
+                                            batch_size=self.batch_size)
+        init_generated = OrderedDict(zip(self.generate_names, init_generated))
```
rizar: There is a `return_dict=True` keyword argument for that.

blocks/search.py
```diff
@@ -0,0 +1,261 @@
...
+        init_cg = ComputationGraph(init_generated.values())
+        init_readouts = VariableFilter(application=generator.readout.emit,
+                                       name='readouts')(init_cg.variables)[-1]
+        init_probs = generator.readout.emitter._probs(init_readouts)
```
rizar: If `probs` is already an application method, why don't you retrieve the output of this application method directly from the graph? Why do you need to fetch readouts first?

rizar: Non-existing `generate_outputs`, the name mismatch between `_probs` here and `probs` in your changes to the sequence generator imply that you did not run that code. Please try to run your code before asking for reviews: that makes the process much smoother. You can train a sequence generator for the "reverse_words" demo (takes ~2 hours on GPU), unpickle and use it.

blocks/search.py
```diff
@@ -0,0 +1,261 @@
...
+            cur_value = tensor.zeros_like(value)
+            cur_value.name = name
+            cur_variables[name] = cur_value
+
+        input_names = (generator.state_names + generator.glimpse_names +
+                       ['outputs'])
+        generator_inputs = {name: val for name, val
+                            in cur_variables.iteritems()
+                            if name in input_names}
+        next_generated = generator.generate(attended=attended,
```
rizar: I do not understand why you need `init_generate` and `next_generated`. The first, as you wrote it, will not give the initial states that you are probably willing to get this way.

For getting the initial states use the `initial_state` method of the `BaseRecurrent`. The names of the states can be accessed by `generate.states`. 

Off the top of my head I think your should need the following functions:

1) One to compute the initial states. I described above how you should be able to build this one.

2) One to compute probabilities for the next words given the current states. You are doing it right here: 
`generator.generate` builds these probabilities under the hood and we can simply retrieve them with the variable filter.

3) One that computes next states given the current states and the chosen outputs. For that you need a more advanced hack. You know that in the `generator.generate` computation graph the sampled outputs are annotated as the outputs of the `readout.emit` computation graph. So what you could to do is replace those with your input Theano variables. Use the power of annotations!

4) That should have been said first, but you also need to compute all the contexts your sequence generator needs, as you most likely don't want to do it multiple times.

blocks/search.py
```diff
@@ -0,0 +1,261 @@
...
+    beam_size : int
+        Size of beam
+    batch_size : int
+        Size of input batch
+    sequence_generator : sequence generator
+        Sequence generator brick
+
+    """
+    def __init__(self, beam_size, batch_size, sequence_generator, attended,
+                 y, attended_mask, y_mask, inputs_dict, inputs):
```
rizar: That's a lot of parameters with no explanation what they should be used for. 

Ideally such this method should accept just two parameters:

- beam size

- the computation graph for sampling built by `generator.generate`

Wonderful thing about blocks is that all the rest can be recovered using `ComputationGraph` and `VariableFilter` methods. Well, to be honest there is one thing missing - it is right now not possible to obtain the original computation graphs of the recurrent methods, before they might be arbitrarily modified by `theano.scan`. But I think I know how to fix that and I will do it ASAP.

dmitriy-serdyuk: Yes, thougth about it, but the current way seemed easier to me. I'll rewrite this part.

blocks/search.py
dmitriy-serdyuk: Hmm, actually I ran it. Should be some issue with PYTHONPATH.

blocks/search.py
dmitriy-serdyuk: Yeah, it should be easier to do it with computational graph.

dmitriy-serdyuk:
I hope, that it should work with reverse words demo. I'll try to run it and inform you about the results.

blocks/search.py
```diff
@@ -0,0 +1,370 @@
...
+            self.compile()
+
+
+class GreedySearch(Search):
+
+    def __init__(self, sequence_generator):
+        super(GreedySearch, self).__init__(sequence_generator)
+
+
+class BeamSearch(Search):
```
rizar: Greedy search is just beam search with beam width equal to one, so I do not see why we need this one and the base class.

dmitriy-serdyuk:
I implemented 4 funcitons as we discussed. The code should work with the reverse word example. I'll use `VariableFilter` and graph transformations when @rizar will finish #330 .

dmitriy-serdyuk:
I added `beam_search` mode to reverse words example.

rizar:
This is nice that it works and thanks for adding to the reverse_words demo! I will now go through the code and give you some hints, and when #330 is done we will be able to finish it.

blocks/search.py
```diff
@@ -0,0 +1,401 @@
...
+from theano import function
+from theano import tensor
+
+from blocks.filter import VariableFilter
+from blocks.graph import ComputationGraph
+
+floatX = config.floatX
+
+
+class Search(object):
```
rizar: I do not think we need this abstract class when no hierarchy is anticipated.

blocks/search.py
```diff
@@ -0,0 +1,401 @@
...
+
+
+class BeamSearch(Search):
+    """Beam search.
+
+    Parameters
+    ----------
+    beam_size : int
+        Size of beam.
+    batch_size : int
```
rizar: I am not sure the beam search needs batch size. This makes the code significantly more complicated. When beam size small it should be fast on its own and when beam size is large that would give you the required degree of parallelization that you need in order to benefit from GPU.

blocks/search.py
```diff
@@ -0,0 +1,401 @@
...
+    def __init__(self, beam_size, batch_size, sequence_generator, attended,
+                 attended_mask, inputs_dict):
+        super(BeamSearch, self).__init__(sequence_generator)
+        self.beam_size = beam_size
+        self.batch_size = batch_size
+        self.sequence_generator = sequence_generator
+        self.attended = attended
+        self.attended_mask = attended_mask
+        self.inputs_dict = inputs_dict
+        self.generate_names = sequence_generator.generate_outputs()
```
rizar: I was wrong that it can't work, but please use `generate.output` instead. The functions that implement logics for properties were not meant to be invoked explicitly.

blocks/search.py
```diff
@@ -0,0 +1,401 @@
...
+            initial_states[name] = generator.initial_state(
+                name,
+                self.real_batch_size,
+                attended=attended)
+
+        self.initial_state_computer = function([attended, attended_mask],
+                                               initial_states.values(),
+                                               on_unused_input='ignore')
+        # Construct initial values
+        init_generated = generator.generate(attended=attended,
```
rizar: I am sorry I still do not understand why you need separate `init_generated` and `next_generated`.

examples/reverse_words/__init__.py
```diff
@@ -169,6 +173,19 @@ def main(mode, save_path, num_batches, data_path=None):
         for brick in model.get_top_bricks():
             brick.initialize()
 
+        # Give an idea of what's going on
+        model = Model(cost)
```
rizar: These are not changes for this PR, which means that probably you should rebase your PR one more time (merge the master branch into it).

rizar:
I merged #330. You can now get your computation graph for the sampling graph, if I am not missing anything.

blocks/search.py
```diff
@@ -0,0 +1,266 @@
...
+        # Create theano function for next values
+        self.next_computer = function([attended, attended_mask,
+                                       cur_variables['states']],
+                                      next_generated.values() + [next_probs])
+        self.compiled = True
+
+    @construct_dict
+    def compute_contexts(self, inputs_dict):
+        contexts = self.attended_computer(*inputs_dict.values())
+        return ["attended", "attended_mask"], contexts
```
rizar: I think that `OrderedDict(zip(*))` is simple enough and does not justify using a decorator. It is a very well-known snippet as well.

blocks/search.py
```diff
@@ -0,0 +1,266 @@
...
+        self.state_names = (sequence_generator.state_names +
+                            sequence_generator.glimpse_names +
+                            ['outputs'])
+        self.compiled = False
+
+    def compile(self, *args, **kwargs):
+        """Compiles functions for beam search."""
+        generator = self.sequence_generator
+        attended = self.attended
+        attended_mask = self.attended_mask
```
rizar: Forgive if I am wrong, but it seems to me that instead of the pair `attended` and `attended_mask` you could as well work with just the list of contexts everywhere. That would kill two rabbits (of course there is no such an idiom in English..): the code will become both simpler and more general.

blocks/search.py
```diff
@@ -0,0 +1,266 @@
...
+        cur_variables = OrderedDict()
+        for name, value in initial_states.iteritems():
+            cur_value = tensor.zeros_like(value)
+            cur_value.name = name
+            cur_variables[name] = cur_value
+
+        next_generated = generator.generate(attended=attended,
+                                            attended_mask=attended_mask,
+                                            iterate=False,
+                                            n_steps=1,
```
rizar: Fortunately you do not have to specify  `n_steps` or `batch_size` anymore after #330. That was a sort of bug that in order to use `iterate=False` one had to provide fake values for them.

blocks/search.py
```diff
@@ -0,0 +1,266 @@
...
+
+        Returns
+        -------
+        Most probable sequences, corresponding probabilities and costs.
+
+        """
+        if not self.compiled:
+            self.compile()
+        # Inputs repeated beam_size times
+        aux_inputs = OrderedDict(
```
rizar: I know I am terrible picky, but "auxiliary" more confuses here than explain what happens. Together with your nice informative comment simply "inputs" will do the job.

blocks/search.py
```diff
@@ -0,0 +1,266 @@
...
+            cur_states['cur_outputs'])
+
+        for i in range(max_length):
+            cur_states.update(self.compute_next(contexts, cur_states))
+            next_probs = (cur_states['cur_probs'][:, :, None] *
+                          cur_states['probs'] *
+                          cur_states['cur_outputs_mask'][-1, :, None])
+
+            # Top probs
+            indexes, top_probs = self._top_probs(next_probs[:, :],
```
rizar: `next_probs[:, :] -> next_probs`?

blocks/search.py
```diff
@@ -0,0 +1,266 @@
...
+                cur_states['cur_outputs_mask'],
+                next_out_mask.copy(),
+                axis=0)
+
+            # All first element in sequences ended
+            if numpy.all(cur_states['cur_outputs'][-1, 0] == eol_symbol):
+                break
+
+        # Select only best
+        best_outputs = cur_states['cur_outputs'][1:, 0]
```
rizar: I am not sure what should be the desired output format for this method. I would say that a list of the candidates sequences without eol symbols and a list of respective costs is the best. That seems more convenient to work with compared to a dense matrix and a mask, apart from the case when we want to feed this data as an input batch. 

@pbrakel , what's your opinion?

dmitriy-serdyuk: Anyway, I need this mask in order to compute the whole batch on GPU. I can't change the batch size in the middle of computation.

rizar: But you do not have to continue search for the sequences that are 
finished. Or if you do not want to do this, there is no need to remember 
all masks as you currently do.

rizar: Wait, I see some sense in storing masks now: they are handled together with the rest by `_rearrange` in this case. I have to read your code a bit more.

dmitriy-serdyuk:
@rizar , seems that the graph still doesn't contain all the variables. In reverse words example:
```python
next_probs = VariableFilter(
            application=generator.readout.emitter.probs
            )(cg.variables)
```
returns `[]`.

dmitriy-serdyuk:
Ok, seems, that I need to use `get_application_call` to retrieve the variable.

rizar:
Sure, because the inner graph is not a part of the normal you typically work with.

blocks/search.py
```diff
@@ -0,0 +1,269 @@
...
+
+        cur_states['cur_outputs'] = cur_states['outputs']
+        cur_states['cur_outputs_mask'] = numpy.ones_like(
+            cur_states['cur_outputs'])
+        cur_states['cur_probs'] = numpy.ones_like(
+            cur_states['cur_outputs'])
+
+        for i in range(max_length):
+            cur_states.update(self.compute_next(contexts, cur_states))
+            next_probs = (cur_states['cur_probs'][:, :, None] *
```
rizar: Why do you make `next_probs` 3-dimensional? It should be indexed by the number of the candidate in the beam and the output, which means 2-dimensions.

dmitriy-serdyuk: It was easier to add a time dimension to have all the variables 3D or 4D. So the time dimension in this case is always 1. I'll check if it is easy to refactor it.

rizar: Please also check it the `top_probs` documentation is up-to-date. I am not sure about that.

rizar:
Do you need any help to switch to using the computation graph?

dmitriy-serdyuk:
Yes, I still don't understand how to extract inputs and outputs of `generate`.

rizar:
Roughly like that: 

```
# Let 'samples' be the output variable of the generation computation graph.
from blocks.filter import get_application_call
generator = get_brick(samples)
application_call = get_application_call(samples)
inner_cg = ComputationGraph(application_call.inner_outputs)
contexts = [VariableFilter(bricks=[generator], name=name, roles=[INPUT])[0] 
                    for name in generator.generate.contexts]
sequences = ... # the same
```

Was it clear?

dmitriy-serdyuk:
Oh, I see. I tried to filter by `application=generator.generate`.

blocks/search.py
```diff
@@ -0,0 +1,269 @@
...
+                                               on_unused_input='ignore')
+
+        # Define inputs for next values computer
+        cur_variables = OrderedDict()
+        for name, value in initial_states.iteritems():
+            cur_value = tensor.zeros_like(value)
+            cur_value.name = name
+            cur_variables[name] = cur_value
+
+        next_generated = generator.generate(iterate=False,
```
rizar: I am afraid this is wrong. Sorry for not saying it in the beginning, I read it not attentively enough.

When you give an output to `generate`, it is not used to get the next states. It is potentially used to to compute the next output, but the next output is anyway sampled from the distribution defined by the emitter. You can see that in the [generate code](https://github.com/bartvm/blocks/blob/master/blocks/bricks/sequence_generators.py#L208).

I proposed a workaround in one of my recent commits: you should substitute the output of `readout.emit` with your variable.

rizar:
And you should mind #340. If possible, do not use filtering by application until this is understood and fixed. But I think you can do without this kind of filtering.

blocks/search.py
dmitriy-serdyuk: Yes, I know. But I'm throwing away the output and using only probabilities.

rizar: Right, here you do, but what about the `next_computer`?

dmitriy-serdyuk:
@rizar , how can I filter initial states? They all have the same name `generator_initial_state_output_0`.

rizar:
True, they can't be currently filtered out :(

But: having extracted the contexts you can call `generator.initial_state` just as you do in the current implementation. Meanwhile I will try to improve their annotations.

dmitriy-serdyuk:
Seems, that I get it to work, but it still returns garbage instead of <S> symbol.

dmitriy-serdyuk:
@rizar , I fixed the issue with `<S>` symbol.

rizar:
Nice, the code became much better, but apparently it does not fully work. 

When I ran `reverse_words` in the beam search mode the costs different from those which I get in the `test` mode.

```
╭─rizar@rizar-Vostro-3360  ~/Dist/blocks  ‹dima*› 
╰─$ ~/Dist/blocks/examples/reverse_words/main.py beam_search reverse_new_model.npz                                     1 ↵
INFO:examples.reverse_words:Model is loaded
WARNING:root:model with multiple output 

Model with multiple outputs are currently only partially supported in Blocks. For instance a call of 'get_objective' will crash. Contact Blocks developers for more details.

/home/rizar/Dist/theano/theano/scan_module/scan_perform_ext.py:133: RuntimeWarning: numpy.ndarray size changed, may indicate binary incompatibility
  from scan_perform.scan_perform import *
Enter a sentence
Does beam search work ?
Encoder input: [42, 3, 14, 4, 18, 41, 1, 4, 0, 12, 41, 18, 4, 0, 17, 2, 7, 41, 22, 14, 17, 10, 41, 39, 43]
Target:  [42, 18, 4, 14, 3, 41, 12, 0, 4, 1, 41, 7, 2, 17, 0, 4, 18, 41, 10, 17, 14, 22, 41, 39, 43]
(0.779674216884)<S>seod maeb hcraes krow ?</S> CORRECT!
(0.779674216884)<S>seod maeb hcraes krow ?</S> CORRECT!
(0.779674216884)<S>seod maeb hcraes krow ?</S> CORRECT!
(0.779674216884)<S>seod maeb hcraes krow ?</S> CORRECT!
(0.779674216884)<S>seod maeb hcraes krow ?</S> CORRECT!
(0.779674216884)<S>seod maeb hcraes krow ?</S> CORRECT!
(0.779674216884)<S>seod maeb hcraes krow ?</S> CORRECT!
(0.779674216884)<S>seod maeb hcraes krow ?</S> CORRECT!
(0.779674216884)<S>seod maeb hcraes krow ?</S> CORRECT!
(0.779674216884)<S>seod maeb hcraes krow ?</S> CORRECT!
Enter a sentence
^CTraceback (most recent call last):
  File "/home/rizar/Dist/blocks/examples/reverse_words/main.py", line 28, in <module>
    main(**vars(args))
  File "/home/rizar/Dist/blocks/examples/reverse_words/__init__.py", line 294, in main
    line = input("Enter a sentence\n")
KeyboardInterrupt
╭─rizar@rizar-Vostro-3360  ~/Dist/blocks  ‹dima*› 
╰─$ ~/Dist/blocks/examples/reverse_words/main.py test reverse_new_model.npz                                            1 ↵
INFO:examples.reverse_words:Model is loaded
WARNING:root:model with multiple output 

Model with multiple outputs are currently only partially supported in Blocks. For instance a call of 'get_objective' will crash. Contact Blocks developers for more details.

D/home/rizar/Dist/theano/theano/scan_module/scan_perform_ext.py:133: RuntimeWarning: numpy.ndarray size changed, may indicate binary incompatibility
  from scan_perform.scan_perform import *
oINFO:root:Sampling function is compiled
Enter a sentence
Does beam search work ?
Enter a number of samples
10
Encoder input: [42, 3, 3, 14, 4, 18, 41, 1, 4, 0, 12, 41, 18, 4, 0, 17, 2, 7, 41, 22, 14, 17, 10, 41, 39, 43]
Target:  [42, 18, 4, 14, 3, 3, 41, 12, 0, 4, 1, 41, 7, 2, 17, 0, 4, 18, 41, 10, 17, 14, 22, 41, 39, 43]
(8.7635188646)wserd maeb hcraes krow ?</S>
(2.93154418945)<S>seodd maeb hcraes krow !</S>
(2.93154418945)<S>seodd maeb hcraes krow !</S>
(2.93154418945)<S>seodd maeb hcraes krow !</S>
(0.249262670523)<S>seodd maeb hcraes krow ?</S> CORRECT!
(0.249262670523)<S>seodd maeb hcraes krow ?</S> CORRECT!
(0.249262670523)<S>seodd maeb hcraes krow ?</S> CORRECT!
(0.249262670523)<S>seodd maeb hcraes krow ?</S> CORRECT!
(0.249262670523)<S>seodd maeb hcraes krow ?</S> CORRECT!
(0.249262670523)<S>seodd maeb hcraes krow ?</S> CORRECT!
Enter a sentence
```

In addition there multiple identical results of search, which for a correct implementation can not happen in principle.

blocks/search.py
```diff
@@ -0,0 +1,280 @@
...
+        for name in generator.generate.states:
+            var = VariableFilter(bricks=[generator], name='^' + name + '$',
+                                 roles=[OUTPUT])(inner_cg)[-1:]
+            next_states.extend(var)
+
+        next_probs = VariableFilter(
+            bricks=[generator.readout.emitter],
+            name='^probs')(inner_cg)[-1]
+        # Create theano function for next states
+        self.next_state_computer = function(list(contexts.values()) + states,
```
rizar: As I have already said before, I believe you have an issue here: you are not overriding the feedback to the network here. Effectively you still some random generation inside, and I have a proof. The quote below shows how your implementation assigns different scores to identical sequences, which is a clear sign of something non-deterministic happening under the hood.

```
Enter a sentence
das ist ein problem .
Encoder input: [42, 3, 0, 18, 41, 8, 18, 19, 41, 4, 8, 13, 41, 15, 17, 14, 1, 11, 4, 12, 41, 37, 43]
Target:  [42, 18, 0, 3, 41, 19, 18, 8, 41, 13, 8, 4, 41, 12, 4, 11, 1, 14, 17, 15, 41, 37, 43]
(0.915070145559)<S>sad tsi nie melborp .</S> CORRECT!
(0.913463181202)<S>sad tsi nie melborp .</S> CORRECT!
(0.913414438598)<S>sad tsi nie melborp .</S> CORRECT!
(0.913414438598)<S>sad tsi nie melborp .</S> CORRECT!
(0.913414438598)<S>sad tsi nie melborp .</S> CORRECT!
(0.913414438598)<S>sad tsi nie melborp .</S> CORRECT!
(0.913414438598)<S>sad tsi nie melborp .</S> CORRECT!
(0.913414438598)<S>sad tsi nie melborp .</S> CORRECT!
(0.913414438598)<S>sad tsi nie melborp .</S> CORRECT!
(0.913414438598)<S>sad tsi nie melborp .</S> CORRECT!
```

blocks/search.py
```diff
@@ -0,0 +1,280 @@
...
+
+        states = self.compute_initial_states(contexts)
+
+        states['cur_outputs'] = states['outputs']
+        states['cur_outputs_mask'] = numpy.ones_like(states['cur_outputs'])
+        states['cur_probs'] = numpy.ones_like(states['cur_outputs'])
+
+        for i in range(max_length):
+            states.update(self.compute_next(contexts, states))
+            next_probs = (states['cur_probs'][:, :, None] * states['probs'] **
```
rizar: Okay, the reason for different scores is clear: you are working with pure probabilities here. Please use logarithms, this much safer.

dmitriy-serdyuk:
It is not costs, it is probabilities. And it is more or less the same.
```python
In [2]: math.log(0.779674216884)
Out[2]: -0.24887911720872388
```

blocks/search.py
dmitriy-serdyuk: We are using a `LookupFeedback` here which is deterministic. So, the problem is somewhere else. Can we have a stochastic feedback?

dmitriy-serdyuk:
Sequences are similar, because in the beginning all the inputs in the beam are the same. I need to break this symmetry.

blocks/search.py
rizar: The problem is here: https://github.com/bartvm/blocks/blob/master/blocks/bricks/sequence_generators.py#L208, and I tried to explain it in first comment. This line is still used by your code.

You can also see it by printing your computation graph:
```
        self.next_state_computer = function(list(contexts.values()) + states,
                                            next_states + [next_probs])
        from theano.printing import pydotprint
        pydotprint(self.next_state_computer, "next_state.png")
```

dmitriy-serdyuk: Oh, I see, that's beacause `fork` depends  on `next_output` and `next_state` depends on `fork`.

rizar: Exactly!

You should use `next_output` as an input of your `next_state_computer` graph. You can extract it with the `VariableFilter`, though for that you definitely need filtering by the application method, which is currently broken after unpickling. I will try to fix it!

rizar:
I still have duplicates:

```
~/Dist/blocks/examples/reverse_words/main.py beam_search reverse_new_model.npz                                     1 ↵
INFO:examples.reverse_words:Model is loaded
WARNING:root:model with multiple output 

Model with multiple outputs are currently only partially supported in Blocks. For instance a call of 'get_objective' will crash. Contact Blocks developers for more details.

/home/rizar/Dist/theano/theano/scan_module/scan_perform_ext.py:133: RuntimeWarning: numpy.ndarray size changed, may indicate binary incompatibility
  from scan_perform.scan_perform import *
Enter a sentence
What is the weather like today ?
Encoder input: [42, 22, 7, 0, 19, 41, 8, 18, 41, 19, 7, 4, 41, 22, 4, 0, 19, 7, 4, 17, 41, 11, 8, 10, 4, 41, 19, 14, 3, 0, 24, 41, 39, 43]
Target:  [42, 19, 0, 7, 22, 41, 18, 8, 41, 4, 7, 19, 41, 17, 4, 7, 19, 0, 4, 22, 41, 4, 10, 8, 11, 41, 24, 0, 3, 14, 19, 41, 39, 43]
(0.149181212877)ot si eht rehtaew ekil yadot ?</S>
(0.149181212877)ot si eht rehtaew ekil yadot ?</S>
(0.149181212877)ot si eht rehtaew ekil yadot ?</S>
(0.149181212877)ot si eht rehtaew ekil yadot ?</S>
(0.149181212877)ot si eht rehtaew ekil yadot ?</S>
(0.149181212877)ot si eht rehtaew ekil yadot ?</S>
(0.149181212877)ot si eht rehtaew ekil yadot ?</S>
(0.149181212877)ot si eht rehtaew ekil yadot ?</S>
(0.149181212877)ot si eht rehtaew ekil yadot ?</S>
(0.131659120438)<S>tahw si eht rehtaew ekil yadot ?</S> CORRECT!
Enter a sentence
^CTraceback (most recent call last):
  File "/home/rizar/Dist/blocks/examples/reverse_words/main.py", line 28, in <module>
    main(**vars(args))
  File "/home/rizar/Dist/blocks/examples/reverse_words/__init__.py", line 294, in main
    line = input("Enter a sentence\n")
KeyboardInterrupt
```

blocks/search.py
```diff
@@ -0,0 +1,303 @@
...
+                                                 states + [outputs]))
+
+        # Add time dimension back
+        next_values = [state.reshape((1,) + state.shape)
+                       for state in next_values]
+        return OrderedDict(zip(self.generate_names, next_values))
+
+    @staticmethod
+    def _top_probs(scores, beam_size, unique=False):
+        """Returns indexes of elements with lowest scores.
```
rizar: You should do something with this function. It is very hard to understand why it receives 3D array of probabilities, to be honest I still do not get it after a week of staring at the code.

blocks/search.py
```diff
@@ -0,0 +1,303 @@
...
+            states['cur_logprobs'] = numpy.array(top_probs).T[None, :]
+            indexes = numpy.array(indexes)  # 2, beam
+            # here we suppose, that we have 2d outputs
+            outputs = indexes[1, :].copy()
+
+            # rearrange outputs
+            rearrange_ind = indexes[0, :]
+            for name in states:
+                states[name] = self._rearrange(states[name], rearrange_ind)
+            for name in contexts:
```
rizar: Contexts are the same for all beam elements, so maybe we should not do this?

blocks/search.py
```diff
@@ -0,0 +1,303 @@
...
+            flatten = scores.flatten()
+        args = numpy.argpartition(flatten, beam_size)[:beam_size]
+        args = args[numpy.argsort(flatten[args])]
+        # convert args back
+        indexes = numpy.unravel_index(args, scores.shape[1:])
+        return indexes, scores[0][indexes]
+
+    @staticmethod
+    def _rearrange(outputs, indexes):
+        new_outputs = outputs[:, indexes.T.flatten()]
```
rizar: Why do you first transpose and then flatten? Isn't `flatten` sufficient?

blocks/search.py
dmitriy-serdyuk: Oh, you are right.

dmitriy-serdyuk:
These duplicates might have been completed in different ways.

rizar:
Closed

PR #281, opened 2015-02-13
added SlidingWindowDataStream

pbrakel:
I implemented this with the future application in mind of keeping track of a pool that is sorted on size to generate batches of sequences of similar lengths. I'm not entirely happy with the name yet because it can be confused with an iterator that moves a sliding window over the frames of a sequence.

### DIFF
--- a/tests/datasets/test_datasets.py
+++ b/tests/datasets/test_datasets.py
@@ -8,7 +8,7 @@ from nose.tools import assert_raises
 from blocks.datasets import ContainerDataset
 from blocks.datasets.streams import (
     CachedDataStream, DataStream, DataStreamMapping, BatchDataStream,
-    PaddingDataStream, DataStreamFilter, ForceFloatX)
+    PaddingDataStream, DataStreamFilter, ForceFloatX, SlidingWindowDataStream)
 from blocks.datasets.schemes import BatchSizeScheme, ConstantScheme
 
 floatX = theano.config.floatX
@@ -215,3 +215,16 @@ def test_padding_data_stream():
         .get_default_stream(),
         ConstantScheme(2)))
     assert len(next(stream3.get_epoch_iterator())) == 4
+
+
+def test_slidin_window_data_stream():
+    stream = ContainerDataset([1, 2, 3, 4, 5]).get_default_stream()
+    batches = list(SlidingWindowDataStream(stream, ConstantScheme(2))
+                   .get_epoch_iterator())
+    expected = [(numpy.array([1, 2]),),
+                (numpy.array([2, 3]),),
+                (numpy.array([3, 4]),),
+                (numpy.array([4, 5]),)]
+    assert len(batches) == len(expected)
+    for b, e in zip(batches, expected):
+        assert (b[0] == e[0]).all()
--- a/blocks/datasets/streams.py
+++ b/blocks/datasets/streams.py
@@ -314,6 +314,50 @@ class CachedDataStream(DataStreamWrapper):
             cache.extend(data)
 
 
+class SlidingWindowDataStream(DataStreamWrapper):
+    """Moves a sliding window over a stream that provides single examples.
+
+    This wrapper initially fills a buffer with the requested number of
+    samples. At every iteration it updates this batch by discarding
+    the first example and appending a new example from the stream it is
+    wrapping. At each iteration, the current state of the buffer is
+    returned.
+
+    Parameters
+    ----------
+    data_stream : :class:`AbstractDataStream` instance
+        The data stream to wrap.
+    iteration_scheme : :class:`.BatchSizeScheme` instance
+        The iteration scheme to use; should return integers representing
+        the size of the batch to return. Only ConstantScheme is supported.
+
+    """
+    def __init__(self, data_stream, iteration_scheme):
+        super(SlidingWindowDataStream, self).__init__(
+            data_stream, iteration_scheme=iteration_scheme)
+        self.buffer = [[] for _ in self.sources]
+
+    def get_data(self, request=None):
+        """Get data from the dataset."""
+        if request is None:
+            raise ValueError
+        if len(self.buffer[0]) < request:
+            for i in range(request):
+                try:
+                    for source_data, example in zip(
+                            self.buffer, next(self.child_epoch_iterator)):
+                        source_data.append(example)
+                except StopIteration:
+                    raise ValueError
+        else:
+            for source_data, example in zip(
+                    self.buffer, next(self.child_epoch_iterator)):
+                source_data.append(example)
+                del source_data[0]
+
+        return tuple(numpy.asarray(source_data) for source_data in self.buffer)
+
+
 class BatchDataStream(DataStreamWrapper):
     """Creates minibatches from data streams providing single examples.
 


bartvm:
Looks good to me. The only comment I would have is that you might want to use a `collections.deque` instead of a list, which would be O(1) instead of O(n) when popping from the front. I guess it should be a relatively simple change.

If you want to make this change, let me know, otherwise I'm happy to merge and leave that change to be made for when someone actually needs the performance.

rizar:
Any publication that uses this way of iteration through data? Don't consecutive batches overlap a lot?

I am not against merging, I just want to make sure I understand what it is and what the motivation behind is.

bartvm:
When you train a language model on n-grams your batches overlap as well. This seem exactly the same to me in the case where your examples are words. I guess that in speech you would use this to have a sliding window over your audio signal?

rizar:
For a sliding window over an audio signal I would rather use your `NGramStream`.

bartvm:
`NGramStream` assumes your data is in the format of variable length lists in which you can slice though. I agree that it would be nice if we could combine the two streams, because they have very similar functionality, but it does depend on what their data looks like, so that's up to @pbrakel

pbrakel:
What I had in mind is that if you want to do some form of sorting of your data samples, you need some sort of pool that allow samples to be added and removed such that another stream can generate sorted batches of samples of a similar size. Now this stream is not suitable for that because that would require the deletion of items from the pool to depend on the request, but I thought it might be useful for similar situations where you want to compute normalization over a larger number of samples than you are actually presenting. I didn't realize NGramStream was so similar so perhaps I should just finish the original idea for generating sorted batches and make a PR for that instead.

rizar:
Regarding sorting the data samples: this can be done very easily in Blocks. I would solve this problem by adding two data stream wrappers:

```
class Sort(DataStreamWrapper):
    """Sorts the examples in a batch according to a given key function."""
    ...

class Unpack(DataStreamWrapper):
     """Frrm a data stream of batches switch to a data stream of samples."""
     ....
```

In this case to have your data consist of sorted portions you add BatchDataStream -> Sort -> Unpack combinations in the beginning of the wrappers chain. That might sound complicated, but the advantage of such scheme is that every wrapper on its own is extremely simple.

Please let me know if you plan to implement sorting in this or another way, because this one of the thing we are definitely missing. It brings very cheap speedup for very time consuming applications.

pbrakel:
So unpack would be very similar to 'chain' from itertools right? 

Anyway, the method you suggest seems like the most logical one but I'd like to implement the case where you want to sort over a large group of sequences (say 2000), then take out enough sequences to have close to a combined sequence length of 1000 frames (say that this takes 10 sequences), remove these from the pool, add 10 new sequences to the pool so you have 2000 again, sort again etcetera. The catch is that the batch over which is sorted is only partially updated and behaves a bit more like a heap with all the sorting in of new samples. So I think your suggested DataStreamWrapper classes would be very useful but for this use case I would also need one that keeps track of a sortable pool/queue from which items can be removed or new items can be added to.

rizar:
Yes, `Unpack` is very much like `itertools.chain`.

So you want overlapping sorted batches right? May I just ask what do you need such strange thing for?

pbrakel:
The reason is to produces batches with samples of a similar size and potentially also adapt the size of the batches such that the total number of frames remains roughly the same. One could simply take a big batch, sort it and iterate over it to generate batches but by sliding the big sorted batch over the dataset before a smaller batch is taken, the sorting of the samples will be more reflective of the true the dataset with the same memory requirements. This is how our current iterator for TIMIT works, if I remember correctly.

tests/datasets/test_datasets.py
```diff
@@ -215,3 +215,16 @@ def test_padding_data_stream():
         .get_default_stream(),
         ConstantScheme(2)))
     assert len(next(stream3.get_epoch_iterator())) == 4
+
+
+def test_slidin_window_data_stream():
```
bartvm: *sliding

rizar:
But wouldn't the large examples simply stay in the batch forever because nobody can push them out of it?

pbrakel:
They would be presented at the very last but that is actually closer to what would happen if the dataset was truly sorted. I admit that it's a bit strange that the shortest sequences don't get a similar treatment but I guess they vary less in length. I also got to say I didn't invent this method but think the sliding batch could perhaps also be used for emulating other operations that would normally require a full pass over the dataset. How did you guys do this for the translation experiments? I thought you did something similar.

rizar:
How large was your sliding batch in practice? 

We did exactly what I described above. That is we sorted every 1000 sentences and formed batches of 100 sentences from them (numbers do not match, but this is roughly the way we did. This seems to be simpler and worked in practice well. Later Jan also put limitation on the maximum total length of sequences taken into a batch.

pbrakel:
I guess it's difficult to compare speech utterances and text sentences directly because the amount of time-steps and features differs so much. For now I'll first implement something locally and think a bit more about whether I think the application is general enough to put in blocks.

rizar:
But I am working with TIMIT right away and do not see any major difference. 

In your case you have a sort of importance sampling, that is you include a longer sequence more often in the batch. But I don't immediately see why it should be helpful.

pbrakel:
But the big batch itself is only used to create smaller mini-batches of sequences that are of similar size. Say the first big batch has many short sequences of about 200 frames and one extreme sequences of 700 frames. This long sequence will never be combined with any of the shorter sequences until longer sequences are getting more common in the big batch and the shortest k that form the next mini-batch includes it. The last couple of mini-batches will only contain the longest sequences of the dataset with the occasional short sequence that happens to be there. I'm not sure if it matters all that much in practice but you'd expect to see fewer occasions in which one long sequence forces a whole batch of much shorter sequences to be padded with zeros.

rizar:
Would it be right to say, that ideally you would like just to traverse the data in sorted order and your method a memory-efficient proxy for that?

pbrakel:
That's a good summary of it :-). @harmdevries89 implemented it this way and I think he based his code on either Jan's or Cho's code but it could also be his idea.

rizar:
Okay I see. I think it is a bit of an overkill, and it would be almost as fast to fetch a large chunk that fits into a memory, sort it and consume. Such a solution would in addition consist of reusable components. Also traversing the whole dataset in the sorted order might be a ideal scenario to aim for, because in this case data the really long sentences are seen really rare.`

pbrakel:
Good points. I guess it depends on the distribution of the lengths and how it affects computation time. Because other people used this method I just wanted to be on the safe side but I can totally see that it introduces a component that doesn't have much use for other situations. I guess at some point I should have a look at the total number of padded frames with both methods and see how much it matters.

harmdevries89:
@pbrakel Although I like your proposal, I didn't implement it this way. I used the approach of Jan and Cho (filling the pool/buffer once in a while with +- 20 batches, and then create batches till the pool is empty). It might be a good idea to benchmark and see if this strategy really results in less zero-padding.

rizar:
@harmdevries89 , sorry, but this was mine code in the Grounghod :)

I remember in summer in had a parameter `k`, which was the number of batches to be merged, sorted, and packed into batches again. That speed up saturated very quickly with growth of `k`. I think I remember the final value - it was 20.

You are right though that the it all depends a lot on the distribution. But on the other hand, no one prevents you from setting `k` to 100: CPU RAM is abundant.

harmdevries89:
@rizar Alright, credits for you, then. :) Only looked at Jan's code at the time.

pbrakel:
Funny how we went full circle and apparently I came up with a method by misinterpreting code that was based on Dima's code :-). In that case I'll close this PR and just use big batches for sorting so the Unpack/Chain and Sort streams would be sufficient.

rizar:
There is however one thing I dislike in the Sort/Unpack solution: if the size of the sorted chunk and the actual batch size are not multiples of each other, or if we have an adaptive grouping strategy, we can find ourselves in the situation where some batches include the longest and the smallest sequences at the same time. For a sufficiently large `k` this will not hurt too much, but it is however slightly ugly. 

On the other hand, the event where one shorted chunk ends  and another one starts can be identified: one can just notice that lengths drop instead of keeping increasing monotonously.

PR #282, opened 2015-02-13
Rename input_dim to image_shape

bartvm:
@dmitriy-serdyuk I think you missed a few variables when renaming in #269. I think this should fix it :) Do you have any code that could easily test `ConvolutionalLayer` so that this bug doesn't come back? If not I'll write a test myself if I have time later.


dmitriy-serdyuk:
Yes, I have. I'll add a test.

bartvm:
@dmitriy-serdyuk I've added a quick test of my own right now, because I wanted something that tests `ConvolutionalLayer` directly. I think that the `ConvNN` class you wrote is definitely something we need to add, but it should go in `bricks/conv.py` instead of the tests, so let's keep that for a later PR for now.

PR #283, opened 2015-02-14
Step rule refactor (fixes #266)

### LINKED ISSUE #266: Refactoring of Step Rules
Long time has passed since they were first introduced, and need for a substantial renaming has accumulated:

* The `SteepestDescent` name is a bit too much for a simple scaling of the step, and it does not play well with new semantics of the step introduced by #243.

* The name of the input argument of `compute_step` is now `gradient`, but very often it is not the actual gradient, but an output of another step rule. We should decide whether we should rename it into `step`.

Please add more concerns regarding terminology in this part of the library.

vdumoulin:
The following is performed in this pull request:

* The term `step` is used instead of `gradient` in `StepRule` and its subclasses.
* The `SteepestDescent` step rule is renamed to `Scale`.
* The `GradientClipping` step rule is renamed to `StepClipping`.
* The `Momentum` step rule is split into `BasicMomentum` and `Momentum`.

### DIFF
--- a/blocks/algorithms/__init__.py
+++ b/blocks/algorithms/__init__.py
@@ -309,14 +309,14 @@ class CompositeRule(StepRule):
 class Scale(StepRule):
     """A step in the direction proportional to the previous step.
 
-    If the previous step is the gradient itself, this step rule is equivalent
-    to steepest ascent.
+    If the previous step is the gradient itself, this step rule is
+    equivalent to steepest ascent.
 
     Parameters
     ----------
     learning_rate : float
-        The learning rate by which the previous step is multiplied to produce
-        the step.
+        The learning rate by which the previous step is multiplied to
+        produce the step.
 
     Attributes
     ----------
--- a/blocks/algorithms/__init__.py
+++ b/blocks/algorithms/__init__.py
@@ -331,13 +331,19 @@ class Scale(StepRule):
         return self.learning_rate * previous_step, []
 
 
-class Momentum(StepRule):
+class BasicMomentum(StepRule):
     """Accumulates step with exponential discount.
 
     Parameters
     ----------
     momentum : float, optional
-        The momentum coefficient.
+        The momentum coefficient. Defaults to 0.
+
+    Notes
+    -----
+    This step rule is intended to be used in conjunction with another
+    step rule, _e.g._ :class:`Scale`. For an all-batteries-included
+    experience, look at :class:`Momentum`.
 
     """
     def __init__(self, momentum=0.):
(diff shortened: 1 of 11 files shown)


vdumoulin:
Fixes #266. Not sure what to do about the coverage decrease failure?

vdumoulin:
Right, I completely forgot about the world outside `blocks.algorithms`. I'll fix the failing tests by tomorrow evening.

bartvm:
Great! Lots of changes, so we should merge quickly before we end up with merge conflicts. I'll give @rizar a chance to have a look, but LGTM

bartvm:
Oh, and don't worry if the coverage tests fail. I enabled it because I like the fact that it reminds people to write tests, and it should warn us if large blocks of codes aren't tested, but wasn't planning on making it a requirement for merging.

blocks/algorithms/__init__.py
```diff
@@ -293,23 +297,26 @@ class CompositeRule(StepRule):
...
+        return steps, updates
+
 
+class Scale(StepRule):
+    """A step in the direction proportional to the previous step.
 
-class SteepestDescent(StepRule):
-    """A step in the direction proportional to the gradient.
+    If the previous step is the gradient itself, this step rule is
+    equivalent to steepest ascent.
```
rizar: Why is it ascent, not descent? Yes, what it returns points in the same direction as the gradient, but what it returns shall be subtracted, which effectively means that it implement steepest descent.

rizar:
Thanks for doing this not the most exciting but super important job of refactoring!

I will merge it right away, we can polish it later. Would you mind if I change "ascent" to "descent" as I suggest in a comment above?

blocks/algorithms/__init__.py
vdumoulin: It's because if you give `Scale` the gradient, the step it will return will have the same orientation as the gradient, thus pointing in the direction of steepest _ascent_.

It's the fact that `GradientDescent` applies a minus sign to the step that makes it become steepest _descent_.

rizar: Okay, this a question of how the documentation of step rules should be written: describing what they actually do or what the effect is when applied in the `GradientDescent`. Ideally both aspects should be attended. I think that in this case the first line of your docstring describes clearly what exactly the rule does. Then, when I see this

> If the previous step is the gradient itself, this step rule is equivalent to steepest ascent. 

, I expect it to be the description of what this step rule effectively does when used in the `GradientDescent`. This is why I was confused when I saw _ascent_.

I suggest to reformulate it like this:

> If used in `GradientDescent` alone, this step rule effectively does steepest descent.

vdumoulin: I agree with your suggestion. I'll make a quick one-liner pull request to fix that.

rizar: No need for that, I can do it myself.

PR #284, opened 2015-02-14
Added test for convolutional layer

dmitriy-serdyuk:


tests/bricks/test_conv.py
```diff
@@ -51,3 +54,65 @@ def test_max_pooling():
...
+
+
+class ConvNN(Sequence, Initializable, Feedforward):
+    """Several convolutional layers
+
+    """
+    def __init__(self, conv_activations, input_dim, filter_sizes,
+                 feature_maps, pooling_sizes, conv_step=None, **kwargs):
+        if conv_step == None:
+            self.conv_step = (1, 1)
```
bartvm: Tuples are immutable, so it's okay to do `conv_step=(1, 1)`, which will make the function signature a bit clearer in the documentation.

tests/bricks/test_conv.py
```diff
@@ -51,3 +54,65 @@ def test_max_pooling():
...
+                                          conv_step=self.conv_step,
+                                          name='conv_pool_{}'.format(i))
+                       for i, (activation, filter_size, num_filter,
+                               pooling_size)
+                       in enumerate(params)]
+
+        application_methods = [brick.apply for brick in list(chain(*zip(
+            self.layers)))
+                               if brick is not None]
+        self.flattener = Flattener()
```
bartvm: This seems unused

tests/bricks/test_conv.py
```diff
@@ -51,3 +54,65 @@ def test_max_pooling():
...
+                                          num_channels=None,
+                                          pooling_size=pooling_size,
+                                          activation=activation.apply,
+                                          conv_step=self.conv_step,
+                                          name='conv_pool_{}'.format(i))
+                       for i, (activation, filter_size, num_filter,
+                               pooling_size)
+                       in enumerate(params)]
+
+        application_methods = [brick.apply for brick in list(chain(*zip(
```
bartvm: I don't think you need this, do you? The `list(chain(*zip(...)))` interleaves lists, but in this case you only have one list, so `application_methods = [brick.apply for brick in self.layers]` should do.

tests/bricks/test_conv.py
```diff
@@ -5,8 +6,10 @@
 from theano import tensor
 from theano import function
 
-from blocks.bricks.conv import Convolutional, MaxPooling
-from blocks.initialization import Constant
+from blocks.bricks import Sequence, Initializable, Feedforward, Rectifier
+from blocks.bricks.conv import (Convolutional, MaxPooling, ConvolutionalLayer,
+                                Flattener)
```
bartvm: Don't forget to get rid of the `Flattener` import as well

tests/bricks/test_conv.py
```diff
@@ -51,3 +54,65 @@ def test_max_pooling():
     pool.input_dim = (x_size, y_size)
     pool.get_dim('output') == (num_channels, x_size / pool_size + 1,
                                y_size / pool_size + 1)
+
+
+class ConvNN(Sequence, Initializable, Feedforward):
```
bartvm: I guess we'll want to move this to `bricks/conv.py`, because this is useful beyond having a test only. I think @harmdevries89 was working in something similar though (a wrapper for multiple convolutional layers), so not sure which implementation has more features?

dmitriy-serdyuk:
Yes, it was a part of a more general class. I fixed it.

bartvm:
Closing this for now, but feel free to reopen as I suggested in https://github.com/bartvm/blocks/pull/282#issuecomment-74393872

PR #285, opened 2015-02-14
WIP: Switch from Dill to cPickle. Fixes #253.

### LINKED ISSUE #253: Review possibility of removing dependency on dill
Resulting from the discussion at https://github.com/bartvm/blocks/issues/236#issuecomment-73904309 we should look into removing the dependency on Dill if possible, mostly because

* Dill seems to user deeper recursion than `cPickle`, quickly resulting in maximum recursion depth errors
* `pickle`/`cPickle` seems anywhere between 3-10x as fast

At the very least this will require

- [ ] Removing all nested functions
- [ ] Removing all lambda functions that are somehow kept as an attribute of an instance
- [ ] Ensuring all attributes to be pickled are available in the global namespace; beyond nested functions this also means that the application/bounded application instances need a reference to the original application method which is available in the global namespace
- [ ] Use a picklable wrapper around file handles, available as part of [picklable_itertools](https://github.com/dwf/picklable_itertools)

bartvm:
After spending a few hours going down the rabbit hole of the subtle differences in Python 2's and 3's data models, I think that I've made a tiny bit of progress... Just wondering what errors this will raise on Travis.

### DIFF
--- a/blocks/bricks/base.py
+++ b/blocks/bricks/base.py
@@ -153,7 +153,7 @@ class Application(object):
             raise ValueError
 
         def wrap_property(application_property):
-            self.properties[name] = application_property
+            self.properties[name] = application_property.__name__
             return application_property
         return wrap_property
 
@@ -193,7 +193,7 @@ class Application(object):
         ['foo', 'bar']
 
         """
-        self.delegate_function = f
+        self.delegate_function = f.__name__
         return f
 
     def __get__(self, instance, owner):
@@ -208,8 +208,8 @@ class Application(object):
     def __getattr__(self, name):
         # Mimic behavior of properties
         if 'properties' in self.__dict__ and name in self.properties:
-            return property(create_unbound_method(self.properties[name],
-                                                  self.brick))
+            return property(create_unbound_method(
+                getattr(self, self.properties[name]), self.brick))
         raise AttributeError
 
     def __setattr__(self, name, value):
@@ -336,14 +336,18 @@ class BoundApplication(object):
         # These always belong to the parent (the unbound application)
         if name in ('delegate_function', 'properties'):
             return getattr(self.application, name)
+        if name in self.properties.values():
+            return getattr(self.application.brick, name)
         if name in self.properties:
-            return self.properties[name](self.brick)
+            return getattr(self, self.properties[name])(self.brick)
         # First try the parent (i.e. class level), before trying the delegate
         try:
             return getattr(self.application, name)
         except AttributeError:
             if self.delegate_function:
-                return getattr(self.delegate_function(self.brick), name)
+                return getattr(getattr(self.brick,
+                                       self.delegate_function)(),
+                               name)
             raise
 
     @property
--- a/blocks/bricks/base.py
+++ b/blocks/bricks/base.py
@@ -115,11 +115,16 @@ class Application(object):
     """
     call_stack = []
 
-    def __init__(self, application):
-        self.application = application
+    def __init__(self, application_function):
+        self._application_function = application_function
+        self.application_name = application_function.__name__
         self.delegate_function = None
         self.properties = {}
         self.bound_applications = {}
+
+    @property
+    def application_function(self):
+        return getattr(self.brick, '_' + self.application_name)
 
     def property(self, name):
         """Decorator to make application properties.
@@ -213,21 +218,21 @@ class Application(object):
             raise AttributeError("can't set attribute")
         super(Application, self).__setattr__(name, value)
 
-    @property_
-    def inputs(self):
-        return self._inputs
+    # @property_
+    # def inputs(self):
+    #     return self._inputs
 
-    @inputs.setter
-    def inputs(self, inputs):
-        args_names, varargs_name, _, _ = inspect.getargspec(
-            self.application)
-        if not all(input_ in args_names + [varargs_name] for input_ in inputs):
-            raise ValueError("Unexpected inputs")
-        self._inputs = inputs
+    # @inputs.setter
+    # def inputs(self, inputs):
+    #     args_names, varargs_name, _, _ = inspect.getargspec(
+    #         self.application_function)
+    #     if not all(input_ in args_names + [varargs_name] for input_ in inputs):
+    #         raise ValueError("Unexpected inputs")
+    #     self._inputs = inputs
 
     @property_
     def name(self):
-        return self.application.__name__
+        return self.application_name
 
     def __call__(self, brick, *args, **kwargs):
         if not isinstance(brick, Brick) and six.PY2:
@@ -245,7 +250,7 @@ class Application(object):
 
         # Find the names of the inputs to the application method
         args_names, varargs_name, _, _ = inspect.getargspec(
-            self.application)
+            self.application_function)
         args_names = args_names[1:]
 
         # Construct the ApplicationCall, used to store data in for this call
@@ -292,7 +297,7 @@ class Application(object):
             raise ValueError
         self.call_stack.append(brick)
         try:
-            outputs = self.application(brick, *args, **kwargs)
+            outputs = self.application_function(brick, *args, **kwargs)
             outputs = pack(outputs)
         finally:
             self.call_stack.pop()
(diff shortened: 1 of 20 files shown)


bartvm:
@rizar @dwf Tests are passing without Dill now! It might require a bit of cleaning up, but after trying to tackle all the serialization issues in 10 different ways, I ended up with going something so stupidly simple it had to work. Would like to know whether you guys think making this switch is a good idea.

* You can't have references to functions outside of the global namespace lying around. This concerns functions before they're bound (i.e. when wrapping functions in decorators), but in Python 3 all class methods are also functions outside of the global namespace (because the "unbound method" doesn't exist anymore). Instead of doing something clever, I ended up storing the name of the function instead of the function itself. Note that this does mean that the function needs to stay part of the class where it was originally defined (so I move `apply` to `_apply` and then replace `apply` with an `Application` that looks up `_apply` when it needs it).
* No more lambdas, anywhere. I personally don't mind that; makes for easier debugging.

blocks/bricks/base.py
```diff
@@ -331,14 +337,18 @@ def __getattr__(self, name):
...
-            return self.properties[name](self.brick)
+            return getattr(self, self.properties[name])(self.brick)
         # First try the parent (i.e. class level), before trying the delegate
         try:
             return getattr(self.application, name)
         except AttributeError:
             if self.delegate_function:
-                return getattr(self.delegate_function(self.brick), name)
+                return getattr(getattr(self.brick,
+                                       self.delegate_function)(),
```
rizar: I am confused: do you assume that the delegate application is always an application of the same brick? But this is not true, see a quote from `sequence_generators.py`:

```python
    @compute_states.delegate
    def compute_states_delegate(self):
        return self.transition.apply
```

I honestly do not know how it works now, and moreover think that we should add a test for that.

rizar:
It is nice that you made it work! (though there something I am sort of suspicious about, see above). But we should weight all the pros and cons:

Pros:

* no need for the users to learn about a new serialization library

* speed (does it _really_ give 5 times speed up on our tasks?)

* cPickle seems to be less recursive: I think this a minor argument because now we have recursion depth in config and beginners do not feel it. The advanced users working with big models will have to mind the stack size anyway.

Cons:

* code clarity

* limitations users have to mind: no lambdas even in user code. However even using `dill` one has to structure his code in a rather special way

I think that speed might be the decisive argument. Do you have any links with `cPickle` vs `dill` comparison?

blocks/bricks/base.py
bartvm: `delegate_function` in this case is `'compute_states_delegate'`, so this part here retrieves the method `compute_states_delegate` and calls it, which returns `self.transition.apply`. It then performs `getattr(self.transition.apply, name)`. Nothing changed from before, it's just that `delegate_function` used to be an object, the unbound method `compute_states_delegate`, now it is the name of the method which needs to be retrieved from the class first (and because we retrieve it from the class instance, it is returned bounded already, which is why we don't need to pass `self.brick` as an argument anymore).

bartvm:
I'll perform some benchmarks on our models. Before I got this working I tried `dill` vs. `cPickle` on basically a long list and a super-simple user-defined class. In those cases, speedups were anywhere between 3x and 20x times, but those toy examples don't say much about our applications. I'll let you know when I have some numbers.

bartvm:
These are timings from my desktop computer:

#### `sqrt` example
```
%timeit dill.dumps(main_loop)
10 loops, best of 3: 155 ms per loop
%timeit cPickle.dumps(main_loop)
10 loops, best of 3: 34.5 ms per loop
```
#### `reverse_words` example
```
%timeit dill.dumps(main_loop)
1 loops, best of 3: 12.1 s per loop
%timeit cPickle.dumps(main_loop)
1 loops, best of 3: 2.73 s per loop
```

I think the 4.4x speedup in the `reverse_words` example is a strong case for making the switch.

dwf:
Holy shit, yes please re: speed. Lambdas do make debugging harder as you
say, since lambda callbacks have no name information attached and you
therefore have no idea which one was actually run.

I'm not sure I can parse "functions outside the global namespace", doesn't
that mainly mean inner functions? Pretty sure dill can't do those anyway.

I'll try and have a look after breakfast  (in San Diego right now) about
the class method business.

bartvm:
Well, not just inner functions. It also means references to methods of other classes. For example:
```python
class A(object):
    def f(self):
        pass

class B(object):
    def __init__(self, func):
        self.func = func

b = B(A().f)  # Same error for B(A.f)
cPickle.dumps(b)
```
```
TypeError: can't pickle instancemethod objects
```

In Python 2 you can fix that using `copy_reg` to register a function that handles the pickling of methods:

```python
def _pickle_method(method):
    func_name = method.im_func.__name__
    obj = method.im_self
    cls = obj.im_class
    return _unpickle_method, (func_name, obj, cls)

def _unpickle_method(func_name, obj, cls):
    for cls in cls.__mro__:
        if func_name in cls.__dict__:
            func = cls[func_name]
            break
    return func.__get__(obj, cls)

 copy_reg.pickle(MethodType, _pickle_method, _unpickle_method)
```
But here's the problem: In Python 3 unbound methods don't exist. Given `A.f` there is no way to know that `f` belongs to `A`, there is no `im_class` equivalent. So `A.f` is just a function, but one that lies outside of the global namespace. There don't seem to be easy ways around this...

You could try hacking something together that parses `A.f.__qualname__` and searches the global namespace for the class to retrieve the function from, but that sounds very hacky. Another option I guess would be to write a metaclass that automatically tags all functions with `im_class` if it's Python 3, so that the same `copy_reg` method works; I haven't tried this but it could actually work quite nicely I guess. The option I've used right now is to not keep a reference to the methods at all (but keep the name of the method, and have the client class retrieve it whenever it needs to).

As far as user-experience goes, the metaclass idea might be worth trying out.

blocks/bricks/base.py
rizar: Okay, I see now.

rizar:
I am fine with going the way this PR proposes. I do not think we should try too hard to support something that Python 3 actually does not support. Current solution with names is not that bad in fact.

bartvm:
Well, it's not that bad if you know of it, but I'm more worried about user-code. It can be quite baffling if you pass `str.lower` as a preprocessing function somewhere, and you get the obscure error:
```
TypeError: can't pickle instancemethod objects
```
Pickling errors are a nightmare to debug, and someone not familiar with the internals of pickling, will have no idea what happened. If we can resolve that by making our base classes like `Brick`, `Dataset`, etc. inherit from a metaclass that solves this problem, I think it's worth considering.

rizar:
Do you mean you can pickle references to class method if those are attributes of `Brick` or `Dataset` classes?

I think it is easier to have a page with an explicit checklist like:

* you must not have lamdas

* you must not have `str.lower` and other class method references

than to invest a lot and make such things work in a few specific cases. I would rather think  about having more informative errors, e.g. `DataStreamMapping` could check that its argument is not a class method, same for `TextFile` and the `preprocess` argument.

bartvm:
Thinking about it, probably writing a metaclass as well as raising errors is quite tricky, because it's easy to know whether you are dealing with a class method or with a normal function (which could potentially be available in the global namespace). The only way is probably to check whether `__name__` and `__qualname__` match, but that only works in Python 3.4, not in 3.0-3.3.

An alternative is to just use a custom `dump` method which basically does this:

```python
try:
    cPickle.dump(main_loop, f)
except TypeError as e:
    if 'isinstancemethod' in e.message:
        reraise_as("You probably have a reference stored to a class method")
    raise
```

rizar:
Sounds good to me. We can have a bunch of if's like that and this should reduce meeting the cPickle's requirements for beginners.

dwf:
Could we catch that particular exception, check for that error message and
reraise a more informative error? Or is that already what is proposed?

bartvm:
Yeah, that's what I was proposing. Should be easy enough, and I also think
it's probably more robust than trying to patch it.

bartvm:
Okay then, if I find the time today I will rebase this, add the custom pickling function with informative errors, and then I think we can make the switch--although there's quite a bit of documentation that needs to be updated.

bartvm:
@rizar This should completely eliminate all references to Dill, and adds informative errors when pickling fails. If the tests pass, let me know if I can merge!

blocks/utils.py
```diff
@@ -405,8 +405,56 @@ def ipdb_breakpoint(x):
...
+Python is unable to serialize lambda functions. Make sure that your code does \
+not rely on them and use normal function definitions instead."""
+
+nested_function_error = """
+
+Python is unable to serialize nested functions. Make sure that all your \
+functions are available in the global namespace."""
+
+
+def pickle_dump(*args, **kwargs):
```
rizar: I think we should move these functions to something like `blocks/serialization.py`, similarly to `blocks/dump.py`. They are too fundamental to be part of a general util stuff.

examples/reverse_words/__init__.py
```diff
@@ -89,6 +90,22 @@ def get_dim(self, name):
         return super(Transition, self).get_dim(name)
 
 
+def lower(s):
```
rizar: Let's follow the pattern of having underscores in front of such dumb helpers similarly to the way you do in `sqrt.py`.

rizar:
I will read it more tomorrow in order to really understand how it works, but you can merge now, because rebasing this monster all the time I guess is extremely inconvenient.

PR #288, opened 2015-02-15
Speed up reverse words demo and add it to tests

rizar:
I want to include the reverse words demo in our automatic tests because the sequence generation framework is about to be refactored and beam search is in development #279. By switching to `SimpleRecurrent` and disabling agressive optimization I managed to get it as fast as 16 seconds on Python 3: that's a lot but I think that short-term benefits of having such a test overweight this slow down.

Also includes some refactoring.

### DIFF
--- a/tests/test_examples.py
+++ b/tests/test_examples.py
@@ -1,3 +1,4 @@
+from __future__ import print_function
 import tempfile
 
 import dill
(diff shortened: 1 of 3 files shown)


tests/test_examples.py
```diff
@@ -1,20 +1,18 @@
...
 from examples.markov_chain.main import main as markov_chain_test
+from examples.reverse_words import main as reverse_words_test
 from tests import silence_printing
 
-
-def setup():
-    # Silence Block's logger
-    logger = logging.getLogger(blocks.__name__)
-    logger.setLevel(logging.ERROR)
+theano.config.optimizer = 'fast_compile'
```
bartvm: We should probably put this in the `setUp` function and restore the default in `tearDown`, because right now this setting will persist for all tests, which might not be what we want, and in the future we might completely forget about it.

rizar: What about putting it into `.theanorc`?

bartvm: Sure, at least that will be more explicit. Or you can just add it to the `THEANO_FLAGS` variable in `.travis.yml`

PR #289, opened 2015-02-15
Improve Scale docstring

vdumoulin:


rizar:
Okay, you were faster :) Thanks!

vdumoulin:
No problem!

PR #290, opened 2015-02-15
ConvNets

harmdevries89:
Here is another ConvNet implementation in blocks. It assumes that you pass convolutional layers (with specified filter_size and num_filter, etc) to the network. It will then automatically configure the right input dimensions (num_channels, input_dim) per layer. Advantage is that you can specify per layer specific details (step_size, etc), thereby avoiding a very complicated constructor of the convnet.

### DIFF
--- a/blocks/bricks/conv.py
+++ b/blocks/bricks/conv.py
@@ -199,12 +199,12 @@ class ConvolutionalLayer(Sequence, Initializable):
     """
     def __init__(self, filter_size, num_filters, num_channels, pooling_size,
                  activation, conv_step=(1, 1), pooling_step=None,
-                 border_mode='valid', input_dim=None, **kwargs):
+                 border_mode='valid', image_shape=None, **kwargs):
         self.convolution = Convolutional(filter_size, num_filters,
-                                         num_channels, input_dim=input_dim,
+                                         num_channels, image_shape=image_shape,
                                          step=conv_step,
                                          border_mode=border_mode)
-        if input_dim is not None:
+        if image_shape is not None:
             pooling_input_dim = self.convolution.get_dim('output')
         else:
             pooling_input_dim = None
@@ -214,11 +214,9 @@ class ConvolutionalLayer(Sequence, Initializable):
             application_methods=[self.convolution.apply, activation,
                                  self.pooling.apply], **kwargs)
 
-        self.input_dim = input_dim
-
     def get_dim(self, name):
         if name == 'input_':
-            return self.input_dim
+            return self.convolution.get_dim('input_')
         if name == 'output':
             return self.pooling.get_dim('output')
         return super(ConvolutionalLayer, self).get_dim(name)
--- a/blocks/bricks/conv.py
+++ b/blocks/bricks/conv.py
@@ -100,7 +100,8 @@ class Convolutional(Initializable):
         output = conv2d(
             input_, W,
             image_shape=(None, self.num_channels) +
-                        (self.image_shape if self.image_shape else (None, None)),
+                        (self.image_shape if self.image_shape else (None,
+                                                                    None)),
             subsample=self.step,
             border_mode=self.border_mode,
             filter_shape=((self.num_filters, self.num_channels) +
--- a/blocks/bricks/conv.py
+++ b/blocks/bricks/conv.py
@@ -38,7 +38,6 @@ class Convolutional(Initializable):
         details. Defaults to 'valid'.
 
     """
-
     @lazy
     def __init__(self, filter_size, num_filters, num_channels, input_dim=None,
                  step=(1, 1), border_mode='valid', **kwargs):
@@ -225,8 +224,8 @@ class Flattener(Brick):
     """Flattens the input.
 
     It may be used to pass multidimensional objects like images or feature
-    maps of convolutional bricks into bricks which allow only two dimensional
-    input (batch, features) like MLP.
+    maps of convolutional bricks into bricks which allow only two
+    dimensional input (batch, features) like MLP.
 
     """
     @application(inputs=['input_'], outputs=['output'])
--- a/blocks/bricks/conv.py
+++ b/blocks/bricks/conv.py
@@ -26,7 +26,7 @@ class Convolutional(Initializable):
         subsequent layers this is equal to the number of filters output by
         the previous convolutional layer. The filters are pooled over the
         channels.
-    input_dim : tuple, optional
+    image_shape : tuple, optional
         The height and width of the input (image or feature map). If given,
         this will be passed to the Theano convolution operator, resulting
         in possibly faster execution times.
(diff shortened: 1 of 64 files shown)


blocks/bricks/conv.py
```diff
@@ -221,7 +273,83 @@ def get_dim(self, name):
...
+    def _push_allocation_config(self):
+        input_dim = self.input_dim
+        num_channels = self.num_channels
+        for layer in self.layers:
+            layer.input_dim = input_dim
+            layer.num_channels = num_channels
+            layer.batch_size = self.batch_size
+
+            # Push input dimensions to children
+            layer._push_allocation_config()
```
bartvm: You shouldn't need to do this, this should be automatically done by `push_allocation_config` (which wraps this method).

bartvm:
I'll have a closer look in a bit, but you shouldn't need to pass bricks in order to fine-tune the configuration of the individual layers. The idea in Blocks is more that you would do it like this:

```python
# uses sensible defaults for the children
network = ConvolutionalNetwork(filter_sizes=[...], num_filters=[...])
# now we configure the children; normally this is done automatically when
# calling network.apply(x), but in this case we want to fine-tune
# (note that for the case of step_size, the parent doesn't change any
# configuration, so we strictly speaking would be fine without calling it)
network.push_allocation_config()
# now we can tweak at will, for example, to change the 3rd layer you can do this:
network.layers[2].step_size = (2, 2)
network.layers.push_allocation_config() 
# or this directly
network.layers[2].convolution.step_size = (2,2)
```

For this particular case, this seems almost more complicated than passing readily configured bricks, but it becomes a big advantage in the case where you are constructing a single brick which needs a dozen children, all of which have children, and you just want to change one tiny setting of one of those. In this case, it's easier to just configure everything with the defaults, and then change the settings you need, which is why we came up with lazy initialization. This just means that you should be able to change the configuration of the bricks you want up to the point that `network.apply(x)` gets called.

blocks/bricks/conv.py
harmdevries89: What do you mean? The MLP class is also using _push_allocation_config in a similar way.

bartvm: Got it, my bad. Normally you don't need to push this manually, because the `push_allocation_config` method automatically does this for you. In this case however you want to do it manually because you want to access `layer.get_dim('output')` directly afterwards to get `output_shape`.

bartvm:
So I noticed that `ConvolutionalLayer` didn't have lazy-initialization enabled yet. Fixed that in #291.

harmdevries89:
I see what you mean, but that wasn't really the setting I had in mind. In general, the architectures of convnets vary a lot, and I wanted to have some flexibility. For example, how would you implement that some layers are followed by pooling, and some are not. I can imagine you would set pool_step=(1, 1) or somehow remove the pooling from the list in ConvNet, but I don't think that's elegant either.

bartvm:
Mmm, I see. So, I'd inclined to say that those kind of cases are so irregular that the question really becomes whether we need a special constructor for them at all. Unlike Pylearn2, where YAML files mean that you *need* to have a single constructor call for any model that you can come up with, for Blocks we can say that if you want a very irregular architecture where some layers have pooling, some don't, etc. then you should just create the layers manually, and string them together by calling `Sequence` directly.

However, for convnets I guess that the need to calculate the dimensions for each layer would be a reason to have a separate brick that does that for you. `ConvolutionalNetwork` is then basically just a small wrapper around `Sequence` that does that for you (and mixes in `Initializable`). I guess that sounds reasonable.

But in that case we could consider having two bricks: `ConvolutionalSequence` could be what you implemented, but we can also have a `ConvolutionalNetwork`, which is more like the one that @dmitriy-serdyuk implemented in #284. That one just assumes a regular convolution-nonlinearity-pooling structure and basically just takes a list of filter sizes as an input.

Also, I like the idea of having a `ConvolutionalActivation` class, but then it seems to make sense that the `ConvolutionalLayer` class uses the `ConvolutionalActivation` followed by pooling, instead of re-implementing the same logic.

blocks/bricks/conv.py
```diff
@@ -40,16 +39,18 @@ class Convolutional(Initializable):
...
+                 border_mode='valid', **kwargs):
         super(Convolutional, self).__init__(**kwargs)
 
         self.filter_size = filter_size
-        self.input_dim = input_dim
-        self.border_mode = border_mode
         self.num_filters = num_filters
+        self.batch_size = batch_size
         self.num_channels = num_channels
+        self.input_dim = input_dim
```
dmitriy-serdyuk: @harmdevries89 We are using `image_shape` for it now.

blocks/bricks/conv.py
```diff
@@ -98,22 +99,27 @@ def apply(self, input_):
...
 
-        output = conv2d(
-            input_, W, image_shape=self.input_dim, subsample=self.step,
-            border_mode=self.border_mode,
-            filter_shape=((self.num_filters, self.num_channels) +
-                          self.filter_size))
+        image_shape = None
+        if all([x is not None for x in [self.batch_size,
+                                        self.num_channels, self.input_dim]]):
+            image_shape = (self.batch_size, self.num_channels) + self.input_dim
```
dmitriy-serdyuk: I think, it's better to rename this (`input_shape`?).

blocks/bricks/conv.py
```diff
@@ -98,22 +99,27 @@ def apply(self, input_):
...
+                        filter_shape=(self.num_filters, self.num_channels)
+                        + self.filter_size)
         if self.use_bias:
             output += b.dimshuffle('x', 0, 'x', 'x')
         return output
 
     def get_dim(self, name):
         if name == 'input_':
-            return self.input_dim
+            return (self.batch_size, self.num_channels, self.input_dim)
```
dmitriy-serdyuk: As I understand, blocks doesn't return batch size in `get_dim`. @bartvm is it true?

bartvm: Yeah, that's correct. Because batch sizes in general can be variable, even though for convolutional networks they might need to be fixed, it probably shouldn't be returned here so that it plays nice with all the other layers.

blocks/bricks/conv.py
harmdevries89: Ok, I will change it.

blocks/bricks/conv.py
harmdevries89: I agree.

blocks/bricks/conv.py
harmdevries89: Alright, but we have to be sure to pass batch_size from ConvolutionalSequence to all Convolutional bricks.

harmdevries89:
I agree with your reasoning @bartvm.

harmdevries89:
I've renamed input_dim to image_size ( and not image_shape) to be consistent with filter_size. I only need to merge with @dmitriy-serdyuk implementation, and then we should be ready to go.

bartvm:
You'll also need a rebase, there are merge conflicts currently.

bartvm:
Was rebased and merged in #304.

PR #296, opened 2015-02-17
Fine-tune error interrupt handling. Fixes #287.

### LINKED ISSUE #287: Fix responding to errors, SIGINT and SIGTERM
This is just a place to collect the discussions/wish list regarding error handling. It has changed a few times now but after having a closer look I'm still not very happy with it, for a few reasons:

* `SIGTERM` is not handled
* `logger.info` calls are used to report errors (lines [121](https://github.com/bartvm/blocks/blob/f440563fd5c4f8078e9c9003dc2412570a29348f/blocks/main_loop.py#L121) and [126](https://github.com/bartvm/blocks/blob/f440563fd5c4f8078e9c9003dc2412570a29348f/blocks/main_loop.py#L126)); these aren't shown by default, so the user has no idea what is going on. They are errors, so they should use `logger.error`, but...
* `logger.error` calls currently result in mysterious `no handler found` errors with no further information to the user; we either need to use `logging.error()` or call `logging.basicConfig()`
* Right now the first `SIGINT` just tries to finish the batch, but the second one quits halfway the batch and then tries to run the `after_training` extensions, which could screw things up more than it could fix.
* The `on_interrupt` callback isn't used so far, and maybe we should remove it until a good use case comes along?

So, better behavior I imagine would be:

* Call `logging.basicConfig()` in `blocks/__init__.py` to make sure warnings and errors get through
* `SIGINT` behavior:
  * The first one should result in a `WARNING` message, saying that we will try to finish the epoch.
  * A second one will raise a message saying that it will try to finish the batch. In both cases, the `after_training` extensions should be run if we manage to finish the epoch/batch successfully.
  * A third `SIGINT` during training ends the program immediately; no `after_training` extensions will be run.
* `SIGTERM` behavior is to finish the batch and quit (as if two `SIGINT`s were sent).
* An error during the main loop should result in `on_error` extensions being called, but *not* in `after_training` being called. The `on_error` behavior right now is good I think; we should try to run each and every extension there to make sure each one gets the chance to do damage control.
* Logging level could be set by a global configuration flag which defaults to `INFO`

bartvm:
This implements I think all of the changes discussed in #287.

### DIFF
--- a/blocks/main_loop.py
+++ b/blocks/main_loop.py
@@ -7,15 +7,33 @@ from blocks import config
 from blocks.log import TrainingLog
 from blocks.utils import reraise_as, unpack, change_recursion_limit
 
+logging.basicConfig()
 logger = logging.getLogger(__name__)
 
-error_message = (
-    """An error occurred during the training.
+error_message = """
 
-Attempting to run `on_error` extensions before exiting...""")
+Blocks will attempt to run `on_error` extensions, potentially saving data, \
+before exiting and reraising the error. Note that the usual `after_training` \
+extensions will *not* be run. The original error will be re-raised and also \
+stored in the training log. Press CTRL + C to halt Blocks immediately."""
 
-error_in_error_handling_message = (
-    """While calling `on_error` extension another error occured.""")
+error_in_error_handling_message = """
+
+Blocks will now exit. The remaining `on_error` extensions will not be run."""
+
+
+epoch_interrupt_message = """
+
+Blocks will complete this epoch iteration of training and run extensions \
+before exiting. If you do not want to complete this epoch, press CTRL + C \
+again to stop training after the current batch."""
+
+batch_interrupt_message = """
+
+Blocks will complete the current batch and run extensions before exiting. If \
+you do not want to complete this batch, press CTRL + C again. WARNING: Note \
+that this will end training immediately, and extensions that e.g. save your \
+training progress won't be run."""
 
 
 class MainLoop(object):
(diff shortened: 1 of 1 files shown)


blocks/main_loop.py
```diff
@@ -7,15 +7,33 @@
...
+Blocks will attempt to run `on_error` extensions, potentially saving data, \
+before exiting and reraising the error. Note that the usual `after_training` \
+extensions will *not* be run. The original error will be re-raised and also \
+stored in the training log. Press CTRL + C to halt Blocks immediately."""
 
-error_in_error_handling_message = (
-    """While calling `on_error` extension another error occured.""")
+error_in_error_handling_message = """
+
+Blocks will continue running the remaining `on_error` extensions."""
```
rizar: I think it is not true: after any of the extensions crashes all the rest will simply not be run.

rizar:
LGTM!

blocks/main_loop.py
bartvm: You're right. I guess we could implement that, but I'm more inclined to say that the `on_error` extensions just need to be super careful.

blocks/main_loop.py
```diff
@@ -178,28 +199,52 @@ def _run_iteration(self):
...
         # the iteration the corresponding log record can be found only in
         # the previous row.
         if (self.log.current_row.training_finish_requested or
-                self.log.current_row.keyboard_interrupt_received or
-                self.log.previous_row.keyboard_interrupt_received):
+                self.log.current_row.batch_interrupt_received or
+                self.log.previous_row.batch_interrupt_received):
+            raise TrainingFinish
+        if (level == 'epoch' and
+                (self.log.current_row.epoch_interrupt_received or
```
rizar: Don't know how I could miss that, but of course that does not work. Assume you make a record `epoch_interrupt_received` at the iteration 234, and your epoch ends at iteration 400. Neither the current nor the previous row can contain this record.

I guess we on the keyboard interrupts we need to make records into both log and status. Into the log, in order to be able to reconstruct what happened later, into the status, in order to for the information to be easily accessible.

Same change could be applied to `batch_interrupt_received` to avoid looking into the previous row.

bartvm: Right, silly me. Not sure if I have time to fix it today, but I'll try to get to it soon.

PR #298, opened 2015-02-17
Model

rizar:
This pull request introduces a `Model` interface and its subclass `DifferentiableCostModel`. The purpose of these classes is to be passed as a `model` argument to the `MainLoop` constructor.

## Why do we need it? 

The shortest answer: because the assumption that there is always a differentiable cost is to heavy. I would not like to see it hard-coded everywhere. 

More verbosely:

* There is a need for an interface to model parameters. For instance the `Dump` extension needs them. If we let it extract them directly from the computation graph we a) can never interfere into this process b) we are bound to differentiable costs only.

* It would be nice to give extension access to the optimization criterion. That would allow very nice defaults, e.g. the `DataStreamMonitoring` in the case of a simple Theano cost could automatically add it to the list of monitored variables. 

I will answer the concern that have been already raised via email in subsequent comments.

### DIFF
--- a/blocks/main_loop.py
+++ b/blocks/main_loop.py
@@ -136,10 +136,10 @@ class MainLoop(object):
             # Sanity check: model and algorithm should be configured
             # similarly.
             if not self._model.get_objective() == self.algorithm.cost:
-                raise ValueError("different costs for model and algorithm")
+                logger.warning("different costs for model and algorithm")
             if not (set(self._model.get_params().values()) ==
                     set(self.algorithm.params)):
-                raise ValueError("different params for model and algorithm")
+                logger.warning("different params for model and algorithm")
 
         with change_recursion_limit(config.recursion_limit):
             self.original_sigint_handler = signal.signal(
--- a/blocks/main_loop.py
+++ b/blocks/main_loop.py
@@ -62,22 +62,22 @@ class MainLoop(object):
 
     Parameters
     ----------
-    model : object
-        The model object. It is entirely transparent for the main loop
-        but may be used by extensions.
-    data_stream : instance of :class:`.DataStream`.
-        The data stream.
     algorithm : object
         The training algorithm.
-    log : instance of :class:`.TrainingLog`
+    data_stream : instance of :class:`.DataStream`.
+        The data stream.
+    model : :class:`.AbstractModel` instance, optional
+        The model object. It is entirely transparent for the main loop
+        but may be used by extensions.
+    log : instance of :class:`.TrainingLog`, optional
         The log. When not given, a :class:`.TrainingLog` is created.
     extensions : list of :class:`.TrainingExtension` instances
         The training extensions. Will be called in the same order as given
         here.
 
     """
-    def __init__(self, model, data_stream, algorithm,
-                 log=None, extensions=None):
+    def __init__(self, algorithm, data_stream,
+                 model=None, log=None, extensions=None):
         self.model = model
         self.data_stream = data_stream
         self.algorithm = algorithm
@@ -123,7 +123,6 @@ class MainLoop(object):
                 if not self.status._training_started:
                     for extension in self.extensions:
                         extension.main_loop = self
-                    self.algorithm.log = self.log
                     self._run_extensions('before_training')
                     self.algorithm.initialize()
                     self.status._training_started = True
--- a/blocks/main_loop.py
+++ b/blocks/main_loop.py
@@ -102,7 +102,8 @@ class MainLoop(object):
     @property
     def model(self):
         if not self._model:
-            raise ValueError("no model in this main loop" + no_model_message)
+            raise AttributeError("no model in this main loop"
+                                 + no_model_message)
         return self._model
 
     @property
--- a/blocks/main_loop.py
+++ b/blocks/main_loop.py
@@ -35,6 +35,11 @@ you do not want to complete this batch, press CTRL + C again. WARNING: Note \
 that this will end training immediately, and extensions that e.g. save your \
 training progress won't be run."""
 
+no_model_message = """
+
+A possible reason: one of your extensions requires the main loop to have \
+a model. Check documentation of your extensions."""
+
 
 class MainLoop(object):
     """The standard main loop of Blocks.
@@ -78,19 +83,26 @@ class MainLoop(object):
     """
     def __init__(self, algorithm, data_stream,
                  model=None, log=None, extensions=None):
-        self.model = model
-        self.data_stream = data_stream
-        self.algorithm = algorithm
-
         if not log:
             log = TrainingLog()
         if not extensions:
             extensions = []
+
+        self.data_stream = data_stream
+        self.algorithm = algorithm
         self.log = log
         self.extensions = extensions
 
+        self._model = model
+
         self.status._training_started = False
         self.status._epoch_started = False
+
+    @property
+    def model(self):
+        if not self._model:
+            raise ValueError("no model in this main loop" + no_model_message)
+        return self._model
 
     @property
     def iteration_state(self):
--- a/blocks/main_loop.py
+++ b/blocks/main_loop.py
@@ -6,6 +6,7 @@ import traceback
 from blocks import config
 from blocks.log import TrainingLog
 from blocks.utils import reraise_as, unpack, change_recursion_limit
+from blocks.algorithms import DifferentiableCostMinimizer
 
 logging.basicConfig()
 logger = logging.getLogger(__name__)
@@ -128,6 +129,17 @@ class MainLoop(object):
         # This should do nothing if the user has already configured
         # logging, and will it least enable error messages otherwise.
         logging.basicConfig()
+
+        if self._model and isinstance(self.algorithm,
+                                      DifferentiableCostMinimizer):
+            # Sanity check: model and algorithm should be configured
+            # similarly.
+            if not self._model.get_objective() == self.algorithm.cost:
+                raise ValueError("different costs for model and algorithm")
+            if not (set(self._model.get_params().values()) ==
+                    set(self.algorithm.params)):
+                raise ValueError("different params for model and algorithm")
+
         with change_recursion_limit(config.recursion_limit):
             self.original_sigint_handler = signal.signal(
                 signal.SIGINT, self._handle_epoch_interrupt)
--- a/blocks/main_loop.py
+++ b/blocks/main_loop.py
@@ -125,6 +125,9 @@ class MainLoop(object):
         a `training_finish_requested` record in the log.
 
         """
+        # This should do nothing if the user has already configured
+        # logging, and will it least enable error messages otherwise.
+        logging.basicConfig()
         with change_recursion_limit(config.recursion_limit):
             self.original_sigint_handler = signal.signal(
                 signal.SIGINT, self._handle_epoch_interrupt)
--- a/blocks/main_loop.py
+++ b/blocks/main_loop.py
@@ -102,8 +102,8 @@ class MainLoop(object):
     @property
     def model(self):
         if not self._model:
-            raise AttributeError("no model in this main loop"
-                                 + no_model_message)
+            raise AttributeError("no model in this main loop" +
+                                 no_model_message)
         return self._model
 
     @property
--- a/blocks/model.py
+++ b/blocks/model.py
@@ -1,0 +1,162 @@
+from abc import ABCMeta, abstractmethod
+from collections import OrderedDict
+from itertools import chain
+
+from six import add_metaclass
+
+from blocks.graph import ComputationGraph
+from blocks.select import Selector
+from blocks.filter import VariableFilter, get_brick
+from blocks.roles import PARAMETER
+
+@add_metaclass(ABCMeta)
+class Model(object):
+    """A parameterized entity trained in the main loop.
+
+    A model is a parameterized entity the user trains a in a main loop.
+    The following are traits of every model:
+
+    * It has parameters and supports a way to access them. In addition
+    to returning handles to parameter objects it can return their values
+    as numpy arrays and set their values to given numpy arrays.
+
+    * It has an optimality criterion.
+
+    * It can be serialized and deserialized by mean of pickling.
+
+    * It might have bricks as its components.
+
+    This class provides an interface for models. For experiments use
+    a subclass, e.g. the :class:`DifferentiableCostModel`.
+
+    """
+
+    @abstractmethod
+    def get_params(self):
+        """Return the model parameters.
+
+        Returns
+        -------
+        params : OrderedDict
+            Dictionary of (name, parameter) pairs.
+
+        """
+        pass
+
+    def get_param_values(self):
+        """Return the values of model parameters.
+
+        The default implementation assumes that parameters are Theano
+        shared variables.
+
+        Returns
+        -------
+        param_values : OrderedDict
+            Dictionary of (parameter name, :class:`~numpy.ndarray`) pairs.
+
+        """
+        return OrderedDict((name, param.get_value())
+                           for name, param in self.get_params().items())
+
+
+    def set_param_values(self, param_values):
+        """Set the values of model parameters.
+
+        The default implementation assumes that parameters are Theano
+        shared variables.
+
+        Parameters
+        ----------
+        param_values : OrderedDict
+            Dictionary of (parameter name, :class:`~numpy.ndarray`) pairs.
+
+        """
+        params = self.get_params()
+        for name, value in param_values:
+            params[name].set_value(value)
+
+
+    @abstractmethod
+    def get_criterion(self):
+        """Return the optimization criterion."""
+        pass
+
+    @abstractmethod
+    def get_top_bricks(self):
+        """Return the top-level bricks that are used in the model.
+
+        Returns
+        -------
+        bricks : list
+            List of bricks.
+
+        """
+        pass
+
+
+class DifferentibleCostModel(Model):
+    """A model for which a differentiable Theano cost is available.
+
+    This model covers the most common case when the criterion is Theano
+    variable, the parameter are shared Theano variables. In such case
+    the parameters can be typically extracted from the computation graph,
+    as well as bricks.
+
+    The parameter names are formed from positions
+    of their owner brick in the bricks hierarchy. The variable names are
+    used for the parameters that do not belong to any brick.
+
+    Parameters
+    ----------
+    cost : `~theano.Varible`
+        The cost variable.
+
+    .. todo::
+
+        Overriding the automatically found parameters and bricks might
+        be needed.
+
+        If there top bricks in scan inner graphs, those will not be found.
+
+    """
+    def __init__(self, cost):
+        self.cost = cost
+
+        self._cg = ComputationGraph(cost)
+
+        bricks = [get_brick(var) for var in self._cg if get_brick(var)]
+        children = set(chain(*(brick.children for brick in bricks)))
+        # Quadratic complexity: we should not have thousands of
+        # top-level bricks.
+        self.top_bricks = []
+        for brick in bricks:
+            if not brick in children and not brick in self.top_bricks:
+                self.top_bricks.append(brick)
+
+        brick_param_names = {
+            v: k for k, v in Selector(self.top_bricks).get_params().items()}
+
+        self.params = []
+        for param in VariableFilter(roles=[PARAMETER])(
+                self._cg.shared_variables):
+            if param in brick_param_names:
+                self.params.append((brick_param_names[param], param))
+            else:
+                self.params.append((param.name, param))
+        self.param = OrderedDict(self.params)
+
+    def get_criterion(self):
+        return self.cost
+
+    def get_params(self):
+        """Get model parameters.
+
+        The parameter names are formed from positions of their owner bricks in
+        the bricks hierarchy. The variable names are used for the parameters
+        that do not belong to any brick.
+
+        """
+        return self.params
+
+    def get_top_bricks(self):
+        return self.top_bricks
--- a/blocks/model.py
+++ b/blocks/model.py
@@ -127,17 +127,18 @@ class Model(AbstractModel, ComputationGraph):
     Due to frequency of this case this class is called simply 'Model'
     and not 'ComputationGraphModel'.
 
-    Parameters
-    ----------
-    outputs : (list of) :class:`~theano.Variable`
-        The output variables of the computation graph.
-
     .. todo::
 
         Overriding the automatically found parameters and bricks might
         be needed.
 
-        If there top bricks in scan inner graphs, those will not be found.
+        If there are top bricks in scan inner graphs, those will not be
+        found.
+
+    Parameters
+    ----------
+    outputs : (list of) :class:`~theano.Variable`
+        The output variables of the computation graph.
 
     """
     def __init__(self, outputs):
--- a/blocks/model.py
+++ b/blocks/model.py
@@ -9,6 +9,7 @@ from blocks.select import Selector
 from blocks.filter import VariableFilter, get_brick
 from blocks.roles import PARAMETER
 
+
 @add_metaclass(ABCMeta)
 class Model(object):
     """A parameterized entity trained in the main loop.
@@ -30,7 +31,6 @@ class Model(object):
     a subclass, e.g. the :class:`DifferentiableCostModel`.
 
     """
-
     @abstractmethod
     def get_params(self):
         """Return the model parameters.
@@ -58,7 +58,6 @@ class Model(object):
         return OrderedDict((name, param.get_value())
                            for name, param in self.get_params().items())
 
-
     def set_param_values(self, param_values):
         """Set the values of model parameters.
 
@@ -72,9 +71,8 @@ class Model(object):
 
         """
         params = self.get_params()
-        for name, value in param_values:
+        for name, value in param_values.items():
             params[name].set_value(value)
-
 
     @abstractmethod
     def get_criterion(self):
@@ -94,7 +92,7 @@ class Model(object):
         pass
 
 
-class DifferentibleCostModel(Model):
+class DifferentiableCostModel(Model):
     """A model for which a differentiable Theano cost is available.
 
     This model covers the most common case when the criterion is Theano
@@ -130,7 +128,7 @@ class DifferentibleCostModel(Model):
         # top-level bricks.
         self.top_bricks = []
         for brick in bricks:
-            if not brick in children and not brick in self.top_bricks:
+            if brick not in children and brick not in self.top_bricks:
                 self.top_bricks.append(brick)
 
         brick_param_names = {
@@ -143,7 +141,7 @@ class DifferentibleCostModel(Model):
                 self.params.append((brick_param_names[param], param))
             else:
                 self.params.append((param.name, param))
-        self.param = OrderedDict(self.params)
+        self.params = OrderedDict(self.params)
 
     def get_criterion(self):
         return self.cost
@@ -151,9 +149,9 @@ class DifferentibleCostModel(Model):
     def get_params(self):
         """Get model parameters.
 
-        The parameter names are formed from positions of their owner bricks in
-        the bricks hierarchy. The variable names are used for the parameters
-        that do not belong to any brick.
+        The parameter names are formed from positions of their owner bricks
+        in the bricks hierarchy. The variable names are used for the
+        parameters that do not belong to any brick.
 
         """
         return self.params
--- a/blocks/model.py
+++ b/blocks/model.py
@@ -1,3 +1,11 @@
+"""Defines models.
+
+A model is a thin layer of abstraction between the user-defined computation
+graph, bricks, parameters and main loop extensions. This module provides
+the basic :class:`AbstractModel` interface as well as its implementations
+(currently only :class:`Model`).
+
+"""
 from abc import ABCMeta, abstractmethod
 from collections import OrderedDict
 from itertools import chain
@@ -11,7 +19,7 @@ from blocks.roles import PARAMETER
 
 
 @add_metaclass(ABCMeta)
-class Model(object):
+class AbstractModel(object):
     """A parameterized entity trained in the main loop.
 
     A model is a parameterized entity the user trains a in a main loop.
@@ -28,7 +36,7 @@ class Model(object):
     * It might have bricks as its components.
 
     This class provides an interface for models. For experiments use
-    a subclass, e.g. the :class:`DifferentiableCostModel`.
+    a subclass, e.g. the :class:`Model`.
 
     """
     @abstractmethod
@@ -92,22 +100,19 @@ class Model(object):
         pass
 
 
-class DifferentiableCostModel(Model):
-    """A model for which a differentiable Theano cost is available.
+class Model(AbstractModel, ComputationGraph):
+    """Wraps a computation graph to support model interface.
 
-    This model covers the most common case when the criterion is Theano
-    variable, the parameter are shared Theano variables. In such case
-    the parameters can be typically extracted from the computation graph,
-    as well as bricks.
-
-    The parameter names are formed from positions
-    of their owner brick in the bricks hierarchy. The variable names are
-    used for the parameters that do not belong to any brick.
+    This model covers the most common case when all information
+    about the model is contained in an annotated computation graph:
+    parameters are identified by the roles, bricks found by annotations.
+    Due to frequency of this case this class is called simply 'Model'
+    and not 'ComputationGraphModel'.
 
     Parameters
     ----------
-    cost : `~theano.Varible`
-        The cost variable.
+    outputs : (list of) :class:`~theano.Variable`
+        The output variables of the computation graph.
 
     .. todo::
 
@@ -117,12 +122,10 @@ class DifferentiableCostModel(Model):
         If there top bricks in scan inner graphs, those will not be found.
 
     """
-    def __init__(self, cost):
-        self.cost = cost
+    def __init__(self, outputs):
+        super(Model, self).__init__(outputs)
 
-        self._cg = ComputationGraph(cost)
-
-        bricks = [get_brick(var) for var in self._cg if get_brick(var)]
+        bricks = [get_brick(var) for var in self if get_brick(var)]
         children = set(chain(*(brick.children for brick in bricks)))
         # Quadratic complexity: we should not have thousands of
         # top-level bricks.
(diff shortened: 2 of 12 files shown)


rizar:
@bartvm says:

> As far as I can tell, the Theano graph holds all the information you need: The cost 
> variable, the parameters. 

Indeed it does. But if we hard-code retrieval of parameters and bricks from the computation graph everywhere, I feel it can sting us so badly later. World is changing rapidly: what if I have a few costs and I alternate optimizing them? Then I can not simply get all parameters from one of them. Think about implementing target propagation in Blocks: would you try to enforce the constraint that there is a single cost for it? 

> I would very much discourage the introduction of a Model abstraction; I currently always 
> pass None and have never had any issues. The only extension that uses it is "Dump", and I >  think that can be resolved in the way I mentioned.

I am curious how you can work by passing `None` as a model. Suppose I want to translate sentences with a trained model. First I have to deserialize the whole main loop pickle in this case just in order to get the cost out of it. This is slow, because unpickling the algorithm involves a lot of hidden Theano activity. Then I have to extract bricks from the computation graph. The respective code should be either copy-pasted or be made a separate routine - but this is what I do here.

Instead if would be better if using a model would be as simple as loading the _model_ pickle and using `get_top_bricks` and `get_criterion`.

rizar:
@vdumoulin says

> I agree that defining a Model class is a hard thing to do, since machine learning is a field 
> that is diverse and is evolving very quickly. I think you made the right decision by stepping 
> out of the question and focusing on a toolset that facilitates building computation graphs.

I agree with almost every word of it. However, there are a few common traits that _all_ models I can think of share:

* they have some parameters, and whatever the parameters are, they typically can be represented as numpy arrays

* they have an associated optimality criterion, no matter what it actually is

* relevant for Blocks - they might have related bricks

I want to establish a standard interface for querying such things, that's it.

rizar:
Answering @bartvm 

> For example, the idea of the model having "a cost" can already be too specific; what if 
> you're alternately minimizing two separate costs? An example of a non-Theano cost could 
> be BLEU score.

This is very true, but in this PR a model has an optimization criterion, which can be anything and therefore is unique. However, when the `model` is an instance of `DiffentiableCostMinimizer`, an extension can already try to do something useful with `get_criterion`, because it simply returns the cost variable.

> Make the Dump extension work on Theano graphs directly as I outlined in my first e-mail

This is exactly what I try to avoid: having the assumption of a single cost graph hard-coded everywhere. There are two places where this is surely true: `DifferentiableCostModel` and `DifferentiableCostMinimizer`, and let's keep all these slippery things done there. 

> Keep the current behavior of the Dump extension, where it pickles a Brick, as an optional feature

`Dump` extension does not pickle bricks, there is a misunderstanding here.

rizar:
Just to sum it up: I propose here a lean and mean model. It is an optional component, which becomes mandatory when you want your extensions do smart things in a compatible way. I will pause working on it waiting for your feedback on what currently is available.

bartvm:
Sorry, still very strongly dislike the idea of introducing this abstraction, and I still fail to see why it is needed.

I know that `Dump` doesn't actually serialize bricks, but it does depend on the presence of them in order to retrieve parameters. I would like to remove that dependency. I agree that the idea of a single *cost* graph is restrictive, and I never meant to propose that. A single *computation graph* isn't restrictive however (because it can have multiple outputs of completely disjoint graphs), and we already have an interface to that through the `ComputationGraph` class.

Variable roles provide can provide you easy access to the loss function, which is something else you want. Lastly, wrapping our model in a single top brick would remove the need for the `Model` abstraction entirely.

```python
# This is a single brick which contains the entire brick hierarchy
model = TopBrick()
# Dumping the parameters however, doesn't rely on these bricks
# it accesses the shared variables directly (note that y can be a list of Theano
# variables, so it can represent *multiple* cost objectives)
y = model.apply(x)
# The dump method simply walks the graph, finds shared variables and stores
# their values in a variable name => value dictionary. This means it works
# if you don't have a single top brick as well
dump('params.npz', y)
# Reloading the parameters requires a computation graph with the same variable names
load_dump('params.npz', y)

# However, you want the simple case where we can reconstruct a Theano
# graph, load its parameters, and run with it e.g. to translate sentences, reuse the VAE
# as part of another model, etc. For this, we simply serialize the top brick as well
save_model(model, y, 'params.npz', 'model.pkl')
# Note: In the background, dump_model attaches the input variables of `y` to `model`
model.model_inputs = ComputationGraph(z).inputs

# Now for the loading
load_model('model.pkl', 'params.npz')
# In the background, load_model does something like
model = cPickle.load(open('model.pkl'))
# We now have the top brick, so we can recreate the Theano graph
model_inputs = model.model_inputs
y = model.apply(*model_inputs)
# And restore the parameter values
load_dump('params.npz', y)
# If we want to access the loss criteria the model was trained with, we do it like usual
objectives = VariableFilter(roles=[OBJECTIVE])(ComputationGraph(y))
# We can simply say that some extensions only work as long as the computation graph
# only has a single variable tagged with this role, which is basically your "DifferentiableModel" case
```

This avoids the need of a new abstraction, a `Model` is simply a `Brick` which has its Theano input variables stored with it. The `Dump` extension can rely on this to save and reload models.

However, extensions should be strongly discouraged from relying on the existence of a single top-brick whenever possible and hence use `VariableFilter(roles=[OBJECTIVE])` instead of `get_criterion`. This way we stick with IMHO is still the core philosophy of the library: Bricks annotate the graph. We then interact with these annotations (not with the bricks!) whenever possible, in order to allow people to mix bricks and raw Theano freely without the need for further abstractions.

rizar:
> I know that Dump doesn't actually serialize bricks, but it does depend on the presence of 
> them in order to retrieve parameters. I would like to remove that dependency. 

The `Model` layer introduced by this PR removes this dependency entirely. Check the current version: no occurrence of `brick` in `dump.py`. And there is no other way to do it.  The hacks with renaming the parameters after creation of the computation graph will face my strong opposition. Otherwise, there should be a way to query the names, and `model.get_params` is a way of doing this.

bartvm:
If that is all you want to do, then I really see no need to introduce a new abstraction with a grandiose name like `Model`. This logic you could simply put in the `Dump` extension (or factor it out into a function that goes in `utils`), and `Dump` will take as an argument the cost variable, with the docstring: "WARNING: This extension will only save parameters that were created by bricks."

It seems minor, but in my mind it's a completely different idea: I'm okay with an extension having the limitation that it makes certain assumptions about whether and how the computation graph is annotated. On the other hand, expecting a user to wrap their computation graph in some sort of `Model` class that depends on bricks, and which is a direct argument for the main loop violates the idea of the main loop being brick-agnostic, model-agnostic, and Theano-first.

rizar:
```python
y = model.apply(*model_inputs)
```

This line has a _hell_ lot of assumptions, of which the main are: all the inputs of the application method are positional, there are no other inputs of the application methods but for the inputs of the CG. What if I used a boolean flag when building my cost computation graph, for instance?

In general there is no easier way to reconstruct the cost graph than to create it from scratch. For this reason I want it, not only bricks, to be saved. It is currently saved, but as a part of algorithm, which I consider very inconvenient.

bartvm:
Okay, so that's definitely one point of confusion. I thought the entire point of the `Dump` extension (and the `Model` you want to use for it) was that you wanted it to be entirely cross-platform, which means that you *can't* save the computation graph, because it won't unpickle between CPU/GPU, etc.

Hence, your only option would be to save a way of *reconstructing* the computation graph, for which you would need to save the top-brick only. However, if you are saying that this is not what you want, but you just want to an easy way to retrieve the model, then you need even less abstraction. 

The `Dump` extension takes a list of Theano output variables as an argument. The extension is in charge of walking the computation graph, finding a mapping between the bricks hierarchy and parameters (potentially through a function that is factored out into `utils.py`), and saving them. Let it take an optional keyword argument `save_graph=True` which pickles the computation graph to a separate file.

If you then want to easily reload your model, you just do something like `DumpManager.load_model` which will unpickle the graph, restore the parameters, and return the user the computation graph. (And if you want to, you can add all sorts of nifty helper functions that retrieve the top-level brick from this computation graph of course, if you want the bricks instead of the Theano variables.)

This basically performs exactly the same logic as you have right now, it's just that instead of passing `DifferentiableCostMost(cost)` to `MainLoop`, you pass `cost` to `Dump`.

rizar:
You say "just pass the cost the to the Dump" as if it is the only expression that will ever need access to the parameters, but this is not so. Even now there is (a rather ugly) `LoadFromDump` extension that needs exactly the same rights. 

And there will be more: for instance I like to have shapes and name of my parameter printed when I run the training script: it helps a lot. So I want to be able to write an extension `ShowParameters` and use it in all my scripts. I do not want to pass the cost to this extension as well. Finally, I think that later we should add the cost to the list of observables automatically, perhaps also scanning for some variable with `Diagnostic` roles. That is, a lot of extensions need the cost and/or parameters and there should be a way to query them from the main loop they belong to. Passing `cost` to all extensions is not an option.

rizar:
> Okay, so that's definitely one point of confusion. I thought the entire point of the Dump 
> extension (and the Model you want to use for it) was that you wanted it to be entirely 
> cross-platform, which means that you can't save the computation graph, because it won't 
> unpickle between CPU/GPU, etc.

> Hence, your only option would be to save a way of reconstructing the computation graph, 
> for which you would need to save the top-brick only. However, if you are saying that this is 
> not what you want, but you just want to an easy way to retrieve the model, then you need 
> even less abstraction. 

This is all true, but I did not say that I will only use the `Dump` extension. In fact in many cases (and when finally CPU-GPU unpickling is solved) I would rather play with the pickled model, the one saved by using the `save_separately` argument of the `SerializeMainLoop` extension. And I would really like this model to contain the cost: it makes my current experiments much harder that it is only available in the algorithm.

bartvm:
Okay, so then you are trying to do something completely different from what I thought you were. It seemed to me that you were adding a lot of unnecessary abstraction just to be able to dump a computation graph, while the whole point of dump was to *not* pickle the computation graph.

But if your goal is to add a thin layer of abstraction that extensions can use to retrieve parameters, that seems reasonable. In that case however, I have an entirely different set of reservations :) Can't make it too easy on you!

* I think `model` shouldn't be a positional argument; this is something that has been bugging me for a while, because I pass `None` in many cases, which clearly shows it should be a keyword argument
    * We can make `model` a property of `MainLoop` in this case that raises an informative error when an extension tries to use a model when none was passed.
* Would it be possible to make `Model` a subclass of `ComputationGraph`? They share so many things in common: They walk to computation graph, retrieve annotations... It seems that the only thing that `DifferentiableCostModel` adds is that it adds a few convenience methods.
* I don't see why `DifferentiableCostModel` is called that way; it doesn't rely on things being differentiable at any point. An assumption it does make is that there is a single cost. However, a lot of the logic in that class would work just fine with multiple outputs as well. So I would propose a `BricksModel` class (because that is an assumption it does make; the existence of bricks in order to name parameters) which deals with multiple outputs, which can have a subclass, `SingleCostBricksModel` (please come up with something better than that) which makes the assumption of having a single cost.

rizar:
> It seemed to me that you were adding a lot of unnecessary abstraction just to be able to 
> dump a computation graph, while the whole point of dump was to not pickle the 
> computation graph.

Sorry, 10k kilometers make communication once in a while quite hard. Just to make it clear: `Dump` needs only `get_param_values` and `set_param_values`.

I am fine with making `model` a keyword argument, though I still do not understand your workflow. Do you actually start from the `algorithm.cost` when you want to explore the trained bricks? Or do are you only interested in the final error and not use the trained bricks for any  other purpose?

As for subclassing from the computation graph: this is a tricky question. I guess I prefer to keep the abstract interface separate from all the rest. The current `DifferentiableCostModel` can indeed inherit from the `ComputationGraph`. 

Regarding renaming: I want to notice that for the current `DifferentiableCostModel` bricks are optional. If a parameter does not belong to any brick, its name will be used (see https://github.com/bartvm/blocks/blob/model/blocks/model.py#L143). So `BricksModel` would be a bit too strong. I chose the name by analogy with the `DiffentiableCostMinimizer`, but here we indeed do not talk much about gradients. Considering the fact that this model should cover 99% cases, we could call it `Model`, leaving a longer `AbstactModel` to the interface.

There is one more question I would like to raise: I do not like that parameter logic will be duplicated in `Model` and `DifferentiableCostMinimizer`. Do you think we should do something with it?

bartvm:
Yeah, in my case I was doing optimization, and I really only care about how the error goes down, no interest in exploring the actual trained model.

I noticed they were optional, so yeah, in that case just calling it `Model` might be a good idea, assuming that we refactor it to support multiple outputs/costs.

Which logic is being duplicated exactly? `DifferentiableCostMinimizer` just retrieves all the shared variables from the graph, right?

rizar:
I am not sure if your case should be consider as the main one. The beginners might find it helpful if we save the model for them by default. We can even hide the `Model` call in the `MainLoop`, as this is clearly the default model, then the MNIST level users will simply give the cost as the first argument to the main loop.

Supporting multiple outputs is super-easy. 

Yes, currently it just retrieves all the shared variables. But this should be changed to selecting all variables with `PARAMETER` roles, otherwise dropout is impossible to support.

bartvm:
It's quite easy for users to get in the habit of passing `model=Model(cost)`, which is a minimal burden considering the large numbers of keyword arguments they need to pass anyway. On the other hand, I don't see any benefit from forcing me to pass it when I don't need it. (See the [docs on live plotting](http://blocks.readthedocs.org/en/latest/plotting.html) by the way for an example where I also don't pass `model`.)

I'm also not sure if we should save the model separately by default; for large models serialization can take a long time, and someone who turns on checkpointing as well as dumping would be waiting for ages, because their parameters effectively get saved three times: Once in the main loop, once in the NumPy parameter dump, and once in the model.

We can factor out that parameter selection logic I guess, but maybe into a separate function that goes into `utils.py`, because the algorithm and the model should have nothing to do with each other I think.

rizar:
Okay, let it be a keyword argument.

I was not planning to turn on separate saving of the model by default. In simple cases when loading the whole main loop pickle is feasible this option is definitely the best.

I have been thinking for two weeks about something like having

```python
algorithm.setup(model.get_criterion(), model.get_params())
```

in the `MainLoop` to get rid of this duplication. In such a way the cost is specified only once, the parameters are extracted only once, and not link whatsoever between algorithms and the rest of the library is introduced. But that would lead to significant complications in the algorithm code: it would have to be able to receive these setting both from the user and from the main loop. So I am inclined not to do that and require double specification of the cost. What's your opinion?

bartvm:
I agree with you; I don't have too much of an issue with the user specifying the cost twice. Considering how the framework is set up, it makes sense: The algorithm is agnostic to the model, and the model is agnostic to the algorithm, but both need the cost to be defined, so both take it as an argument.

rizar:
Okay. The only thing I am worried about is that once we might at some point have a silent and unexpected mismatch between algorithm parameters and model parameters. 
I will check that in the `MainLoop`: it knows interfaces of both and can identify such a situation. The warning raised in such cases will be always present when parameters of the algorithm are provided manually, for instance when we want to train only a part of parameters. But I think people can live with this warning when they start doing such dangerous things.

rizar:
Okay, I think I addressed all your comments. In addition

* I removed this provocative reference from the `algorithm` to the `log`. 

* I swapped `algorithm` and `data_stream` in the `MainLoop` argument least. I think that the algorithm is the main and should go first.

rizar:
Okay, I think I am done with this PR. @bartvm , let me know if you are minding merging this.

blocks/model.py
```diff
@@ -0,0 +1,181 @@
...
+        param_values : OrderedDict
+            Dictionary of (parameter name, :class:`~numpy.ndarray`) pairs.
+
+        """
+        params = self.get_params()
+
+        unknown = set(param_values) - set(params)
+        missing = set(params) - set(param_values)
+        if len(unknown):
+            logger.error("unknown parameter names:\n".format(unknown))
```
bartvm: You need to call `logging.basicConfig()` at the beginning of the file to make sure that this call doesn't crash with a "no handler" message. I think we should just move `logger.basicConfig()` to `blocks/__init__.py` though.

blocks/model.py
```diff
@@ -0,0 +1,181 @@
...
+        if len(missing):
+            logger.error("missing values for parameters:\n".format(missing))
+
+        for name, value in param_values.items():
+            if name in params:
+                params[name].set_value(value)
+
+    @abstractmethod
+    def get_criterion(self):
```
bartvm: Maybe we can use `objective` instead of criterion? Unlike cost/loss, objective doesn't necessarily need to be minimized, but it's clearer to me than criterion, which sounds like it would return the standard of measurement (i.e. 'maximum likelihood') instead of its value.

blocks/model.py
```diff
@@ -0,0 +1,181 @@
...
+        """Return the optimization criterion."""
+        pass
+
+    @abstractmethod
```
bartvm: Does this need to be an abstract method? That kind of suggests that a model *must* have bricks, which isn't the case, right?

blocks/model.py
```diff
@@ -0,0 +1,181 @@
...
+    parameters are identified by the roles, bricks found by annotations.
+    Due to frequency of this case this class is called simply 'Model'
+    and not 'ComputationGraphModel'.
+
+    Parameters
+    ----------
+    outputs : (list of) :class:`~theano.Variable`
+        The output variables of the computation graph.
+
+    .. todo::
```
bartvm: Sphinx complains if you have the todo boxes after the parameters heading, just move them to right above the heading.

blocks/model.py
```diff
@@ -0,0 +1,181 @@
...
+        self.params = OrderedDict(self.params)
+
+    def get_criterion(self):
+        """Return the output variable, if there is a single one.
+
+        If there is only one output variable, it is a reasonable default
+        setting to assume that it is the optimization objective.
+
+        """
+        if len(self.outputs) == 1:
```
bartvm: There's the `unpack` function that does this if you want.

blocks/model.py
```diff
@@ -0,0 +1,181 @@
...
+        The default implementation assumes that parameters are Theano
+        shared variables.
+
+        Returns
+        -------
+        param_values : OrderedDict
+            Dictionary of (parameter name, :class:`~numpy.ndarray`) pairs.
+
+        """
+        return OrderedDict((name, param.get_value())
```
bartvm: Why does this need to be an `OrderedDict`? Every name is unique I'm guessing, in which case the order doesn't matter, or does it?

blocks/model.py
```diff
@@ -0,0 +1,181 @@
...
+
+
+class Model(AbstractModel, ComputationGraph):
+    """Wraps a computation graph to support model interface.
+
+    This model covers the most common case when all information
+    about the model is contained in an annotated computation graph:
+    parameters are identified by the roles, bricks found by annotations.
+    Due to frequency of this case this class is called simply 'Model'
+    and not 'ComputationGraphModel'.
```
bartvm: Don't need to do this for this PR, but eventually I think we will want to extend this documentation a bit, to show how a user would go about making their Theano code compatible i.e. they would need to give their parameters a name and add the `PARAMETER` role if they want them to be picked up.

I also feel it might be worthwhile to eventually look into giving a warning if it picks up a shared variable which is not a parameter (the algorithm can shared the code for giving that warning) so that a user gets a hint in case he forgets to do any of this, and ends up with a parameter not being trained.

blocks/model.py
```diff
@@ -0,0 +1,181 @@
...
+        for brick in bricks:
+            if brick not in children and brick not in self.top_bricks:
+                self.top_bricks.append(brick)
+
+        brick_param_names = {
+            v: k for k, v in Selector(self.top_bricks).get_params().items()}
+        self.params = []
+        for param in VariableFilter(roles=[PARAMETER])(self.shared_variables):
+            if param in brick_param_names:
+                self.params.append((brick_param_names[param], param))
```
bartvm: It feels like it would be very easy to create parameters with the same names e.g. `mlp1 = MLP(...), mlp2=MLP(...) y = mlp2.apply(mlp1.apply(x))` will have duplicate names already.

Shouldn't we raise an error in this case? Or is this why you are relying on `OrderedDict`? If so, how sure are we that `toposort` returns exactly the same sorting in all cases? There are many topological sortings of a DAG...

bartvm:
Overall looks good to me, but since it's a pretty important part I guess I made a bunch of smaller comments here and there.

blocks/model.py
rizar: I thought we have discussed this a number of times: having `logging.basicConfig` in the beginning of a file is a bad practice because the user can not run later run `basicConfig` himself, set the desired level, destination and format in one  call, as the Python documentation encourages to do. I do not mind putting `logging.basicConfig()` in `MainLoop.run` though.

blocks/model.py
rizar: As for the first part: that's true. I would say that one a of huge number of documentation tasks that we have, and the highest-priority one, because I just guess that most people will use bricks.

Giving a warning every time random streams are used is too much. When we learn how to filter them out it will make more sense to put a warning here.

blocks/model.py
rizar: Good point, we should raise an error whenever there are (a) two top bricks with the same name (b) two children of a brick with the same name. I will implement (a), we should think about the right place for (b) to be done.

blocks/model.py
rizar: Not exactly. It does not raise an error when there is more than a single element.

blocks/model.py
rizar: This is to keep the original sort-of-topological order.

blocks/model.py
bartvm: Okay then, at least add a `NullHandler` for the main module. That would make these errors pass silently instead of crash with a mysterious "no handler found" message if they occur outside of the main loop. Just put this in `blocks/__init__.py`:

```python
import logging
logger = logging.getLogger(__name__)
logger.addHandler(logging.NullHandler())
```

blocks/model.py
bartvm: When you walk the graph, you can just keep a flag to see what the shared variable is an input to. If it's an input to some sort of random op, then you know it stores random stream data. It's not as easy as an `isinstance` call with the GPU random streams, but should nonetheless be quite straightforward.

Also, keep in mind that right now Blocks doesn't produce any RNG's that could be mistaken for `SharedVariable`; we use MRG_RNG everywhere, no?

blocks/model.py
rizar: That's reasonable, will do that.

blocks/model.py
bartvm: Okay, just for user convenience of being presented the parameters in order I guess?

rizar: Yes, except that I am not happy with the order they are returned in my test. Will investigate that.

blocks/model.py
bartvm: Mm, I see, but maybe we should let the user set the cost as an optional keyword argument instead? It's just kind of strange to support multiple outputs, but not support retrieving the cost in that case. Right now the user would be forced to implement an entirely new `Model` class just if he wants to have two outputs of which the first one is the cost. It's also awkward that you can still instantiate the model, but that it will crash only when you do `get_criterion`.

You could do something like:

```python
"""
Parameters
------------
outputs : list
    A list of :class:`~tensor.TensorVariable` representing the outputs of your model.
cost : :class:`~tensor.TensorVariable`, optional
    If not given, `outputs` is expected to have only a single element (the cost). If
    outputs has multiple elements, you must pass the cost separately. If the cost
    is not passed as part of outputs, it will be added to it.
"""
def __init__(self, outputs, cost=None):
    if len(pack(outputs)) > 1 and cost is None:
        raise ValueError
    if cost is not None:
        if cost not in outputs:
            outputs.append(cost)
        self.cost = cost
    else:
        self.cost = unpack(outputs)
    self.outputs = outputs
```

rizar: The other day you proposed to have the `OBJECTIVE` role instead, and I like that proposal more. But so far I do not think we should care too much about this case until we found the users with concrete needs. So far the only audience I can think of are the target propagation people.

So I do not think it is bad to let the user set parameters and cost manually, but those finetunings of the `Model` code can wait until we see real use-cases, I think. "Premature optimization is ..." (c) you_know_who

blocks/model.py
bartvm: If we really want to be correct then I guess we should call `basicConfig` from within `MainLoop.run` so that we don't configure root handlers the moment import happens. This allows users to do `import blocks.main_loop` and call `logging.basicConfig()` with whatever configuration they like (i.e. we only configure the root handler if the user hasn't done so by the time we get to calling `MainLoop.run()`).

blocks/model.py
bartvm: Yeah, long term I still think we should introduce an `OBJECTIVE` role, but here it just seemed strange to wait with crashing till an extension calls `get_criteration`, if we know that's going to happen in `__init__` already.

blocks/model.py
rizar: Oops, I am taking my words back. `basicConfig` works only when `if len(root.handlers) == 0:`

In fact this "no handler found" thing is a legacy from Python 2, Python 3 automatically calls `basicConfig` when `logger.debug/info/error` is called.

blocks/model.py
rizar: Sorry, I don't understand. For the 99% situation with a single cost `get_criterion` will not crash. When somebody actually comes up with a model which necessarily has a different objective, we should think what to do in this case.

bartvm: All I'm saying is that in that 1% case where someone passes multiple outputs to `__init__`, we know at that point in time that calls to `get_criterion` are going to fail in the future. Yet, we allow `Model` to be initialized, potentially 100 epochs of training being run, before some extension calls `get_criterion` and crashes. All I'm saying is that it might be safer to just raise an error in `__init__` then, or to support passing multiple outputs fully.

blocks/model.py
bartvm: Haha, that's what I meant earlier when I said `basicConfig` is a no-op. However, I think you're still right in saying that you shouldn't just call `basicConfig` on import, because then if a user does this:

```python
import blocks.model
logging.basicConfig(level='DEBUG')
```
Nothing happens, which is confusing. This is because *libraries* are not expected to configure logging, but *applications* are. The thing is that we fall in either category depending on what part of the library people are using (e.g. we should set a handler for the main loop, but not when importing bricks). So basically, for Python 2 I think we only want to call `basicConfig` at the point that we really need a handler but there still isn't one. That allows users to configure the handlers when they want to.

So, maybe we should just call `logging.basicConfig()` right before `logger.error` calls?

blocks/model.py
rizar: Okay, I see your point. But if I check multiple outputs in `__init__` it will make it impossible to inherit a `TwoCostModel` from the `Model` and simply override `get_criterion`. 

Python is a dynamic language and errors inevitably emerge during runtime.

blocks/main_loop.py
```diff
@@ -62,37 +67,44 @@ class MainLoop(object):
...
 
+        self._model = model
+
         self.status._training_started = False
         self.status._epoch_started = False
 
     @property
+    def model(self):
+        if not self._model:
+            raise ValueError("no model in this main loop" + no_model_message)
```
bartvm: This should be an `AttributeError`, I think that would also allow for `hasattr(main_loop, 'model')` checks.

blocks/model.py
bartvm: Well, another option would be to rename `Model` to something else, eh, I don't know, `TheanoModel`? And make it an abstract class from which `Model` inherits, very simply:

```python
def __init__(self, cost):  # cost is a single variable here
    super(Model, self).__init__(cost)

def get_criteration(self):
    return unpack(self.outputs)
```

bartvm: That's just a suggestion though, I'm also happy with a warning in the docstring saying that if a user passes multiple outputs, he can expect `get_criterion` to crash.

blocks/model.py
```diff
@@ -0,0 +1,182 @@
...
+            else:
+                self.params.append((param.name, param))
+        self.params = OrderedDict(self.params)
+
+    def get_objective(self):
```
bartvm: By the way, maybe we can turn all of these into properties? I don't see why they should be methods i.e. `objective`, `params`, `top_bricks`, etc.

blocks/model.py
rizar: On 18.02.2015 16:19, Bart van Merriënboer wrote:
>
> In blocks/model.py 
> <https://github.com/bartvm/blocks/pull/298#discussion_r24908298>:
>
> > +
> > +        The default implementation assumes that parameters are Theano
> > +        shared variables.
> > +
> > +        Parameters
> > +        ----------
> > +        param_values : OrderedDict
> > +            Dictionary of (parameter name, :class:`~numpy.ndarray`) pairs.
> > +
> > +        """
> > +        params = self.get_params()
> > +
> > +        unknown = set(param_values) - set(params)
> > +        missing = set(params) - set(param_values)
> > +        if len(unknown):
> > +            logger.error("unknown parameter names:\n".format(unknown))
>
> Haha, that's what I meant earlier when I said |basicConfig| is a 
> no-op. However, I think you're still right in saying that you 
> shouldn't just call |basicConfig| on import, because then if a user 
> does this:
>
> import  blocks.model
> logging.basicConfig(level='DEBUG')
>
> Nothing happens, which is confusing. This is because /libraries/ are 
> not expected to configure logging, but /applications/ are. The thing 
> is that we fall in either category depending on what part of the 
> library people are using (e.g. we should set a handler for the main 
> loop, but not when importing bricks). So basically, for Python 2 I 
> think we only want to call |basicConfig| at the point that we really 
> need a handler but there still isn't one.
>
Exactly.
>
> That allows users to configure the handlers when they want to.
>
> So, maybe we should just call |logging.basicConfig()| right before 
> |logger.error| calls?
>
That's something that should be done automatically, for which purpose we 
need something like ```logger = blocks_get_logger(__name__)``` in the 
beginning of each file. Out of the scope of this PR, I think.

blocks/model.py
rizar: I think that it is appropriate to have getters here. I like properties, but only in cases when they are sematically "almost" attributes. Here there is a whole graph traversal and exclusion of children underneath the `get_top_bricks` method, and basically any other logic in a case of a more advanced models. 

That is model too abstract to have properties, in my view.

blocks/model.py
bartvm: Indeed out of the scope of this PR, but I would prefer just adding a `NullHandler` in Python 2. Seems slightly overkill to start patching all over Python 2's strange logging behaviour. A `NullHandler()` to the `blocks` logger works isn't great because it makes errors pass silently, but at least it doesn't crash, and it doesn't interfere with user logging configuration:

```python
import logging
logger = logging.getLogger(__name__)
logger.addHandler(logging.NullHandler())
logger.error('Error!')  # Passes silently
logging.basicConfig()
logger.error('Error!')  # Propagates up and prints to stderr
```

blocks/model.py
rizar: I put a warning so far, but we should think it through later. I almost like the idea of having one more class in the chain: `AbstractModel` -> `ComputationGraphModel` -> `Model`.

blocks/model.py
rizar: On 18.02.2015 16:57, Bart van Merriënboer wrote:
>
> In blocks/model.py 
> <https://github.com/bartvm/blocks/pull/298#discussion_r24911935>:
>
> > +
> > +        The default implementation assumes that parameters are Theano
> > +        shared variables.
> > +
> > +        Parameters
> > +        ----------
> > +        param_values : OrderedDict
> > +            Dictionary of (parameter name, :class:`~numpy.ndarray`) pairs.
> > +
> > +        """
> > +        params = self.get_params()
> > +
> > +        unknown = set(param_values) - set(params)
> > +        missing = set(params) - set(param_values)
> > +        if len(unknown):
> > +            logger.error("unknown parameter names:\n".format(unknown))
>
> Indeed out of the scope of this PR, but I would prefer just adding a 
> |NullHandler| in Python 2. Seems slightly overkill to start patching 
> all over Python 2's strange logging behaviour. A |NullHandler()| to 
> the |blocks| logger works isn't great because it makes errors pass 
> silently, but at least it doesn't crash, and it doesn't interfere with 
> user logging configuration:
>
> import  logging
> logger=  logging.getLogger(__name__)
> logger.addHandler(logging.NullHandler())
> logger.error('Error!')# Passes silently
> logging.basicConfig()
> logger.error('Error!')# Propagates up and prints to stderr
I think that will not work. I looked in the `basicConfig` 
implementation, and it is all in the branch corresponding to `if 
len(root.handlers) == 0`. That is we can not add a dummy handler without 
disabling `basicConfig` calls.

It seems we can have `blocks_get_logger` simply return a thin wrapper 
over the logger, that would ensure `basicConfig` is called when no 
handlers are available.

blocks/model.py
rizar: No we do not, but when we are targeting the usecase when bricks are not used, we can not assume that this RNG is always used.

blocks/model.py
bartvm: But we're not adding a `NullHandler` to the root logger! I'm adding it to the logger of the `blocks` namespace. The hierarchy looks like this:
```
root
   blocks
       blocks.main_loop
       blocks.model
```
If you add a `NullHandler` to `blocks` (as in my example, which works, try it) then messages will go to die there if there is no handler for `root`. However, if you call `basicConfig` later on, the messages will be handled in two places: They will go to the `NullHandler`, but they will also be propagated further up the hierarchy to the `root` logger.

bartvm: So just to clarify, what you are saying only applies if you do this:
```python
logging.getLogger().addHandler(logging.NullHandler())
```
While I'm suggesting to do this in `blocks/__init__.py`:
```python
logging.getLogger(__name__).addHandler(logging.NullHandler())
```

rizar: Got it! Sounds like a really nice solution.

rizar:
I think in this version most of your concerns are addressed, @bartvm . Would you mind me merging this PR?

blocks/main_loop.py
```diff
@@ -113,6 +127,20 @@ def run(self):
...
+
+        if self._model and isinstance(self.algorithm,
+                                      DifferentiableCostMinimizer):
+            # Sanity check: model and algorithm should be configured
+            # similarly.
+            if not self._model.get_objective() == self.algorithm.cost:
+                raise ValueError("different costs for model and algorithm")
+            if not (set(self._model.get_params().values()) ==
+                    set(self.algorithm.params)):
+                raise ValueError("different params for model and algorithm")
```
bartvm: Why is this an error, and not a warning? What if I am training just part of the parameters? I thought there would be a warning using `logger.warning`. With this approach I don't see how I'm supposed to implement freezing parameters...

bartvm:
LGTM, except for that one last comment about raising an error when the parameters don't match; I don't understand how parameter freezing is supposed to work now.

blocks/main_loop.py
rizar: Sure, it must be warning! My fingers are living a separate live from my brain.. :)

PR #299, opened 2015-02-17
Allow Sequence bricks to have multiple outputs

vdumoulin:


vdumoulin:
@rizar Here's a fix for the `Sequence` bug.

blocks/bricks/__init__.py
```diff
@@ -464,6 +464,10 @@ def apply(self, input_):
             child_input = output
         return output
 
+    @apply.property('outputs')
```
rizar: That's right, but your should remove `outputs` from the `@application` declaration. It would be also nice to permit multiple input in the same way as you did with outputs.

vdumoulin: Sure, I'll commit something for that.

vdumoulin: Adding multiple inputs support looks a little more complicated, though. At least, with my limited knowledge I don't have a clear idea of how to implement it. Could you give me some starting tips?

rizar: It seems that all you need is a) make `@apply.property('inputs')` b) accept `*args` in `apply c) remove `inputs` from `@application` declaration.

vdumoulin:
@rizar Regarding your suggestion for multiple inputs, I implemented it but I'm getting an error (see the failing unit test in my latest commit).

rizar:
Oops, that's because the `Parallel` brick does not support positional arguments :(

There are two ways to fix this: a) make `Sequence` work with keyword arguments b) make `Parallel` accept positional arguments. (a) seems a very slippery way, (b) is very doable and should be helpful later. You can as well test your changes with a custom fake brick.

vdumoulin:
@rizar I pushed a fix allowing positional arguments in `Parallel`, but I'm not 100% confident that I did it right, could you validate?

rizar:
Everything seems okay, merging!

PR #300, opened 2015-02-17
Introduce blocks-plot

jbornschein:
This certainly needs more polishing before it can be merged.

But I want to use this pull-request to get some feedback regarding whether we want such a script and what features it should provide.

### DIFF
--- a/bin/blocks-plot
+++ b/bin/blocks-plot
@@ -4,6 +4,113 @@ import logging
 import argparse
 
 
+PROMPT_HEADER = """
+
+Type a comma separated list of channels to plot or [q]uit.
+
+Channels may be prefixed by <number>: to refer to a specific experiment \
+and may contain '*' or '?' characters to match multiple channels at once.
+
+"""
+
+IPYTHON_HEADER = """
+
+Complete DataFrames for all experiments can be found in the \
+'experiments[fname]' dictionary. The DataFrame containing only \
+the selected channels can be access as 'matched'
+
+"""
+
+def main(args):
+    import readline
+    import blocks.scripts.plot as plot
+
+    from six import iteritems
+    from six.moves import input
+    from collections import OrderedDict
+    from functools import reduce
+
+    # Load and convert experiments into DataFrames...
+    experiments = OrderedDict()
+    for fname in args.experiments:
+        log = plot.load_log(fname)
+
+        rows_to_keep = [0] + log._status._epoch_ends
+        data_frame = log.to_dataframe()
+        data_frame = data_frame.iloc[rows_to_keep]
+
+        experiments[fname] = data_frame
+
+    # Gain some overview...
+    n_experiments = len(experiments)
+    channels_per_experiment = OrderedDict(
+        [(fname, set(df.columns)) for fname, df in iteritems(experiments)]
+    )
+    all_channels = reduce(set.union, channels_per_experiment.values())
+
+    if args.list:
+        print()
+        plot.print_column_summary(experiments)
+        return
+
+    if args.channels:
+        column_specs = args.channels.split(',')
+        matched = plot.match_column_specs(experiments, column_specs)
+
+        if args.ipython:
+            from IPython import embed
+            print(IPYTHON_HEADER)
+            embed()
+            return
+
+        plot.plot_dataframe(matched)
+        return
+    else:
+        # Interactive mode
+        def completer(text, state):
+            """Completion callback function for readline library."""
+            options = []
+            if text == "":
+                options += [str(i)+":" for i in range(n_experiments)]
+            if ":" in text:
+                exp_id, text = text.split(":")
+                channels = channels_per_experiment.values()[int(exp_id)]
+                options += [
+                    exp_id+":"+ch for ch in channels if ch.startswith(text)
+                ]
+            else:
+                options += [
+                    ch for ch in all_channels if ch.startswith(text)
+                ]
+
+            if state >= len(options):
+                return None
+            return sorted(options)[state]
+
+        readline.parse_and_bind("tab: complete")
+        readline.set_completer_delims(" \t,;")
+        readline.set_completer(completer)
+
+        print()
+        plot.print_column_summary(experiments)
+        print(PROMPT_HEADER)
+        while 1:
+            # Note: input() uses python3 semantics (provided by six)
+            column_spec = input("blocks-plot> ")
+            if column_spec in ["q", "quit", "exit", "e"]:
+                break
+
+            column_specs = column_spec.split(',')
+            matched = plot.match_column_specs(experiments, column_specs)
+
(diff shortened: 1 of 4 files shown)


bin/blocks-plot
```diff
@@ -0,0 +1,256 @@
...
+
+def log_channels(log):
+    """Return all channels contained in blocks.Log object
+    """
+    channels = set()
+    for t, channel, value in log:
+        channels.add(channel)
+    return channels
+
+def log_to_dataframe(log, channels=None, epochs_only=True):
```
rizar: Doesn't it duplicate with https://github.com/bartvm/blocks/blob/master/blocks/log.py#L211? If yes, then it would be nice to move your additional functionality there.

bin/blocks-plot
```diff
@@ -0,0 +1,256 @@
...
+    for t, channel, value in log:
+        # TODO: ensure value is something we actually want to keep.
+        # We probably have to discard all non-scalars as they can not
+        # be stored in a DataFrame.
+        # TODO: ensure that the channels have "homogeneous" data types
+        channels[channel] [t] = value
+
+    df = pandas.DataFrame(channels)
+
+    if epochs_only:
```
rizar: I think your should consider `log.status._epoch_ends`. This is a list of end iterations of all epochs.

bin/blocks-plot
jbornschein: Yes, it does, and I will move it. I didn't replace the dummy implementation in blocks/log.py yet because for a moment it was easier to have it all in one place.

bin/blocks-plot
```diff
@@ -0,0 +1,256 @@
...
+            if len(matched.columns) == 0:
+                print("Your specification did not match any channels.")
+                continue
+
+            plot_dataframe(matched)
+            print()
+
+       
+
+if __name__ == "__main__":
```
rizar: We have a sort of standard way of adding scripts: only the parameter parsing is located in `blocks/bin`, and the main function is located in `blocks/scripts` with some sensible name, so that it could be called also in some other way. See `blocks-continue` for an instance.

bin/blocks-plot
```diff
@@ -0,0 +1,256 @@
...
+Complete DataFrames for all experiments can be found in the 'experiments[fname]'
+dictionary. The DataFrame containing only the selected channels can be access
+as 'matched'
+
+"""
+
+def main(args):
+    # Load and convert experiments into DataFrames...
+    experiments = OrderedDict()
+    for fname in args.experiments:
```
rizar: That might be implemented in future revisions, but ideally I would like to be able to load pickles consisting of a log only as well, and maybe even dataframes.

rizar:
I am still working my way through the code, but it will be very nice to have such a script!

bin/blocks-plot
jbornschein: Oh, there was a working implementation! I removed mine.

bin/blocks-plot
jbornschein: Done. I manually add [0] because for many monitored values it seems reasonable to plot their value before the first iteration.

bin/blocks-plot
jbornschein: Done. 

Is it easier to write unit test if it's done like this? Or what's the rationale behind this pattern?

bartvm: When you install using pip or setup.py it will copy the files in bin/ in a
folder on the user's path, so you can't import from it anymore, so you
won't be able to share code.

I haven't looked at this PR yet, but I'm not 100% sure you'll need to put
the code in blocks/scripts. It depends on whether you think the code could
ever be imported by another script or function. If this script is
interactive, this might not be the case. On the other hand, if the plotting
logic could be useful for other scripts, then put it in blocks/scripts.
Hope that made sense.

blocks/scripts/plot.py
```diff
@@ -0,0 +1,188 @@
...
+    for fname in exp_fnames:
+        logging.info("Loading '{0}'...".format(fname))
+        try:
+            # TODO: Load "dumped" experiments
+            with change_recursion_limit(config.recursion_limit):
+                main_loop = dill.load(open(fname, "rb"))
+        except EnvironmentError:
+            logging.error("Could not open '{}'".format(fname))
+
+        rows_to_keep = [0] + main_loop.log._status._epoch_ends
```
rizar: Here many people would also need plotting all iterations and some sort of rolling average, because on certain datasets an epoch takes a day. But I do not have to implement this, somebody else (very likely me, because I need that a lot) can do it.

blocks/scripts/plot.py
```diff
@@ -0,0 +1,188 @@
...
+from six import iteritems
+from six.moves import input
+from collections import OrderedDict
+from functools import reduce
+
+from blocks import config
+from blocks.utils import change_recursion_limit
+
+
+def match_channel_specs(experiments, channel_specs):
```
rizar: This is a very nice function which might be useful outside this script. Can you write a few tests for it?

rizar:
So this is a new script, and there is absolutely no problem in merging it before reaching state of perfection. Let me know when you want to get it merged.

bin/blocks-plot
jbornschein: Ok, that sounds good. 

I will move some of the argument handling and most of the interactive loop back to bin/. I don't see any benefit in having it in blocks/scripts. And it decreases the coverage noticeable. 
Everything that might be of use and can be easily unit-tested will go to to blocks/scripts/plot.py

tests/scripts/test_plot.py
```diff
@@ -0,0 +1,43 @@
...
+    experiments['exp0']['col0'] = (0,1,2)
+    experiments['exp0']['col1'] = (3,4,5)
+    experiments['exp1'] = DataFrame()
+    experiments['exp1']['col0'] = (6,7,8)
+    experiments['exp1']['col1'] = (9,9,9)
+    return experiments
+
+
+#@silence_printing
+@skipIf(not PANDAS_AVAILABLE, "pandas not installed")
```
bartvm: I actually wrote a helper function for this yesterday :) You can use `from testing import skip_if_not_available` and then `skip_if_not_available(modules=['pandas'])`

jbornschein: Ok! I was a bit surprised, that it isn't a decorator ...

bartvm: Good point, I should probably turn it into one. Just habit I guess, it wasn't a decorator in Pylearn2. I'll change it in a bit.

bartvm: But I'll wait till this is merged, so that you don't have to rebase again!

jbornschein:
I guess it's in a acceptable state now and can be merged.

Most important missing features:
 - load experiment dumps
 - specify/change x-axis labeling

PR #301, opened 2015-02-17
Installation. Again.

bartvm:
So installation through `pip` is a real nightmare if you have dependencies on packages that are only on GitHub. The `dependency_links` argument is deprecated and requires you to pass the `--process-dependency-links` (which was nearly removed in pip 1.6), and results in a long list of warnings and errors that you are using deprecated features and downloading from insecure sources...

See [this](https://caremad.io/2013/07/setup-vs-requirement/) for a long overview, but apparently the correct way to specify these dependencies is through a `requirements.txt` file. Although this means that we use a slightly different `pip install` command, it actually works like a charm. It automatically installs the latest version of Theano and `picklable_itertools` from GitHub without warnings while keeping the `setup.py` file looking the way it should.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -35,12 +35,11 @@ before_install:
 install:
   # Install all Python dependencies
   - |
-      conda install -q --yes python=$TRAVIS_PYTHON_VERSION mkl bokeh nose numpy pip coverage six scipy
-      pip install -q --no-deps git+git://github.com/Theano/Theano.git # Development version
+      conda install -q --yes python=$TRAVIS_PYTHON_VERSION mkl bokeh nose numpy pip coverage six scipy toolz
       pip install -q nose2[coverage-plugin] coveralls
 script:
   - |
-      python setup.py install # Tests setup.py (should also install dill)
+      pip install . -r requirements.txt # Tests setup.py (should also install dill)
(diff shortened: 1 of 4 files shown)


bartvm:
@rizar @arasmus This seems to work for me, but please let me know if this fixes all the problems you ran into

arasmus:
On commit 5fc453ee. Cmd used:
```pip install -e git+git://github.com/bartvm/blocks.git#egg=blocks\[test,plot,docs\] --src=foo  -r https://raw.githubusercontent.com/bartvm/blocks/install_again/requirements.txt```

Our lab's system pip version is 1.4.1, so it doesn't recognize those --allow-unverified things in requirements.txt. So I had to upgrade it first in a virtualenv by ```pip install pip --upgrade```. Maybe an issue to write to the mailing list, faq or troubleshooting section.

After the upgrade it failed at the installation of scipy, not sure why and I don't have time dig deeper right now:
```ImportError: No module named numpy.distutils.core```

I also tried on my laptop; worked like a charm. Though, I had my scipy up-to-date already..

bartvm:
Right, that error remains; you can't install numpy and scipy from scratch
with pip but it's an error that we can't do anything about. I'll add it to
the docs though, including the instructions about needing a new pip
version. Thanks!

PR #303, opened 2015-02-18
Adam optimizer

memimo:
Haven't tested the convergence rate yet. Just putting it here in case there is any comment about format or anyone catches a bug.

### DIFF
--- a/blocks/algorithms/__init__.py
+++ b/blocks/algorithms/__init__.py
@@ -575,11 +575,12 @@ class Adam(StepRule):
 
         t1 = time + 1
         learning_rate = self.learning_rate * \
-                        tensor.sqrt((1. - (1. - self.beta2)**t1)) / \
-                        (1. - (1. - self.beta1)**t1)
-        beta_1t = 1 - (1 - self.beta1) * self.decay_factor**(t1 -1)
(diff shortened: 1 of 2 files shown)


memimo:
Seems to be converging more or less as expected. Tested on mnist with single layer fully connected model vs sgd.

blocks/algorithms/__init__.py
```diff
@@ -534,3 +534,59 @@ def compute_steps(self, previous_steps):
...
+        self.epsilon = epsilon
+        self.decay_factor = decay_factor
+
+    def compute_step(self, param, previous_step):
+        mean = shared_floatx(param.get_value() * 0., 'mean')
+        variance = shared_floatx(param.get_value() * 0., 'variance')
+        time = shared_floatx(0., 'time')
+
+        t1 = time + 1
+        learning_rate = self.learning_rate * \
```
rizar: The line breaks with "\" are discouraged in the Python style guide except for a few cases, so we in Blocks try not to use them. Can you replace those with parentheses?

memimo: Sure, removed them.

rizar:
I do not know much about this optimizer, but if you say it works, let's just merge it.

PR #304, opened 2015-02-18
Conv harm

bartvm:
@harmdevries89 This is a rebase of your PR #290; not sure what went wrong there, but it shows nearly 200 commits and edits in 59 files :p

I made a few changes to your original PR:

* `num_channels` shouldn't be optional; even though it's optional for Theano, we are initializing the filters as a shared variable, so we need it
* I removed the `num_channels` logic from the `MaxPooling` operation. I think it worked without as well, and this makes the brick slightly more general and applicable in other contexts as well (e.g. in cases where you want to apply it to a 2D tensor)

I'll merge after adding a test for `ConvolutionalSequence`

### DIFF
--- a/blocks/bricks/conv.py
+++ b/blocks/bricks/conv.py
@@ -10,10 +10,6 @@ from blocks.utils import shared_floatx_zeros
 class Convolutional(Initializable):
     """Performs a 2D convolution.
 
-    .. todo::
-
-       Allow passing of image shapes for faster execution.
-
     Parameters
     ----------
     filter_size : tuple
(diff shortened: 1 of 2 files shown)


harmdevries89:
I also don't know what went wrong there... Anyway, I'm running some experiments with the convolutional code now, so let me know if this could help in writing the tests. 

I was also wondering where the ConvNet implementation of @dmitriy-serdyuk is.

bartvm:
@dmitriy-serdyuk's version is here: https://github.com/dmitriy-serdyuk/blocks/blob/add-test/tests/bricks/test_conv.py#L58-L91 (there might be a newer one somewhere, but that's the one I found)

I've added a test, but it fails for very mysterious reasons. I'm probably just overlooking something, but the first convolution should return just 41 in my mind, but instead it returns 41 mixed in with some other random numbers, and I can't figure out where they come from.

bartvm:
nvm, I had defined `num_channels` to be 4, but gave only 3 channels in the numerical input, which means that the last output channel was gibberish. Should work now.

PR #307, opened 2015-02-19
WIP: Refactoring Attention

rizar:
Addresses #200.

This PR separates the `AttentionRecurrent` (former `AttentionTransition`) from the sequence generation framework and documents it. It is also made much more usable: now it does not require the wrapped transition to always have the attended context, now it also creates a basic version of `Distribute` automatically. 

In addition I introduce and document the common interface for attention bricks.

Altogether this should much easier to understand and read then the previous version of the code. @pbrakel , @dmitriy-serdyuk , can you as potential users of this give me some feedback?

By the way, the new `AttentionRecurrent` would make a great building block for hierarchical attention: a few of such bricks should be just chained.

For some reason the output values in the `test_attention_recurrent` test are non-deterministic and tests are either failing or not depending on how many of them I run. I am trying to figure out why.

### DIFF
--- a/blocks/bricks/attention.py
+++ b/blocks/bricks/attention.py
@@ -88,7 +88,7 @@ class SequenceContentAttention(Initializable):
         if not sequence_transformer:
             sequence_transformer = MLP([Identity()], name="seq_trans")
         if not energy_computer:
-            energy_computer = EnergyComputer(name="energy_comp")
+            energy_computer = ShallowEnergyComputer(name="energy_comp")
         self.sequence_transformer = sequence_transformer
         self.energy_computer = energy_computer
 
@@ -182,10 +182,11 @@ class SequenceContentAttention(Initializable):
         return super(SequenceContentAttention, self).get_dim(name)
 
 
-class EnergyComputer(Sequence, Initializable, Feedforward):
+class ShallowEnergyComputer(Sequence, Initializable, Feedforward):
+    """A simple energy computer: First tanh, then weighted sum."""
     @lazy
     def __init__(self, **kwargs):
-        super(EnergyComputer, self).__init__(
+        super(ShallowEnergyComputer, self).__init__(
             [Tanh().apply, MLP([Identity()]).apply], **kwargs)
 
     @property
--- a/blocks/bricks/attention.py
+++ b/blocks/bricks/attention.py
@@ -428,7 +428,7 @@ class AttentionRecurrent(AbstractAttentionRecurrent, Initializable):
             attended_mask_name = self.context_names[1]
         if not distribute:
             normal_inputs = [name for name in self.sequence_names
-                             if not 'mask' in name]
+                             if 'mask' not in name]
             distribute = Distribute(normal_inputs,
                                     attention.take_look.outputs[0])
 
--- a/blocks/bricks/attention.py
+++ b/blocks/bricks/attention.py
@@ -13,7 +13,7 @@ from abc import ABCMeta, abstractmethod
 from theano import tensor
 from six import add_metaclass
 
-from blocks.bricks import (MLP, Identity, Initializable, Sequence,
+from blocks.bricks import (MLP, Identity, Brick, Initializable, Sequence,
                            Feedforward, Tanh)
 from blocks.bricks.base import lazy, application
 from blocks.bricks.parallel import Parallel, Distribute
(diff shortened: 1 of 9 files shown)


rizar:
Okay, that was an issue with initialization.

@bartvm , should we change `Initializable`to take `rng` as a keyword argument?

bartvm:
It takes `seed` as a keyword argument, which seemed slightly more safe. If you pass an RNG, the RNG can be shared between a variety of objects, so the numbers you get depend on the execution order and could change between versions of Blocks, while passing a seed should ensure that you always get exactly the same numbers.

rizar:
But (a) seed is not propagated downwards in `push_initialization_config` (b) if it was, all parameter matrices with similar shapes would have same random values. I am just trying to figure out the right way to choose a fixed initialization for a hierarchy of bricks.

bartvm:
Right, you can make an issue for that, because there's another issue with `Initializable` that I ran into yesterday: `use_bias` is propagated downwards but overrides all the children in the process. This is annoying for e.g. `ConvolutionalSequence` (see #304) where the only way to turn biases of for a particular layer right now is doing something like:

```python
convolutional_layers = [ConvolutionalLayer(...), ConvolutionalLayer(...)]
convnet = ConvolutionalSequence(convolutional_layers, weights_init=...)
convnet.push_allocation_config()
convnet.layers[3].convolution.use_bias = False
```
Instead of the more intuitive
```python
convolutional_layers = [ConvolutionalLayer(...), ConvolutionalLayer(..., use_bias=False)]
convnet = ConvolutionalSequence(convolutional_layers, weights_init=...)
```

bartvm:
More directly related to your question: I guess we can keep `seed` as a keyword argument, but pass down the `rng` value. Although that could theoretically still change the initialization values, at least the RNG will be shared only within a hierarchy of bricks, and a user won't be able to use a single RNG all over the place (unless he manually overrides the `rng` attribute before the configuration is pushed down).

rizar:
Okay, this is the way it works currently. But I suggest to also make the `rng` a keyword argument. It is more convenient than `brick.rng = rng`.

bartvm:
But in that case I don't get it. If it pushes down the RNG, the initialization should be fixed if you pass `seed`. That seed should be used to initialize one RNG, which is then shared across all layers. That is, `Brick(seed=1)` should give exactly the same effect as `Brick(rng=numpy.random.Randomstate(1))`... I'm not too keen on making `rng` a keyword argument as well, because it's confusing if a user can pass *both* `rng` and `seed`.

dmitriy-serdyuk:
It looks cool! As I can see, an encoder (or something, that produces contexts from inputs) should be a separate brick. Is it right?

blocks/bricks/attention.py
```diff
@@ -8,15 +8,140 @@
...
+from blocks.bricks.recurrent import recurrent, BaseRecurrent
+from blocks.utils import dict_union, dict_subset
+
+
+@add_metaclass(ABCMeta)
+class AbstractAttention(Brick):
+    """The common interface for attention bricks.
+
+    A generic attention mechanism functions as follows. Its inputs are the
+    state of the agent and the object of attention, which we call here
```
bartvm: So, this might just be me because I'm not familiar enough with the usual vocabulary, but at first sight these docstrings are still really hard to read for me. Agent? Object of attention? Attended? Composite state of the agent? For me, it's unclear from the docstring what any of these things are.

I guess there's a variety of options:

1. Assume less familiarity with these concepts, and explain more explicitly what things like "agent" and "object of attention" mean. Use examples, not abstractions! My suggestion of a beginning would be something like:

> An attention mechanism decides what part of the input to pay attention to at a given point in time. It does so by taking a particular part of input (e.g. a word in a sentence, or a patch of larger image) and the "state" of the neural network at that point in time, assigning each partial input a score of "importance".

> For example, in a neural translation model [ref] the "state" is the hidden state of the RNN that outputs our translation, representing our translation so far. The partial inputs in this case are the words in the source sentence. Based on what the network has translated so far the attention mechanism now decides how much attention the network should pay to each word in the source sentence.

> The partial inputs and the scores given to each are combined into a "glimpse". As such, a glimpse is a single representation of what is important to the neural network (the agent) in its current state.

> A simple example of a glimpse would be taking the weighted sum of all the partial inputs, weighted by their importance.

2. Intersperse with pseudocode/docstrings and/or graphviz diagrams; maybe that is easier than textual explanation
3. Leave it, and eventually make a tutorial which introduces these concepts in a more gentle manner

rizar:
@bartvm, my bad, I now see that if we fix the seed in the top-most brick, we effectively choose initialization for all children. But: switching to `seed=1234` from `recurrent.rng = numpy.random.RandomState(1234)` does not work for me for some reason, the outputs change.

In addition what  I find strange is that in `_push_initialization_config` rngs are pushed always to all children, that is I can not override initialization of a subtree by `brick.some_brick.another_brick.seed = 4321`.

rizar:
Okay, that was a bug in `Initializable`, fixed in b1fef06

rizar:
Addressing you comment @bartvm: you are absolutely right that we need as well a simpler example-based text. But this is too much for the docstring, I think. Here the purpose is to fix the general terminology which is very close to the code. 

In general I think there should always be a specification and a tutorial. This text tries to be exhaustive specification. Python specification is also extremely hard to read.

So I will write a tutorial as well when I have time, or maybe there will be volunteers.

rizar:
I am not sure I understand your question. Nice thing about the new 
`AttentionRecurrent` that in most cases it is enough to simply give it a 
recurrent brick and an attention, all the rest is done automagically.

bartvm:
I personally have a lot less trouble with the Python documentation than these particular docstrings though. I don't agree that all explanations should be deferred to a tutorial. Tutorials are not only there to give an easy introduction, I actually think they are mostly there to synthesise a set of concepts and ideas that would otherwise be spread about in multiple docstrings. The docstring itself should at the very least clearly explain all the terminology used within, in case it can't be assumed common knowledge.

And to give a quick example from the Python documentation on its data model:

> Every object has an identity, a type and a value. An object’s identity never changes once it has been created; **you may think of it as the object’s address in memory**. The ‘is‘ operator compares the identity of two objects; the id() function returns an integer representing its identity (**currently implemented as its address**). An object’s type is also unchangeable. An object’s type determines the operations that the object supports (**e.g., “does it have a length?”**) and also defines the possible values for objects of that type.

You'll see that they use analogies, concrete as well as abstract explanations, and examples, and I think that you shouldn't shy away from using those in these docstrings as well. All the abstract terminology (a "hypothetical agent" using "structured input", processing the "input signal" while being allowed to have "composite states") really completely loses people...

Anyway, this PR is mostly a refactoring, so we can relegate the documentation to another PR.

pbrakel:
Hey @rizar, I just wanted to say I really like the changes and think you did a nice job!

rizar:
Thanks! Looking forward for your sorted batches ;)

dmitriy-serdyuk:
@rizar I don't understand where `attended` comes from. Is it the same as `contexts` in `SequenceGenerator`?

rizar:
There currently two options:

1) The attended is one of the contexts of the wrapped transition. Before 
this refactoring it was the only option. The disadvantage is, that 
normally one has to create a separate custom ancestor of a recurrent 
brick to add contexts to it in order to make this work.

2) The attended is a new context of the `AttentionRecurrent`. It does 
not pass the attended to the wrapped transition: the wrapped knows 
nothing about the attended. It assumes that the attention mechanism was 
configured correctly and the dimension of the attended can be retrieved 
from there (see `get_dim`) method.

I hope I answered your question.

Issue #167, opened 2015-01-27
Epoch progress extension


bartvm:
This is very much a gimmicky extension, but one that I always wanted in Pylearn2 because it doesn't do monitoring in between epochs. Would love an extension which shows a nice progress bar (using e.g. https://code.google.com/p/python-progressbar/) so that you can wait in anticipation till your epoch is finished. This would require the extension to know the number of examples in a single epoch though, which might not always be available (and right now, even when it technically could be available, it's not passed up the data stream hierarchy) so it would need a bit of work.

bartvm:
Fixed by #308.

PR #308, opened 2015-02-19
ProgressBar extension

jbornschein:
Here is a small extension that will display a progress-bar during training.

Cons: 
- introduces a new dependency to progressbar2 (https://github.com/WoLpH/python-progressbar) . It is a really small, well maintained package though.
- does not work with all iteration schemes as it needs to compute the number of iterations per epoch to display anything.

Pros:
- displays a fancy progress bar. Including  -- and that might be the only useful aspect of it -- the estimated walltime until the end of this epoch:

```
    Epoch 4, step 170 ( 34%) |#######               | Elapsed Time: 0:00:03 ETA:  0:00:07
```

### DIFF
--- a/blocks/extensions/__init__.py
+++ b/blocks/extensions/__init__.py
@@ -343,6 +343,59 @@ class Printing(SimpleExtension):
         print()
 
 
+class ProgressBar(TrainingExtension):
+    """Display a progress bar during training.
+
+    This extension tries to obtain the number of mini-batches processed per
+    epoch and will display progress bar. Not all :class:`IterationScheme`
+    classes provide the necessary atributes.
+
+    Notes
+    -----
+    This extension should be run before other extensions that print to the screen
+    at the end or at the beginning of the epoch (e.g. the Printing extension).
+    Placing ProgressBar before these extension will ensure you won't get
+    intermingled output on your terminal.
+
+    """
+    def __init__(self, **kwargs):
+        super(ProgressBar, self).__init__(**kwargs)
+
+        self.bar = None
+        self.epoch_counter = 0
+        self.batch_counter = 0
+
+    def before_epoch(self):
+        self.epoch_counter += 1
+        self.batch_counter = 0
+
+        iteration_scheme = self.main_loop.data_stream.iteration_scheme
+        if hasattr(iteration_scheme, 'batches_per_epoch'):
+            batches_per_epoch = iteration_scheme.batches_per_epoch
+        elif hasattr(iteration_scheme, 'num_examples') and
+                hasattr(iteration_scheme, 'batch_size'):
+            batches_per_epoch = iteration_scheme.num_examples // iteration_scheme.batch_size
+        else:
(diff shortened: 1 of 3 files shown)


bartvm:
Yay, I opened #167 so I'm clearly for merging this extension. No problem with the dependency; it's tiny and on PyPI. I'll try it out in a bit.

bartvm:
Oh, and although out of the scope of this PR, it's still something that we need to consider eventually: This only works if the final data stream has an iteration scheme attached. However, in many cases you could be using data stream wrappers that don't have an iteration scheme but just pass on the data. Ideally the information on `num_examples`, `num_batches` etc. should be passed along these wrappers.

jbornschein:
Yes, providing ````num_examples, num_batches``` (etc.) is an issue:

I just noticed that my unittest is nearly useless because ```CotainerDataset``` does not provide the necessary attributes and it thus only tests the "no progressbar can be displayed" case.

rizar:
@jborgschein, indeed currently `ContainerDataser` provides no information whatsoever about the number of examples in it. Looks like if we equip it with a `num_examples` property and teach to handle requests consisting of sample indices, it will be possible to use the `SequentialScheme` and your extension. 

But if you don't want to do this, another option is to write a custom small datastream wrapper for test purposes, so that your progress bar code was automatically tested. It should it be quite easy.

rizar:
A brutal suggestion on how to pull the `num_examples` and other attributes through a chain of wrappers: we can make a subclass `TransparentWrapper` that overrides `__getattr__` to pass requests of unknown attributes to the wrapped datastream. `DatastreamMapping` can inherit from it, for instance.

bartvm:
Keep in mind that `ContainerDataset` could potentially wrap (infinite) iterators, in which case indexing (and `num_examples`) is still not available. So you would need to handle both cases.

Passing on attributes might be a good idea, although I would be on the safe side and only pass on attributes that we define (else we could end up with some very strange bugs, because it could start e.g. calling methods of the wrapped data streams instead of the parent class's).

However, maybe this is a problem that's worth putting into some more thought first. For example, the logic about calculating `num_batches` from `num_examples` and `batch_size` should probably be moved to an iteration scheme or data stream somewhere. More importantly: We need to distinguish between the number of examples a data stream can provide, and the number an iteration scheme can provide. Right now a `ConstantScheme(1)` doesn't give you `num_examples`, but we know it will be equal to `num_examples` provided by the data stream. However, the iteration schemes and request iterators are currently agnostic to what data streams they provide requests for (and we should keep it that way I think), so if/how do we nicely implement this logic?

jbornschein:
I updated the code to 1) display a simplified progress bar when `num_iter_per_epoch` can't be inferred and 2) to ensure that it is picklable. 

1) also solves the the test coverage issue due to `ContainerDataset` not providing `num_examples`. 2) was necessary because the progressbar extension keeps an open file descriptor around.

I guess this is now in a mergable state and we deal with DataStreams/Datasets not providing `num_examples` another time?

rizar:
Yeap, the problem of pushing this information through a chain of wrappers is definitely not in the scope of this PR.

PR #309, opened 2015-02-20
User-level documentation for recurrent bricks

vdumoulin:
It's not 100% complete, but it's a good improvement over nothing!


docs/rnn.rst
```diff
@@ -0,0 +1,211 @@
...
+
+This is very useful when you're trying to combine multiple recurrent layers in
+a network.
+
+Imagine you'd like to build a network with two recurrent layers. The second
+layer accumulates the output of the first layer, while the first layer
+accumulates the input of the network and the output of the second layer (see
+figure below).
+
+.. figure:: _static/feedback_rnn.svg
```
bartvm: Awesome to have images! Just FYI, I also enabled GraphViz support (see e.g. the [datasets documentation](http://blocks.readthedocs.org/en/latest/datasets.html)). The main advantage of that is that it will be easier for people to edit in the future than SVG.

vdumoulin:
The tests are failing in a weird way: the `FeedbackRNN` class I defined in the user-level documentation is making unrelated tests fail.

docs/rnn.rst
```diff
@@ -0,0 +1,211 @@
...
+
+We will modify the starting example so that it accumulates the input it
+receives, but starting from one instead of zero:
+
+.. math:: \mathbf{h}_t = \mathbf{h}_{t-1} + \mathbf{x}_t, \quad \mathbf{h}_0 = 1
+
+>>> h0 = tensor.matrix('h0')
+>>> h = rnn.apply(inputs=x, states=h0)
+>>> f = theano.function([x, h0], h)
+>>> print f(numpy.ones((3, 2, 3)), numpy.ones((2, 3))) # doctest: +SKIP
```
bartvm: Is there any reason you're skipping these tests? If the problem is the `dtype` flag appearing when `floatX = 'float32'`, then you can also add `+ELLIPSIS ` and change the last line to `  [ 4.  4.  4.]]...`. Doctests are always good sanity tests to have!

docs/rnn.rst
```diff
@@ -0,0 +1,211 @@
...
+
+The class has a ``get_dim`` method whose purpose is to tell the dimensionality
+of each input to the brick's ``apply`` method.
+
+The core of the class resides in its ``apply`` method. The ``@recurrent``
+decorator is used to specify which of the arguments to the method are sequences
+to iterate over, which are recurrent states and which are returned when the
+method is called.
+
+Notice how no call to ``theano.scan`` is being made. This is because the
```
bartvm: You can also link to Theano documentation if you want e.g. ```:func:`theano.scan` ```

docs/rnn.rst
vdumoulin: I'll check that out, it probably isn't that much effort to replace my SVG figure with a `digraph` block.

bartvm:
They're failing in a very, very weird indeed... At first sight I don't have slightest clue as to what could be going on :/

vdumoulin:
I guess defining a class in an example doesn't play nice with the doctests?

bartvm:
Defining a class itself shouldn't be an issue; I've done that before (e.g. in the developer guidelines). What's also strange is that I am unable to reproduce the error. All the tests run just fine on my system.

bartvm:
New levels of absurdity when trying to fix it: `NameError: global name 'False' is not defined`. How the hell can `False` be undefined?

docs/rnn.rst
```diff
@@ -0,0 +1,211 @@
+Recurrent neural networks
+=========================
+
```
rizar: I think it would be nice to put something like

```
This tutorial explains recurrent bricks in Blocks. If you are new to the brick concept, we recommend to start with https://blocks.readthedocs.org/en/latest/bricks_overview.html first and continue with this tutorial afterwards
```
(that was a bad and very quick wording and you can change it as you like).

docs/rnn.rst
```diff
@@ -0,0 +1,211 @@
...
+initializes two :class:`.bricks.recurrent.SimpleRecurrent` bricks as its
+children.
+
+The class has a ``get_dim`` method whose purpose is to tell the dimensionality
+of each input to the brick's ``apply`` method.
+
+The core of the class resides in its ``apply`` method. The ``@recurrent``
+decorator is used to specify which of the arguments to the method are sequences
+to iterate over, which are recurrent states and which are returned when the
+method is called.
```
rizar: I would rephrase just a bit here to indicate that the states are returned as well, and that the order of the returned values must be always states first, outputs last.

rizar:
It is super cool to have this!

vdumoulin:
@bartvm @rizar I've addressed all your comments except for the graphviz replacement of the SVG figure (I'm fighting against the dot engine to keep things aligned).

I'll try to fix the weird test failure (which I can reproduce on my machine) as well.

vdumoulin:
Okay, I've managed to produce a decent-looking figure with graphviz. I also changed the output format to SVG, because why not? Scalable graphics look much better than PNG on a high-res display.

For some reason Scrutinizer has started erroring this morning, and I'm not sure if it's related to something I've changed.

The only thing remaining is to fix the weird test failures.

bartvm:
Meh, Scrutinizer can just be a bit temperamental. I'll restart it later.

vdumoulin:
That's weird. I'm getting syntax errors on `rnn.rst` when I'm running the tests which I don't get when I use doctest directly on the file.

bartvm:
It's clearly cursed.

vdumoulin:
@rizar I have a question that's somewhat related to this pull request, as the answer may incite me to clarify part of the tutorial: imagine that I build a recurrent brick that produces an additional output in addition to the recurrent states, how should I call the `@recurrent` decorator?

Here's an example: say I have a recurrent brick that accepts 'inputs' as its sequences and produces a 'states' recurrent state, and suppose I also want to output the square of 'states' when the apply method is called. Is the following the only right way to call `@recurrent`?

```python
@recurrent(sequences=['inputs'], contexts=[], states=['states'],
           outputs=['states', 'squares'])
```

rizar:
I am pretty sure you are right. See for instance this quote from the code:

```python
outputs_info = (list(states_given.values()) +
      [None] * (len(application.outputs) - len(application.states)))
```

It corresponds to what you guessed and to my memories of how I wanted it to be done.

vdumoulin:
Is there a reason for the redundancy? In other words, is there a reason that prevents the `outputs` argument to be reserved only for outputs that are not themselves recurrent states?

rizar:
There are two consumers for the properties declared in `@recurrent`. The `recurrent` decorator itself is the first one and it needs `sequence`, `contexts`, `states` and `outputs` to call `theano.scan` with the right arguments. On the other hand the external consumers that use recurrent methods with `iterate=True` needs `inputs` and `outputs` properties in their usual sense. That is, `outputs` is used by both. The recurrent decorator is the one who adapts in this situation and ignores first `len(states)` outputs.

It is not entirely senseless, but I am very open to suggestions on how to make it clearer.

vdumoulin:
I see. It makes perfect sense when explained this way.

There are two sensible ways to see the interface, IMO:

* `states` and `outputs` form two disjoint sets whose union is what's returned by `apply`, and the inputs to that `apply` expects is the union of `inputs` and `states`.
* `outputs` is the set of everything returned by `apply`, and `states` is a subset of `outputs` which is fed to `apply` along with `inputs`.

The second one has the advantage that it is consistent with the application interface, it's just a matter of explaining it well. I'll refine the explanation in the tutorial based on my improved understanding.

vdumoulin:
I've solved the doctest issues (apparently the `print` statement isn't liked very much), but I'm still banging my head against the wall for the weird pickling bug.

rizar:
Can you reproduce the pickling bug locally?

vdumoulin:
Yes, I get the same errors as Travis.

rizar:
Okay, this is not as hopeless :)

I will try to reproduce it at my machine today.

vdumoulin:
One weird thing is that the pickling error I'm getting looks like

    PicklingError: Can't pickle <class 'FeedbackRNN'>: attribute lookup __builtin__.FeedbackRNN failed

but `FeedbackRNN` is defined in `docs/rnn.rst` inside an example! When removing the code block where it is defined as an example, the tests pass.

vdumoulin:
Another thing to note: if I use `# doctest: +SKIP` on the line where the class is instantiated and all subsequent lines, the tests pass.

bartvm:
Does runnig the doctest separately from all the other tests fix it? I'd be happy to run them separately; as @lamblin mentioned, nose can apparently really fuck up the global namespace if run in the same interpreter as the other tests.

bartvm:
Oh, and yeah, doctests need to be written in Python 3. Sorry that I missed that!

vdumoulin:
I haven't looked into that, I'll check it right away.

bartvm:
Seems to work, right? Hurrah!

We need to combine the coverage reports though: http://coverage.readthedocs.org/en/coverage-4.0a5/cmd.html#combining-data-files

I think it should be as simple as adding `-p` to the coverage command, and then call `coverage combine` after the two runs.

vdumoulin:
The tests pass, but I had a failure in the doctests that wasn't reported by Travis, weirdly enough.

bartvm:
Strange, can't see immediately why Travis would have missed that...

vdumoulin:
The tests now pass, except for the coverage test.

bartvm:
LGTM! About time to solve #315 though so that it will actually show up :/ @rizar Are you okay with merging; it seems a vast improvement over what we have so far.

rizar:
I am more than okay! The thought about this pretty doc hanging here without  being pushed did not let me sleep!

bartvm:
#315 was fixed and the docs are at http://blocks.readthedocs.org/en/latest/rnn.html :)

vdumoulin:
Great! Looks pretty good!

rizar:
I have this issue with `FeedbackRNN` back, testing on Python 3.4. What was the solution?

vdumoulin:
Running the doctests in a separate test suite solved the problem for me.

rizar:
You mean that I should run tests with two commands, not with one, right?

vdumoulin:
Yes, like [this](https://github.com/bartvm/blocks/blob/master/.travis.yml#L46-L48)

rizar:
Okay, now I see, thanks!

PR #311, opened 2015-02-20
Basic Scan Support in Computation Graph

rizar:
This PR equips the `ComputationGraph` with two properties: `scans` and `scan_variables`. I did not provide access to the nested inner scan graphs so far: I hesitate what the right interface should be. 

The first consumer of this new feature is the `Model`.

### DIFF
--- a/blocks/graph.py
+++ b/blocks/graph.py
@@ -53,7 +53,7 @@ class ComputationGraph(object):
     scans : list of :class:`~theano.scan_module.scan_op.Scan`
         All Scan ops used in this computation graph.
     scan_variables : list of :class:`~tensor.TensorVariable`
-        All variables of inner graphs of Scan op.
+        All variables of the inner graphs of Scan ops.
     updates : :class:`~tensor.TensorSharedVariable` updates
         All the updates found attached to the annotations.
 
--- a/blocks/graph.py
+++ b/blocks/graph.py
@@ -7,6 +7,8 @@ import theano
 from theano import Variable
 from theano.gof import graph
 from theano.sandbox.rng_mrg import MRG_RandomStreams
+from theano.scan_module.scan_op import Scan
+from toolz import unique
 
 from blocks import config
 from blocks.roles import add_role, AUXILIARY
@@ -48,6 +50,10 @@ class ComputationGraph(object):
         Any variable that is not part of :attr:`inputs` or :attr:`outputs`.
     variables : list of :class:`~tensor.TensorVariable`
         All variables (including auxiliary) in the managed graph.
+    scans : list of :class:`~theano.scan_module.scan_op.Scan`
+        All Scan ops used in this computation graph.
+    scan_variables : list of :class:`~tensor.TensorVariable`
+        All variables of inner graphs of Scan op.
     updates : :class:`~tensor.TensorSharedVariable` updates
         All the updates found attached to the annotations.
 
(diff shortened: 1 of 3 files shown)


jbornschein:
Just a comment -- not sure if this is related and should be considered, if it is a totally orthogonal issue, or if it is already handled somewhere and I just didn't find the feature yet:

I have a model (not implemented with the new ```Model``` hierarchy, but just a computational graph with a None model, as far as the MainLoop is concerned) that has a scan inside. The scan-function in turn uses a random stream. In this setting scan will thus return updates for shared variables.

The problem: I could not find a way to supply these updates to DataStreamMonitoring. As a result the DataStreamMonitoring compilation will fail. For TrainingDataMonitoring this is not an issue because I supply these updates to the GradientDescent algorithm.

rizar:
There is indeed right now no support for updates in the monitoring, but 
that should be relatively easy to fix.

The way additional updates are handled in Blocks is the following. Some 
variables of the computations have the `tag.annotations` attribute. Each 
of these annotation is an instance of `Annotation` and has
the `updates` attribute. That is updates are attached to annotations of 
the computation graph. For compiling the graph correctly we have the 
`get_theano_function` method of the `ComputationGraph`: it takes care of 
all updates.

Unfortunately, neither the training algorithm nor data 
`DataStreamMonitoring` are currently using this feature. But they 
should, so feel free to change this in a PR.

I should probably mention where the annotations come from: if you use 
recurrent bricks from `blocks/bricks/recurrent` they are added 
automatically. Otherwise, you can attach your own annotations to the 
computation graph.

bartvm:
I don't think they should! If we're not careful we'll lose our original vision of training algorithms being Theano-only, and we start mixing all sorts of logic into them that relies on annotated graphs, the `Model` interface, etc.

I guess we have a variety of options:

* Keep the algorithms completely modular like right now i.e. you should be able to import the algorithms folder by itself, and everything should work. This is the option that has my strong preference.
* Raise warnings when the values passed don't match with the annotated graph e.g. warn if there are shared variables without a `PARAMETER` role, or warn if no updates were added but there were updates in the annotations. This means that algorithms are *no longer stand-alone*; they can't be re-used without installing the entirety of Blocks, and they'll start relying on the interface of e.g. `ComputationGraph`, which introduces dependencies between parts of the framework.
* Only use `ComputationGraph` logic as a default i.e. use it to extract parameters and updates if none were given. Same shortcomings as above, but it comes with a nicer user interface (i.e. things happen automagically). Downside is that we start moving to scary automagic behavior that is likely to bite us in the ass later on.

Classical usability problem, but considering the philosophy of the framework, I have zero issue with users being expected to pass e.g. `GradientDescent(params=model.get_parameters(), updates=cg.updates)` instead of everything happening magically behind the scenes if that means we don't end up introducing interdependencies throughout the framework.

rizar:
First, algorithms are already using the `ComputationGraph`. So "modular like right now" is a false statement. 

But if you want, we can go this way. We can remove the default search of shared variables from the the algorithm. Note, this will cause certain amount of boiler-plate code dealing with creation of the algorithm object. But I can live with that to have clear conscience. 

But the monitoring extension is an internal Blocks thingy and do not see a problem if it uses `get_theano_function`. Do you object this as well?

rizar:
I would also like to hear what do you think about the code of this PR :)

bartvm:
The PR looks good!

PR #317, opened 2015-02-20
Plot Every Iteration

rizar:
@jbornschein , waiting for your approval!

### DIFF
--- a/bin/blocks-plot
+++ b/bin/blocks-plot
@@ -47,7 +47,7 @@ def main(args):
         log = plot.load_log(fname)
 
         rows_to_keep = ([0] + log.status._epoch_ends
-                        if args.every == "epochs"
+                        if args.every == "epoch"
                         else range(log.status.iterations_done))
         data_frame = log.to_dataframe()
         data_frame = data_frame.iloc[rows_to_keep]
--- a/bin/blocks-plot
+++ b/bin/blocks-plot
@@ -1,4 +1,4 @@
-#!/usr/bin/env python 
+#!/usr/bin/env python
 
 from __future__ import print_function, division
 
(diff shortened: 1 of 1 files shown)


bin/blocks-plot
```diff
@@ -131,16 +133,17 @@ if __name__ == "__main__":
...
+    parser.add_argument("--channels", "-c", type=str,
         help="A comma separated list of channels to plot. " + \
              "The wildcards '*' and '?' can be used to select multiple " + \
              "channels at once. Prefix a channel name with <num>: to " + \
              "refer to a channel of a particular experiment")
-    #parser.add_argument("--xaxis", "-x", choices=['iterations', 'epochs', 'time'],
-    #    default='iterations', help="Label/scale to use on the x-axis of the plot")
-    parser.add_argument("--ipython", action="store_true", 
+    parser.add_argument("--every", "-e",
+        choices=["epochs", "iters"], default="iters",
```
bartvm: Since it's "every epoch" or "every iteration" I would make the choise `epoch` and `iter` instead of `epochs` and `iters`

bin/blocks-plot
```diff
@@ -131,16 +133,17 @@ if __name__ == "__main__":
     logging.basicConfig()
 
     parser = argparse.ArgumentParser()
-    parser.add_argument("--list", "-l", action="store_true", 
+    parser.add_argument("--list", "-l", action="store_true",
         help="List all channels contained in the experiments and exit")
-    parser.add_argument("--channels", "-c", type=str, 
+    parser.add_argument("--channels", "-c", type=str,
         help="A comma separated list of channels to plot. " + \
```
bartvm: Even though this isn't part of this PR, there's no need for the `+` and the `\` here. Strings are automatically concatenated within parentheses.

PR #324, opened 2015-02-21
Change default activation functions

rizar:
Does anybody mind?

### DIFF
--- a/blocks/bricks/recurrent.py
+++ b/blocks/bricks/recurrent.py
@@ -234,10 +234,8 @@ class SimpleRecurrent(BaseRecurrent, Initializable):
 
     """
     @lazy
-    def __init__(self, dim, activation=None, **kwargs):
+    def __init__(self, dim, activation, **kwargs):
         super(SimpleRecurrent, self).__init__(**kwargs)
-        if activation is None:
-            activation = Tanh()
         self.dim = dim
         self.activation = activation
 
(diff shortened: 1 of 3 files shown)


dwf:
Seems a bit of a weird default to me, but I am not as familiar with recurrent nets.

rizar:
@pbrakel , @bartvm , what do you think guys?

bartvm:
I think `Tanh` is  better default than `Identity`, because I've already heard of people not realising that they weren't using an activation function, making their models untrainable. However, maybe it's better to not provide a default at all?

mohammadpz:
I think that the problem with ```Identity``` is that after a few time-steps, states' values can go to infinity because the recurrent weights are shared and they get multiply together. There's a same reason for not using ```ReLU``` in RNNs.

bartvm:
Yeah, `Identity` is there more because of principal reasons: The idea is that we don't want to decide for the user what activation to use, so by default it doesn't use any. In practice that doesn't work very well though, users just tend to forget setting it, so we either need to switch to `Tanh` as a default, or just not provide a default at all (and crash if the user doesn't provide an activation function).

mohammadpz:
Yes, it's important to make models as flexible as possible for users. On the other hand, for Simple Vanilla RNN, I've never seen using something other than ```Tanh```. However this is not the case for Feedforward Neural Networks. Thus, I myself prefer a default activation function but with a message to the user.

dwf:
Not providing a default seems perfectly reasonable.

rizar:
Okay, for safety reasons we can provide no default.

But then I would like to make another change: make activation in LSTM configurable with `tanh` as default. I would like to see these recurrent bricks simply interchangeable, and for that purpose they should have similar interfaces when possible. Any reasons against such a change?

bartvm:
No objections. This ready to merge then?

PR #325, opened 2015-02-21
RNN toy example

mohammadpz:
Hi,

Here's a super simple example of Vanilla RNN applied on a toy example dataset.
The task is a simple memorization task. The network input has a shape of ```(15, 2)``` and the output is another sequence with a shape of ```(15, 2)```. So, we have 2 input signals and 2 output signals (each with a length of 15). Output signals are same as input signals but with 2 and 4 time-steps delay respectively.

At the end of training I also plot these 4 signals.


examples/RNN_toy_example.py
```diff
@@ -0,0 +1,142 @@
...
+from blocks.algorithms import (GradientDescent, Scale,
+                               StepClipping, CompositeRule)
+from blocks.extensions.monitoring import TrainingDataMonitoring
+from blocks.main_loop import MainLoop
+from blocks.extensions import FinishAfter, Printing
+from blocks.bricks.recurrent import SimpleRecurrent
+import matplotlib.pyplot as plt
+
+
+class ToyDataset(Dataset):
```
rizar: I believe it is not used later in the code.

examples/RNN_toy_example.py
```diff
@@ -0,0 +1,142 @@
+import numpy as np
```
rizar: Can you please write here a text description of the learning task solved by the script, just like you explained it in the PR description?

examples/RNN_toy_example.py
```diff
@@ -0,0 +1,142 @@
...
+                                                         Scale(0.01)]))
+    monitor = TrainingDataMonitoring([cost],
+                                     prefix="train",
+                                     after_every_epoch=True)
+    main_loop = MainLoop(data_stream=stream, algorithm=algorithm,
+                         extensions=[monitor,
+                                     FinishAfter(after_n_epochs=n_epochs),
+                                     Printing()])
+
+    main_loop.run()
```
rizar: How long does the whole thing take?

examples/RNN_toy_example.py
```diff
@@ -0,0 +1,142 @@
...
+
+    predict = theano.function(inputs=[u, ], outputs=y_hat)
+
+    # Cost
+    y = T.tensor3('target_sequence')
+    cost = SquaredError().apply(y, y_hat)
+    cost.name = 'MSE'
+
+    # Initialization
+    RNN.weights_init = state_to_output.weights_init = \
```
rizar: Please use parentheses instead of `\`: this is a part of Blocks code style.

examples/RNN_toy_example.py
rizar: Please do not use `import something as st`, this is a part of Blocks code style.

You can find the list of whole restrictions like this one in http://blocks.readthedocs.org/en/latest/development/index.html

examples/RNN_toy_example.py
mohammadpz: Less than 1 minute.

rizar:
Nice! It is absolutely self-contained and should be a very good thing to start with. 

You should also add it to our test system, see `tests/test_examples.py`. You can either test that training runs, or, if it is fast enough, you can even check the training error.

examples/RNN_toy_example.py
rizar: Okay, one minute is a bit too much for a test. I was just thinking what is the best way to test this script. I think we should just run for fewer epochs, so that it took not more than, let's say, 5 seconds.

mohammadpz: Yeah, now the network should learn 2 and 4 time-steps delay. Learning 2 time-steps happens after 150 epochs which is less than 5 seconds.

examples/RNN_toy_example.py
mohammadpz: Sorry, I don't know how to do this!

mohammadpz:
I'm surprised that how fast it is.  I have a same version with pure theano which is 5 times slower than this.

examples/RNN_toy_example.py
bartvm: Parentheses don't work for multiple assignments. I guess you could leave it, but I think this is clearer (and 3 lines shorter):
```python
for brick in (RNN, state_to_output, input_to_state):
    brick.weights_init = IsotropicGaussian(0.01)
    brick.biases_init = Constant(0)
    brick.initialize()
```

bartvm:
I don't think we want the image as part of the commit, right? And if it's going to be part of the set suite then it shouldn't save images (or try to show them, because that will crash on any headless server). 

Few other things: Please use `num_` instead of `n_`, this is just a bit clearer IMO and is what is used throughout Blocks.

examples/RNN_toy_example.py
```diff
@@ -0,0 +1,143 @@
+"""
```
bartvm: You'll see that Scrutinizer already complained about this, the formatting for docstrings (including module-level ones) is:
```
"""A summary line.

A length explanation here.

"""
```
So you'll need a summary line, and a blank line before closing. Also make sure that the docstring never exceeds 75 characters.

rizar:
@mohammadpz , do you plan to ever finish this?

mohammadpz:
For now I close this.

rizar:
As you wish, though it was only about making it tested and removing an image from the committed files.

mohammadpz:
I don't know how to make it tested.
Besides, I don't really like struggling with python, theano, documentation, and git while I have 3 projects and 2 courses at the same time.
I'll get back to this at the appropriate time.

PR #326, opened 2015-02-22
Computationally stable softmax cost

yingzha:


blocks/bricks/cost.py
```diff
@@ -18,6 +18,19 @@ def apply(self, y, y_hat):
...
 @add_metaclass(ABCMeta)
+class Softmax(Brick):
+    @application
+    def apply(self, x):
+        return tensor.nnet.softmax(x)
+
+    @application
+    def categorical_cross_entropy(self, x, y_hat):
+        x =  x - x.max(axis=1).dimshuffle(0, 'x')
+        log_prob = x - T.log(T.exp(x).sum(axis=1).dimshuffle(0, 'x'))
```
bartvm: `T` should probably read `tensor`

blocks/bricks/cost.py
```diff
@@ -18,6 +18,19 @@ def apply(self, y, y_hat):
 
 
 @add_metaclass(ABCMeta)
+class Softmax(Brick):
+    @application
+    def apply(self, x):
+        return tensor.nnet.softmax(x)
+
+    @application
+    def categorical_cross_entropy(self, x, y_hat):
```
bartvm: The input is `y` I think (`y_hat` would be the output of the softmax, while `y` is the actual distribution i.e. the labels)

bartvm: Also, could you add a docstring that describes what the inputs should be e.g. `y` can be of the same dimension (in which case it needs to be a full distribution) or it needs to have one less dimension (in which case it needs to have integers representing the target classes).

yingzha:
I'll add the docstring.

bartvm:
Great! Could you also add a short tests in `tests/bricks/test_cost.py`? Just two test cases (one with labels as targets, one with distributions as targets) to make sure that the output is numerically correct.

yingzha:
No problem!

rizar:
Shouldn't this new softmax be also used in the `BinaryCrossEntropy` and `CategoricalCrossEntropy` bricks?

rizar:
Okay, I see `categorical_cross_entropy` became an application method, which is okay, but then we should get rid from the old `CategoricalCrossEntropy` brick and replace it in the examples. And also the old `Softmax` from `blocks/bricks/__init__.py` becomes also obsolete, right?

bartvm:
Well, theoretically there might cases where you want to calculate the cross
entropy of a probability distribution that did not come out of a softmax.
So strictly speaking it's not obsolete.

I think this Softmax should go in bricks/__init__.py though, and not in
bricks/cost.py.

dwf:
One important use case is to be able to easily calculate the cross-entropy
between two discrete distributions, neither of which is completely one-hot.

yingzha:
Should I delete the old version Softmax in __init__.py?

bartvm:
Yeah, that will be replaced by this one.

yingzha:
what does this error "coverage/coveralls — Coverage decreased (-0.14%) to 90.73%" mean?

bartvm:
Don't worry about it :) It means that fewer lines were covered by unit
tests. We can still merge even if that fails.

yingzha:
Ok! :)

yingzha:
The new softmax class has completed.

blocks/bricks/__init__.py
```diff
@@ -444,6 +444,32 @@ class Softmax(Activation):
...
         return tensor.nnet.softmax(input_)
 
+    @application
+    def categorical_cross_entropy(self, x, y):
+        """
+        Return computationally stable softmax cost.
+
+        Parameters
+        ----------
+        :type x: matrix
```
bartvm: Please have a look at the [documentation guidelines](http://blocks.readthedocs.org/en/latest/development/docs.html#docstrings) of Blocks. We use NumPy docstrings, not rST ones. Just look at any of the other bricks for an example.

blocks/bricks/__init__.py
```diff
@@ -444,6 +444,32 @@ class Softmax(Activation):
     def apply(self, input_):
         return tensor.nnet.softmax(input_)
 
+    @application
+    def categorical_cross_entropy(self, x, y):
```
bartvm: I was just using this method, and realised that the arguments are in the reverse order from `CategoricalCrossEntropy`. It would be nicer if they are the same i.e. `self, y, x`. Could you make that change?

yingzha: fixed it.

bartvm:
Sorry to keep making comments! I just noticed that you changed the crossentropy function two days ago, and that you're now using: `cost = -tensor.mean(log_prob[tensor.arange(y.shape[0]), y])`. This can be problematic, because Theano doesn't support advanced indexing on GPU, which means that this will force Theano to move all the data from GPU to CPU (and potentially back again).

Have a look at Pylearn2 to see how this is handled: https://github.com/lisa-lab/pylearn2/blob/master/pylearn2/models/mlp.py#L1429-L1438

yingzha:
Sorry I didn't realize it.. I'll fixed it

dwf:
Is this still true @bartvm? I was under the impression from reading the
docs last night that 0.6rc4 onwards had support for advanced indexing. I'll
have another look.

bartvm:
I actually tried before posting that, and noticed there's still a `HostFromGpu` op before `AdvancedSubtensor`, so I think it's still true, unless this is a bug in Theano.
```python
x = tensor.matrix('features')
y = tensor.lvector('targets')
from blocks.bricks import Softmax
cost = Softmax().categorical_cross_entropy(y, x)
f = theano.function([x, y], cost)
from theano.printing import debugprint as dp
dp(f)
```
```
Elemwise{Composite{(-(i0 / Cast{float32}(i1)))}}[(0, 0)] [@A] 'softmax_categorical_cross_entropy_output_0'   14
 |Sum{acc_dtype=float64} [@B] ''   13
 | |AdvancedSubtensor [@C] ''   12
 |   |HostFromGpu [@D] ''   11
 |   | |GpuElemwise{Sub}[(0, 0)] [@E] ''   10
 |   |   |GpuElemwise{Sub}[(0, 0)] [@F] ''   5
 |   |   | |GpuFromHost [@G] ''   1
 |   |   | | |features [@H]
 |   |   | |GpuDimShuffle{0,x} [@I] ''   4
 |   |   |   |GpuCAReduce{maximum}{0,1} [@J] ''   3
 |   |   |     |GpuFromHost [@G] ''   1
 |   |   |GpuElemwise{Log}[(0, 0)] [@K] ''   9
 |   |     |GpuDimShuffle{0,x} [@L] ''   8
 |   |       |GpuCAReduce{add}{0,1} [@M] ''   7
 |   |         |GpuElemwise{exp,no_inplace} [@N] ''   6
 |   |           |GpuElemwise{Sub}[(0, 0)] [@F] ''   5
 |   |ARange [@O] ''   2
 |   | |TensorConstant{0} [@P]
 |   | |Shape_i{0} [@Q] ''   0
 |   | | |targets [@R]
 |   | |TensorConstant{1} [@S]
 |   |targets [@R]
 |Shape_i{0} [@Q] ''   0
```

yingzha:
I used flatten indice as pylearn2, hope it would be okay.

bartvm:
LGTM, but Travis is offline, so I'll just wait for the tests to run later before merging, thanks!

rizar:
Can anybody explain me why we need both `Softmax.categorical_cross_entropy` and `CategoricalCrossEntropy` brick.

bartvm:
I did in https://github.com/bartvm/blocks/pull/326#issuecomment-75457998. The softmax isn't the only way of producing a probability distribution, so if you want to calculate the cross-entropy between two distributions which aren't the output of a softmax, you'll still need `CategoricalCrossEntropy`.

blocks/bricks/__init__.py
```diff
@@ -444,6 +444,38 @@ class Softmax(Activation):
...
+        """Return computationally stable softmax cost.
+
+        Parameters
+        ----------
+        y : :class:`~tensor.TensorVariable`
+            In the case of a matrix argument, each slice along
+            axis represents one distribution. In the vector case, each
+            element represents the position of the '1' in a one hot-vector.
+        x : :class:`~tensor.TensorVariable`
+            Each slice along axis represents one distribution.
```
rizar: Hmm, but `x` is not distribution but rather the energies that have to be transformed into a distribution!

yingzha: right, I'll fixed it!

rizar:
I am sorry for not reading your answers! This confusion of my was however a consequence of a documentation issue that I have already commented on.

PR #330, opened 2015-02-22
Save the original inner scan graph

rizar:
While thinking about the best way to implement beam search for the sequence generator I realized that there should be a way to access the original inner graphs built by brick application methods before they were arbitrarily modified by `theano.scan` internals. That turned out to be a matter of two lines. I kind of dislike that the input and output variables of the graph saved like I do it here are not properly annotated.

@dmitriy-serdyuk , that should be all you need to build none computation graphs in the beam search and only use one provided by the user.

### DIFF
--- a/blocks/bricks/recurrent.py
+++ b/blocks/bricks/recurrent.py
@@ -183,9 +183,8 @@ def recurrent(*args, **kwargs):
                 # We want to save the computation graph returned by the
                 # `application_function` when it is called inside the
                 # `theano.scan`.
-                application_call.inner_inputs = kwargs
-                application_call.inner_outputs = OrderedDict(
-                    zip(application.outputs, pack(outputs)))
+                application_call.inner_inputs = args
+                application_call.inner_outputs = outputs
                 return outputs
             outputs_info = (list(states_given.values()) +
                             [None] * (len(application.outputs) -
--- a/blocks/bricks/recurrent.py
+++ b/blocks/bricks/recurrent.py
@@ -177,9 +177,16 @@ def recurrent(*args, **kwargs):
                 args = list(args)
                 arg_names = (list(sequences_given) + list(states_given) +
                              list(contexts_given))
-                kwargs = dict(zip(arg_names, args))
+                kwargs = OrderedDict(zip(arg_names, args))
                 kwargs.update(rest_kwargs)
-                return application_function(brick, **kwargs)
+                outputs = application_function(brick, **kwargs)
+                # We want to save the computation graph returned by the
+                # `application_function` when it is called inside the
+                # `theano.scan`.
+                application_call.inner_inputs = kwargs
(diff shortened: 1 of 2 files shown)


blocks/bricks/recurrent.py
```diff
@@ -177,9 +177,16 @@ def scan_function(*args):
                 args = list(args)
                 arg_names = (list(sequences_given) + list(states_given) +
                              list(contexts_given))
-                kwargs = dict(zip(arg_names, args))
+                kwargs = OrderedDict(zip(arg_names, args))
                 kwargs.update(rest_kwargs)
-                return application_function(brick, **kwargs)
+                outputs = application_function(brick, **kwargs)
```
rizar: I have just got a crazy idea: what if instead we do
```python
getattr(brick, application_function.__name__)(**dict_union(kwargs, iterate=False))
```
here? That makes the inner graph to look exactly like if the user built it manually!

rizar:
So, two super important changes:

1. The `iterate=False` call used to have quite a lot of logic in common with `iterate=True`. For instance one could omit state arguments and those would be replaced by the initial values from the `initial_state` application method. I doubt whether anyone will ever want to use it. For this reason I move it to the top of `recurrent_apply` and `iterate=False` calls the application method as if it had `application` instead of `recurrent` decorator.

2. Instead of calling the user's code directly in the `scan_function` I am emulating an `iterate=False` called of the wrapped iteration method. The benefit is that the saved original inner graph is now annotated just as a normal `iterate=False` code, which makes it convenient to work with.

blocks/bricks/recurrent.py
```diff
@@ -169,17 +172,20 @@ def only_given(arg_names):
...
-                return application_function(brick, **kwargs)
-
             def scan_function(*args):
                 args = list(args)
                 arg_names = (list(sequences_given) + list(states_given) +
                              list(contexts_given))
                 kwargs = dict(zip(arg_names, args))
                 kwargs.update(rest_kwargs)
-                return application_function(brick, **kwargs)
+                outputs = getattr(brick, application_function.__name__)(
```
bartvm: Why the `getattr(brick, application_function.__name__)` instead of `application_function(brick, ...)`? I find the first one clearer (and shorter), actually, but maybe that's just a matter of taste.

rizar: In fact those are different!

`getattr` calls the decorated version, and the resulting computation graph is properly annotated. I will add an example of how one can use it to the test.

rizar:
@nouiz, can you please recommend a Theano routine that given input and outputs of two computation graphs would check that they are identical? By identical I mean that all variables have the same type and are linked with the same ops.

nouiz:
theano.gof.graph.is_same_graph(var1, var2, givens=None, debug=False).

See its docstring for more information

Issue #323, opened 2015-02-21
Propagating Updates to DataStreamMonitoring


rizar:
There is currently no way to push updates to the monitoring mechanism. This makes observing certain quantities impossible, if I am not missing anything, only in the case when the computation graph contains a scan with randomness in the inner graph. @jbornschein, was that your case?

Proposed solutions

1) `DataStreamMonitoring` uses the `ComputationGraph` interface to retrieve updates. In this case it can just use the `get_theano_function` method which compiles a function with the updates from annotations automatically.

2) The `Model` gets an additional method `get_updates`, to which the request from `DataStreamMonitoring` is made.

At the first thought I liked the second way more, because thanks to this indirection the user can intercept `get_updates` request and change the default logic. We can make a keyword argument for the model that will override the updates extracted in the default way. 

But on the other hand, `DataStreamMonitoring` might be used to monitor something very simple like average input magnitude, and in this case the updates required for the whole computation graph are not required. The solution 2 will simply result into compilation failure.

So (1) seems to be the only feasible way I see.

jbornschein:
I agree that (1) seems to be the preferred. This would also ensure that ```DataStreamMonitoring``` does not need any notion of  ```Model``` at all, which might be good for composability (some subsystem building a ```DataStreamMonitor``` for a sub-graph without knowing about the final top-level model). 

But to stay flexible and explicit, ```DataStreamMonotoring``` could have an additional ```updates=None``` argument. The docstring for this argument would IMHO also be a 
good place to describe that it would scan the computational graph for update-annotations when no explicit updates are supplied.

rizar:
Sounds good!

That sounds like a relatively easy fix, if you wish you could do it.

PR #331, opened 2015-02-23
Add update argument to DatasetEvaluator and to DatastreamMonitoring

jbornschein:
This patch introduces a new `updates` argument to the `DatasetEvaluator` and to `DatastreamMonitoring`. This is a first, incomplete fix for issue #323. We probably want to add an automatism that scans the computational graph for necessary updates when this argument is `None`.


memimo:
This is a great fix. Perhaps it would be helpful to have a mention of this problem in recurrent documentation as well, something like a note. I personally wasted a lot of time figuring out the scan problem with random variables till I found this patch.
I'd be happy to make a PR if you agree.

rizar:
I more than agree that it would be nice to have it mentioned in 
@recurrent docs!

Issue #339, opened 2015-02-24
Random Bricks / extensions


lukemetz:
Hello,
In the process of converting over an existing model to blocks I created a number of bricks and extensions that might be of use. (as well as a few that have already been added to blocks as I am slow)

They are all at: https://github.com/lukemetz/cuboid
They are mainly meant for me, but if any of them would be useful to the repo I will be happy to send a pullrequest(s) with the additions.

I have the following in a finished-ish state:
bricks.FilterPool,
bricks.Dropout (I saw the dropout thread already, but I wanted a separate brick for various reasons).
algorithms.NAG -- nesterov momentum
datasets.DataStreamBackground -- which runs a datastream in the background.

Totally fine if you don't want anything too. Thanks for the awesome library either way!

rizar:
Hi,

first of all thanks for your kind words!

Exactly for the cases like yours we plan to have another repository called `blocks-contrib`. This will host components from the users which can be useful but for one or another reason can not or should not be part of Blocks. I hope creation of `blocks-contrib` is a matter of a few days.

I think we would like to have Nesterov momentum in Blocks. Take a look at `BaseMomentum` and `Momentum` classes as example of how such things are done in Blocks. Or maybe better first wait until we launch `blocks-contrib` and push it there. We will much less picky for `blocks-contrib` regarding naming variables and docs, we will only ask for a unit-test to keep track of compatibility of the contributed code with cutting edge Blocks.

bartvm:
@lukemetz Two questions: What's your motivation for wanting a separate `Dropout` brick? If other people would want one for the same reasons, we could consider adding it.

Secondly, I'd be curious to know whether you tested your `DataStreamBackground` under heavy loads, on GPU, etc. and if so, how did it do? I've been thinking about something similar, but came up with something far more complicated (using a separately started server which pushes NumPy arrays as byte streams through sockets using ZeroMQ, https://github.com/bartvm/fuel/issues/7).

The main problems I'm worried about in your approach (but I might just be paranoid):

1. I've heard stories about CUDA going haywire if you fork a process after binding Theano to a GPU already. Did you experience any of this? This was the main reason I considered starting a separate server instead of using `multiprocessing`.
2. Using a `Queue` requires your data to be pickled and unpickled, which I imagine can be relatively slow compared to just pushing the raw byte stream. I'm guessing you might not notice any of this if your training isn't I/O bound i.e. if your computations take much longer than your data preprocessing. In your case, did you run into this?
[...]

PR #342, opened 2015-02-24
Adding wer

mohammadpz:
Hi @bartvm ,
I added Word Error Rate to the cost.py but I feel it does not look like your own codes at all :)
Could you please check it and let me know what's missing?


bartvm:
I'll read the code when I get to the lab, but what are you trying to measure exactly? AFAIK WER is simply how many times you predict the correct word in e.g. an LM. From what I saw, this calculates the edit distance between two sequences instead, which seems something very different to me?

rizar:
Your commit history suggests that you continued your RNN Toy Example. When you want to start another PR before finishing an ongoing one, create another branch by `git branch`.

From the code it looks like it is not a differentiable Theano cost and it does not make sense to wrap in a Brick. Also it is not entirely Block-ish thing: I could imagine such a routine in `blocks-contrib` (see #339) or even more likely in a separate speech recognition repo.

mohammadpz:
@bartvm , so, word error rate in speech recognition is different.
@rizar , my previous PR was on my ```master``` branch. I made another branch but it seems it still have the track of previous commits. Sorry by the way. Let's ask @pbrakel  about it too. @pbrakel , should I add it into another different repo?

bartvm:
Okay, but in that case the problem as @rizar points out is that none of this is differentiable. What you're looking for us a non-Theano monitor channel. This is something that we've kept in mind, but never implemented. This would be the right time to do so :)

Basically, I think you'll want to extend DataStreamMonitoring to execute arbitrary callbacks that take the output of a Theano function and the data stream as an argument, and writing the result to the log. This would also allow for monitoring e.g. BLUE score. The tricky part is that I guess the results of these callbacks also need to be aggregated in a particular way...

mohammadpz:
@bartvm , yes exactly. We don't use ```wer``` for training. We want it to monitor for validation. I can start adding it to the ```DataStreamMonitoring ``` but since I'm new I need your help for better understanding of Blocks :)

pbrakel:
I agree it doesn't really fit in the core of blocks very well. A public speech repo might be an option but the metric might still be common enough for other groups. I originally had contrib in mind for this but a generic monitor for non-theano variables would indeed be needed first (although @dmitriy-serdyuk already wrote something along those lines for our current implementation of the beamsearch).
@mohammadpz sorry for giving you a task that I hadn't thought through enough but I hope you still learned a bit about blocks in the process. At least we know what to do next.

mohammadpz:
@pbrakel , we may talk about this in our speech meeting today. BTW, it's perfectly okay and I still need to learn more about Blocks inner structure. :+1:

rizar:
@mohammadpz , a technical hint: before creating a new branch checkout your master:

```
git checkout master
git branch lets_try_something_new
```
This way I will not have old changes in your new branch.

@pbrakel, `blocks-contrib` is a good place for WER.

Monitoring non-Theano stuff should be included into 0.1 TODO list. And we need a ticket for that, of course. I would argue that Blocks should not care about averaging in this case, since some ways of averaging are just waaay to complex, e.g. BLEU. We should simple call one function for every batch and another to get an aggregated value.

bartvm:
Closing this for now, we'll see it back as a monitoring function later on...

rizar:
Opened #342.

Issue #162, opened 2015-01-27
Catch NaN in training algorithm


rizar:
There should be a possibility to avoid updating the parameter with `NaN` gradient.

rizar:
See discussion in #343.

bartvm:
http://deeplearning.net/software/pylearn2/library/devtools.html#module-pylearn2.devtools.nan_guard

bartvm:
And if you want to avoid updating instead of stopping, the new plugin by
@memimo does that, no?

rizar:
I want to avoid updating but raise an exception or warning. The new step rule just hides the NaNs.

dwf:
Relevant:
http://deeplearning.net/software/pylearn2/library/devtools.html#pylearn2.devtools.nan_guard.NanGuardMode

bartvm:
The new step rule avoids updating of you set the scaling to 1, so that + NanGuardMode might be enough?

dmitriy-serdyuk:
I don't like that by default blocks overrides the model file with parameters corrupted with NaNs and I have to retrain the model from scratch. So throwing an exception seems reasonable in some cases.

dmitriy-serdyuk:
Is it better to thow an exception in `Dump` class when trying to serialize NaN?

dwf:
You will probably want to be very careful about that to make sure you avoid
partially serializing something before you throw that exception.

dmitriy-serdyuk:
@dwf Yes, sure. It can be solved with a temp file.

rizar:
While this is true that a temp file would do the job, I feel it much more naturally to handle these things in the algorithm. This way when you get NaNs an exception will be raised and the main loop from the point just before you got NaN will be saved in a separate file.

rizar:
Is the `NanGuardMode` as fast as `fast_run`?

bartvm:
Theoretically, no solution is ever going to be as fast as not checking, right? Even if you could optimize the hell out of it, at the very least you still need to make an element-wise comparison for each parameter which you otherwise wouldn't have to.

You won't want to handle this for each possible algorithm, but you can probably add it to the `MainLoop`. If the error occurs, catch it, print an informative warning, and go straight to `after_training` extensions (instead of `on_error` ones).

rizar:
I did not mean "exactly as fast" as `fast_run`, I meant that I would not like to break optimizations or use less optimization compared to `fast_run`.

bartvm:
[...]

PR #343, opened 2015-02-24
RemoveNotFinite step rule

memimo:
Simple trick to remove nan, inf from gradients. Thanks to @jych 
Not the best name chosen, and can change if anyone has better suggestion.

### DIFF
--- a/blocks/algorithms/__init__.py
+++ b/blocks/algorithms/__init__.py
@@ -590,3 +590,29 @@ class Adam(StepRule):
                    (time, t1)]
 
         return step, updates
+
+
+class RemoveNotFinite(StepRule):
+    """A step rule that replaces non-finite gradients.
+
+    Replaces steps with non-finite norm (`inf` or `NaN`) with
+    a scaled version of the parameter being updated.
+
+    source of the trick: https://github.com/lisa-groundhog/GroundHog
+
+    Parameters
+    ----------
+    scaler : float, optional
+        default value set to 0.1
+
+    """
+    def __init__(self, scaler=0.1):
+        self.scaler = scaler
+
+    def compute_step(self, param, previous_step):
+        grad_norm = l2_norm([previous_step])
+        not_finite = tensor.or_(tensor.isnan(grad_norm),
+                                tensor.isinf(grad_norm))
+        step = tensor.switch(not_finite, self.scaler * param, previous_step)
+
+        return step, []
--- a/tests/algorithms/test_algorithms.py
+++ b/tests/algorithms/test_algorithms.py
@@ -7,7 +7,8 @@ from theano import tensor
 
 from blocks.algorithms import (GradientDescent, StepClipping, CompositeRule,
                                Scale, StepRule, BasicMomentum, Momentum,
-                               AdaDelta, BasicRMSProp, RMSProp, Adam)
+                               AdaDelta, BasicRMSProp, RMSProp, Adam,
+                               RemoveNotFinite)
 from blocks.utils import shared_floatx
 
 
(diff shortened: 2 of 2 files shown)


blocks/algorithms/__init__.py
```diff
@@ -590,3 +590,28 @@ def compute_step(self, param, previous_step):
                    (time, t1)]
 
         return step, updates
+
+
+class RemoveNotFinite(StepRule):
+    """If gradients norm is inf or nan.
```
bartvm: You'll need a blank line after this... Sorry, the docstring checker is picky!

blocks/algorithms/__init__.py
```diff
@@ -590,3 +590,29 @@ def compute_step(self, param, previous_step):
                    (time, t1)]
 
         return step, updates
+
+
+class RemoveNotFinite(StepRule):
+    """A step rule that replaces non-finite elements.
+
+    Replaces non-finite elements (`inf` or `NaN`) in the step with
```
bartvm: Is this really what it does? If I read the code it seems like it replaces the *entire* step with the scaled parameter, instead of doing it entry-wise (which is how I read this docstring).

memimo: Yeah, this is not correct. @dwf suggested this docstring. (For some odd reason, his comment was on my fork not here.)
How about replacing first line with:
``Replaces steps with non-finite norm (`inf` or `NaN`) with``

dwf: `tensor.switch` operates element-wise.

dwf: Oh, I misread. Sorry.

rizar:
This overlaps with #162, but is different, I guess we should have both.

More in detail: I would also like to be able to avoid updating parameters when the step contains `NaN`, so that the last valid parameters were available for analysis. For this I guess we will have to split the single Theano function of the `DifferentiableCostMinimizer` into two, one computing step and another one subtracting it from parameters, similarly like it was done in Groundhog.

bartvm:
You want to compile two separate Theano functions, one which calculates the step and another one which applies it (using the output of the first as an input)? That sounds woefully inefficient.

rizar:
... and another one that subtracts it. What is so inefficient about it? Computation-wise it is the same, memory-wise it is harder to tell.

bartvm:
You could be stopping Theano from performing a whole bunch of optimizations, for example memory-wise as you mention (perhaps Theano likes updating parameters one by one, not needing all of the steps to be available simultaneously). But most importantly, wouldn't this force Theano to push the step from GPU to CPU (as output from the first function) and then back onto the GPU (as input to the second function)? That's a massive bandwidth issue.

rizar:
Why would it force to do such an awful thing? 

I assume that the step is stored in shared variables on GPU, which guarantees that both function will be performed without copying to the main memory. In fact this is exactly the way it is done in Groundhog.

Memory-wise it might be indeed inefficient.

bartvm:
If you keep it on GPU, that could potentially be almost a doubling in memory, because you need to keep a step for each parameter in your model, while Theano could otherwise do them weight matrix by weight matrix. If that's really how it was done in GroundHog, I'm surprised the models fit on GPUs at all!

rizar:
I doubt that Theano is smart enough to do that! However if it is not as smart, which I would expect, the inefficiency is still there. 

I guess it the time to ask Fred what is the recommended way in Theano to skip an update if it would lead to corruption with NaNs...

dwf:
Why not just switch on any(isnan(update)) ? Or take the disjunction across
all proposed updates.

bartvm:
I guess because there's no `any` operator in Theano, is there? I guess that `l2_norm` is a bit strange though, a simple reduce like `sum` would have the same effect.

dwf:
`theano.tensor.basic.any` seems to exist for me.

bartvm:
Ah, great, my bad! I was looking for it in the documentation with other logical functions like `and_` and `or_`...

bartvm:
#352 better?

PR #344, opened 2015-02-24
before_training as string

bartvm:
@rizar Came across this when reading the code. The name of a function is always `'before_training'` (if not, `self.before_training` would return an `AttributeError`) so not sure why it was done like this.

### DIFF
--- a/blocks/extensions/monitoring.py
+++ b/blocks/extensions/monitoring.py
@@ -124,7 +124,7 @@ class TrainingDataMonitoring(SimpleExtension, MonitoringExtension):
         aggregated values of the monitored variables to the log.
 
         """
-        if callback_name == self.before_training.__name__:
+        if callback_name == 'before_training':
             if not isinstance(self.main_loop.algorithm,
                               DifferentiableCostMinimizer):
                 raise ValueError


rizar:
I was just trying to be typo-proof. I think instead of using plain strings we should be using constants wherever possible. We could automatically add upper-cased callback names as constants to the `TrainingExtension` and write `self.BEFORE_TRAINING` - looks more readable.

bartvm:
Okay, I see what you mean, but it feels a bit strange to have in committed code.

We could change everything to using constants and throw away the use of strings to identify callbacks entirely. Although technically safer, I wonder whether we're not just uselessly complicating code. I find `self.BEFORE_TRAINING` awkward, and would prefer `TrainingExtension.BEFORE_TRAINING` then, but I'm wondering if we're not just complicating things. My vote is for just using strings. Maybe one day a single typo will sneak into the master branch, but I'm confident we'll find it soon enough :)

rizar:
Not sure I agree with you. The bugs when something is not called because of a typo are sometimes  really unpleasant to catch. I would rather switch to using upper-cased module-level or class-level constants and add the corresponding "search-and-replace" task to the CCW pool.

bartvm:
I could live with that, but I do still think it needlessly complicates
things and is quite old school, defensive programming :) reducing code
legibility and increasing complexity all in the name of security. It's like
NumPy not letting you use `dtype == 'float32'`, which I would find
frustrating. I think it's okay to accept that users will make mistakes;
they'll make less if we don't expect them up use custom constants for the
simplest string comparisons.

rizar:
I do not think that constants are old-school.

Assume Blocks become super popular, have 10000 users, and all of a sudden we decide to change a name of a callback. If all 10000 users use upper-cased constants, their code fails in a predictable way: no such attribute. If they all have `before_training` hard-coded, every code will fail in its own hard-to-predict way. Of course good code should have some protection for his case like an exception "oops! unknown callback name", but our users will be researchers, who are not always good coders.

bartvm:
You could also solve that by making the `callback` that is passed to `do` not a string but something with a custom equality operator, e.g.

```python
class Callback(object):
    def __init__(self, callback):
        self.callback = callback

    def __eq__(self, other_callback):
        if other_callback not in ['before_epoch', 'after_epoch', ...]:
            raise ValueError("This callback doesn't exist")
        return other_callback == self.callback
```
This is what NumPy does as well:
```python
>>> numpy.array([1, 2, 3]).dtype == 'int63'
TypeError: data type "int63" not understood
```

rizar:
This is fine with me. As long as the situation I describe above is not possible, I am happy.

PR #346, opened 2015-02-24
Orthogonal initialization for non-squared matrices. 

jbornschein:
Before this patch Orthogonal initialization would only work for square matrices. 

In fact, it would silently(!) set shared weight variables to be square even though their shape was set to some other aspect ratio. As such I think the old behavior was even erroneous.

### DIFF
--- a/blocks/initialization.py
+++ b/blocks/initialization.py
@@ -156,16 +156,26 @@ class Identity(NdarrayInitialization):
 class Orthogonal(NdarrayInitialization):
     """Initialize a random orthogonal matrix.
 
-    Only works for 2D, square arrays.
+    Only works for 2D arrays.
 
     """
     def generate(self, rng, shape):
-        M = rng.randn(*shape).astype(theano.config.floatX)
+        if len(shape) != 2:
+            raise ValueError
+        M1 = rng.randn(shape[0], shape[0]).astype(theano.config.floatX)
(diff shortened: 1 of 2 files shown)


bartvm:
I should probably have checked the shape, apologies :) This LGTM, but do the `atol` and `rtol` really need to be that high? A matrix with 1.3 on the diagonal and 0.2 everywhere else seems hardly equal to the identity.

jbornschein:
rtol can basically be removed.

But atol needs to stay around 0.2 -- at least with the matrix shapes I used in those test cases. Especially in the non-square cases we find some matrix elements that exceed 0.1.

bartvm:
The mean doesn't seem much better, and is probably still quite high. The problem occurs when `W` is `(n x m)` or `(m x n)` and your output matrix `WWT` or `WTW` is of size `(n x n)` with `n > m`. It makes sense that you can't reconstruct a good identity in this case. All the other cases you can probably do with much, much lower `atol`.

Then again, we're not testing the numerical accuracy of the procedure here, just whether a blatant mistake was made or not. So if you can't be bothered, fine by me.

I don't think the special case actually will increase the accuracy enough to be significant, but keep it anyway, because I know that for machine translation scripts simply initializing all the weights (and there are a lot) to be orthogonal takes a while, so a 2x speedup is worthwhile.

PR #352, opened 2015-02-25
Simplify NaN-check

bartvm:


dwf:
LGTM

However, I noticed something else: the docstring is a little ambiguous. It says "replaces steps"; it could be read to suggest that it replaces the *entire step* if any of the updates go NaN or Inf, when in fact it does so on a parameter by parameter basis. This is what's done in the clipping step, after all.

bartvm:
Mmm, it's tricky though, because it doesn't do it on an actual parameter by parameter basis. It does it on a weight matrix by weight matrix basis :/

dwf:
Yeah, by "parameter" I mean "shared variable", sorry.

dwf:
I've been thinking about this in other contexts too, i.e. that it
would be nice to be able to compose step rules on a subset of the
parameter tensors in a model, i.e. to add learning rate scalers on
convolution maps, or clip the gradient of only the recurrent weights
(to be honest, that's what I thought the standard thing was -- am I
wrong?).

bartvm:
Yeah, that does sound reasonable. Not sure what the best way of implementing it is. I guess there are two ways: Either we give all step rules automatic support for e.g. `exclude` or `filter` keyword arguments, which makes them pass on excluded parameters/all parameters that don't match the filter without changing them.

The other way would be to create a a `FilterStep` which works like `step_rule=FilterStep(StepClipping(threshold=100), VariableFilter(roles=[RECURRENT_WEIGHTS])(cg.shared_variables))`. I guess this could be more powerful because you could easily filter weights for an entire sequence by doing something crazy like `step_rule=FilterStep(CompositeRule(...), VariableFilter(roles=[BIASES])(cg.shared_variables))`.

dwf:
Rephrasing sounds good to my ear. Filed the other discussion as #353.

PR #355, opened 2015-02-25
Add support for Theano known_grads in GradientDescent.

dwf:
See the docstring diff for an explanation. (I am in a situation where I will need this if I ever want to port a certain model to Blocks, so I figured it makes sense to add it now.)


dwf:
@bartvm Any comments on this/does this Travis build need restarting?

rizar:
Have no idea why Travis is so mean.

PR #356, opened 2015-02-25
Added references for shared parameters of learning algorithms

dmitriy-serdyuk:
I added references to parameters of inner components of learning algorithms. It should help to use `SharedParameterModifier`, for example one should write
```python
SharedParameterModifier(momentum.momentum)
```
instead of
```python
SharedParameterModifier(momentum.components[0].momentum)
```

### DIFF
--- a/blocks/algorithms/__init__.py
+++ b/blocks/algorithms/__init__.py
@@ -376,7 +376,7 @@ class Momentum(CompositeRule):
     momentum : :class:`~tensor.SharedVariable`
         A variable for momentum.
 
-    See also
+    See Also
     --------
     :class:`SharedVariableModifier` for a parameter decay during the
     training.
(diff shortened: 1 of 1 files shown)


bartvm:
Cool, but could you document the attributes in an Attributes section (see the NumPy docstring standards for examples).

bartvm:
Cool :) The error is there because it's `See Also` and not `See also`. Sorry, the docstring checker is picky!

blocks/algorithms/__init__.py
```diff
@@ -369,10 +369,25 @@ class Momentum(CompositeRule):
...
+    --------
+    :class:`SharedVariableModifier` for a parameter decay during the
+    training.
+
     """
     def __init__(self, learning_rate=1.0, momentum=0.):
-        self.components = [Scale(learning_rate=learning_rate),
-                           BasicMomentum(momentum=momentum)]
+        scale = Scale(learning_rate=learning_rate)
+        momentum_alg = BasicMomentum(momentum=momentum)
```
rizar: Please rename `momentum_alg` to `basic_momentum`. Sorry for being that picky, but we allow only a few very popular abbreviations in blocks, such as `init`, `param`, `var` and so on. `alg` does not belong to this category.

I will merge right after you fix that, honestly!

PR #365, opened 2015-02-26
Skip tests if there is no Bokeh server running. Fixes #358.

### LINKED ISSUE #358: Skip bokeh tests if there is no server
Currently it skips if the module isn't available, but if the module is available but there is no server running, it still crashes. It should skip in this case.

bartvm:


tests/__init__.py
```diff
@@ -54,6 +54,15 @@ def skip_if_not_available(modules=None, datasets=None, configurations=None):
             import_module(module)
         except Exception:
             raise SkipTest
+        if module == 'bokeh':
+            ConnectionError = \
```
rizar: Parentheses?

Sorry, I am a real line break hater!

rizar:
LGTM!

PR #367, opened 2015-02-26
MLP now works with bricks and application functions

### LINKED ISSUE #220: Allow passing of application methods to MLP
Currently the MLP class only takes Bricks and makes makes the assumption that the application method to be used is `apply`. This currently prevents you from using any brick application with a different name.

dmitriy-serdyuk:
FIxes #220 .

### DIFF
--- a/blocks/bricks/__init__.py
+++ b/blocks/bricks/__init__.py
@@ -518,7 +518,8 @@ class MLP(Sequence, Initializable, Feedforward):
 
     Parameters
     ----------
-    activations : list of :class:`.Brick` or ``None``
+    activations : list of :class:`.Brick`, :class:`ApplicationCall`,
+                  or ``None``
         A list of activations to apply after each linear transformation.
         Give ``None`` to not apply any activation. It is assumed that the
         application method to use is ``apply``. Required for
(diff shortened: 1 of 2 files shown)


blocks/bricks/__init__.py
```diff
@@ -553,8 +554,14 @@ def __init__(self, activations, dims, **kwargs):
...
                                        for i in range(len(activations))]
         # Interleave the transformations and activations
-        application_methods = [brick.apply for brick in interleave(
-            [self.linear_transformations, activations]) if brick is not None]
+        application_methods = []
+        for entity in interleave([self.linear_transformations, activations]):
+            if entity is None:
+                continue
+            if isinstance(entity, Brick):
+                application_methods += [entity.apply]
```
bartvm: Please use `.append`, which should be much faster; I think list concatenation causes a new list to be created instead of the old one being appended to.

blocks/bricks/__init__.py
```diff
@@ -518,7 +518,8 @@ class MLP(Sequence, Initializable, Feedforward):
 
     Parameters
     ----------
-    activations : list of :class:`.Brick` or ``None``
+    activations : list of :class:`.Brick`, :class:`ApplicationCall`,
```
rizar: :class:`BoundApplication`?

bartvm: Oops, missed that one, thanks! Fixed in aa0dfdf.

PR #372, opened 2015-02-27
1D initialization for Identity and Orthogonal 

### LINKED ISSUE #366: Non 2d weight parameters
The LSTM class uses the WEIGHTS role for a 1d weight vector:

https://github.com/bartvm/blocks/blob/master/blocks/bricks/recurrent.py#L352

while some of the initialization objects only operate on 2d matrices. E.g. Identity and Orthogonal.

Currently the code will just crash when using such an initialization. I see two ways to approach this issue: a) A brick allocating 1d weight matrices has to ensure that the user can supply a different, suitable initialization method for those parameters, or b) all initialization classes meant for matrices should provide a reasonable 1d initialization. 

a) is already implemented because we can of course alter the initialization scheme for different parameters after object creation... but this issue should probably be mentioned in the LSTM documentation. Maybe the constructor should take an additional 'weights_1d_init' argument?

b) seems preferable from a 'things should just work' perspective. But maybe the user should better be aware of these parameters that don't fit the chosen initialization scheme?

jbornschein:
Fixes #366 by allowing 1D initialization.

### DIFF
--- a/blocks/initialization.py
+++ b/blocks/initialization.py
@@ -134,8 +134,9 @@ class Uniform(NdarrayInitialization):
 class Identity(NdarrayInitialization):
     """Initialize to the identity matrix.
 
-    Only works for 2D arrays. If the number of columns is not equal to the
-    number of rows, the array will be truncated or padded with zeros.
+    Only works for 1D and 2D variables. If the number of columns is not
+    equal to the number of rows, the array will be truncated or padded with
+    zeros. For 1D variables all elements but the first one will be zero.
 
     Parameters
     ----------
@@ -147,21 +148,40 @@ class Identity(NdarrayInitialization):
         self.mult = mult
 
     def generate(self, rng, shape):
-        if len(shape) != 2:
-            raise ValueError
-        rows, cols = shape
-        return self.mult * numpy.eye(rows, cols, dtype=theano.config.floatX)
+        if len(shape) == 1:
+            eye_matrix = numpy.eye(1, shape[0], dtype=theano.config.floatX)
+            return self.mult * eye_matrix[0]
+        elif len(shape) == 2:
+            rows, cols = shape
+            eye_matrix = numpy.eye(rows, cols, dtype=theano.config.floatX)
+            return self.mult * eye_matrix
+        else:
+            raise ValueError("Identity initialization only works for 1D and"        
+                             " 2D variables")
 
 
 class Orthogonal(NdarrayInitialization):
-    """Initialize a random orthogonal matrix.
+    """Initialize a random, approximately orthogonal matrix.
 
-    Only works for 2D arrays.
+    Only works for 1D and 2D variables.
 
+    When used to initialize a 1D parameter vector the elements will
+    be drawn i.i.d. from a normal distribution.
+
+    For 2D matrices we initialalize such that
+    :math:`M M^T \approx I` and :math:`M^T M \approx I` where
+    :math:`I` is an identity matrix of appropriate size.
+
+        
     """
     def generate(self, rng, shape):
-        if len(shape) != 2:
-            raise ValueError
+        if len(shape) not in (1, 2):
+            raise ValueError("Orthogonal initialization only works for 1D and"        
+                             " 2D variables")
+
+        if len(shape) == 1:
+            # Special case for 1D variables
+            return rng.randn(*shape).astype(theano.config.floatX)
 
         if shape[0] == shape[1]:
             # For square weight matrices we can simplify the logic
--- a/blocks/initialization.py
+++ b/blocks/initialization.py
@@ -156,12 +156,12 @@ class Identity(NdarrayInitialization):
             eye_matrix = numpy.eye(rows, cols, dtype=theano.config.floatX)
             return self.mult * eye_matrix
         else:
-            raise ValueError("Identity initialization only works for 1D and"        
+            raise ValueError("Identity initialization only works for 1D and"
                              " 2D variables")
 
 
 class Orthogonal(NdarrayInitialization):
-    """Initialize a random, approximately orthogonal matrix.
+    r"""Initialize a random, approximately orthogonal matrix.
 
     Only works for 1D and 2D variables.
 
@@ -172,11 +172,10 @@ class Orthogonal(NdarrayInitialization):
     :math:`M M^T \approx I` and :math:`M^T M \approx I` where
     :math:`I` is an identity matrix of appropriate size.
 
-        
     """
     def generate(self, rng, shape):
         if len(shape) not in (1, 2):
-            raise ValueError("Orthogonal initialization only works for 1D and"        
+            raise ValueError("Orthogonal initialization only works for 1D and"
                              " 2D variables")
 
         if len(shape) == 1:
--- a/tests/test_initialization.py
+++ b/tests/test_initialization.py
@@ -4,7 +4,7 @@ import theano
 from numpy.testing import assert_equal, assert_allclose, assert_raises
 
 from blocks.initialization import Constant, IsotropicGaussian, Sparse
-from blocks.initialization import Uniform, Orthogonal
+from blocks.initialization import Uniform, Identity, Orthogonal
 
 
 def test_constant():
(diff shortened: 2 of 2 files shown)


blocks/initialization.py
```diff
@@ -147,21 +148,39 @@ def __init__(self, mult=1):
         self.mult = mult
 
     def generate(self, rng, shape):
-        if len(shape) != 2:
-            raise ValueError
-        rows, cols = shape
-        return self.mult * numpy.eye(rows, cols, dtype=theano.config.floatX)
+        if len(shape) == 1:
+            eye_matrix = numpy.eye(1, shape[0], dtype=theano.config.floatX)
```
rizar: `self.mult * numpy.ones((shape[0]))`?

rizar:
I do not mind merging this, but I think that the right way to solve the issue with `LSTM`, is to add let's say `multipliers_init` for 1-D vectors which are not biases. Your choice of "orthogonal" initialization for 1-D vectors is rather arbitrary and might cause some issues, the only way of resolving which would be using this `multipliers_init`.

blocks/initialization.py
bartvm: That's not the same thing, this command basically creates a vector of zeros with a 1 in the beginning.

blocks/initialization.py
```diff
@@ -147,21 +148,39 @@ def __init__(self, mult=1):
...
     def generate(self, rng, shape):
-        if len(shape) != 2:
-            raise ValueError
+        if len(shape) not in (1, 2):
+            raise ValueError("Orthogonal initialization only works for 1D and"
+                             " 2D variables")
+
+        if len(shape) == 1:
+            # Special case for 1D variables
+            return rng.randn(*shape).astype(theano.config.floatX)
```
bartvm: Wouldn't it make more sense to normalize this vector so that the dot product is 1? That's the property of all orthonormal vectors in an orthogonal matrix after all.

blocks/initialization.py
rizar: Do you think it is a nice initialization? I am not sure if it makes any sense.

bartvm: No idea whether it makes sense as an initialization, but it's the only correct extension to 1D cases. A `(n, 1)` or `(1, n)` truncated identity matrix would look the same.

blocks/initialization.py
rizar: Hmm, it is not that clear! I would expect that if I multiple an orthogonally initialized vector element-wise with another vector, the expected norm of the results should not change. In this sense the current initialization is better.

rizar:
I am sorry, I joined too late to this discussion, but from comments on the code we can already see that we are entering very shady area when we expand originally matrix-only initialization schemes on non-matrices. Let's better extend the LSTM interface and keep these schemes matrix-only.

blocks/initialization.py
bartvm: I don't see why? The norm of a random vector like this with another random vector is in expectation `sqrt(n) * ||other vector|| * cos(theta)``, meaning that the expected norm of results is dependent on the size of this random vector. If you don't want that, you need to normalize as I suggested.

The expected norm of the final result is always zero though, because you're projecting an *n*-dimensional vector into a 1-dimensional space (i.e. the expected value of `cos(theta)` is 0).

rizar: You are considering the case of a proper dot-product, and in this case you are right. But as far as I can read `LSTM` code there 1D vectors are applied for elemwise multiplication.

bartvm: Perhaps, but the scheme is `Orthogonal` not `RNNInitialization` :) Like with the identity matrix, there is only one correct way IMHO to extend orthogonal initializations into 1D, which is to have a vector of unit norm. If you want another use case, maybe there should be a special `LSTM` scheme that basically uses orthogonal for 2D matrices, but does something else for 1D cases.

rizar: > Perhaps, but the scheme is `Orthogonal` not `RNNInitialization`

So what? People will plug it into LSTM and feel frustrated because training does not happen! They will do it because for other recurrent networks the orthogonal initialization scheme is quite popular. I think we do no want that.

The fact that these schemes did not work for 2-matrices was a nice protection against extremely weird bugs. @jbornschein, do you have any other usecase in mind except for initialization of the weight vectors in LSTM?

rizar: I gave it another thought, and here is an idea: we make this behavior optional under a flag "allow_non_squares". 

```python
lstm = LSTM(dim=100, weights_init=Orthogonal(), biases_init=Constant(0))
lstm.initialize()
```

would print

```
Exception: can not initialize vector.

By default the Orthogonal initialization scheme works only for square matrices. If you want to allow
vector initialization, use Orthogonal(allow_non_squares=True), in which case an approximation 
for true orthogonality is used. See documentation for more details. ```

Comment (bartvm) [inline blocks/initialization.py]: > So what?

If the name has no relation to what is output, I suggest `Uniform` should from now on output a Weibull distribution on Thursdays, but only if the weather is nice.

If you want to special case things for LSTM, I don't see the problem of writing:
```python
OrthogonalLSTM(NdarrayInitialization):
    """Orthogonal initialization on weight matrices, random initialization for the gates."""
    def generate(self, rng, shape):
        if len(shape) == 1:
            return IsotropicGaussian().generate(rng, shape)
        if len(shape) == 2:
            return Orthogonal().generate(rng, shape)
```

jbornschein: Ok, just to recap: 

The way we initialize 1D vectors clearly depends on their usage. If we initialize a vector that is effectively used as a row or column of a matrix, the Orthogonal initialization should be normalized to one. 
If we use it as the diagonal of a connectivity matrix, the (most strict) generalization would  be numpy.ones(shape[0]). I tested this on some small toy tests -- and it gives neither good results, nor is it very intuitive. From that perspective (1. + randn(shape[0])) also looks like a reasonable candidate. 

So when initializing 1D 'weight-vectors' it's neither obvious if they are used coumn/row-wise or as diagonal, Nor is it obvious that the same initialization scheme is well defined and meaningfull even when we know how it is used.

Given all this ambiguity I would propose to keep Identity and Orthogonal as they are: only working for 2D vars, and give an additional argument to LSTM constructor

```
lstm = LSTM(dim=100, weights_init=Orthogonal(), 
diagonal_weights_init=IsotropicGaussian(), biases_init=Constant(0))
```

The default for diagonal_weights_init could be to use the same as weights_init.

rizar: If we stick to the exact correspondence between the name and action taken, then only square matrices can be orthogonal.

But I am not sure that you are right, @bartvm. If the initialized object is a `1 x n` matrix, then indeed the expected operation with it is the dot product, and the logical extension of the orthogonality concept would be to make it have norm 1. But if it is a Theano vector, then most likely nobody's going to compute dot products with it. Element-wise multiplication is a more typical usage of such a tensor. And if we ask ourselves a question how a vector should be initialized so that element-wise multiplication with this vector would not change its norm, one of the possible answers is to make his elements i.i.d with zero mean and unit variance. In fact this would generalize to any tensor with number of dimensions other than two.

Apart from that, my main practical consideration: we should not make it possible that switching from `SimpleRecurrent(weights_init=Orthogonal())` to `LSTM(weights_init=Orthogonal())` yields a badly initialized network. That's super dangerous.

I propose that regardless of what initialization for 1-D vector we converge on, it should be explicitly activated by a switch, forcing the user to read the docs after performing a switch like I described above. After that I do not care too much what exactly we choose for those degenerate cases, but maybe if there are debates we should not do anything.

rizar: Cross-posting!

I agree with you, @jbornschein. We could make `diagonal_weights_init` an optional parameter, for which `weights_init` would be used if no value is given. We can also create a new roles subclassing from `WEIGHTS`: `DIAGONAL_WEIGHTS`.

jbornschein: We already merged code that extends Orthogonal to non-square matrices. In the non-quare case it approximates orthogonality by ensuring M M^T ~= I and M^T M ~= I . Do you think that this should be activated by flag only too? Or that non-sqared orthogonal initialization should be called ApproximateOrthogonal ?

rizar: I am not 100% sure, but am I inclined to say that yes, it should go under another name or with a flag for security reasons.

bartvm: Although not as clearly defined, many people refer to a rectangular matrix where the columns are orthonormal as being an "orthogonal matrix". If you then consider a vector as a matrix with one row or column, it should have a dot product of one. The distinction of a dot product of a `1 x n` matrix with its transpose and the inner product of a vector of length `n` is a purely programmatic one; mathematically they are equivalent (and I don't like making the assumption that no one is going to compute dot products with vectors).

I understand the danger of having a badly initialized network, but shoehorning an initialization scheme to be mathematically incorrect based on an assumption of what the scheme will be used for is not the way to go. And it is just that, an assumption, because who is to say that someone doesn't want a vector that really is like a column from an orthogonal matrix?

I like @jbornschein's idea much better. This makes the initialization explicit to the user instead of being hidden away inside a magical initialization scheme.

bartvm: Cross-posting, but I don't think you would need a flag for this. Calling rectangular matrices orthogonal when referring to the fact that their columns are orthonormal is common practice.

jbornschein: I agree with @bartvm, initializing non-square matrices with approximate orthogonal ones seems to be so common, that adding an additional flag (allow_the_common_thing=True)  is maybe too much.

The 1D case on the other hand seems so degenerate, that it should probably not done automatically.

rizar: I can live with the 2D case handled automatically. Although I am not aware of any paper doing like that and saying that they use orthogonal matrices. Maybe it a new jargon of LISA which has been developed after I left it :)

Should we close the PR?

jbornschein: Yes, closing this PR.

I will create a new one with diagonal_weight_init added to the LSTM class.

PR #373, opened 2015-02-27
Fuel: ContainerDataset name change and no more get_default_stream

bartvm:
Parallel PR to https://github.com/bartvm/fuel/pull/20, see list of full changes there.

Most important for Blocks: `ContainerDataset` is now `IterableDataset`, `get_default_stream` is now `get_example_stream` and no longer returns batches for some datasets (only examples).

### DIFF
--- a/docs/plotting.rst
+++ b/docs/plotting.rst
@@ -41,9 +41,9 @@ We train on a 150 random points in :math:`[0, 1]`.
 
 >>> import numpy
 >>> from fuel.streams import DataStream
->>> from fuel.datasets import ContainerDataset
+>>> from fuel.datasets import IterableDataset
 >>> sample = theano.tensor.scalar('data')
->>> data_stream = DataStream(ContainerDataset(
+>>> data_stream = DataStream(IterableDataset(
 ...     numpy.random.rand(150).astype(theano.config.floatX)))
 
 Now let's train with gradient descent and plot the results.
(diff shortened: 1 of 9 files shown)


docs/plotting.rst
```diff
@@ -41,9 +41,9 @@ We train on a 150 random points in :math:`[0, 1]`.
 
 >>> import numpy
 >>> from fuel.streams import DataStream
->>> from fuel.datasets import ContainerDataset
+>>> from fuel.datasets import IterableDataset
 >>> sample = theano.tensor.scalar('data')
->>> data_stream = DataStream(ContainerDataset(
+>>> data_stream = DataStream(IterableDataset(
```
rizar: Perhaps I a bit late with this, but I can't help reading `IterableDataset` as "a dataset that supports iteration" and not "dataset made from iterables"

rizar:
`get_example_stream` reads definitely easier than `get_default_stream`

docs/plotting.rst
bartvm: Mm, I see your point, but I hate the idea of throwing the name around for this one case and using "DatasetOfIterables" or something weird. "Iterables" are the only correct name IMHO, because the only property it relies on is being able to iterate over things. `ContainerDataset` doesn't work, because there is no requirement for any of the sources to be containers.

Any  suggestions? I'd be fine with keeping `IterableDataset` though; it should be clear from reading the summary line alone what it is already.

rizar: Well, in fact "a dataset that supports iteration" is also entirely true, so let's keep it list it currently is.

PR #374, opened 2015-02-27
Streamline parameter passing to algorithms

### LINKED ISSUE #364: Default filtering of graph in TrainingAlgorithm: maybe use roles?
Currently ```DifferentiableCostMinimizer``` finds parameters for gradient calculation simply by taking all ```shared_variables``` by default . Is this a design decision? To me this sounds like it should filter based on ```PARAMETER``` role. Maybe use ```shared_variables``` only if no item with role ```PARAMETER``` was found in case someone has created their custom graph and didn't bother with roles.

I bumped into this after adding a noise source by manually creating RandomStream. Apparently RandomStream internally uses shared variables and injects some to the graph causing ```theano.tensor.grad``` to fail

```      
 self.params = (params if params
                       else ComputationGraph(cost).shared_variables)
```

arasmus:
DifferentiableCostMinimizer needs to know which parameters are being
optimized and currently it gets all shared variables from the graph.
This is bad because assumes that all shared variables are trainable
that is not true for constants such as ones in MRG_RandomStream.

Having user always specify parameters improves transparency and
keeps framework modular. For convenience, add new property to
ComputationGraph. To avoid circular dependencies, implement role lookup
in roles.py and make both VariableFilter and ComputationGraph use it.

Fixes #364

### DIFF
--- a/blocks/roles.py
+++ b/blocks/roles.py
@@ -39,6 +39,26 @@ def add_role(var, role):
     var.tag.roles = roles
 
 
+def has_roles(var, roles, match_all=False):
+    r"""Test if a variable has given roles taking subroles into account.
+
+    Parameters
+    ----------
+    var : :class:`~tensor.TensorVariable`
+        Variable being queried.
+    roles : an iterable of :class:`.VariableRole` instances.
+    match_all : bool, optional
+        If ``True``, checks if the variable has all given roles.
+        If ``False``, any of the roles is sufficient.
+        ``False`` by default.
+
+    """
+    var_roles = getattr(var.tag, 'roles', [])
+    matches = (any(isinstance(var_role, role.__class__) for
+                   var_role in var_roles) for role in roles)
+    return all(matches) if match_all else any(matches)
+
+
 class VariableRole(object):
     """Base class for all variable roles."""
     def __eq__(self, other):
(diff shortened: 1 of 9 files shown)


blocks/roles.py
```diff
@@ -39,6 +39,24 @@ def add_role(var, role):
...
 
 
+def has_roles(var, roles, match_all=False):
+    r"""Test if a variable has given roles taking subroles into account
+
+    Parameters
+    ----------
+    var : :class:`~tensor.TensorVariable`
+        Variable being queried.
+    role : an iterable of :class:`.VariableRole` instances
```
rizar: `roles`?

rizar:
It look very nice! All I found is a typo + scrutinizer is unhappy with a line in your docs: https://scrutinizer-ci.com/g/bartvm/blocks/inspections/e2618438-fb86-4991-8443-93c11467abbc

arasmus:
New version pushed. Commit was amended and force updated. What do you say now?

rizar:
Despite the Travis failure I think we can merge this, it seems completely unrelated to the content of the PR.

rizar:
Thanks for fixing this!

That was quite a major interface change about which we should notify our users. Would you like to write an email to the mail list?

PR #377, opened 2015-02-27
Development mode installation

rizar:


bartvm:
It's already here: http://blocks.readthedocs.org/en/latest/setup.html#development However, it let's `pip` do the cloning for you instead. You could add your version there as an alternative, and put a link to the installation page from the development page.

rizar:
Seems like you have already done the required changes. I am closing the PR.

Issue #244, opened 2015-02-11
NaN for uninitialized parameters?


dwf:
A common trick when writing numerical code where you expect some array to be initialized is to initialize all the values to be NaN. If you've got a bug, then, you have a very easy canary in the coal mine. What about this as a way to quickly detect uninitialized parts of a model?

rizar:
Sounds good. Might be confused with "true" nans that emerge during training. But current zeros are even harder to detect.

We could as well use inf, any arguments against or in favour?

dwf:
Well, usually it will be obvious that no training has occurred. We could
also add checks for all(isnan(param)) in some error handling pathway,
conceivably.

rizar:
As soon as we have a NaN in the model, the cost becomes one, the gradients become NaN, everything becomes NaN. 

But anyway I like your idea.

bartvm:
Closed via https://github.com/bartvm/blocks/pull/379

PR #378, opened 2015-02-27
NaN value for uninitialized parameters

Thrandis:
Hey,
Here is my CCW pull request. I added the function shared_floatx_nans in utils.py to solve the issue: https://github.com/bartvm/blocks/issues/244
Now the uninitialized parameters have a nan value instead of a 0 value.
César

### DIFF
--- a/blocks/bricks/conv.py
+++ b/blocks/bricks/conv.py
@@ -4,7 +4,7 @@ from theano.tensor.signal.downsample import max_pool_2d, DownsampleFactorMax
 from blocks.bricks import Initializable, Feedforward, Sequence
 from blocks.bricks.base import application, Brick, lazy
 from blocks.roles import add_role, FILTERS, BIASES
-from blocks.utils import shared_floatx_zeros
+from blocks.utils import shared_floatx_nans
 
 
 class Convolutional(Initializable):
@@ -52,13 +52,13 @@ class Convolutional(Initializable):
         self.border_mode = border_mode
 
     def _allocate(self):
-        W = shared_floatx_zeros((self.num_filters, self.num_channels) +
+        W = shared_floatx_nans((self.num_filters, self.num_channels) +
                                 self.filter_size, name='W')
         add_role(W, FILTERS)
         self.params.append(W)
         self.add_auxiliary_variable(W.norm(2), name='W_norm')
         if self.use_bias:
-            b = shared_floatx_zeros(self.get_dim('output'), name='b')
+            b = shared_floatx_nans(self.get_dim('output'), name='b')
             add_role(b, BIASES)
             self.params.append(b)
             self.add_auxiliary_variable(b.norm(2), name='b_norm')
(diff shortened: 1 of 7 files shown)


blocks/utils.py
```diff
@@ -71,9 +71,43 @@ def unpack(arg, singleton=False):
 
 
 def shared_floatx_zeros(shape, **kwargs):
+    """Creates a shared variable array filled with zeros.
```
bartvm: If the docstring contains backslashes, you'll need to make the string raw. This just means it needs to start with `r"""` instead of `"""`

blocks/utils.py
```diff
@@ -71,9 +71,43 @@ def unpack(arg, singleton=False):
...
+    -------
+    :class:'tensor.TensorSharedVariable'
+        A Theano shared variable filled with zeros.
+
+    """
     return shared_floatx(numpy.zeros(shape), **kwargs)
 
 
+def shared_floatx_nans(shape, **kwargs):
+    """Creates a shared variable array filled with nans.
```
bartvm: Same here, `r"""`

blocks/utils.py
```diff
@@ -71,9 +71,43 @@ def unpack(arg, singleton=False):
 
 
 def shared_floatx_zeros(shape, **kwargs):
+    """Creates a shared variable array filled with zeros.
+
+    Parameters
+    ----------
+    shape : tuple
+        A tuple of integers representing the shape of the array.
+    \*\*kwargs :
```
bartvm: No need for a colon

blocks/utils.py
```diff
@@ -71,9 +71,43 @@ def unpack(arg, singleton=False):
...
 
 
+def shared_floatx_nans(shape, **kwargs):
+    """Creates a shared variable array filled with nans.
+
+    Parameters
+    ----------
+    shape : tuple
+         A tuple of integers representing the shape of the array.
+    \*\*kwargs :
```
bartvm: No colon here either.

blocks/bricks/conv.py
```diff
@@ -52,13 +52,13 @@ def __init__(self, filter_size, num_filters, num_channels, batch_size=None,
         self.border_mode = border_mode
 
     def _allocate(self):
-        W = shared_floatx_zeros((self.num_filters, self.num_channels) +
+        W = shared_floatx_nans((self.num_filters, self.num_channels) +
                                 self.filter_size, name='W')
```
bartvm: You'll need to remove one space in this line, and in many of the others, so that the indention matches again.

bartvm:
@rizar This PR caused this test to fail, and I'm not sure why:

```
======================================================================
FAIL: tests.bricks.test_attention.FunctionTestCase (test_sequence_content_attention)
  File "/home/travis/build/bartvm/blocks/tests/bricks/test_attention.py", line 52, in test_sequence_content_attention
    assert numpy.all(weight_values >= 0)
AssertionError
```

rizar:
Apparently because I do not initialize the tested bricks here. Some initialization schemes should be use and `initialize` should be called. Sorry for that!

PR #379, opened 2015-02-27
Ccw

Thrandis:
Rebase of #378


bartvm:
@Thrandis Everything looks good! The reason the tests are failing is because of the bug I asked @rizar about. We'll just have to wait for that to be fixed, but that's okay, there's no real rush to merge this.

PR #381, opened 2015-02-27
Should fix filtering after unpickling in a new environment

### LINKED ISSUE #340: Filtering by Application does not Work after Unpickling
Need to think more what happens here, but looks like it is yet another issue with singletons and pickling: the `Application` objects are singletons.

rizar:
Should fix #340, but this is impossible to test with nose! I will check manually if it works.

### DIFF
--- a/blocks/bricks/base.py
+++ b/blocks/bricks/base.py
@@ -90,6 +90,8 @@ class Application(object):
     call_stack : :obj:`list` of :class:`Brick`
         The call stack of brick application methods. Used to check whether
         the current call was made by a parent brick.
+    brick : type
+        The brick class to which this instance belongs.
 
     Raises
     ------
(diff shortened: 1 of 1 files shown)


rizar:
It works! From a log of working with an unpickled model:

```
In [11]: VariableFilter(bricks=[mlp])(ComputationGraph(cost))
Out[11]: [mlp_apply_args_0, mlp_apply_output]

In [12]: VariableFilter(application=mlp.apply)(ComputationGraph(cost))
Out[12]: [mlp_apply_args_0, mlp_apply_output]
```

If do not have any idea how to test and will just merge when tests pass to clear the way for @dmitriy-serdyuk

rizar:
I will merge this right away. @bartvm , you might want to take a look later as this is a core change, supposedly a harmless one.

bartvm:
Ah, already did but forgot to comment I guess, LGTM.

PR #387, opened 2015-03-01
WIP: Finishing Beam Search

rizar:
This is a substantial cleanup of #279 with a few bugs fixed. Now it seems to work properly. #279 will be closed.

@dmitriy-serdyuk, we now switch roles, I am the coder, you are the reviewer :)


rizar:
This PR seems ready! I leave it open for possible feedback for today and will merge it tomorrow.

blocks/search.py
```diff
@@ -0,0 +1,304 @@
...
+
+        # Fetching names from the sequence generator
+        self.context_names = self.generator.generate.contexts
+        self.state_names = self.generator.generate.states
+
+        # Parsing the inner computation graph of sampling scan
+        self.contexts = [
+            VariableFilter(bricks=[self.generator], name='^' + name + '$',
+                           roles=[INPUT])(self.inner_cg)[0]
+            for name in self.context_names]
```
dmitriy-serdyuk: Not all the contexts may be used, so `[0]` may crash.

examples/reverse_words/__init__.py
```diff
@@ -220,24 +252,60 @@ def main(mode, save_path, num_batches, data_path=None):
...
+    elif mode == "sample" or mode == "beam_search":
+        chars = tensor.lmatrix("input")
+        generated = reverser.generate(chars)
         model = Model(generated)
+        logger.info("Loading the model..")
         model.set_param_values(load_parameter_values(save_path))
-        sample_function = model.get_theano_function()
-        logging.info("Sampling function is compiled")
+
+        def generate(input_):
```
dmitriy-serdyuk: Is it better to define it in the global scope? It is non-pickable and users who are going to study this example may have problems.

examples/reverse_words/__init__.py
```diff
@@ -220,24 +252,60 @@ def main(mode, save_path, num_batches, data_path=None):
...
+                The negative log-likelihood of generating the respective
+                sequences.
+
+            """
+            if mode == "beam_search":
+                samples, = VariableFilter(
+                    bricks=[reverser.generator], name="outputs")(
+                        ComputationGraph(generated[1]))
+                beam_search = BeamSearch(input_.shape[1], samples)
+                outputs, _, costs = beam_search.search(
```
dmitriy-serdyuk: `BeamSearch` is going to be constructed and compiled every time `generate` is called. It doesn't really matter in this example, but may be misleading for readers.

rizar: Right, this is not efficient, but currently the beam size is required for compilation, and the beam size is given by the user. So I think in this context it is okay. I will put a comment though, warning the user that this is slow.

examples/reverse_words/__init__.py
rizar: Here is does not matter because nothing is pickled. The problem only arises when there is a reference from the main loop to an unpicklable object.

blocks/search.py
rizar: Are are right, but I will leave it for a refactoring that should follow after #391.

dmitriy-serdyuk:
Everything else looks good for me. 

Is it okey to give a user freedom to duplicate the input? The beam search returns garbage if user provides different inputs.

rizar:
I am afraid that in general we do not have another choice: there is no always working way to duplicate inputs. 

We might add something for convenience later, e.g. give `BeamSearch` the axis number as a cue and trust it to do duplication.

rizar:
The error below is apparently not related to the PR, merging.

```
FAIL: tests.bricks.test_cost.FunctionTestCase (test_softmax_vector)

File "/home/travis/build/bartvm/blocks/tests/bricks/test_cost.py", line 34, in test_softmax_vector

assert_allclose(softmax_cost, softmax_cost_stable)

File "/home/travis/miniconda/lib/python2.7/site-packages/numpy/testing/utils.py", line 1297, in assert_allclose

verbose=verbose, header=header)

File "/home/travis/miniconda/lib/python2.7/site-packages/numpy/testing/utils.py", line 665, in assert_array_compare

raise AssertionError(msg)

AssertionError:

Not equal to tolerance rtol=1e-07, atol=0

(mismatch 100.0%)

x: array(2.6921029090881348, dtype=float32)

y: array(2.692103385925293, dtype=float32)
```

bartvm:
I had that the other day as well. We should move to a fixed seed and maybe
increase the tolerance a bit.

Issue #338, opened 2015-02-23
'Building with bricks'  tutorial typos and improvements


fvisin:
Hey,

I have a few comments on the _Building with bricks tutorial_:

1. In the _lazy initialization_ section you claim:

    > In the example above we configured the Linear brick during initialization.

    If I am not wrong, this should be 'during configuration'
2. In the sentence:

    > which is quite common: We want to take the output of one model,

    'we' should be lowercase
3. It would be nice to explicit the lifecycle of a brick object earlier in the tutorial, maybe with a very simple graph (`construction/configuration` --> `further configuration` --> `allocation` --> `initialization`). Also, I think that an introductory part briefly explaining the meaning of each life phase would be of help to comprehend the rest.

4. Finally, in the Brick documentation

    > Allocation configuration of children

    is not very clear.

bartvm:
1. There's a confusion here between *parameter initialization* and *object initialization* (which is what the `__init__` method does). "Configuration" is also ambiguous though, because you can configure a brick at any point *after* initialization (but before calling `Brick.apply` or `Brick.initialize`).
2. Depends on which style guide you choose. Chicago agrees with you, AP doesn't for example. I don't think we've gotten to the point yet where we have decided which style guide to stick with...
3. Feel free to make a PR which you think clarifies things. It's probably better to get editing suggestions from someone who's going through the tutorials for the first time, than for me to edit it.
4. Ditto.

fvisin:
Addressed point 3 in https://github.com/bartvm/blocks/pull/473

PR #388, opened 2015-03-01
Correct a few typos and reword a few sentences in the doc

adbrebs:


docs/bricks_overview.rst
```diff
@@ -66,9 +66,9 @@ Lazy initialization
 In the example above we configured the :class:`.Linear` brick during
 initialization. We specified input and output dimensions, and specified the
 way in which weight matrices should be initialized. But consider the
-following case, which is quite common: We want to take the output of one
+following case, which is quite common: we want to take the output of one
```
bartvm: Haha, @fvisin mentioned this in https://github.com/bartvm/blocks/issues/338 so I will say the same thing. I know that in French it needs to be a small letter, but not in English. It's arguably a slightly ambiguous case, because some style guides (e.g. Chicago) will argue that it should be small letter if there's only a single full sentence following the colon, but in this case the following sentence contains several clauses, so IMHO it really needs to be a capital :)

docs/bricks_overview.rst
```diff
@@ -154,8 +154,8 @@ brick:
 
 When dealing with children, the life cycle actually becomes a bit more
 complicated. (The full life cycle is documented as part of the
-:class:`.Brick` class.) Before allocating or initializing parameters, the
-parent brick will call its :meth:`.Brick.push_allocation_config` and
+:class:`.Brick` class). Before allocating or initializing parameters, the
```
bartvm: Periods go inside the parentheses when it encloses a full sentence, as is the case here.

docs/tutorial.rst
```diff
@@ -186,7 +184,8 @@ of size 256.
 >>> data_stream = DataStream(mnist, iteration_scheme=SequentialScheme(
 ...     num_examples=mnist.num_examples, batch_size=256))
 
-As our algorithm we will use straightforward SGD with a fixed learning rate.
+Regarding the training algorithm, we will use straightforward SGD with a fixed
```
bartvm: Not sure I find this an improvement. Simpler is better, so I would go with "The training algorithm we will use is straightforward SGD with a fixed learning rate."

docs/bricks_overview.rst
adbrebs: Interesting, I wasn't aware of that!

docs/bricks_overview.rst
adbrebs: Right, I missed the period just before the first parenthesis.

PR #389, opened 2015-03-01
Fix ConvolutionalSequence 

harmdevries89:
The push_allocation_config didn't work if there was a separate MaxPooling in layers of ConvolutionalSequence. The problem is that MaxPooling doesn't have an image_size, but input_dim instead. For now, I just check if a layer is an instance of MaxPooling, but we might also consider renaming input_dim to image_size for cleaner code.

### DIFF
--- a/blocks/bricks/conv.py
+++ b/blocks/bricks/conv.py
@@ -336,7 +336,7 @@ class ConvolutionalSequence(Sequence, Initializable, Feedforward):
         image_size = self.image_size
         for layer in self.layers:
             if isinstance(layer, MaxPooling):
-                layer.input_dim = image_size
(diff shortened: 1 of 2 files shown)


bartvm:
What's the use case for adding a `MaxPooling` layer to `ConvolutionalSequence`? The only thing I can think of is when you want a convolution without an activation function, followed by pooling. Is that a common use case? If not, I would argue the user could just pass a `ConvolutionalLayer` with `Identity` as the activation function.

harmdevries89:
I was passing a separate ConvolutionalActivation and MaxPooling layer to ConvolutionalSequence and thought that it should be able to handle that. I forgot that you can simply pass one ConvolutionalLayer in that case. I'll close the PR.

PR #393, opened 2015-03-02
Added support to remote bokeh-server

EderSantana:
If you started your bokeh-server using an --ip flag that is not the default, blocks won't feed its update and will crash. This PR adds support to any bokeh-server url. The default value keeps live plotting exactly as it was before. Thus, there is nothing to lose in here.

### DIFF
--- a/blocks/extensions/plot.py
+++ b/blocks/extensions/plot.py
@@ -59,9 +59,9 @@ class Plot(SimpleExtension):
         start the server manually using the ``bokeh-server`` command. Also
         see the warning above.
     server_url : str, optional
-        Url that the bokeh-server is listening to. Ex: when starting the
-        bokeh-server with ``bokeh-server --ip your.i.p.here``, server_url
-        should be ``http://your.i.p.here:5006``.
+        Url of the bokeh-server is listening to. Ex: when starting the
+        bokeh-server with ``bokeh-server --ip 0.0.0.0`` at ``alice``, server_url
+        should be ``http://alice:5006``.
 
     """
     # Tableau 10 colors
--- a/blocks/extensions/plot.py
+++ b/blocks/extensions/plot.py
@@ -59,7 +59,7 @@ class Plot(SimpleExtension):
         start the server manually using the ``bokeh-server`` command. Also
         see the warning above.
     server_url : str, optional
-        Url of the bokeh-server is listening to. Ex: when starting the
+        Url of the bokeh-server. Ex: when starting the
         bokeh-server with ``bokeh-server --ip 0.0.0.0`` at ``alice``, server_url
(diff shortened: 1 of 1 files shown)


EderSantana:
Also, now you can check your live plots from a different computer. You just need to pass your ip to the bokeh-server and set your server_url accordingly

blocks/extensions/plot.py
```diff
@@ -58,6 +58,10 @@ class Plot(SimpleExtension):
         it down you will lose your plots. If you want to store your plots,
         start the server manually using the ``bokeh-server`` command. Also
         see the warning above.
+    server_url : str, optional
+        Url that the bokeh-server is listening to. Ex: when starting the
+        bokeh-server with ``bokeh-server --ip your.i.p.here``, server_url
+        should be ``http://your.i.p.here:5006``.
```
bartvm: Are you sure this is correct? I thought that `--ip` specified the IP being listened on. So that means that if I start `bokeh-server --ip 0.0.0.0` on `alice`, I will have to pass `server_url=alice` on `bob` (and not `0.0.0.0`, which simply means that `alice` will respond to requests originating from any IP address, unlike the default `127.0.0.1`, which listens only to `localhost`).

EderSantana:
Sorry, "your.i.p.here" were not the best words, change it for XXX.XXX's or something . This is how it should be:
```alice: bokeh-server --ip XXX.XXX.XXX.XXX```
```bob: server_url= http://XXX.XXX.XXX.XXX:5006```
It doesn't work if you forget `http://`, I tried that.

bartvm:
And it's not this? That's why I thought it was:

`alice: bokeh-server --ip 0.0.0.0` (or `bob`'s IP address)
`bob: server_url='http://alice:5006'`

jbornschein:
Do you think this would be something where we would like the default to be configurable in ~/.blocksrc?

EderSantana:
I got what you mean @bartvm . You are right. I just tested that. Started ```alice: bokeh-server --ip 0.0.0.0``` and ```bob: server_url=http://alice:5006```. I thought the --ip flag needed to be ```alice``` herself.

@jbornschein It may be useful to set your server_url on ~/.blocksrc and use the ```default``` if it doesn't exist.

EderSantana:
Also, what is the right thing to do now? Make another PR with a better description for server_url? Or is there a way to update this one?

bartvm:
I see you figured out how to update your PR already.

Please have a look at the Scrutinizer errors, they're related to the formatting guidelines we follow i.e. Python lines need to be < 80 characters and docstrings <= 75 characters.

EderSantana:
I had to google for "trailing whitespace"...

PR #397, opened 2015-03-03
Use the right glimpses for computing readout

rizar:
Seems like I have found yet another bug in the sequence generator. I will have to train a new model to test if am right, meanwhile, if you guys are interested, you can try to check it by reading. 

 @dmitriy-serdyuk , @bartvm , @pbrakel

### DIFF
--- a/blocks/bricks/sequence_generators.py
+++ b/blocks/bricks/sequence_generators.py
@@ -106,7 +106,10 @@ class BaseSequenceGenerator(Initializable):
         self.children = [self.readout, self.fork, self.transition]
 
     def _push_allocation_config(self):
-        # Configure readout
+        # Configure readout. That involves `get_dim` requests
+        # to the transition. To make sure that it answers
+        # correctly we should finish its configuration first.
+        self.transition.push_allocation_config()
         state_dims = self.transition.get_dims(self.state_names)
         context_dims = self.transition.get_dims(self.context_names)
         self.glimpse_dims = self.transition.get_dims(self.glimpse_names)
@@ -114,7 +117,9 @@ class BaseSequenceGenerator(Initializable):
             state_dims, context_dims, self.glimpse_dims,
             feedback=self.readout.get_dim('feedback'))
 
-        # Configure fork
+        # Configure fork. For similar reasons as outlined above,
+        # first push `readout` configuration.
+        self.readout.push_allocation_config()
         feedback_names = self.readout.feedback.outputs
         if not len(feedback_names) == 1:
             raise ValueError
@@ -140,7 +145,9 @@ class BaseSequenceGenerator(Initializable):
         The contexts are expected as keyword arguments.
 
         """
-        batch_size = outputs.shape[-2]  # TODO Assumes only 1 features dim
+        # Assume the the last dimension stands for the feature number and
+        # the second last for the number in the batch.
+        batch_size = outputs.shape[-2]
 
         # Prepare input for the iterative part
         states = dict_subset(kwargs, self.state_names, must_have=False)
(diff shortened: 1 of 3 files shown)


blocks/bricks/sequence_generators.py
```diff
@@ -162,7 +162,7 @@ def cost(self, application_call, outputs, mask=None, **kwargs):
 
         # Separate the deliverables
         states = {name: results[name][:-1] for name in self.state_names}
-        glimpses = {name: results[name] for name in self.glimpse_names}
+        glimpses = {name: results[name][1:] for name in self.glimpse_names}
```
dmitriy-serdyuk: As for me, it needs more explanation. Why are you not using the last state and the first glimpse?

rizar: I use `return_initial_state=True` here when calling `transition.apply`. That means that I get my initial states and initial glimpses, which are typically all zero. But the glimpses are computed _before_ the word generation. That is 
I need the state number zero and the glimpses number one to compute the probability of the output number one.

dmitriy-serdyuk: Okey, I see. Can you add it to the comment on the code?

bartvm: +1 for more comments in this file; it's not the easiest code to read because of how tightly integrated all the bricks here are

rizar: Okay, np, I will explain that in the code.

blocks/bricks/sequence_generators.py
```diff
@@ -436,6 +436,9 @@ def emit(self, readouts):
 
     @application
     def cost(self, readouts, outputs):
+        # WARNING: unfortunately this application method works
+        # just fine when `readouts` and `outputs` have
+        # different dimensions. Be careful!
         probs = self.probs(readouts)
```
pbrakel: I guess it would be useful to have error checking nodes in theano or blocks for cases like this where you want to raise an error during execution rather than compilation time without writing something cryptic that uses a computation that breaks on purpose when the shapes aren't the same. Does something like this already exist?

bartvm: There's the `Assert` op, but reading the documentation I'm not sure how user-friendly it is: https://github.com/Theano/Theano/blob/master/theano/tensor/opt.py#L1486-L1495

If you would want to integrate this into Blocks, I guess the way to go would be: Use the `assert_` function in a Brick, add a useful message and tag the output variable with an `ASSERTION` role. We could then have an extension that basically compiles a function that returns all the `ASSERTION` variables, and runs a single batch of a data stream before training starts. If one of the assertions fails, it would give a useful error before training starts, otherwise it could just claim success and start training.

pbrakel: I didn't know about the `Assert` op and it kind of looks like what I had in mind. I don't think I'll have much time to look into this in detail now but I'll think about it. I was mainly curious to know whether I was missing out on some well-known feature. Thanks for the info!

rizar: Asserts are really nice, but they break optimization. One of Theano guys told me, that they have an idea to delete them from the graph in `fast_run`, something like that. 

I like your idea @bartvm! Can you create an issue so that we do not forget about it?

rizar:
I wrote more comments, changed the test to cover this case.

In addition I stumbled upon a funny bug: `SequenceGenerator` asked `AttentionRecurrent` for dimension of glimpses, `AttentionRecurrent` redirected this query to `SequenceContentAttention`, but the latter needed more information from `AttentionRecurrent` to answer this query! Solved by an early configuration push.

PR #402, opened 2015-03-03
Added dummy cost to TrivialEmitter

dmitriy-serdyuk:
`TrivialEmitter` shouldn't be abstract.


blocks/bricks/sequence_generators.py
```diff
@@ -405,6 +405,10 @@ def emit(self, readouts):
         return readouts
 
     @application
+    def cost(self, readouts, outputs):
+        return tensor.zeros_like(outputs)
```
rizar: Can you please say in the documentation of this class that by default the `cost` method returns zero tensor?

Issue #401, opened 2015-03-03
N-D Adapter for 2-D only bricks.


rizar:
A brick that would wrap another brick by flattening inputs before application of the wrapped and converting the output back to the original shape. If we had this, I could significantly refactor the `SoftmaxEmitter` from `blocks/sequence_generators.py`.

rizar:
@vdumoulin , at some point you wanted to implement such a thing, as far as I remember.

bartvm:
This is basically the same as #101 I believe, so I'll close that one.

rizar:
Right, forgot about that one, sorry!

vdumoulin:
Right, I can do that in the next couple days if you'd like.

rizar:
Okay, when you have time.

vdumoulin:
@rizar see #404

vdumoulin:
If you also need to wrap an application call with `dimshuffle` calls, I can also implement a `WithAxesSwapped` brick in the same kind of fashion.

rizar:
Closed by #404

dwf:
Strangely it didn't actually close.

PR #404, opened 2015-03-03
Wrapper bricks

vdumoulin:
Solves #401.

### DIFF
--- a/blocks/bricks/wrapper.py
+++ b/blocks/bricks/wrapper.py
@@ -1,0 +1,36 @@
+from theano import tensor
+
+from blocks.bricks.base import Brick, application
+
+
+class As2D(Brick):
+    """Wraps a 2D application around reshapes to accomodate N-D tensors.
+
+    The input is first flattened across all but its last axis, then the
+    application method is called on the flattened input, and finally
+    the output is reshaped so that it has the same number of dimensions
+    as the input.
+
+    Parameters
+    ----------
+    application_method : callable
+        Some brick's :meth:`apply` method
+
+    """
+    def __init__(self, application_method, **kwargs):
+        super(As2D, self).__init__(**kwargs)
+        self.application_method = application_method
+        self.children = [self.application_method.brick]
+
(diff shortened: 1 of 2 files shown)


blocks/bricks/wrapper.py
```diff
@@ -0,0 +1,36 @@
...
+    application_method : callable
+        Some brick's :meth:`apply` method
+
+    """
+    def __init__(self, application_method, **kwargs):
+        super(As2D, self).__init__(**kwargs)
+        self.application_method = application_method
+        self.children = [self.application_method.brick]
+
+    @application(inputs=['input_'], outputs=['output'])
```
rizar: When a brick is one input and one output they are typically never used. So I would add it later if it turns out necessary.

rizar:
Everything looks nice! `WithAxesSwapped` would be also nice to have.

blocks/bricks/wrapper.py
bartvm: Really? I would have added them myself as well, if for nothing else, then for consistency.

rizar: Hmm, I have just seen that someone has already added lot's of trivial input and output names for many bricks... could it be you? :)

rizar:
Would anybody mind moving `bricks/wrapper.py` to `bricks/wrappers.py` or `bricks/wrap.py`?

blocks/bricks/wrapper.py
bartvm: I guess? Not in any conscious effort though, I probably just figured it was more consistent.

bartvm:
No objections, although I'd prefer `wrappers.py`

vdumoulin:
@rizar @bartvm I'll implement `WithAxesSwapped` and rename the module to `wrappers` in the same pull request.

vdumoulin:
Just found out the renaming has already been done, so I'll just implement `WithAxesSwapped`.

rizar:
I guess you should open a new one now.

PR #405, opened 2015-03-03
callback added

yingzha:


blocks/main_loop.py
```diff
@@ -231,6 +231,12 @@ def _run_iteration(self):
         return True
 
     def _run_extensions(self, method_name, *args):
+        class Callback(str):
+            def __eq__(self, other):
+                if other not in ['before_batch', 'after_batch']:
```
bartvm: This list needs to be extended to include all possible callbacks.

blocks/main_loop.py
```diff
@@ -231,6 +231,12 @@ def _run_iteration(self):
...
+                    raise ValueError("I'm being compared to something \
+                                      I don't know!")
```
bartvm: When `other` is a known callback, you need to check here whether it is the same one using e.g. `str(self) == other`

blocks/main_loop.py
```diff
@@ -231,6 +231,12 @@ def _run_iteration(self):
...
+
         for extension in self.extensions:
             extension.dispatch(method_name, *args)
```
bartvm: `method_name` here is a string, which should now be a custom `Callback` instance e.g. use `Callback(method_name)`

bartvm:
Great, I see you made a start! Gave a few pointers about what needs to happen next.

blocks/main_loop.py
yingzha: Thanks for your comment! I understood the rest part except this one. If other is a known callback. It should look like this: Callback(method_name),  right? so now I have something like Callback(Callback(method_name)), after having line 235/236, I need another "if". However, what's the meaning of str(self)? why not use self directly?

rizar:
I thank that the right place for this class is in `blocks/extensions/__init__.py`, where the callback names are declared. I would also call it `CallbackName`. It could also collect the names of callbacks automatically. For that we need to decorate them, e.g.

```
def callback(func):
    func._is_callback = True
    return func

class TrainingExtension(object):
    ...
    @callback
    def after_batch(self, batch):
        pass

class CallbackName(object):
    # Traverse TrainingExtension methods and choose names of those 
    #  having `_is_callback` attribute
```

blocks/main_loop.py
bartvm: No, the comparison is made between a Callback instance and a string. So other is just a string, self == other would work too I guess.

bartvm:
@yingzha Just to clarify the behaviour we're looking for:
```python
class Callback(str):
    def __eq__(self, other):
        if other not in ['a', 'b']:
            raise ValueError("I can only be compared to 'a' and 'b'")
        return self == other
```
```python
>>> Callback('a') == 'a'
True
>>> Callback('a') == 'b'
False
>>> Callback('a') == 'c'
ValueError: I can only be compared to 'a' and 'b'
```
As @rizar suggested, it might be a good idea to collect the callback names automatically using a decorator. You can do that using something like this:
```python
def callback(f):
    f._is_callback = True
    return f

class Foo(object):
    @callback
    def bar(self):
        pass

    def baz(self):
        pass
```
```python
>>> foo = Foo()
>>> methods = [method for method in dir(foo) if callable(getattr(object, method))]
>>> callback = [callback for callback in methods if getattr(callback, '_is_callback', False)]
```

yingzha:
@bartvm,  `self == other` occurs maximum recursion depth problem, as least for me. I don't know why so I just use `str(self) == other`. To make sure, I don't need to build a CallbackName class, the statements you gave are enough to extract all the possible callback names.

bartvm:
Ah, that makes sense, now I remember why I used `str(self)` in the first place :) `self == other` calls `self.__eq__(other)`, which is the function we are already in, so that would result in a recursion depth. Calling `str(self)` converts `self` from `Callback` instance to a `str` instance, and simply tells you whether the strings are equal.

You shouldn't need another class.

bartvm:
@yingzha The errors you're getting in Travis are because of a bug that was in the master branch, and have nothing to do with your code. Just rebase your current branch on the current master to get rid of them.

PS I noticed you're working in your master branch, this is a bad idea! Especially now that you need to rebase. This is what you could do to fix it:
```bash
git checkout master  # This is to go to the branch you are working on now
git checkout -b ccw_callback  # This copies the branch to a new one called ccw_callback
# Add the central (main) repo, so that you can get the latest updates, if you don't have it
# already (you can check with `git remote -v`
git remote add upstream git@github.com:bartvm/blocks.git  
git fetch upstream  # Get the latest data from the main repo
# Normally you could use `git merge upstream/master` to update your master to the
# main master, but you made changes to your master branch so we need to reset instead
git reset --hard upstream/master
git checkout ccw_callback # This is the copy we made of the  branch you made the changes in
git rebase -i master  # We now re-apply the changes you made on a *new* version of the master
# Rebasing changes the history of your branch, so we need force push using `-f`
git push -f
```
That should be it!

blocks/main_loop.py
```diff
@@ -231,8 +231,19 @@ def _run_iteration(self):
         return True
 
     def _run_extensions(self, method_name, *args):
+        class Callback(str):
+            def __eq__(self, other):
+                if other not in ['before_batch', 'after_batch',
```
bartvm: Now that you added the `callback` decorator, you don't need to list the callbacks here anymore. You can automatically retrieve them! You could use e.g.
```python
callbacks = [key for key, value in TrainingExtension.__dict__.items()
             if getattr(value, '_is_callback', False)]
if other not in callbacks:
    ...
```

blocks/main_loop.py
```diff
@@ -231,8 +231,19 @@ def _run_iteration(self):
...
+                                 'on_resumption', 'on_error',
+                                 'before_training', 'before_epoch',
+                                 'after_epoch', 'after_training',
+                                 'on_interrupt']:
+                    raise ValueError("I'm being compared to \
```
bartvm: Don't forget to update this message to something more informative e.g. `"`{}` is not a valid callback name`. It should probably also be a `TypeError` instead of a `ValueError`.

yingzha: Thanks for your detailed comments. I made some changes to the code and push it again. As you see, it has the same mistake as before. Did I do sth wrong with it? Sorry for keeping asking so many questions. :)

yingzha: I rebased it by making another branch --> reset master --> rebase --> push -f, did I miss sth?

bartvm:
It seems like you didn't reset your master branch, because your `master` is still 9 commits ahead of my `master`. I see you create a branch `callback` though, so that's good. You should now reset your master branch (`git checkout master` and `git reset --hard upstream/master`), close this PR, and make a new PR from the `callback` branch instead.

Try changing line 69 (`getattr(self, callback_name)(*args)`) to `getattr(self, str(callback_name))(*args)` to see if that helps with Python 3.

blocks/main_loop.py
```diff
@@ -233,8 +234,20 @@ def _run_iteration(self):
...
 
     def _run_extensions(self, method_name, *args):
+        class Callback(str):
+            def __eq__(self, other):
+                callbackname = [key for key, value
+                                in TrainingExtension.__dict__.items()
+                                if getattr(value, '_is_callback', False)]
+                if other not in callbackname:
+                    if isinstance(other, str):
+                        raise ValueError(str(other) +
```
bartvm: It's better to use Python's new string formatting, so `"{} is not a valid callback".format(other)`. The casting to a string will happen automatically.

blocks/main_loop.py
```diff
@@ -233,8 +234,20 @@ def _run_iteration(self):
         return True
 
     def _run_extensions(self, method_name, *args):
+        class Callback(str):
+            def __eq__(self, other):
+                callbackname = [key for key, value
+                                in TrainingExtension.__dict__.items()
+                                if getattr(value, '_is_callback', False)]
+                if other not in callbackname:
+                    if isinstance(other, str):
```
bartvm: This is bad practice, because `str` doesn't include unicode strings in Python 3, while `basestring` doesn't exist anymore in Python 3. If you really need to check for a string, you need to use `six.string_types`. In this case however, I don't think you need to check specifically for a string. If it's not a string it will just say `"<SomeObject> is not a valid callback"`, which is perfectly fine.

blocks/main_loop.py
```diff
@@ -233,8 +234,20 @@ def _run_iteration(self):
         return True
 
     def _run_extensions(self, method_name, *args):
+        class Callback(str):
+            def __eq__(self, other):
+                callbackname = [key for key, value
```
bartvm: Maybe `callback_names` is a better variable name.

blocks/main_loop.py
bartvm: Oh, I understand now where the confusion comes from: It's okay to raise a `TypeError("{} is not a valid callback.format(other)")` even if `other` is a string. This is because theoretically, the error is that the `str` other can't be cast to a `Callback` type, hence making it a `TypeError` and not a `ValueError`. I guess this kind of depends on your interpretation, but you'll see that e.g. NumPy raises `TypeError`s in these situations as well.

yingzha: got it. So now I just use TypeError for all possible errors. Like:
if other not in callback_names:
raise TypeError("{} is not a valid callback.".format(other))
return str(self) == other
And I will close this pr and make a new one.

PR #407, opened 2015-03-03
WIP: Dropout

dmitriy-serdyuk:
- [x] Make graph replacement work with dependent variables.
- [x] Implement dropout itself.
- [x] Add tags to variables.
- [x] Test dropout

### DIFF
--- a/blocks/graph.py
+++ b/blocks/graph.py
@@ -161,8 +161,37 @@ class ComputationGraph(object):
             substitutes.
 
         """
-        return ComputationGraph(theano.clone(self.outputs,
-                                             replace=replacements))
+        # Due to bug in theano we have to make one replacement in time
+        replacements = OrderedDict(replacements)
+        # Replacements with previous replacements
+        replacements_cur = replacements
+        # Links to current keys
+        replacement_keys = OrderedDict(zip(replacements.keys(),
+                                           replacements_cur.keys()))
+        vars_sorted = theano.gof.graph.io_toposort(self.inputs, self.outputs)
+        outputs_cur = self.outputs
+        # Replace step-by-step in topological order
+        for var in vars_sorted:
+            for input in var.inputs:
+                if input not in replacements:
+                    continue
+                # We also want to make changes in future replacements
+                new_vals = theano.clone(
+                    outputs_cur +
+                    list(replacements_cur.keys()) +
+                    list(replacements_cur.values()),
+                    replace={replacement_keys[input]:
+                             replacements_cur[replacement_keys[input]]})
+                # Reconstruct outputs, keys, and values
+                outputs_cur = new_vals[:len(outputs_cur)]
+                keys = new_vals[len(outputs_cur):
+                                len(outputs_cur) + len(replacements)]
+                vals = new_vals[len(outputs_cur) + len(replacements):]
+                # Update maps
+                replacements_cur = OrderedDict(zip(keys, vals))
+                replacement_keys = OrderedDict(zip(replacements.keys(),
+                                                   replacements_cur.keys()))
+        return ComputationGraph(outputs_cur)
 
     def get_theano_function(self, additional_updates=None):
         """Create Theano function from the graph contained."""
--- a/blocks/graph.py
+++ b/blocks/graph.py
@@ -11,11 +11,13 @@ from theano.scan_module.scan_op import Scan
 from toolz import unique
 
 from blocks import config
-from blocks.roles import add_role, has_roles, AUXILIARY, PARAMETER
+from blocks.roles import (add_role, has_roles, AUXILIARY, OUTPUT, PARAMETER,
+                          WEIGHTS)
 from blocks.utils import (is_graph_input, is_shared_variable, dict_union,
                           shared_like)
 
 logger = logging.getLogger(__name__)
+floatX = theano.config.floatX
 
 
 class ComputationGraph(object):
@@ -161,7 +163,7 @@ class ComputationGraph(object):
             substitutes.
 
         """
-        # Due to bug in theano we have to make one replacement in time
+        # Due to theano specifics we have to make one replacement in time
         replacements = OrderedDict(replacements)
         # Replacements with previous replacements
         replacements_cur = replacements
@@ -383,3 +385,41 @@ def apply_noise(computation_graph, variables, level, seed=None):
         replace[variable] = (variable +
                              rng.normal(variable.shape, std=level))
     return computation_graph.replace(replace)
+
+
+def apply_dropout(computation_graph, variables, drop_prob=0.5, rng=None,
+                  seed=None):
+    """Returns a graph to variables in a computational graph.
+
+    Parameters
+    ----------
+    computation_graph : instance of :class:`ComputationGraph`
+        The computation graph.
+    variables : list of :class:`~tensor.TensorVariable`
+        Variables to be dropped out.
+    drop_prob : float
+        Probability of dropping out. If you want to apply the dropout
+        with different probabilities for different layers, call it
+        several times.
+    rng : :class:`~theano.sandbox.rng_mrg.MRG_RandomStreams`
+        Random number generator.
+    seed : int
+        Random seed to be used if `rng` was not specified.
+
+    Notes
+    -----
+    For more information, see [DROPOUT]_.
+
+    .. [DROPOUT] Hinton et al. *Rate MethodImproving neural networks
+    by preventing co-adaptation of feature detectors*, arXiv:1207.0580.
+
+    """
+    if not rng and not seed:
+        seed = config.default_seed
+    if not rng:
+        rng = MRG_RandomStreams(seed)
+    replacements = {var: var * rng.binomial(var.shape, p=1 - drop_prob,
+                                            dtype=floatX) / (1 - drop_prob)
+                    for var in variables}
+
+    return computation_graph.replace(replacements)
--- a/blocks/graph.py
+++ b/blocks/graph.py
@@ -11,8 +11,7 @@ from theano.scan_module.scan_op import Scan
 from toolz import unique
 
 from blocks import config
-from blocks.roles import (add_role, has_roles, AUXILIARY, OUTPUT, PARAMETER,
-                          WEIGHTS)
+from blocks.roles import (add_role, has_roles, AUXILIARY, PARAMETER, DROPOUT)
 from blocks.utils import (is_graph_input, is_shared_variable, dict_union,
                           shared_like)
 
@@ -418,8 +417,12 @@ def apply_dropout(computation_graph, variables, drop_prob=0.5, rng=None,
         seed = config.default_seed
     if not rng:
         rng = MRG_RandomStreams(seed)
-    replacements = {var: var * rng.binomial(var.shape, p=1 - drop_prob,
-                                            dtype=floatX) / (1 - drop_prob)
-                    for var in variables}
+
+    replacements = [(var, var * rng.binomial(var.shape, p=1 - drop_prob,
+                                             dtype=floatX) / (1 - drop_prob))
+                    for var in variables]
+    for variable, replacement in replacements:
+        replacement.tag.roles = variable.tag.roles + [DROPOUT]
+        replacement.tag.replacement_of = variable
 
     return computation_graph.replace(replacements)
--- a/blocks/graph.py
+++ b/blocks/graph.py
@@ -465,7 +465,7 @@ def apply_dropout(computation_graph, variables, drop_prob, rng=None,
 
     Here we apply dropout with default setting to our computation graph
 
-    >>> cg_dropout = apply_dropout(cg, inputs)
+    >>> cg_dropout = apply_dropout(cg, inputs, 0.5)
 
     Dropped out variables have role `DROPOUT` and are tagged with
     `replacement_of` tag. Let's filter these variables and check if they
--- a/blocks/graph.py
+++ b/blocks/graph.py
@@ -420,22 +420,59 @@ def apply_dropout(computation_graph, variables, drop_prob=0.5, rng=None,
 
     Examples
     --------
-    >>> from theano import tensor
+    >>> import numpy
+    >>> from theano import tensor, function
     >>> from blocks.bricks import MLP, Identity
     >>> from blocks.filter import VariableFilter
     >>> from blocks.initialization import Constant
     >>> from blocks.roles import INPUT
-    >>> linear = MLP([Identity(), Identity()], [10, 10, 10],
+    >>> linear = MLP([Identity(), Identity()], [2, 10, 2],
     ...              weights_init=Constant(1), biases_init=Constant(2))
     >>> x = tensor.matrix('x')
     >>> y = linear.apply(x)
     >>> cg = ComputationGraph(y)
+
+    We are going to drop out all the input variables
+
     >>> inputs = VariableFilter(roles=[INPUT])(cg.variables)
+
+    Here we apply dropout with default setting to our computation graph
+
     >>> cg_dropout = apply_dropout(cg, inputs)
+
+    Dropped out variables have role `DROPOUT` and are tagged with
+    `replacement_of` tag. Let's filter these variables and check if they
+    have the links to original ones.
+
     >>> dropped_out = VariableFilter(roles=[DROPOUT])(cg_dropout.variables)
     >>> inputs_referenced = [var.tag.replacement_of for var in dropped_out]
     >>> set(inputs) == set(inputs_referenced)
     True
+
+    Compiling theano functions to forward propagate in original and dropped
+    out graphs
+
+    >>> fprop = function(cg.inputs, cg.outputs[0])
+    >>> fprop_dropout = function(cg_dropout.inputs, cg_dropout.outputs[0])
+
+    Initialize an MLP and apply these functions
+
+    >>> linear.initialize()
+    >>> fprop(numpy.ones((3, 2)))
+    array([[42., 42.],
+           [42., 42.],
+           [42., 42.]])
+    >>> fprop_dropout(numpy.ones((3, 2)))
+    array([[0., 0.],
+           [0., 0.],
+           [0., 0.]])
+
+    And after the second run answer is different
+
+    >>> fprop_dropout(numpy.ones((3, 2)))
+    array([[0., 52.],
+           [100., 0.],
+           [0., 0.]])
 
     """
     if not rng and not seed:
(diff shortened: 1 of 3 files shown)


blocks/graph.py
```diff
@@ -161,8 +161,37 @@ def replace(self, replacements):
             substitutes.
 
         """
-        return ComputationGraph(theano.clone(self.outputs,
-                                             replace=replacements))
+        # Due to bug in theano we have to make one replacement in time
```
rizar: Indeed, I had problems with `theano.clone` and it is nice that you are also aware of that. Is there a Theano issue for the bug? It would be very nice to put a reference here.

bartvm: As far I understood when discussing this with @harmdevries89, this is actually expected behavior; it's the doc string that is wrong I guess? Anyway, there is no Theano issue that I am aware of. Maybe @nouiz would be interested in moving this functionality of being able to replace multiple, dependent variables from Blocks to Theano eventually though?

blocks/graph.py
```diff
@@ -354,3 +384,51 @@ def apply_noise(computation_graph, variables, level, seed=None):
...
+    variables_set = set(variables)
+    weights = []
+    for var in computation_graph.variables:
+        if hasattr(var.tag, 'roles') and OUTPUT in var.tag.roles and \
+                set(var.owner.inputs).intersection(variables_set):
+            for input in var.owner.inputs:
+                if WEIGHTS in input.tag.roles:
+                    weights.append(input)
+
+    weights_replace = {weight: weight / prob for weight in weights}
```
bartvm: The way Pylearn2 implements this might be easier. I believe they just scale the inputs instead of the weights, but @dwf can probably confirm that.

dwf: Yes. Scale the inputs by 1/include_prob (i.e. 1 - dropout_prob) and you
achieve the same effect, but you needn't modify the state of the model at
test time.

dwf: As written, I think this would be not only more complicated but incorrect, since at test time you want to scale by the inverse of the *inclusion* probability rather than the drop probability. It helps to think about a probability other than 0.5; if you are dropping 20% of the time then each input is appearing, on average, 80% of the time, so you want to scale it to 80% of it's actual value at test time (or 125% at training time, or whatever).

But rescaling the inputs is far less messy, I think (you still want to use `1 - prob` though).

blocks/graph.py
```diff
@@ -354,3 +384,51 @@ def apply_noise(computation_graph, variables, level, seed=None):
...
+def apply_dropout(computation_graph, variables, prob=0.5, rng=None, seed=None):
+    """Returns a graph to variables in a computational graph.
+
+    Parameters
+    ----------
+    computation_graph : instance of :class:`ComputationGraph`
+        The computation graph.
+    variables : list of :class:`~tensor.TensorVariable`
+        Variables to be dropped out.
+    prob : float
```
dwf: I'd recommend naming this more descriptively, such as `drop_prob` or `dropout_prob`, as one can easily slip into thinking about inclusion probabilities, especially with the mistake I've noted below.

blocks/graph.py
```diff
@@ -161,8 +162,37 @@ def replace(self, replacements):
...
-        return ComputationGraph(theano.clone(self.outputs,
-                                             replace=replacements))
+        # Due to theano specifics we have to make one replacement in time
+        replacements = OrderedDict(replacements)
+        # Replacements with previous replacements
+        replacements_cur = replacements
+        # Links to current keys
+        replacement_keys = OrderedDict(zip(replacements.keys(),
+                                           replacements_cur.keys()))
+        vars_sorted = theano.gof.graph.io_toposort(self.inputs, self.outputs)
```
rizar: Those are not variables, but apply nodes.

blocks/graph.py
```diff
@@ -161,8 +162,37 @@ def replace(self, replacements):
...
+        outputs_cur = self.outputs
+        # Replace step-by-step in topological order
+        for var in vars_sorted:
+            for input in var.inputs:
+                if input not in replacements:
```
rizar: Please use `input_`.

blocks/graph.py
```diff
@@ -161,8 +162,37 @@ def replace(self, replacements):
...
+                    continue
+                # We also want to make changes in future replacements
+                new_vals = theano.clone(
```
rizar: Naming: `new_outputs`? Values is a bad name for Theano variables.

blocks/graph.py
```diff
@@ -161,8 +162,37 @@ def replace(self, replacements):
             substitutes.
 
         """
-        return ComputationGraph(theano.clone(self.outputs,
-                                             replace=replacements))
+        # Due to theano specifics we have to make one replacement in time
+        replacements = OrderedDict(replacements)
+        # Replacements with previous replacements
```
rizar: Please, explain more what `replacement_cur` and `replacement_keys` actually contain. Be more generous when you write comments to non-trivial code.

blocks/graph.py
```diff
@@ -161,8 +162,40 @@ def replace(self, replacements):
...
+        replacements_cur = replacements
+        # Links to current keys. Maps original replacement keys to
+        # new, replaced earlier, keys.
+        replacement_keys = OrderedDict(zip(replacements.keys(),
+                                           replacements_cur.keys()))
+        apply_nodes_sorted = theano.gof.graph.io_toposort(
+            self.inputs, self.outputs)
+        outputs_cur = self.outputs
+        # Replace step-by-step in topological order
+        for var in apply_nodes_sorted:
```
rizar: You keep using `var` for the elements of the list though.

blocks/graph.py
```diff
@@ -161,8 +162,40 @@ def replace(self, replacements):
             substitutes.
 
         """
-        return ComputationGraph(theano.clone(self.outputs,
-                                             replace=replacements))
+        # Due to theano specifics we have to make one replacement in time
+        replacements = OrderedDict(replacements)
+        # Replacements with previous replacements. We have to track
+        # variables in the new graph corresponding to original replacements.
```
rizar: I am sorry, I still do not understand what you mean. You use the word "replacement" a lot and it is hard to understand in which sense you use it in every given case. 

`Replacements with previous replacements` - this does not tell me anything at all.

`We have to track variables in the new graph corresponding to original replacements.` - what are original replacments? Do you mean values of the `replacements` argument? If yes, then why do we have to track them, what happens to them?

I am sorry for being so critical, but the core code is written once and read thousand times. You are definitely allowed to use 4-5 lines to explain what the issue is and how you handle it. You can even introduce your own terminology in the docstring.

bartvm:
@dmitriy-serdyuk @rizar The deadline for this is supposed to be 1PM this afternoon. Just wondering how much work is still needed. From the comments it looks like the main issue right now is the choice of variable names and comments? And a functional test seems to be missing. How much more time do you guys think is needed?

rizar:
@bartvm, I can not confirm that the only problem left are variable names, because I failed to understand the code of new `replace` routine. I made requests for more comments at the location where I was lost, and can not say anything about the rest.

blocks/graph.py
dmitriy-serdyuk: I wanted to explain, that when I'm doing one replacement a time, I need to do `replace` in both cost computation graph and all expressions in future replacements (in keys and values of dict `replacements`).

rizar: Thanks, this was helpful and now I understand a bit more. Will try to 
read your code again.

blocks/graph.py
```diff
@@ -161,8 +162,45 @@ def replace(self, replacements):
...
+                    list(replacements_cur.values()),
+                    replace={replace_what: replace_by})
+                # Reconstruct outputs, keys, and values
+                outputs_cur = outputs_new[:len(outputs_cur)]
+                keys = outputs_new[len(outputs_cur):
+                                   len(outputs_cur) + len(replacements)]
+                vals = outputs_new[len(outputs_cur) + len(replacements):]
+                # Update mappings
+                replacements_cur = OrderedDict(zip(keys, vals))
+                key_to_key_cur = OrderedDict(zip(keys,
```
rizar: First of all, thanks for your comments and renamings, it is much easier to read this code now.

It seems to me that `key_to_key_cur` will always have identical keys and values from these two lines.

dmitriy-serdyuk: Yes, I just spotted this problem as well.

dmitriy-serdyuk:
@rizar I found a cleaner way to do `replace`, but it is probably slower. I sort the `replacements` first and apply them in this order. Please check it.

dmitriy-serdyuk:
Do I need to add more tests?

BTW, is it going to work with inner graphs of scan operations?

blocks/graph.py
```diff
@@ -161,8 +162,39 @@ def replace(self, replacements):
...
+        # Sort `replacements` in topological order
+        replacements_sorted = []
+        for node in apply_nodes_sorted:
+            for input_ in node.inputs:
+                if input_ not in replacements:
+                    continue
+                replacements_sorted.append((input_, replacements[input_]))
+        # `replacements` with previous replacements applied. We have to track
+        # variables in the new graph corresponding to original replacements.
+        replacements_cur = replacements_sorted
```
rizar: Now it looks correct to me. This iterative logic should also be documented, it would be nice to use your test as a doctest.

Suggestions:

- remove `replacements_sorted` because it is not used anywhere except this assignment

- use an `OrderedDict` for `replacements_cur` as you did. Why did you switch to a list of tuples?

blocks/graph.py
```diff
@@ -354,3 +386,45 @@ def apply_noise(computation_graph, variables, level, seed=None):
...
     return computation_graph.replace(replace)
+
+
+def apply_dropout(computation_graph, variables, drop_prob=0.5, rng=None,
+                  seed=None):
+    """Returns a graph to variables in a computational graph.
+
+    Parameters
+    ----------
+    computation_graph : instance of :class:`ComputationGraph`
```
rizar: This is a bit out the scope of this PR, but we should decide whethers regularization inserters accepts computation graph or cost as parameters. I guess it is a good idea to pass `ComputationGraph`, because it works just as a fine in the case of multiple outputs.

blocks/graph.py
```diff
@@ -354,3 +386,45 @@ def apply_noise(computation_graph, variables, level, seed=None):
...
+    if not rng and not seed:
+        seed = config.default_seed
+    if not rng:
+        rng = MRG_RandomStreams(seed)
+
+    replacements = [(var, var * rng.binomial(var.shape, p=1 - drop_prob,
+                                             dtype=floatX) / (1 - drop_prob))
+                    for var in variables]
+    for variable, replacement in replacements:
+        replacement.tag.roles = variable.tag.roles + [DROPOUT]
```
rizar: This is role copying, which as far as I understood we decided not to do. @bartvm correct me I am wrong.

bartvm: Yeah, that's what we settled on. So `replacement.add_role(DROPOUT)` should suffice here. The rest of the roles can be found by following the `replacement_of` link.

blocks/graph.py
bartvm: I can swear we discussed this before (for `apply_noise`) and settled on `ComputationGraph`. Can't remember why, but seems sensible to me.

blocks/graph.py
dmitriy-serdyuk: I cannot index `OrderedDict` as easy as an array. Do you think it would be better to use it here?

rizar: I think you could use `for replace_what, replace_by in replacements_cur.items()` here.

rizar:
This PR already looks quite nice and the only thing I would do before merging is make the test of new replacement logic a doc test, so that it was crystal clear how we handle interdependencies between replacements.

blocks/graph.py
```diff
@@ -161,8 +162,40 @@ def replace(self, replacements):
...
+        replacements_cur = OrderedDict()
+        # Sort `replacements` in topological order
+        for node in apply_nodes_sorted:
+            for input_ in node.inputs:
+                if input_ not in replacements:
+                    continue
+                replacements_cur[input_] = replacements[input_]
+
+        # Replace step-by-step in topological order
+        for i in range(len(replacements_cur)):
```
dwf: This line and the following line could be replaced by the more idiomatic

```python
for i, replacement in enumerate(replacements_cur):
    ...
```
This should also be more efficient as it's not recreating the items list on every iteration.

dwf: Hmm... maybe not, in fact. Just a second.

rizar: My bad, missed that! By the way, I do not see this index `i` used anywhere else, why not `for replacement in replacements_cur.values()`?

dmitriy-serdyuk: I'm changing `replacements_cur` every iteration. I'd like to have access to a changed version.

rizar: What about

```
while len(replacements_cur):
    replacement = next(iter(replacements_cur.values()))
     ...
     list(replacements_cur.keys())[1:],
     list(replacements_cur.values())[1:],
```

?

dmitriy-serdyuk: @rizar cool! It should work.

dwf: Yes, I see that now. Although it seems like it'd be easier to just do 
```python
keys, values = replacements_cur.keys(), replacements_cur.values()
```
only once before the loop, and replace the current `replacement = ` line with
```python
replacement = (keys[i], values[i])
```
Then you can operate on the key and value lists directly, since you don't seem to need it to be a dictionary within the loop.

bartvm: Haha, I find the original a lot clearer! We've gone from a simple `for i in ...` to creating an iterator, getting the first element from it and then left-popping keys and values...

dwf: (missed some `list()` calls, updated my comment)

dwf: @bartvm @dmitriy-serdyuk I think the original is clearer too, although it currently needlessly creates a lot of `OrderedDicts` which it could just unpack into key and value lists once.

blocks/graph.py
```diff
@@ -354,3 +387,64 @@ def apply_noise(computation_graph, variables, level, seed=None):
...
+    Notes
+    -----
+    For more information, see [DROPOUT]_.
+
+    .. [DROPOUT] Hinton et al. *Improving neural networks by preventing
+       co-adaptation of feature detectors*, arXiv:1207.0580.
+
+    Examples
+    --------
+    >>> from theano import tensor
```
dwf: This doctest is great stuff, though it might be good to interleave a few `>>> # Comments` to help the exposition of what you're demonstrating.

dwf:
Very nice work! I suggest rebasing on top of current master, getting rid of the needless merge commits and maybe squashing some of the smaller stuff.

blocks/graph.py
bartvm: You should be able to just interrupt the code as well i.e.
```python
>>> x = 1

I just defined ``x``

>>> y = 1

Now I defined ``y``. Their sum is

>>> x + y
2

Which is the correct answer
```

blocks/graph.py
```diff
@@ -11,11 +11,12 @@
 from toolz import unique
 
 from blocks import config
-from blocks.roles import add_role, has_roles, AUXILIARY, PARAMETER
+from blocks.roles import (add_role, has_roles, AUXILIARY, PARAMETER, DROPOUT)
```
bartvm: Why the parentheses btw?

blocks/graph.py
dmitriy-serdyuk: I hope that the new version looks better. I separated keys and values and got rid of for loop.

blocks/graph.py
```diff
@@ -354,3 +414,104 @@ def apply_noise(computation_graph, variables, level, seed=None):
         replace[variable] = (variable +
                              rng.normal(variable.shape, std=level))
     return computation_graph.replace(replace)
+
+
+def apply_dropout(computation_graph, variables, drop_prob=0.5, rng=None,
```
rizar: I don't think there is such thing as the default dropout probability and would make the user always give it explicitly.

bartvm: Don't fully agree, almost everyone uses 0.5. However, there are actually cases where people use a scaling which is not directly inverse to the dropout probability, so you might want allow users to (optionally) pas that too.

blocks/graph.py
```diff
@@ -160,9 +161,68 @@ def replace(self, replacements):
...
+        >>> y = x + 2
+        >>> z = y + 3
+        >>> a = z + 5
+
+        Let's suppose we have dependent replacements like
+
+        >>> replacements = {y: x * 2, z: y * 3}
+        >>> cg = ComputationGraph([a])
+        >>> theano.pprint(a)
+        '(((x + TensorConstant{2}) + TensorConstant{3}) + TensorConstant{5})'
```
dmitriy-serdyuk: Scrutinizer complains about this line, how can I fix it?

blocks/graph.py
rizar: I swear I saw a lot of different numbers in the literature. Similarly to the precedent with the transition of `SimpleRecurrent` I would not provide a default value here.

blocks/graph.py
bartvm: Eh... Good question. Two options that come to mind:
```python
>>> textwrap.wrap(theano.pprint(a), 67)  # doctest: +ELLIPSIS
'(((x + TensorConstant{2}) + TensorConstant{3}) + TensorConstant...
```
```python
>>> textwrap.wrap(theano.pprint(a), 67)
(((x + TensorConstant{2}) + TensorConstant{3}) +
TensorConstant{5})
```

blocks/graph.py
bartvm: I'm fine with removing it, asking the user to type 0.5 isn't too much of a burden, but the original paper does say

> We have tried various dropout probabilities and almost all of them improve the generalization
performance of the network. For fully connected layers, dropout in all hidden layers works
better than dropout in only one hidden layer and more extreme probabilities tend to be worse,
which is why we have used 0.5 throughout this paper.

And from Larochelle's course:

> Could use a different dropout probability, but 0.5 usually works well

blocks/graph.py
bartvm: Actually, does this work?
```python
>>> theano.pprint(a)  # doctest: +NORMALIZE_WHITESPACE
(((x + TensorConstant{2}) + TensorConstant{3}) +
TensorConstant{5})
```

blocks/graph.py
rizar: From the journal version

>  For the input units, however, the optimal probability of retention is usually closer to 1 than to 0.5

bartvm: For the input the probability is much more data-dependent I believe, and you might not want to drop out at all. I am just saying that 0.5 is a very reasonable default that works in 90% of the cases. Pylearn2, Caffe, Lasagne, Torch all have it.

rizar: Okay, this impressive list of names makes me surrender :)

bartvm: Like I said, I'm actually fine with not giving a default :p I was just playing devil's advocate haha Let's just leave it up to @dmitriy-serdyuk

rizar:
This PR looked really nice and I could not help merging it :)

PR #410, opened 2015-03-04
Increase default learning rate Adam from 0.0002 to 0.002

bartvm:
As suggested by @dpkingma. @memimo, you committed the original version, any particular reason you chose 0.0002?


memimo:
No particular reason, just a mistake.

PR #411, opened 2015-03-04
Add Merge brick and tests

bartvm:
@rizar As discussed.

Working with the parallel bricks made me curious: Why all the dictionaries and keyword arguments? It seems really redundant to pass: `Parallel(['a', 'b'], {'a': 10, 'b': 20}, {'a': 5, 'b': 7})` instead of `Parallel(['a', 'b'], [10, 20], [5, 7])`. Is there any particular reason for doing it like this?

### DIFF
--- a/blocks/bricks/parallel.py
+++ b/blocks/bricks/parallel.py
@@ -1,6 +1,8 @@
 """Generic transformations with multiple inputs and/or outputs."""
 import copy
 
+from theano import tensor
+
 from blocks.bricks import Initializable, Linear
 from blocks.bricks.base import lazy, application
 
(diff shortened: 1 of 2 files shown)


rizar:
I did not think too much about the constructor interface, because I am not using it... Seems like you are right and indeed one can switch to lists here.

I would keep `dict` as the internal representation for dimension configuration, because it is much easier to be modified for the parent.

bartvm:
Yeah, agreed that dict is easier for the internals, just seemed redundant for the constructor. I'll make an issue about it I guess.

tests/bricks/test_bricks.py
```diff
@@ -469,3 +469,22 @@ def test_linear_nan_allocation():
     b2 = linear.params[1].get_value()
     numpy.testing.assert_equal(w1, w2)
     numpy.testing.assert_equal(b1, b2)
+
+
+def test_merge():
```
rizar: That can be just as well made a doctest. All other parallel brick now have  a doctest, so I think we should stick to this tradition.

blocks/bricks/parallel.py
```diff
@@ -261,3 +263,31 @@ def apply_inputs(self):
...
+        The dictionary of input dimensions, keys are input names, values
+        are dimensions.
+    output_dim : int
+        The output dimension of the merged variables.
+    prototype : :class:`~blocks.bricks.Feedforward`
+        A transformation prototype. A copy will be created for every
+        input.  If ``None``, a linear transformation without bias is used.
+    child_prefix : str, optional
+        A prefix for children names. By default "transform" is used.
+
```
rizar: When you not specify attributes, that is not explain how to configure your class after the construction, you effectively make it non-lazy only.

I guess we should add a property `output_dim`, and document this property and the `input_dims` dictionary as attributes.

rizar:
Ideally I think we should support both dict and list by an `isinstance` check.

tests/bricks/test_bricks.py
bartvm: Sure.

bartvm:
I don't really like unnecessary polymorphic functions though. Just makes for more confusing code as well as documentation.

blocks/bricks/parallel.py
bartvm: Yeah, forgot to copy paste the attributes part. I'll turn `output_dim` into a property that sets `output_dims`.

rizar:
Okay, let's make it list only. If one has a dict, he can just set it right after construction.

bartvm:
@rizar Made the requested changes

rizar:
LGTM!

PR #412, opened 2015-03-04
Interrupting and resuming training

harmdevries89:
As observed by @capybaralet and me, interrupting and resuming training doesn't work properly. After interrupting the current epoch, resuming training will do only one epoch. I think the problem is that ```_epoch_interrupt_received``` and ```_batch_interrupt_received``` are not reset whenever we resume training.

### DIFF
--- a/blocks/main_loop.py
+++ b/blocks/main_loop.py
@@ -160,6 +160,8 @@ class MainLoop(object):
                 # of the main loop.
                 if self.log.status.iterations_done > 0:
                     self._run_extensions('on_resumption')
+                    self.status._epoch_interrupt_received = False
+                    self.status._batch_interrupt_received = False
                 while self._run_epoch():
                     pass
             except TrainingFinish:


dwf:
Hmm, @rizar, can you think of a way to add a test for this?

(As a side note: why are these marked soft-private with a leading '_'? They don't seem to be principally accessed by the 'status' object itself, but by other client code. Scrutinizer noticed this and I think it's a good point.)

bartvm:
I know that things with leading underscores aren't printed to the log, might have something to do with that?

At first sight, fix looks good. I guess it got introduced when we changed from using `log.current_row` to `status`. I'll just merge and we can make adding a test a new issue, because this is pretty annoying bug to have in master!

rizar:
@dwf , to suppress printing them in `Printing`. Currently it just prints everything without underscore.

@bartvm , agreed to just merged it now and create an issue for covering this with a test.

rizar:
@dwf , I've just noticed your first question: testing signal handling would be a very nice thing to do, but as @bartvm said it is rather nasty. I guess the only proper way is to create a thread and send it SIGINTs and SIGTERMs. Can a Python thread receive signal or only a process?

bartvm:
You're probably better off using `multiprocessing` instead of `threading`

rizar:
Created #413.

PR #414, opened 2015-03-04
WIP: Support post-merge transformation in Readout brick

bartvm:


blocks/bricks/__init__.py
```diff
@@ -345,15 +345,18 @@ def __init__(self, input_dim, output_dim, num_pieces, **kwargs):
         self.output_dim = output_dim
         self.num_pieces = num_pieces
 
-        self.linear_transformation = Linear(
-            name=self.name + '_linear_to_maxout', input_dim=input_dim,
-            output_dim=output_dim * num_pieces, weights_init=self.weights_init,
-            biases_init=self.biases_init, use_bias=self.use_bias)
-        self.maxout_transformation = Maxout(name=self.name + '_maxout',
-                                            num_pieces=num_pieces)
+        self.linear_transformation = Linear(name=self.name +
```
rizar: Doesn't this kind of naming break our hierarchical principles? I would remove `self.name` from here.

blocks/bricks/sequence_generators.py
```diff
@@ -290,38 +290,81 @@ def readout(self, **kwargs):
...
+    source_names : list
+        A list of the source names (outputs) that are needed for the
+        readout part e.g. ``['states']`` or ``['states', 'glimpses']``.
     readout_dim : int
         The dimension of the readout.
     emitter : an instance of :class:`AbstractEmitter`
         The emitter component.
     feedback_brick : an instance of :class:`AbstractFeedback`
         The feedback component.
+    merger : :class:`.Brick`, optional
```
rizar: It's a bit inconsistent to have `merger` and `post_merge`. They should by identical except for the prefix, IMHO.

blocks/bricks/sequence_generators.py
```diff
@@ -290,38 +290,81 @@ def readout(self, **kwargs):
...
+        cannot be given.
+    merge_prototype : :class:`.FeedForward`, optional
+        If `merger` isn't given, the transformation given by
+        `merge_prototype` is applied to each input before being summed. By
+        default a :class:`.Linear` transformation without biases is used.
+        If given, `merger` cannot be given.
+    post_merge : :class:`.Feedforward`, optional
+        If given, this transformation is applied to the merged inputs to
+        the readout model.
+    post_merge_input_dim : int, optional
```
rizar: `merged_dim`?

blocks/bricks/sequence_generators.py
```diff
@@ -290,38 +290,81 @@ def readout(self, **kwargs):
...
+        if self.post_merge:
+            self.post_merge.output_dim = self.readout_dim
+            self.post_merge.input_dim = self.post_merge_input_dim
+            self.merger.output_dim = self.post_merge_input_dim
+        else:
+            self.merger.output_dim = self.readout_dim
+
+    @application
+    def readout(self, **kwargs):
+        if set(kwargs.keys()) != set(self.merger.input_names):
```
rizar: This is a nice thing to do in general, but unfortunately it will receive all possible sources as `kwargs`. In early days of Blocks it was not clear to me who should filter necessary input, the calling brick of the callee. Now I think that the calling brick.

Could you please use `dict_dubset` where `readout` is used, that is in `cost` and `generate` methods?

bartvm: Okay, I noticed that some functions seemed to expect more arguments than they used, but when I implemented this check I got no errors in my specific case. But sure, I'll add filters in calling functions; I definitely agree that it should be the calling brick that filters.

blocks/bricks/__init__.py
bartvm: Sure. Now we're commenting on this code anyway, I never know what to do here: Don't configure the child brick at all, and only do it in `_push_allocation_config`, or do it in both places. The advantage if only doing it in `_push_allocation_config` is that there is less code duplication. Plus, I think it makes it clearer to the user that the child brick isn't configured until `push_allocation_config` was called (if the user changes anything before that, it will just be overridden).

The advantage of doing it in both places is that sometimes you might want to delegate properties to child bricks, and you will want these to work before `push_allocation_config` was called. For example, `MLP.input_dim` should really return `MLP.children[0].input_dim` (it doesn't now, but it probably should). For that to work as early as possible, it would have to configure its first and last child's dimensions before `push_allocation_config` was called.

rizar: I would stick to doing all configuration in `_push_allocation_config` for the purpose of simplicity and leave the current MLP brick as is. I am just scared of all this variety of possible behaviours we will have to deal with.

But when you create your children in the constructor as here, I think it is okay to pass the constructor arguments downwards. What I would not like is to configure the created child in the constructor. That would overlap with `_push_allocation_config`

bartvm: What "variety of possible behaviours" are you referring to?

So you're saying that as long as the configuration is limited to passing things down directly, it's okay to push configuration both in `__init__` and in `_push_allocation_config`? That seems a bit arbitrary for me. I would just push configuration down as long as its used by the parent brick (as in my example).

Why wouldn't you want to change the current MLP brick? I find its current behaviour annoying e.g. you can do ```python
>>> mlp.input_dim
50
>>> mlp.push_allocation_config()
>>> mlp.children[0].input_dim = 100
>>> mlp.input_dim == mlp.children[0].input_dim
False
>>> mlp.allocate()
>>> mlp.apply(x).eval({x: np.random.rand(10, mlp.input_dim)})  # Will crash 
>>> mlp.apply(x).eval({x: np.random.rand(10, mlp.children[0].input_dim)})  # Won't
```

Comment (rizar) [inline blocks/bricks/__init__.py]: Sorry, I got confused :)

I like what you propose for MLP. We should not create new configuration attributes when we can use those of the children. In this case configuration will effectively be propagated in the constructor. In other cases I guess we should not propagate the configuration, as this is code duplication.

Comment (rizar) [inline blocks/bricks/__init__.py]: Just looked at your example of bad behaviour one more time and found it not so surprising. One can not expect Blocks to be so smart! It is natural that assignments to a child's configuration are not reflected in the parent's configuration. So maybe we should not change anything here altogether.

Comment (bartvm) [inline blocks/bricks/__init__.py]: After thinking about it a bit more I'm starting to think the same. We need to remember to put a warning about this somewhere in the tutorials though, because it can be very counter-intuitive if you don't fully realize this limitation.

File: blocks/bricks/sequence_generators.py [hunk 5]
```diff
@@ -290,38 +290,80 @@ def readout(self, **kwargs):
...
-@add_metaclass(ABCMeta)
-class Readout(AbstractReadout):
+class Readout(AbstractReadout, Initializable):
     """Readout brick with separated emitting and feedback parts.
 
     Parameters
     ----------
+    source_names : list
+        A list of the source names (outputs) that are needed for the
+        readout part e.g. ``['states']`` or ``['states', 'glimpses']``.
```
rizar: `glimpses` have already become `weighted_averages` in #420

rizar:
LGTM

blocks/bricks/sequence_generators.py
bartvm: Ah, but you skipped this file in that refactor, because `glimpses` occurs nearly 30 times in this file. Do you want me to change them all to `weighted_averages` before merging?

rizar: There used to be a double-usage of term glimpse. Documentation of `attention.py` introduces it a name for a deliverable of an attention mechanism. Also the only available attention mechanism had one glimpse named "glimpse". I renamed the latter to stop this confusing collision. 

The 30 times you are talking about are most likely usages in the first sense, and those are okay. But there you refer to a name of a particular glimpse, which is since #420 called 'weighted_averages`.

bartvm:
@rizar The current failing test is because the old `LinearReadout` block used biases, while the `Merge` brick used now doesn't (so the beam search result changes). Although I was thinking about maybe adding a `use_bias` option to `Merge` at one point, I think the current situation is actually better than the old one, because `LinearReadout` had a bias for each input, making all but one of them redundant.

bartvm:
Waiting for https://github.com/Theano/Theano/pull/2602 to be merged, which should allow me to simplify the `Merge` code a little bit.

rizar:
I guess we still want the `Readout` created by default to have biases, otherwise it is useless and the `reverse_words` demo will be immediately broken (you can't predict characters with a softmax without biases I believe).

I would rather make adding biases a default post-merge operation. For this I guess we would need a `Bias` brick, that would simply add a vector to its input.

bartvm:
Sure, that works, and immediately introduces the `use_bias` option that I mentioned. It makes for cleaner code than having to do `merge.children[0].use_bias = True` after constructing the `Merge` brick I guess.

rizar:
Not sure I understand about the `use_bias` option: where exactly do you want to have it?

bartvm:
For the `Merge` brick. Basically, if `use_bias` is `True` then it will apply the `Bias` brick after merging, otherwise it will behave as it currently does. There might be cases after all where you don't want to use a bias, so we should keep it optional somehow.

rizar:
Right, but I am not sure that monkey patching `Merge` for this particular case is a good idea. It's a too generic one. I would rather used a default `Bias` brick in `Readout` when `post_merge` brick is not given.

bartvm:
That works too for me, not really opinionated. It does cause a problem I've run into a few times now: The `post_merge` brick is supposed to adhere to the `Feedforward` interface, for which I introduced the `FeedforwardSequence` brick, but this becomes useless when the first or last bricks in the sequence of bricks are dimensionless. So how do you make something like `Sequence([Bias, Linear])` conform to the `Feedforward` interface cleanly? Or should we just expect the user to handle these cases (i.e. set the input_dim for `Linear` manually)?

bartvm:
Oh, and to complicate things, keep in mind that even if bricks are dimension-less, that doesn't mean they don't change the dimension e.g. `Sequence([Maxout, Linear])` needs `Linear.input_dim = merged_dim / 2`

rizar:
I had exactly the same problem with `ShallowEnergyComputer` in `attention.py`, wrote a `FeedforwardSequence` and deleted it.

Seems like providing the numbers of the first and the last configurable child to `FeedforwardSequence` would do the job in some 95% cases, and where it does not, I guess a custom brick should be written.

rizar:
To your crosspost: but we have `LinearMaxout` for that.

bartvm:
No, `LinearMaxout` does a linear transformation first, followed by `Maxout`. In this case I wanted to do it the other way around, apply `Maxout` first and then perform the linear transformation.

rizar:
haha, maybe `MaxoutLinear`? :)

Or, we could try and introduce a new abstract brick class `ComplexNonlinearity`:

```
class ComplexActivation(object):

    @staticmethod
    def get_output_dim(input_dim):
        pass
   
    @staticmethod
    def get_input_dim(output_dim):
        pass
```

The `FeedforwardSequence` could try to handle this one and `Activation` gracefully.

bartvm:
Meh, I'm inclined to just let the user configure it for now. Not everything
needs to fully automatic after all, need to leave something for them to do!

rizar:
At least it is not urgent.

bartvm:
@rizar Now with the new `Bias` brick

blocks/bricks/sequence_generators.py
rizar: So what about this? There is currently no glimpse named `glimpses`, only `weighted_averages`.

rizar:
Everything looks good! Anything else you want to do in this PR?

bartvm:
Nope, let me finally merge this after 9 days.

@pbrakel Heads up! This will break your code. Depending on what parameters you were passing to `LinearReadout` you might just be able to rename `LinearReadout` to `Readout`.

PR #418, opened 2015-03-05
Beam search output

rizar:
A few improvements for beam search:

- now it can return costs of all generated symbols. This is helpful for debugging.

- now mask is accumulated in the same way as outputs, because after `ignore_first_eol` options was introduced the old way of computing mask became invalid

- added `as_lists` flag to return list of results instead of masked numpy arrays

@dmitriy-serdyuk , will you take a look?

### DIFF
--- a/blocks/search.py
+++ b/blocks/search.py
@@ -239,7 +239,7 @@ class BeamSearch(object):
         return numpy.unravel_index(args, matrix.shape), flatten[args]
 
     def search(self, input_values, eol_symbol, max_length,
-               ignore_first_eol=False):
+               ignore_first_eol=False, as_lists=False):
         """Performs beam search.
 
         If the beam search was not compiled, it also compiles it.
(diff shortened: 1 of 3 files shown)


blocks/search.py
```diff
@@ -263,10 +263,19 @@ def search(self, input_values, eol_symbol, max_length,
...
+            their costs are returned, instead of the default output format.
 
         Returns
         -------
-        Sequences in the beam, masks, and corresponding log-probabilities.
+        outputs : :class:`numpy.ndarray`
+            The generated sequences, time is the first axis, the number in
+            the beam is the second.
+        mask : :class:`numpy.ndarray`
+            The mask for the outputs, same layout as `outputs`
```
dmitriy-serdyuk: I think it is a good idea to repeat here, that the `mask` is returned only if `as_lists=False`.

rizar: The thing is, the whole output format is changed when `as_lists=True`. I would hate describing all the differences here.

@bartvm , what's the right way to document outputs when those can be delivered into two different formats?

bartvm: I don't think there is a format for it. I can't think of any NumPy function that would have a different number or different order of return values, and my guess is that they don't allow that (which is a reasonable thing to do...)

I agree with @dmitriy-serdyuk though, I think it's worthwhile to repeat. People are likely to look at the *Returns* section without having read every single parameter description.

rizar: It is not different order though. And if asked to choose between two formats, I would choose the new one.  I could make it default and leave a flag to return the internal representation, that is meant to be used only by knowledgeable people.

dmitriy-serdyuk: @rizar I agree, let's return the second format by default and the internal representation (additionally?) if requested.

PR #423, opened 2015-03-05
Update sequence_generators.py

bartvm:
@rizar This makes things work for me. Should any other changes be made?


rizar:
There is also a wrong doc right above the line you have fixed. Otherwise everything is good. Fix it and you can merge.

PR #425, opened 2015-03-05
LinearMaxout uses configuration from children

bartvm:
@rizar @dwf 

So this approach is slightly more wordy than what it was, but it removes the problems we have of configuration of parents and children diverging i.e. the behaviour I talked about in https://github.com/bartvm/blocks/pull/414/files#r25870737

I can do a similar thing for e.g. the MLP brick.

### DIFF
--- a/blocks/bricks/__init__.py
+++ b/blocks/bricks/__init__.py
@@ -358,6 +358,8 @@ class LinearMaxout(Initializable, Feedforward):
 
     @property
     def output_dim(self):
+        if self.linear_transformation.output_dim and self.num_pieces:
+            return self.linear_transformation.output_dim // self.num_pieces
         return self._output_dim
 
     @output_dim.setter
--- a/blocks/bricks/__init__.py
+++ b/blocks/bricks/__init__.py
@@ -358,11 +358,13 @@ class LinearMaxout(Initializable, Feedforward):
 
     @property
     def output_dim(self):
-        return self.linear_transformation.output_dim // self.num_pieces
+        return self._output_dim
 
     @output_dim.setter
     def output_dim(self, value):
-        self.linear_transformation.output_dim = value * self.num_pieces
+        self._output_dim = value
+        if value and self.num_pieces:
+            self.linear_transformation.output_dim = value * self.num_pieces
 
     @property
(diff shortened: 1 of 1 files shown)


bartvm:
Actually, this doesn't work nicely at all... The problem is that two values (`output_dim` and `num_pieces`) are needed in order to configure the `output_dim` of the linear transformation. Keeping them in sync means that changes to both `output_dim` and `num_pieces` must trigger an update to `output_dim` of the linear transformation... I guess that trying to keep the parent brick and child bricks in sync quickly becomes a nightmare if the relationship isn't trivial.

Ideas are welcome, because I'm not really happy with the current status, but this isn't really a solution either.

rizar:
I guess we should delegate storing the configuration to the children only in the case when for each of the parent configuration attributes there is an _exact_ correspondence. Here it is not exact, so let's better keep parent configuration separate, i.e. just make it how it was.

blocks/bricks/__init__.py
```diff
@@ -335,24 +335,48 @@ class LinearMaxout(Initializable):
...
+    @property
+    def output_dim(self):
+        if self.linear_transformation.output_dim and self.num_pieces:
+            return self.linear_transformation.output_dim // self.num_pieces
+        return self._output_dim
+
+    @output_dim.setter
+    def output_dim(self, value):
+        self._output_dim = value
+        if value and self.num_pieces:
```
dwf: A little confused by the 'value' on this line.

blocks/bricks/__init__.py
```diff
@@ -335,24 +335,30 @@ class LinearMaxout(Initializable):
     -----
     See :class:`Initializable` for initialization parameters.
 
-    .. todo:: Name of :attr:`linear_transformation` shouldn't be hardcoded.
-
     """
     @lazy
     def __init__(self, input_dim, output_dim, num_pieces, **kwargs):
         super(LinearMaxout, self).__init__(**kwargs)
+        self.linear_transformation = Linear()
```
rizar: It is 100% up to, but I find `transformation` super long to type. `linear_brick` or just `linear`?

rizar:
LGTM!

PR #426, opened 2015-03-05
A "Restrict" StepRule.

### LINKED ISSUE #353: Ability to restrict application of a StepRule to a subset of parameter tensors
I brought up in the comments on #352 that it would be useful to be able to restrict the application of a `StepRule` to only certain parameters.

A useful trick with convolutional nets is to scale the steepest descent/momentum by some factor for the convolution parameters only. Also, clipping the norm of only the update to the recurrent weights seems like a sane thing to do (if it's not currently what's done in the literature; I thought it was).

Quoth @bartvm:
> Yeah, that does sound reasonable. Not sure what the best way of implementing it is. I guess there are two ways: Either we give all step rules automatic support for e.g. exclude or filter keyword arguments, which makes them pass on excluded parameters/all parameters that don't match the filter without changing them.

> The other way would be to create a a FilterStep which works like `step_rule=FilterStep(StepClipping(threshold=100), VariableFilter(roles=[RECURRENT_WEIGHTS])(cg.shared_variables))`. I guess this could be more powerful because you could easily filter weights for an entire sequence by doing something crazy like `step_rule=FilterStep(CompositeRule(...), VariableFilter(roles=[BIASES])(cg.shared_variables))`.

@rizar was the main architect of the current iteration of `StepRules` so it'd be good to have his input before taking any action.

dwf:
Fixes #353.

Still needs a test. @rizar any comments (particularly on the handling of `updates`) appreciated.

### DIFF
--- a/blocks/algorithms/__init__.py
+++ b/blocks/algorithms/__init__.py
@@ -674,8 +674,7 @@ class Restrict(StepRule):
     variables : iterable
         A collection of Theano variables on which to apply `step_rule`.
         Variables not appearing in this collection will not have
-        `step_rule` applied to them; their updates will be determined
-        solely by any previously chained :class:`StepRule`s.
+        `step_rule` applied to them.
 
     """
     def __init__(self, step_rule, variables):
--- a/blocks/algorithms/__init__.py
+++ b/blocks/algorithms/__init__.py
@@ -9,7 +9,7 @@ from six import add_metaclass
 from theano import tensor
 
 from blocks.graph import ComputationGraph
-from blocks.utils import named_copy, shared_floatx
+from blocks.utils import dict_subset, named_copy, shared_floatx
 from blocks.theano_expressions import l2_norm
 
 logger = logging.getLogger(__name__)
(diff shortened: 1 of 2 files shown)


dwf:
Alright, I think this is ready.

blocks/algorithms/__init__.py
```diff
@@ -657,3 +657,36 @@ def compute_step(self, param, previous_step):
...
+    filters).
+
+    Parameters
+    ----------
+    step_rule : :class:`StepRule`
+        The :class:`StepRule` to be applied on the given variables.
+    variables : iterable
+        A collection of Theano variables on which to apply `step_rule`.
+        Variables not appearing in this collection will not have
+        `step_rule` applied to them; their updates will be determined
```
rizar: The last phrase of this docstring is confusing a bit since it refers to some sort of chaining which does not happen in this class. I understood that you had `CompositeRule` in mind, but I think that for a newcomer it would do more harm than good. I would remove this last phrase.

rizar:
Looks good! You handle updates in the only correct way.

blocks/algorithms/__init__.py
dwf: Fixed in 15a69e1.

dwf:
With @rizar's only comment address I suppose I shall merge this.

PR #430, opened 2015-03-07
Wrapper brick to swap axes

vdumoulin:


vdumoulin:
@dwf I pushed additional tests to verify that, do they cover what you meant to cover?

dwf:
Yeah.There is some funny business with bound methods and picklability but I honestly forget what, and judging by the lack of errors I think we haven't hit it.

I'll give this one more look.

bartvm:
Bound methods can't be pickled (although this can be patched; unbound methods on the other hand are impossible to pickle in Python 3 entirely).

Application "methods" are fine because they're not methods. Application methods on bricks are silently replaced with `Application` objects when the class is constructed, and with `BoundApplication` objects when the class is instantiated. So instead of methods they are actually callable classes stored in attributes.

vdumoulin:
That clarifies a couple things, thanks for explaining!

blocks/bricks/wrappers.py
```diff
@@ -1,3 +1,4 @@
+from six.moves import range
```
dwf: I think we use `xrange` in most other places (they are exactly the same, functionally, but the 'x' is a nice reminder that it's a fake-range i.e. Python 2 `xrange` a.k.a. Python 3 `range`).

blocks/bricks/wrappers.py
```diff
@@ -34,3 +35,35 @@ def apply(self, input_):
...
         return output
+
+
+class WithAxesSwapped(Brick):
+    """Wraps an application around dimshuffles.
+
+    Parameters
+    ----------
+    application_method : callable
+        Some brick's :meth:`apply` method
```
dwf: Bricks can have application methods that are not necessarily named `apply`.

dwf: It might also be a good idea to specify that it expects the application to have exactly one input and one output.

blocks/bricks/wrappers.py
```diff
@@ -34,3 +35,35 @@ def apply(self, input_):
         else:
             output = self.application_method(input_)
         return output
+
+
+class WithAxesSwapped(Brick):
+    """Wraps an application around dimshuffles.
```
dwf: I'd say something more like "Swaps axes in both the input and output of an application."

blocks/bricks/wrappers.py
```diff
@@ -34,3 +35,35 @@ def apply(self, input_):
...
+
+    Parameters
+    ----------
+    application_method : callable
+        Some brick's :meth:`apply` method
+    dim0 : int
+        First dimension to swap
```
dwf: Looks slightly better in the rendered docs (and more consistent with the rest of the documentation) to terminate these descriptions with a period.

dwf:
Alright, other than the documentation fixes I've suggested I'd say this is ready.

vdumoulin:
I made the requested changes.

dwf:
Thanks! I'll merge as soon as Travis passes.

tests/bricks/test_wrappers.py
```diff
@@ -30,3 +31,47 @@ def test_as2d_ndim_leq_2():
...
+    X = tensor.matrix('X')
+    brick = Linear(input_dim=2, output_dim=2, weights_init=Constant(1),
+                   biases_init=Constant(0))
+    wrapper = WithAxesSwapped(brick.apply, 0, 0)
+    wrapper.initialize()
+    brick.W.set_value(
+        numpy.asarray([[1, 2], [1, 1]], dtype=theano.config.floatX))
+    f = theano.function([X], wrapper.apply(X))
+    assert_allclose(f(numpy.arange(
+                        4, dtype=theano.config.floatX).reshape((2, 2))),
```
rizar: This is apparently a syntax error. @bartvm , any idea why scrutinizer skipped this?

bartvm: Flake8 got updated yesterday, and seems like the old version didn't catch it.

PR #431, opened 2015-03-08
Rename Lookup.lookup to Lookup.apply. Fixes #394.

### LINKED ISSUE #394: [naming] LookupTable.lookup to apply
Throughout the library we seem to use the convention that the primary application is called `apply`, and that alternative names are only used for secondary applications or in the case that there are multiple appliations. However, the `LookupTable` brick's main and only application method is confusingly called `lookup`. Maybe we should change this to `apply` as well?

bartvm:


dwf:
Flake errors, otherwise LGTM

bartvm:
Flake errors are actually unrelated (@rizar had them in #432 as well). Seems like `flake8` released a new version yesterday. Fixed them here.

PR #432, opened 2015-03-08
Multiply costs with mask

rizar:
Fixes the bug @bartvm  reported.


rizar:
Syntax errors here are apparently unrelated to the content of the PR, so I will just merge it.

PR #433, opened 2015-03-08
Two fixes

rizar:


blocks/extensions/__init__.py
```diff
@@ -128,6 +128,24 @@ def on_interrupt(self):
         pass
 
 
+class CallbackName(str):
```
bartvm: I preferred `Callback`, considering that a callback in our case is nothing but a string anyway, adding the `Name` is redundant IMHO and just seems to suggest that there is an actual `Callback` class floating around somewhere, which there isn't.

rizar: In a way `Callback` class is just a function, because we do not need any additional functionality for them. I think it is confusing when a class named `Callback` does not even have `__call__` method.

rizar:
Ping! I merged this PR, sorry for inconvenience!

blocks/extensions/__init__.py
bartvm: Why would it need to be callable? Right now, the `callback` argument is a string. This class just adds an equality operator, but it still *is* the callback as we pass it to `do`, it's not the name of something else out there. If you call it `CallbackName` you would need to call the arguments `callback_name ` as well. But like I said, a callback in our case *is* a string, so adding "name" just confuses me.

rizar: This is a question of terminology I guess. For me a callback is a function which is called, not the string which is passed to `dispatch`. That is `def before_training, def after_training` and so on. That's why I was confused. Seems like you call the string message passed to `dispatch` a callback, right?

bartvm: I see your point, but yeah, we need to be consistent then and rename the arguments to callback_name. Currently it's confusing that the callback argument is a CallbackName instance (although I'm also fine with just callback, because it's shorter and the distinction isn't very important).

rizar: But [here]( https://github.com/bartvm/blocks/blob/master/blocks/extensions/__init__.py#L59) it is already `callback_name`. Which location in the code are you talking about?

bartvm: The do methods if I remember correctly (on my phone, so hard to check)

rizar: There it's `which_callback`.

Anyway, if there are `callback` occurrences, we should consider switching to `callback_name`.

PR #435, opened 2015-03-09
Non theano monitoring

mohammadpz:
An example of numpy cross-entropy for non_theano_monitoring.


blocks/extensions/monitoring.py
```diff
@@ -139,3 +139,23 @@ def do(self, callback_name, *args):
...
+
+class MonitoredQuantity(object):
+    def __init__(self):
+       pass
+
+
+class CrossEntropy(MonitoredQuantity):
+    def __init__(self, **kwargs):
+        super(CrossEntropy, self).__init__(**kwargs)
+        self.total_CrossEntropy, self.examples_seen = 0, 0
```
rizar: I would suggest reading [this](https://www.python.org/dev/peps/pep-0008/#naming-conventions) to learn about python naming conventions. 

In brief, it should be `total_cross_entropy`.

blocks/monitoring/evaluators.py
```diff
@@ -172,7 +172,14 @@ class DatasetEvaluator(object):
 
     """
     def __init__(self, variables, updates=None):
-        self.buffer_ = AggregationBuffer(variables)
+        theano_variables = []
+        non_theano_variables = []
+        for variable in variables:
+            if variable is MonitoredQuantity:
```
bartvm: `if isinstance(variable, MonitoredQuantity)`

blocks/extensions/monitoring.py
```diff
@@ -139,3 +139,31 @@ def do(self, callback_name, *args):
             self.add_records(self.main_loop.log,
                              self._buffer.get_aggregated_values().items())
             self._buffer.initialize_aggregators()
+
+
+class MonitoredQuantity(object):
+    def __init__(self):
+        pass
+
+    @abstractmethod
```
bartvm: `from abc import abstractmethod, ABCMeta`

blocks/extensions/monitoring.py
```diff
@@ -139,3 +139,31 @@ def do(self, callback_name, *args):
             self.add_records(self.main_loop.log,
                              self._buffer.get_aggregated_values().items())
             self._buffer.initialize_aggregators()
+
+
+class MonitoredQuantity(object):
```
bartvm: `@six.add_metaclass(ABCMeta)`

mohammadpz:
Now I get this error:
```python
  File "blocks/monitoring/evaluators.py", line 11, in <module>
    from blocks.extensions.monitoring import MonitoredQuantity
  File "/Users/Mohammadpz/Desktop/blocks/blocks/extensions/monitoring.py", line 6, in <module>
    from blocks.monitoring.evaluators import AggregationBuffer, DatasetEvaluator
  File "/Users/Mohammadpz/Desktop/blocks/blocks/monitoring/evaluators.py", line 11, in <module>
    from blocks.extensions.monitoring import MonitoredQuantity
ImportError: cannot import name MonitoredQuantity
```
I'm thinking maybe it's because of importing loop between ```evaluators.py``` and ```monitoring.py```.

bartvm:
Then just move the class to evaluators.py :)

blocks/extensions/monitoring.py
```diff
@@ -139,3 +141,32 @@ def do(self, callback_name, *args):
             self.add_records(self.main_loop.log,
                              self._buffer.get_aggregated_values().items())
             self._buffer.initialize_aggregators()
+
+
+@six.add_metaclass(ABCMeta)
+class MonitoredQuantity(object):
```
rizar: I think that a reasonable place for `MonitoredQuantity` is `blocks/bricks/monitoring/aggregation.py`.

blocks/extensions/monitoring.py
```diff
@@ -139,3 +140,19 @@ def do(self, callback_name, *args):
...
+class CrossEntropy(MonitoredQuantity):
+    def __init__(self, **kwargs):
+        super(CrossEntropy, self).__init__(**kwargs)
+        self.name = "non_thenao_cross_entropy"
+        self.total_cross_entropy, self.examples_seen = 0.0, 0
+
+    def accumulate(self, target, predicted):
+        import numpy
+        self.total_cross_entropy += -(target * numpy.log(predicted)).sum()
+        #self.total_cross_entropy += 1.1
```
mohammadpz: I'll remove all comments later.

blocks/monitoring/evaluators.py
```diff
@@ -211,8 +248,23 @@ def process_batch(self, batch):
...
+        if self._accumulate_fun2 is not None:
+            self.non_theano_variables[-1].accumulate(self._accumulate_fun2(**batch)[0], self._accumulate_fun2(**batch)[1])
+            
+
     def get_aggregated_values(self):
-        return self.buffer_.get_aggregated_values()
+        a = self.buffer_.get_aggregated_values()
+        b = self.buffer2_.get_aggregated_values()
+        c = a.copy()
+        b.update(c)
```
mohammadpz: I did this to concatenate results from both theano and non-theano variables.

tests/monitoring/test_non_theano_monitoring.py
```diff
@@ -0,0 +1,33 @@
...
+
+
+def test_dataset_evaluators():
+    X = theano.tensor.vector('X')
+    Y = theano.tensor.vector('Y')
+
+
+    data = [numpy.arange(1, 4, dtype=floatX).reshape(3, 1) / 7.0,
+            numpy.arange(11, 14, dtype=floatX).reshape(3, 1) / 17.0]
+
```
mohammadpz: So here the number of batches is 1. right?
But when I check ```examples_seen``` in ```monitoring.py``` it is 6!

blocks/monitoring/evaluators.py
```diff
@@ -211,8 +248,23 @@ def process_batch(self, batch):
...
 
+        try:
+            batch = dict_subset(batch, self.buffer2_.input_names)
+        except KeyError:
+            reraise_as(
+                "Not all data sources required for monitoring were"
+                " provided. The list of required data sources:"
+                " {}.".format(self.buffer2_.input_names))
+        if self._accumulate_fun2 is not None:
+            self.non_theano_variables[-1].accumulate(self._accumulate_fun2(**batch)[0], self._accumulate_fun2(**batch)[1])
```
mohammadpz: Here I should iterate over variables and call their ```accumulate```. Right?
I'm still not sure how to keep track of ```requires``` for each ```MonitoredQuantity```. @bartvm , could you let me know where should I create that dictionary which maps from variable to its requires?

bartvm: There are many ways of doing it, just chose whichever one is easiest for you. A useful method to keep in mind is `list.index(item)`, which would allow you to do something like 
```python
monitored_quantity.accumulate(*[numerical_values[unique_theano_variables.index(requirement)]
                                for requirement in monitored_quantity.requires])
```
and avoid using a lookup dictionary entirely.

blocks/monitoring/evaluators.py
```diff
@@ -211,8 +248,23 @@ def process_batch(self, batch):
...
+            
+
     def get_aggregated_values(self):
-        return self.buffer_.get_aggregated_values()
+        a = self.buffer_.get_aggregated_values()
+        b = self.buffer2_.get_aggregated_values()
+        c = a.copy()
```
bartvm: Why make a copy? `b.update(a)` has the same effect, no?

tests/monitoring/test_non_theano_monitoring.py
bartvm: It's 3, because the first axis is the batch-axis and you're reshaping to (3, 1) i.e. 3 batches with 1 element.

tests/monitoring/test_non_theano_monitoring.py
```diff
@@ -0,0 +1,33 @@
...
+    data_stream = IterableDataset(dict(X=data[0], Y=data[1])).get_example_stream()
+
+    print data
+
+    validator = DatasetEvaluator([CrossEntropy(requires=[X, Y]), CategoricalCrossEntropy().apply(X, Y), ])
```
bartvm: You'll need names for all the monitored values, else you won't know what key to use in your final dictionary of results.

blocks/monitoring/evaluators.py
```diff
@@ -197,6 +230,10 @@ def _compile(self):
         else:
             self._accumulate_fun = None
 
+        self._accumulate_fun2 = theano.function(
```
rizar: A tip: these two `accumulate_fun`'s can (and should) be merged in one.

blocks/monitoring/evaluators.py
```diff
@@ -238,4 +290,12 @@ def evaluate(self, data_stream):
                 'Only data independent variables were given,'
                 'will not iterate the over data!')
 
+        if self._accumulate_fun2 is not None:
+            for batch in data_stream.get_epoch_iterator(as_dict=True):
```
rizar: You can loop only once over the data stream. Updating the non-Theano monitored quantities should be done in `process_batch`.

blocks/monitoring/evaluators.py
```diff
@@ -5,13 +5,36 @@
...
 
 
+class NonTheanoVariablesBuffer(object):
+
+    def __init__(self, variables):
+        self.variables = variables
+        requires = []
+        for variable in variables:
+            requires += variable.requires
+        self.requires = requires
```
rizar: Eliminate the duplicates by doing `self.requires = list(set(requires))`.

blocks/monitoring/aggregation.py
```diff
@@ -151,3 +151,17 @@ def get_aggregator(self):
                               (self.storage, tensor.zeros_like(self.storage))],
                           accumulation_updates=[(self.storage, self.variable)],
                           readout_variable=self.storage)
+
+@add_metaclass(ABCMeta)
+class MonitoredQuantity(object):
+    def __init__(self, requires=None):
```
rizar: Guess you will want to add a `name` argument to this constructor.

blocks/monitoring/evaluators.py
```diff
@@ -172,7 +195,17 @@ class DatasetEvaluator(object):
 
     """
     def __init__(self, variables, updates=None):
-        self.buffer_ = AggregationBuffer(variables)
+        theano_variables = []
```
rizar: We should also check that all the names are unique. We have such a check in `AggregationBuffer`, but now it is not sufficient, because there also non-theano monitored quantities.

blocks/monitoring/evaluators.py
```diff
@@ -211,8 +248,23 @@ def process_batch(self, batch):
...
             self._accumulate_fun(**batch)
 
+        try:
+            batch = dict_subset(batch, self.buffer2_.input_names)
+        except KeyError:
+            reraise_as(
+                "Not all data sources required for monitoring were"
+                " provided. The list of required data sources:"
+                " {}.".format(self.buffer2_.input_names))
+        if self._accumulate_fun2 is not None:
```
rizar: I would move this logic to the `NonTheanoBuffer`. That is `DatasetEvaluator` calls the functions, gets the numerical values and passes it the `NonTheanoBuffer` that knows better what to do with them.

blocks/monitoring/evaluators.py
```diff
@@ -172,7 +196,20 @@ class DatasetEvaluator(object):
...
+        theano_variables = []
+        non_theano_variables = []
+        for variable in variables:
+            if isinstance(variable, MonitoredQuantity):
+                non_theano_variables += [variable]
+            else:
+                theano_variables += [variable]
+        self.theano_variables = theano_variables
+        self.non_theano_variables = non_theano_variables
+        variable_names = [v.name for v in theano_variables + non_theano_variables]
```
rizar: I guess you are missing `set` here, list of names will always be just as long as the list of variables.

tests/monitoring/test_non_theano_monitoring.py
```diff
@@ -0,0 +1,35 @@
...
+    X = theano.tensor.vector('X')
+    Y = theano.tensor.vector('Y')
+
+    # Three batches, each with 2 elements
+    data = [numpy.arange(1, 7, dtype=floatX).reshape(3, 2) / 7.0,
+            numpy.arange(11, 17, dtype=floatX).reshape(3, 2) / 17.0]
+
+    data_stream = IterableDataset(dict(X=data[0], Y=data[1])).get_example_stream()
+
+    print data
```
rizar: I know you used printing for debugging purposes, but as it now works, please switch to `assert_allclose`, see any test where numpy arrays are compared for examples.

blocks/extensions/monitoring.py
```diff
@@ -139,3 +140,17 @@ def do(self, callback_name, *args):
             self.add_records(self.main_loop.log,
                              self._buffer.get_aggregated_values().items())
             self._buffer.initialize_aggregators()
+
+
+class CrossEntropy(MonitoredQuantity):
```
rizar: If this is going to be used for test purposes only, move it to the test file please.

blocks/monitoring/evaluators.py
```diff
@@ -50,8 +76,6 @@ def __init__(self, variables, use_take_last=False):
         self.use_take_last = use_take_last
 
         self.variable_names = [v.name for v in self.variables]
-        if len(self.variable_names) < len(self.variables):
```
rizar: Leave this check as well: sometimes `AggergationBuffer` is used on its own, e.g. in `TrainingDataMonitoring`.

blocks/monitoring/evaluators.py
```diff
@@ -172,7 +196,20 @@ class DatasetEvaluator(object):
...
+            if isinstance(variable, MonitoredQuantity):
+                non_theano_variables += [variable]
+            else:
+                theano_variables += [variable]
+        self.theano_variables = theano_variables
+        self.non_theano_variables = non_theano_variables
+        variable_names = [v.name for v in theano_variables + non_theano_variables]
+        if len(variable_names) < len(theano_variables + non_theano_variables):
+            raise ValueError("variables should have different names")
+        self.theano_buffer_ = AggregationBuffer(theano_variables)
```
rizar: I don't remember why there was underscore in the name `buffer_`, maybe there was a conflict with some python global function. Since the names are now longer, you can remove it.

blocks/monitoring/evaluators.py
```diff
@@ -186,33 +223,51 @@ def _compile(self):
...
+        if self.theano_buffer_.accumulation_updates:
             updates = OrderedDict()
-            updates.update(self.buffer_.accumulation_updates)
+            updates.update(self.theano_buffer_.accumulation_updates)
             if self.updates:
                 updates.update(self.updates)
-            self._accumulate_fun = theano.function(
-                self.buffer_.inputs, [],
-                updates=updates)
+            # append?
```
rizar: `extend` is equivalent to `+=` and most like faster.

blocks/monitoring/evaluators.py
```diff
@@ -186,33 +223,51 @@ def _compile(self):
...
+            inputs += self.theano_buffer_.inputs
+
+        if self.non_theano_buffer_.variables != []:
```
rizar: Why do we need this `if`? += or `extend` work fine with empty lists

rizar:
Nice, seems like the code is mostly there! I will check it on PER today.

However a lot of cleaning up and of course docs are yet to be done.

blocks/monitoring/evaluators.py
bartvm: @mohammadpz I probably caused a bit of confusion here because of something I said yesterday or the day before, so let me just clarify:
```python
l = l + [new_element]  # Slow! Allocates a new list, I believe this was what I told you to avoid
l.extend(new_element)  # Better, because it extends the list in-place
l += [new_element]  # Equivalent to above, but without the overhead of a function call, and looks nicer
l.append(new_element)  # This is the fastest if you're adding just one element
```
Note that we're often talking about nanoseconds for small lists, so it's not that important, but when dealing with large lists it's worth keeping in mind.

blocks/extensions/monitoring.py
```diff
@@ -139,3 +140,17 @@ def do(self, callback_name, *args):
...
+    def __init__(self, **kwargs):
+        super(CrossEntropy, self).__init__(**kwargs)
+        self.total_cross_entropy, self.examples_seen = 0.0, 0
+
```
rizar: When trying to play with your code I realized that a `reset` (or `initialize` step is missing). Every time before monitoring the value of `MonitoredQuantity` should be somehow initialized.

blocks/monitoring/evaluators.py
mohammadpz: @bartvm , Thanks for this good comparison.

blocks/extensions/monitoring.py
mohammadpz: @rizar , hmmm. Could you please give me more clear description?

bartvm: Right now you keep adding and adding and adding to `self.examples_seen` in each epoch. You should reset the counters to 0 at the start of each epoch.

blocks/monitoring/aggregation.py
```diff
@@ -151,3 +151,19 @@ def get_aggregator(self):
...
                           readout_variable=self.storage)
+
+
+@add_metaclass(ABCMeta)
+class MonitoredQuantity(object):
+    def __init__(self, requires=None, name=None):
+        self.requires = requires
+        self.name = name
+        pass
+
```
rizar: Sorry, reseting the counters in the `readout` method is not enough. I am not a fan of it because it will be very hard to debug: as soon as you call `readout` your state is lost!

I would rather add an `initialize` method to the `MonitoredQuanity` class that would reset the state, that is set the counters to 0, etc.. `initialize` methods of all observed quantities can be called in `NonTheanoBuffer`'s `initialize` method, which in turn would be triggered by `DatasetEvaluator.initialize_aggregators`.

Let me know if you have any questions.

mohammadpz: I made appropriate changes.

rizar: I am afraid you never call the `initialize` method.

Please try change you test to cover this case, that is make sure that the test fails if reseting does not work (like now).

Also I guess you can start working on documentation. This abstract class should be well documented, the documentation should show what the inherit methods should do. Also don't forget to update docs of `DatasetEvaluator` and `DataStreamMonitoring` to reflect the fact that not only Theano variables are supported.

mohammadpz: Oh. Actually I initialized them in evaluators but I forgot to commit that file!

mohammadpz:
Could you let me know what Coverage is and how it is decreased? I don't understand the log of ```coveralls```.

rizar:
See Bart's explanation from  a fuel PR

> Coverage is the number of lines of code that are run by unit tests divided by the total number of lines of code. If it decreases it means that this percentage dropped. In this case it's because you added 9 new lines of code to transformers/__init__.py, but only 7 of those are tested, so the coverage decreases slightly. There's no need to worry about this too much, it's mostly to remind people that their code needs to be tested, and if the drop is very large then it is a sign that tests need to be written. You can actually click on the details and see exactly which line is currently untested.

Don't worry about it too much when you are sure that your code is well tested. Currently your test does not check `initialize` method, by the way, you should simulate the typical situation that monitoring is run many times with the same quantities monitored.

blocks/extensions/monitoring.py
```diff
@@ -41,23 +41,24 @@ def add_records(self, log, record_tuples):
 
 
 class DataStreamMonitoring(SimpleExtension, MonitoringExtension):
-    """Monitors values of Theano variables on a data stream.
+    """Monitors values of Theano and non-Theano variables on a data stream.
```
rizar: I saw it before but did not mention because there were important issues, but now I think it's the time to decide on terminology. I am not a fan of `non-Theano variables` because those are not variables. If we call them monitored _quantities_, please make sure that this term is used throughout the documentation.

blocks/extensions/monitoring.py
```diff
@@ -41,23 +41,24 @@ def add_records(self, log, record_tuples):
...
         :class:`~tensor.TensorSharedVariable` updates to be performed
-        during evaluation. Be careful not to update any model parameters
-        as this is not intended to alter your model in any meaningfull
-        way. A typical use case of this option arises when the theano
-        function used for evaluation contains a call to
-        :func:`~theano.scan` which might have returned shared
-        variable updates.
+        during evaluation. This parameter is only for Theano variables.
+        Be careful not to update any model parameters as this is not
+        intended to alter your model in any meaningfull way. A typical
```
rizar: The last sentence is very nice, the others confuse more than help, I think.

blocks/monitoring/aggregation.py
```diff
@@ -151,3 +151,41 @@ def get_aggregator(self):
...
+    """
+    def __init__(self, requires=None, name=None):
+        self.requires = requires
+        self.name = name
+        pass
+
+    @abstractmethod
+    def initialize(self):
+        """Initialize parameters for this monitored quantity."""
```
rizar: Not parameters, but rather accumulators?

blocks/monitoring/aggregation.py
```diff
@@ -151,3 +151,41 @@ def get_aggregator(self):
...
+    requires : list
+        List of Theano variables needed to calculate this quantity.
+    name : str
+        The name of monitored quantity which appears in the log.
+
+    """
+    def __init__(self, requires=None, name=None):
+        self.requires = requires
+        self.name = name
+        pass
```
rizar: Remove this `pass`, please.

blocks/monitoring/aggregation.py
```diff
@@ -151,3 +151,41 @@ def get_aggregator(self):
...
                           accumulation_updates=[(self.storage, self.variable)],
                           readout_variable=self.storage)
+
+
+@add_metaclass(ABCMeta)
+class MonitoredQuantity(object):
+    """The base class for monitored-quantities.
+
+    To calculate quantities which are not Theano variables.
+    The calculations are done in an iterative procedure for
```
rizar: > The calculations are done in an iterative procedure for every single batch

This sentence does not make sense to me, and in general this docstring should be rewritten, e.g.

```python
"""
To monitor a non-Theano quanity in Blocks you have to implement this interface for it. 
The initialize method should do this, accumulate method that, and finallly readout is called to 
do this. 

See Also
------------
   :class:`~blocks.monitoring.evaluators.DatasetEvaluator`, 
   :class:`~blocks.extensions.DataStreamMonitoring`

"""
```

blocks/monitoring/evaluators.py
```diff
@@ -5,13 +5,80 @@
...
 
 logger = logging.getLogger()
 
 
+class MonitoredQuantityBuffer(object):
+    """Intermediate results of aggregating values of monitored-quantity.
+
+    Accumulate results for a list of monitored-quantity for every
+    single batch. Provides initialization and readout routines to
+    initialize each variable and capture its accumulated results.
```
rizar: quantity

blocks/monitoring/evaluators.py
```diff
@@ -5,13 +5,80 @@
...
+
+
+    Parameters
+    ----------
+    variables : list of :class:`MonitoredQuantity`
```
rizar: Could you please make sure that you have quantities everywhere, and not here `variable` and there `quanity`?

blocks/monitoring/evaluators.py
```diff
@@ -5,13 +5,80 @@
...
+        The variable names are used as record names in the logs. Hence, all
+        the variable names must be different.
+
+    Attributes
+    ----------
+    requires : list of :class:`~tensor.TensorVariable`
+        Needed to calculate monitored-quantity.
+    variable_names : list of str
+        The name of monitored-quantity.
```
rizar: names, quantities

blocks/monitoring/evaluators.py
```diff
@@ -5,13 +5,80 @@
...
+    inputs : list of :class:`~tensor.TensorVariable`
+        The list of inputs needed for variables in `requires`.
+    input_names : list of str
+        The name of the inputs needed for variables in `requires`.
```
rizar: names

blocks/monitoring/evaluators.py
```diff
@@ -137,7 +204,7 @@ def get_aggregated_values(self):
 
 
 class DatasetEvaluator(object):
-    """A DatasetEvaluator evaluates many Theano variables on a dataset.
+    """A DatasetEvaluator evaluates many Theano varialbes and monitored-quantities.
```
rizar: No need more hyphen here, varialbes -> variables.

blocks/monitoring/evaluators.py
```diff
@@ -163,16 +231,31 @@ class DatasetEvaluator(object):
...
+        theano_variables = []
+        monitored_quantities = []
+        for variable in variables:
+            if isinstance(variable, MonitoredQuantity):
+                monitored_quantities.append(variable)
+            else:
+                theano_variables.append(variable)
+        self.theano_variables = theano_variables
+        self.monitored_quantities = monitored_quantities
+        all_variables = theano_variables + monitored_quantities
```
rizar: What do you need `all_variables` for when you already have `variables`?

blocks/monitoring/evaluators.py
```diff
@@ -186,33 +269,51 @@ def _compile(self):
...
+            numerical_values = self._accumulate_fun(**batch)
+            self.monitored_quantities_buffer.accumulate_variables(
+                numerical_values)
 
     def get_aggregated_values(self):
-        return self.buffer_.get_aggregated_values()
+        values = self.theano_buffer.get_aggregated_values()
+        monitored_quantities_values = self.monitored_quantities_buffer.\
+            get_aggregated_values()
+        values.update(monitored_quantities_values)
```
rizar:
```python

values.update(
    self.monitored_quantities_buffer.get_aggregated_values())

```

blocks/monitoring/evaluators.py
```diff
@@ -186,33 +269,51 @@ def _compile(self):
...
 
     def initialize_aggregators(self):
-        self.buffer_.initialize_aggregators()
+        self.theano_buffer.initialize_aggregators()
+        self.monitored_quantities_buffer.initialize()
 
     def process_batch(self, batch):
         try:
-            batch = dict_subset(batch, self.buffer_.input_names)
+            input_names = self.theano_buffer.input_names + \
```
rizar: I feel bad about this code, because you will have duplicates in `input_names`. Can you make `unique_inputs` from 285 an attribute, i.e. `self.unique_inputs`, and reuse it here?

rizar:
After you address the new comments I've made, please rebase your PR and I believe tomorrow we will merge it.

blocks/extensions/monitoring.py
mohammadpz: @rizar ,
The only thing I added is this:
```his parameter is only for Theano variables.```
Which line do you mean is confusing?

File: blocks/monitoring/evaluators.py (hunk 37 above)
Comment (mohammadpz) [inline blocks/monitoring/evaluators.py]: Very stupid thing actually :D

**Comment by mohammadpz** [comment]
I don't know why it's like this. Here's what I did:
```python
git checkout master
git pull upstream master
git checkout non_theano_monitoring
git rebase master
git push origin non_theano_monitoring
```

rizar:
I guess your local master branch was not clean. 

```
git branch -f master upstream/master
```

would give you clean master.

Now you need to recover your branch before rebase. @bartvm , @dwf, I don't know how to heal consequences of of a wrong rebase, can you please help?

bartvm:
`git reflog` and `git fsck --full` are your friend if you need to revert to commits that aren't part of your history anymore. I'll try to find @mohammadpz to see if he needs help.

rizar:
A future hint anyway. When you are unsure before doing a git operation, create a backup branch:

```
git branch backup-non-theano-monitoring
```

Then after a screwed up rebase you would not have anything lost.

mohammadpz:
Sorry for the mistake.
I'll get into it tonight.

bartvm:
@mohammadpz Ping! Keep in mind that the deadline is tomorrow. Did you manage to resolve the merge issues last night?

mohammadpz:
@bartvm Pong! Is everything right?

rizar:
Apparently not, because the tests are not passing. Click on `Details` button to see the error messages.

bartvm:
But also, a lot of commits don't actually seem to be yours (e.g. `VariableClipping` shows up as new code, but that's from #469). It seems like you merged instead of rebased at several points. Please have a look at http://blocks.readthedocs.org/en/latest/development/pull_request.html#rebasing for a the command you need.

blocks/extensions/monitoring.py
rizar: Well, I do not like this a doc as whole, but as it's most written not by you, we will see later what to do with it.

tests/monitoring/test_monitored_quantity.py
```diff
@@ -0,0 +1,48 @@
...
+
+    data = [numpy.arange(1, 7, dtype=floatX).reshape(3, 2),
+            numpy.arange(11, 17, dtype=floatX).reshape(3, 2)]
+    data_stream = IterableDataset(dict(X=data[0],
+                                       Y=data[1])).get_example_stream()
+
+    validator = DatasetEvaluator([
+        CrossEntropy(requires=[X, Y],
+                     name="monitored_cross_entropy0"),
+        # to test two same quantities and make sure that state will be reset
```
rizar: This does not test that the state will be reset, because you are using two different objects, each having its own state.

rizar:
Commits history is dirty again, we can not merge it. :( I don't know why the proper diff is shown.

mohammadpz:
So, what should I do?

rizar:
Well, ideally you should read a bit about git, try to google your issue, and so on. 

Ask somebody for help in the lab, because I don't have access to your repo and don't know what its current state is.

PR #443, opened 2015-03-10
Use better protocol defaults. Fixes #442.

### LINKED ISSUE #442: Pickling is insanely slow and space-inefficient
My current model refuses to pickle because it says it runs out of space on `/tmp`... Turns out that my `/tmp` is 8GB, while the pickle file ends up being 18GB+!

The main issue seems to be that we're using the default protocol, which is 0 Python 2. This is insanely inefficient. Using protocol 2 takes about 1/10th of the time, and results in a file which is "only" 4.6GB.

bartvm:


blocks/serialization.py
```diff
@@ -36,6 +41,7 @@
 
 def pickle_dump(*args, **kwargs):
     """A wrapper around pickle's dump that provides informative errors."""
+    kwargs.setdefault('protocol', DEFAULT_PROTOCOL)
```
rizar: This is more memory efficient and slower, right?

bartvm: No, much faster! On python 2 it switches from protocol 0 to 2, on Python 3
nothing changes (keeps using protocol 3). Pickling time went from 5 minutes
to 10 seconds for me

rizar:
Any reasons why you don't merge this?

bartvm:
Nope, made the PR just before I left for the gym

Issue #347, opened 2015-02-24
Check whether all bricks where initialized


bartvm:
Following the discussion at #337, it would be a good idea to add another check to see whether weights were initialized (besides #244). Namely, in the main loop we can simply walk the graph, retrieve the bricks, and check whether all of them were initialized. If that's not the case, we can raise a warning (just as is currently done when e.g. parameters don't match between the model and the algorithm).

rizar:
That's really easy to do with a model, but you want it to be as optional as possible.

bartvm:
Mm, the thing is, that although I see the usefulness of `Model`, for things like this, you shouldn't need to pass a `Model` instance, because all of this works perfectly fine by just walking the computation graph (using e.g. `ComputationGraph`) without making assumptions (like only having a single cost).

(I'm also not a fan of how `Model` is really just a `ComputationGraph` with a few assumptions built in/some extra features, but that the uses of the two are entirely different, but I haven't thought of a better alternative.)

rizar:
But how do you want to get this computation graph? `algorithm.cost`?

bartvm:
Well, that's why I said I didn't have a better solution :) because I don't want to rely on `algorithm` having a particular interface either. But my thinking is generally more supportive of not passing a `Model` class, but somehow passing an annotated Theano graph instead e.g. something like `cg=ComputationGraph(cost)` instead of `model=Model(cost)` and moving `get_parameters` from `Model` to something like `ComputationGraph.get_parameters(brick_naming=True)`.

One of the reasons is because I've had several people be really confused about the `Model` class, asking me whether they needed to subclass it, etc. When I tell them it's really just providing a way to rename parameters, and tell extensions what the cost is, and that you could just as well pass `None` in many cases, they're even more confused. All in all, I think the `Model` abstraction is basically just `ComputationGraph` + assumption that there are bricks and a single objective, so I wonder if it's not possible to just use `ComputationGraph`, an `OBJECTIVE` variable role, and a `get_parameters(brick_naming=True)` method. If that is the case, it's one very confusing abstraction less, and simplifcation is always a good thing. Then again, as I said, I haven't thought this through too much (assignment due in a few hours!).

rizar:
Okay, it is not nice that our understanding of what `Model` is are different. Let me try to explain once more why I came to the thought that we need such a thing.

### Terminology

[...]

PR #444, opened 2015-03-10
WIP: Model refactoring

rizar:
This is very much work in progress but I guess is enough to initiate discussion.

### DIFF
--- a/blocks/brickgraph.py
+++ b/blocks/brickgraph.py
@@ -1,0 +1,32 @@
+from blocks.graph import ComputationGraph
+from blocks.filter import get_brick
+from picklable_itertools import chain
+
+class BrickComputationGraph(ComputationGraph):
+    """Computation graph annotated with bricks.
+
+    Attributes
+    ----------
+    top_bricks : list of :class:`~blocks.bricks.base.Brick`
+        The top bricks from the computation graph, that is those
+        that are not children of other bricks.
+
+    """
+    def __init__(self, outputs):
+        super(BrickComputationGraph, self).__init__(outputs)
+        self._get_bricks()
+
+    def _get_bricks(self):
+        bricks = [get_brick(var) for var
+                  in self.variables + self.scan_variables if get_brick(var)]
+        children = set(chain(*(brick.children for brick in bricks)))
+        # Quadratic complexity: we should not have thousands of
+        # top-level bricks.
+        self.top_bricks = []
+        for brick in bricks:
+            if brick not in children and brick not in self.top_bricks:
+                self.top_bricks.append(brick)
+        if len(set(b.name for b in self.top_bricks)) < len(self.top_bricks):
+            raise ValueError("top bricks with the same name")
+
+
--- a/blocks/brickgraph.py
+++ b/blocks/brickgraph.py
@@ -2,6 +2,7 @@ from blocks.graph import ComputationGraph
 from blocks.filter import get_brick
 from picklable_itertools import chain
 
+
 class BrickComputationGraph(ComputationGraph):
     """Computation graph annotated with bricks.
 
@@ -28,5 +29,3 @@ class BrickComputationGraph(ComputationGraph):
                 self.top_bricks.append(brick)
         if len(set(b.name for b in self.top_bricks)) < len(self.top_bricks):
             raise ValueError("top bricks with the same name")
-
-
--- a/blocks/main_loop.py
+++ b/blocks/main_loop.py
@@ -139,7 +139,7 @@ class MainLoop(object):
             # similarly.
             if not self._model.get_objective() == self.algorithm.cost:
                 logger.warning("different costs for model and algorithm")
-            if not (set(self._model.get_params().values()) ==
+            if not (set(self._model.get_parameters()) ==
                     set(self.algorithm.params)):
                 logger.warning("different params for model and algorithm")
 
--- a/blocks/main_loop.py
+++ b/blocks/main_loop.py
@@ -72,9 +72,9 @@ class MainLoop(object):
         The training algorithm.
     data_stream : instance of :class:`.DataStream`.
         The data stream.
-    model : :class:`.AbstractModel` instance, optional
+    model : :class:`~blocks.Model` instance, optional
         The model object. It is entirely transparent for the main loop
-        but may be used by extensions.
+        and used only by extensions.
     log : instance of :class:`.TrainingLog`, optional
         The log. When not given, a :class:`.TrainingLog` is created.
     extensions : list of :class:`.TrainingExtension` instances
--- a/blocks/main_loop.py
+++ b/blocks/main_loop.py
@@ -72,7 +72,7 @@ class MainLoop(object):
         The training algorithm.
     data_stream : instance of :class:`.DataStream`.
         The data stream.
-    model : :class:`~blocks.Model` instance, optional
+    model : :class:`~blocks.Model` instance
         The model object. It is entirely transparent for the main loop
         and used only by extensions.
     log : instance of :class:`.TrainingLog`, optional
@@ -82,8 +82,8 @@ class MainLoop(object):
         here.
 
     """
-    def __init__(self, algorithm, data_stream,
-                 model=None, log=None, extensions=None):
+    def __init__(self, algorithm, data_stream, model,
+                 log=None, extensions=None):
         if not log:
             log = TrainingLog()
         if not extensions:
--- a/blocks/model.py
+++ b/blocks/model.py
@@ -6,42 +6,13 @@ graph, bricks, parameters and main loop extensions.
 """
 import logging
 from collections import OrderedDict
-from itertools import chain
 
-from blocks.graph import ComputationGraph
+from blocks.brickgraph import BrickComputationGraph
 from blocks.select import Selector
-from blocks.filter import VariableFilter, get_brick
+from blocks.filter import VariableFilter
 from blocks.roles import OBJECTIVE
 
 logger = logging.getLogger()
-
-
-class BrickComputationGraph(ComputationGraph):
-    """Computation graph annotated with bricks.
-
-    Attributes
-    ----------
-    top_bricks : list of :class:`~blocks.bricks.base.Brick`
-        The top bricks from the computation graph, that is those
-        that are not children of other bricks.
-
-    """
-    def __init__(self, outputs):
-        super(BrickComputationGraph, self).__init__(outputs)
-        self._get_bricks()
-
-    def _get_bricks(self):
-        bricks = [get_brick(var) for var
-                  in self.variables + self.scan_variables if get_brick(var)]
-        children = set(chain(*(brick.children for brick in bricks)))
-        # Quadratic complexity: we should not have thousands of
-        # top-level bricks.
-        self.top_bricks = []
-        for brick in bricks:
-            if brick not in children and brick not in self.top_bricks:
-                self.top_bricks.append(brick)
-        if len(set(b.name for b in self.top_bricks)) < len(self.top_bricks):
-            raise ValueError("top bricks with the same name")
 
 
 class Model(object):
--- a/blocks/model.py
+++ b/blocks/model.py
@@ -17,7 +17,15 @@ logger = logging.getLogger()
 
 
 class BrickComputationGraph(ComputationGraph):
+    """Computation graph annotated with bricks.
 
+    Attributes
+    ----------
+    top_bricks : list of :class:`~blocks.bricks.base.Brick`
+        The top bricks from the computation graph, that is those
+        that are not children of other bricks.
+
+    """
     def __init__(self, outputs):
         super(BrickComputationGraph, self).__init__(outputs)
         self._get_bricks()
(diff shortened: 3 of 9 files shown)


rizar:
At this point I would like to hear your comments @bartvm. Especially because this conflicts with the your point of view in #347. 

My motivation for going this way (`Model` _is not_ a computation graph but _has_ a computation graph) is to make a `Model` a good flexible base class and to keep interfaces of `ComputationGraph` and `BrickComputationGraph` crystal clear. I also made it mandatory, otherwise many people will have this moment of confusion when it turns out that the results of their training are not accessible easily.

rizar:
By the way, seems like `index.html` is not tested automatically currently, isn't it?

bartvm:
`index.rst` is tested all the way, except for actually running the main loop. We should probably change `# doctest: +SKIP` to `# doctest: +ELLIPSIS` and add a few dots, so that it actually runs it.

I completely forgot to look at this today, sorry! I'll have to see how the `Model` does on its own tomorrow, but at the very least I still think that `Model` should be able to do completely without needing bricks i.e. in theory the user should still be able to use the the main loop after deleting `blocks/bricks/*` and `BrickComputationGraph` (bar a few import lines, maybe).

rizar:
Ping!

bartvm:
Mmm, I first thought that this could work, as long as the `Model.computation_graph` can be either a `BrickComputationGraph` or a `ComputationGraph`, so that users aren't forced into bricks-world in order to use the main loop.

After reading the code though, I changed my mind again and I still don't see the need. It seems like an extra abstraction that adds almost nothing. The `get_objective` logic should really be in `ComputationGraph` I think, it's similar to e.g. the `parameters` property and can be useful in general, not just for brick-annotated graphs. The `get_top_bricks` method could go in `BrickComputationGraph`, since there too there is nothing `Model`-specific. That just leaves the `get_parameters` method, and that too could be largely absorbed by `BrickComputationGraph`.

I also think there's a lot of "feature envy". The `get_parameters` method for example is just a wrapper of the computation graph when `hierarchical_names=False`. Same thing for `get_top_bricks`. All of that feature envy to me just shows again that `Model` is really just a `MutableBrickComputationGraph`, so it should just be in that class hierarchy instead of artificially being put outside of it, awkwardly dividing overlapping functionality:

* `BrickComputationGraph.top_brick` or `Model.get_top_bricks`?
* `ComputationGraph.parameters` or `Model.get_parameters`?
* `ComputationGraph.objective` or `Model.get_objective`?

Going back to our discussion and your objections to `Model` inheriting from `BrickComputationGraph`, in particular https://github.com/bartvm/blocks/issues/347#issuecomment-77687355, I think that this particular use case shouldn't be covered by the `Model` class. If you want to create an all-powerful object that knows the specific of your model, then by all means write it and pass it to the extensions that need it, but it shouldn't be available to all extensions by default. If you're going to write a custom object, then its custom methods will only be known to your custom extensions anyway.

I agree more with https://github.com/bartvm/blocks/issues/347#issuecomment-77685862, but the argument "I don't like having a method mixed in with properties" is a matter of stylistic preference that shouldn't be dictating the way we organise our class hierarchy IMHO. However, I think it's a very good point that someone should be able to use some of the properties currently reserved to `Model` (like `set_parameters`) without the need to use bricks. However, I would like to see this solved by a hierarchy where we have a `ComputationGraph`, a `BrickComputationGraph` and a mix-in class `MutableComputationGraph` which results in `Model(ComputationGraph, MutableComputationGraph)` and `BricksModel(BricksComputationGraph, MutableComputationGraph)`, because I still think that the only relevant feature that `Model` adds is the ability to override the `params`, `objective`, etc. values.

I'll do my best this weekend or maybe next week to put money where my mouth is and code up my version. (Bear with me! I'm also juggling a bunch of other projects.)

rizar:
> Going back to our discussion and your objections to Model inheriting from BrickComputationGraph, in particular #347 (comment), I think that this particular use case shouldn't be covered by the Model class. If you want to create an all-powerful object that knows the specific of your model, then by all means write it and pass it to the extensions that need it, but it shouldn't be available to all extensions by default. If you're going to write a custom object, then its custom methods will only be known to your custom extensions anyway.

Disagreed for the following reasons

- the following code is just ugly:

```python
extensions=[Extension1(model), Extension2(model), Extensions3(model)...
```

and makes the information flow more complicated. I definitely consider passing all-powerful object as the model as the best way to provide my extensions with what they need.

- the user typically tends to need exactly same things that extensions need. And I want to do the following:

```python
>>> model = cPickle.load(open('model.pkl'))
>>> model.translate("To be or not to be", beam_size=10)
```

This is just convenient and model in this example is a way more than a `ComputationGraph`. That's way `Model` should be an new interface.

rizar:
> The get_objective logic should really be in ComputationGraph

Not sure about that? Do you think the assumption that a single output should be the objective can be done at the `ComputationGraph` level of abstraction? I would not say that, and for this reason I think the right place for this logic is at a higher level of abstraction, which is the `Model`. It is okay for it to strive to be user-friendly, whereas `ComputationGraph` is austere, it is a blue-collar worker, it simply does his tasks. 

> The get_top_bricks method could go in BrickComputationGraph, since there too there is nothing Model-specific

As I said before, I would like to have direct control of the set of bricks given to the extensions. You argued that I should not want it, you arguments were not convincing enough for me. `Model` is a central component, it must be future-proof and this is a bad place for disputable assumptions, IMHO. However it is not the point I care about most, because this can be easily overridden in a subclass. 

> All of that feature envy to me just shows again that `Model` is really just a `MutableBrickComputationGraph`

The main reason why I think it should not is because I consider inheriting from a `Model` as a decent way to start a new project, and if `Model` is a `ComputationGraph` and all exceptions are aware of that and use this fact, this might cause certain inconveniences. The assumption that my `CustomModel` _must_ be created after I fully built all computation graphs involved seems dangerous for me.


All I am trying to do is to ensure that Blocks is 100% future-proof and flexible. Once hard-coded, the assumption that `self.main_loop.model` supports all the all `ComputationGraph` features will be impossible to get rid of.

bartvm:
> As I said before, I would like to have direct control of the set of bricks given to the extensions.

I'm not saying this shouldn't be possible. *Getting* the bricks can be part of `BrickComputationGraph`, *setting* the bricks can be part of `MutableBrickComputationGraph` = `Model`.

> `extensions=[Extension1(model), Extension2(model), Extensions3(model)...`

I don't think those examples are realistic. There will still be a `Model`, but it will simply be more like a `MutableBrickComputationGraph`, which will be enough for 99% of extensions. So the example will be more like:
```python
extensions=[Extension1(), Extension2(), ..., Extension50(my_custom_monster_object)]
```
If you dislike that, you could even subclass `MutableBrickComputationGraph`. I don't see why this would suddenly become impossible now that it is a direct subclass of the computation graph instead of delegating everything to it.

> `>>> model.translate("To be or not to be", beam_size=10)`

This is just a convenient hack, and it's not a reason to decide on a particular class hierarchy. I see no issue with expecting the user to

(a) Simply subclass `Model` (even while it's a subclass of `BrickComputationGraph`)
(b) Expect the user to do `main_loop.find_extension('Translator').translate('To be or not to be')

> The main reason why I think it should not is because I consider inheriting from a Model as a decent way to start a new project, and if Model is a ComputationGraph and all exceptions are aware of that and use this fact, this might cause certain inconveniences.

But this is a work flow that is very particular to you. I don't use it, and I don't see why I should be forced into the same workflow by a magical `Model` class which is basically just a pre-made wrapper around `BrickComputationGraph`. As you're pointing out, it is inconvenient that extensions make assumptions about how `Model` behaves. Yet you are proposing feeding every single one of them a custom version of it! So again, I think that this workflow should be optional and be more like

```python
# This is what 99% of extensions use, it has a well-defined interface
# that the user doesn't mess with, and it behaves mostly like a
# MutableBrickComputationGraph. Users can also just pass
# model = ComputationGraph(cost) for a lot of extensions
# which keeps the dependencies in Blocks nice and clean 
model = Model(cost)

# Now there's a custom object here that is particular to your work
# flow. You basically want a single object that provides a series
# of convenience functions. That's great, but not everyone does this
# and I don't see a reason why we should provide a template for it,
# and then make it compulsory.
class TranslationModel(object):
    def __init__(self, model):
        ...

   def translate(self, sentence, beam_size=10):
       ...

translation_model = TranslationModel(model)

main_loop = MainLoop(
    # 90% of extensions do not care about your custom model
    # the MutableBrickComputationGraph object gives you the
    # ability of setting custom parameters, objectives, etc.
    # but forcing things that are completely specific to your project
    # into the same class just increases the possibility of errors
    extensions=[Printing(), ..., CustomTranslationExtension(translation_model)]
)

# Now you want easy access to things when you unpickle.
# Convenience is important, but it shouldn't dictate how we 
# structure our code entirely, and it's okay for a user to do either this
main_loop.translation_model = translation_model
main_loop.run()

# Or this
main_loop.run()
main_loop.find_extension('CustomTranslationExtension').translation_model.translate
```

Anyway, I think that the you have a particular workflow, which is inheriting from `Model` and stringing all the pieces together in this one class so that you have a series of conveniently accessible methods. But I don't like the idea of introducing a string of dependencies and compulsory arguments just so that this one particular workflow becomes the norm. It's not as minimalistic as it could be, it introduces a new abstraction which is feature-envy heavy, and all the problems it addresses can be solved by simply requiring the user to set an attribute on the `Model` class instead. We are a framework after all, not a plug-and-play solution.

rizar:
> Yet you are proposing feeding every single one of them a custom version of it!

@bartvm, sorry, but you are turning the whole situation upside-down. It is not me who is trying to force everybody my workflow, it's you! You say `model = Model(cost)`: please take the effort to look in the code and you will see that this is supported right out of the box! 

> But I don't like the idea of introducing a string of dependencies and compulsory arguments just so that this one particular workflow becomes the norm.

What compulsory arguments, I do not get it. To make the workflow we are dicussing now _a_ norm (it is you who's trying to say what it is _the_ norm) the only thing I need is that `ComputationGraph` is an attribute of the `Model`, not the `Model` itself, and that it can be substituted after creation. And saying that I am the only one who will ever need this functionality is a huge exaggeration! `Blocks` was designed to host large projects, and something tells me quite a few of them would benefit from concentrating building computation graphs in a `Model` object. Fetching it from an extension requires that I load the whole main loop pickle, which is much longer than to load just a model.


Let's put it like this: we must find a compromise. Currently I am not  going to accept any  
solution in which `Model` is a subclass of `ComputationGraph`, because this ties things too close together, because it does not allow to create a `Model` first and set `ComputationGraph` next. This is my main point.

bartvm:
> You say model = Model(cost): please take the effort to look in the code and you will see that this is supported right out of the box!

I know that. And it would be in my suggestion as well. I am not imposing a workflow, I am simply saying that the `model` keyword argument to `MainLoop` should effectively be a `ComputationGraph`+ instead of a new class whose goal it is to be subclassed with a set of project-specific functions (which is a workflow).

I think that in practice, our suggestions are nearly identical to the user: In both cases they would simply pass `model=Model(cost)`, in both cases they can subclass `Model` if they want to.

Could you give me an example of a case where you would want to instantiate a `Model` but then change the computation graph attribute afterwards?

> What compulsory arguments, I do not get it.

The compulsory argument being a `Model` object, whose instantiation relies on being able to import `BrickComputationGraph`. I am agreeing to it being a compulsory argument, I don't like the `model` keyword argument relying on the existence of `BrickComputationGraph`.

> And saying that I am the only one who will ever need this functionality is a huge exaggeration! 

I never said that. I am saying that your approach is a workflow which is currently supported quite easily by adding an attribute to the `main_loop`. It's a bit slow to unpickle maybe, but there are other ways around that (e.g. attach it as an attribute to the `Model` class instead, subclass `Model`, etc.) A hallmark of a flexible framework is that workflows are supported, not imposed. I am simply questioning whether your solution isn't made with having a particular workflow in mind, but in the process complicating the code and dependency graph of the framework for the cases where this workflow isn't used. My other concern is simply that `Model` seems to be nothing but a feature-envy wrapper.

Anyway, like I said earlier, let me code things up, let me toy with both your example and my idea, and then we'll see. There's no rush, and perhaps we should ask some of the other users what they think as well. It would be interesting to see what workflows they've adopted and how those would fit in too.

rizar:
> Could you give me an example of a case where you would want to instantiate a Model but then change the computation graph attribute afterwards?

Something like that:

```python
class MachineTranslator(Model):

    def __init__(self, encoder, decoder):
         super(MachineTranslator, self).__init__(top_bricks=[encoder, decoder])
         self.source_words = tensor.lmatrix('source_words')
         self.target_words = tensor.lmatrix('target_words')

    def get_cost_graph(self):
         # use self.encoder and self.decoder to build a basic cost graph with 
         # self.source_words and self.target_words as input variables


encoder = Sequence(Lookup(), Fork(), GatedRecurrent()) # off-topic: we should rename `Sequence` to `Stack`!
decoder = SequenceGenerator(...)
model = MachineTranslator(encoder, decoder)
cost = model.get_cost_graph()
regularized_cost = apply_dropout(cost, VariableFilter(bricks=[encoder], roles=[OUTPUT])[0], 0.5)
add_role(regularized_cost, OBJECTIVE)
model.computation_graph = ComputationGraph(cost.outputs + regularized_cost.outputs)
```

In fact I realized that the model I want can be attached to the model you want as an attribute, which makes both functionally equivalent. So this is more a question of personal preference. I agree that we should ask people what workflow they use know and what workflows they would like to use in the future.

PR #445, opened 2015-03-10
Use fixed seed for softmax test

bartvm:


bartvm:
Merging because the errors are unrelated.

PR #447, opened 2015-03-10
Allowing images display in Bokeh

laurent-dinh:
I modified the MNIST example so that the functionality can be tested.

### DIFF
--- a/blocks/extensions/plot.py
+++ b/blocks/extensions/plot.py
@@ -3,20 +3,20 @@ import signal
 import time
 from subprocess import Popen, PIPE
 
+import numpy
 try:
     from bokeh.plotting import (curdoc, cursession, figure, output_server,
                                 push, show)
     BOKEH_AVAILABLE = True
 except ImportError:
     BOKEH_AVAILABLE = False
-
 from blocks.extensions import SimpleExtension
 
 logger = logging.getLogger(__name__)
 
 
-class Plot(SimpleExtension):
-    """Live plotting of monitoring channels.
+class PlotManager(SimpleExtension):
+    """Live plotting.
 
     In most cases it is preferable to start the Bokeh plotting server
     manually, so that your plots are stored permanently.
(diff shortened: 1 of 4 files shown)


dwf:
Docstring build failures:

```
blocks/extensions/plot.py:129 in public method `call`:
        D400: First line should end with '.', not 's'
blocks/extensions/plot.py:205 in public class `DisplayImage`:
        D209: Put multi-line docstring closing quotes on separate line preceeded by a blank line
blocks/extensions/plot.py:205 in public class `DisplayImage`:
        D202: No blank lines allowed *after* class docstring, found 1
blocks/extensions/display.py:76 in public function `reshape_image`:
        D202: No blank lines allowed *after* function docstring, found 1
blocks/extensions/display.py:173 in public class `IndexableImageDatasetDisplay`:
        D202: No blank lines allowed *after* class docstring, found 1
blocks/extensions/display.py:206 in public class `WeightDisplay`:
        D202: No blank lines allowed *after* class docstring, found 1
```

dwf:
And your tests are failing with 

```
from blocks.extensions.plot import Plot
ImportError: cannot import name Plot
```

Sooooo that's interesting.

vdumoulin:
@dwf Looks like `Plot` was renamed `PlotManager`.

blocks/extensions/plot.py
```diff
@@ -139,3 +110,157 @@ def __setstate__(self, state):
...
+            fig.renderers = []
+
+            fig.image_rgba(image=[rgba_image],
+                           x=[0], y=[0],
+                           dw=[min(img_w * 1. / img_h, 1)],
+                           dh=[min(img_h * 1. / img_w, 1)])
+
+
+def format_image_to_rgba(image_0255):
+    # norm_image = image / max(abs(image).max(), 1)
```
vdumoulin: The commented code should be removed.

laurent-dinh:
Yes, I'll need to tie loose ends.

rizar:
Do I understand that the application area for this would be generative modelling of images?

vdumoulin:
I think it's more general than that. Basically anything that has to do with visualization, like looking at weights, for instance.

bartvm:
As I mentioned to @laurent-dinh this afternoon: I think this is a nice feature, but I think advanced plotting (and maybe plotting in its entirety) would be a good candidated for `blocks-contrib`. I'll finally try and find some time to set that up!

bartvm:
`blocks-extras` finally exists, so closing this PR here

Issue #327, opened 2015-02-22
Early Stopping


rizar:
Blocks miss an early stopping extension. This involves figuring out the right time to stop and/or save the training process.

It could be done like that: an extension tracks the given quantity in the log and makes a record when (a) it has reached its minimum/maximum so far (b) when it thinks that no further improvement can be expected. In the case (b) the termination request can be written to the log. The record from the case (a) can be handled by another extension configured to do so.

Altogether it should look like that:

```python
MainLoop(...,
    extensions=[
        ..., 
        EarlyStopping("validation_cost"),
        SerializeMainLoop(path=path_for_the_best)
             .set_conditions() # we have to clear the default condition
             .add_condition(
            "after_epoch", OnRecord(EarlyStopping.BEST_SO_FAR)),
        ...
    ]
)
```

The `OnRecord` predicate is of course to be implemented but this is trivial.

rizar:
There's not so much left here after #448, so I put CCW label.

vikkamath:
Probably irrelevant for this issue, but what does 'CCW' stand for?

dwf:
@rizar What precisely do you think is left to do?

rizar:
An extension that decides when to stop. Implements one of these heuristic rules, smth like "if we did twice more iterations and did not get better, halt". I am not a specialist in these ones, perhaps it would make sense to implement the same one as in PyLearn2.

dwf:
I didn't mean to close this, sorry.

PR #448, opened 2015-03-11
WIP: Track the best

rizar:
Is related to #327 

- [x] document attributes
- [x] add `OnLogRecord` predicate
- [x] test saving the best model using `OnLogRecord`, `TrackTheBest` and `SerializeMainLoop`

### DIFF
--- a/blocks/log.py
+++ b/blocks/log.py
@@ -49,6 +49,12 @@ class AbstractTrainingStatus(object):
         """
         pass
 
+    def __getitem__(self, key):
+        return getattr(self, key)
+
+    def __setitem__(self, key, value):
+        return setattr(self, key, value)
+
 
 class TrainingLogRow(object):
     """A convenience interface for a row of the training log.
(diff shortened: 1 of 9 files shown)


blocks/log.py
```diff
@@ -221,6 +227,9 @@ def _to_dataframe(self):
 
 class TrainingStatus(AbstractTrainingStatus):
     """A simple training status."""
+    def __getattr__(self, key):
```
rizar: Yes, I know it looks strange, but at least it is consistent with the current log behaviour. :)

rizar: No, that turned out to be too much and even breaks the serialzation. Will get rid of that.

rizar:
Hooray, saving the best model works! 

As far as I can judge only documentation editing is left.

rizar:
I will merge it now since it can not break anything, but later feedback would be welcome.

PR #451, opened 2015-03-11
Added error message when brick calls not a child

### LINKED ISSUE #436: Better error message when an application method of not a child is called

dmitriy-serdyuk:
Fixes #436.


bartvm:
Thanks! In the future try to use `.format`, it's a bit cleaner/more powerful than string concatenation, and it's what is used in most other places.

PR #457, opened 2015-03-12
Switch from dicts to lists for dimensions

### LINKED ISSUE #453: get_dims should return OrderedDict
That makes it easier to do things like `input_dims = brick.get_dims(input_names).values()`

bartvm:
Fixes #453. Also see https://github.com/bartvm/blocks/issues/453#issuecomment-78380525.

This PR switches from using dicts to store dimension information to using lists, for two principal reasons:

* Shorter function/constructor calls
* No redundant data; this is safer since there was no guarantee that the names in e.g. `self.source_names` and `self.source_dims` matched at all. There is still a chance that the lists aren't of the same length right now, but I think that eventually we should move to a `safe_zip` implementation like in Pylearn2 which would catch those errors.

### DIFF
--- a/blocks/bricks/parallel.py
+++ b/blocks/bricks/parallel.py
@@ -19,8 +19,7 @@ class Parallel(Initializable):
     >>> from blocks.initialization import Constant
     >>> x, y = tensor.matrix('x'), tensor.matrix('y')
     >>> parallel = Parallel(
-    ...     input_names=['x', 'y'],
-    ...     input_dims=dict(x=2, y=3), output_dims=dict(x=4, y=5),
+    ...     input_names=['x', 'y'], input_dims=[2, 3], output_dims=[4, 5],
     ...     weights_init=Constant(2))
     >>> parallel.initialize()
     >>> new_x, new_y = parallel.apply(x=x, y=y)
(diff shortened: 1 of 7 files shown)


blocks/bricks/sequence_generators.py
```diff
@@ -94,36 +94,43 @@ class BaseSequenceGenerator(Initializable):
...
         self.readout = readout
         self.transition = transition
         self.fork = fork
 
-        self.state_names = transition.compute_states.outputs
-        self.context_names = transition.apply.contexts
-        self.glimpse_names = transition.take_glimpses.outputs
         self.children = [self.readout, self.fork, self.transition]
 
+    @property
```
rizar: I you want these to be properties, better make them private in order to force the user use `generate.states`, `generate.contexts`, etc.. There should not be two alternative interfaces to query such things, and this was merely a shortcut.

blocks/bricks/parallel.py
```diff
@@ -73,24 +70,23 @@ def __init__(self, input_names, input_dims, output_dims,
...
-                "{}_{}".format(child_prefix, name))
-        self.children = self.transforms
+            self.children.append(copy.deepcopy(self.prototype))
+            self.children[-1].name = "{}_{}".format(child_prefix, name)
 
     def _push_allocation_config(self):
-        for name, transform in zip(self.input_names, self.transforms):
-            transform.input_dim = self.input_dims[name]
-            transform.output_dim = self.output_dims[name]
+        for name, input_dim, output_dim, child in \
```
rizar: `name` is not used.

rizar:
At first I was going to protest, but that looks not bad. What we should mention always very explicitly in this approach is which list of names corresponds to a list of dimensions.

blocks/bricks/sequence_generators.py
bartvm: I'm not sure what you mean by private? I just made them properties because right now the user was able to change them, which would just lead to errors because they wouldn't match up with the transition inputs anymore. It seemed like they were just used as shorthands more than anything.

bartvm: Nevermind, it's early here :) Made them private.

bartvm:
@rizar Fixed merge conflicts and addressed your comments. Let me know if you have any more comments

rizar:
Go ahead and merge!

PR #458, opened 2015-03-12
Equizip

bartvm:
Behaves like (i)zip but raises an error if two iterables aren't of the same length. Name shamelessly stolen from the Python mailing list (https://mail.python.org/pipermail/python-3000/2006-March/000160.html), but I think it sounds better than Pylearn2's `safe_zip`.

Right now I get two errors in `test_sequence_generator` and `test_integer_sequence_generator` because they request outputs `as_dict`, but there are more outputs than are named. This raises another question: Should we check if the outputs are named, and if they are, should we raise an error if the lengths don't match? Right now this error would only occur if `as_dict=True` is passed, but we might as well check it by default.

### DIFF
--- a/blocks/algorithms/__init__.py
+++ b/blocks/algorithms/__init__.py
@@ -9,7 +9,7 @@ from six import add_metaclass
 from theano import tensor
 
 from blocks.graph import ComputationGraph
-from blocks.utils import dict_subset, named_copy, shared_floatx
+from blocks.utils import dict_subset, named_copy, shared_floatx, equizip
 from blocks.theano_expressions import l2_norm
 
 logger = logging.getLogger(__name__)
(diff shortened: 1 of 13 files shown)


bartvm:
Just a note that his PR will have merge conflicts with #457 and whichever one gets merged last will need a rebase.

rizar:
I think that we should not make output naming mandatory and only start giving errors when the names are requested, e.g. by `as_dict=True`.

To fix this bug you should change the outputs of `FakeAttention.take_glimpses`, now they are `[]`.

bartvm:
Mm, if I add an output I get an error that there is no dimension, if I then add a dimension of 0 I get an error because it expects `apply` to return glimpses as well, but since it doesn't the lookup in `results` dictionary [here](https://github.com/bartvm/blocks/blob/master/blocks/bricks/sequence_generators.py#L169) crashes... Maybe it's about time to get rid of `FakeAttentionRecurrent`, I've never been a fan to be honest. However, that will be quite a bit more refactoring, so I'll merge this for now without using `equizip` in when `as_dict=True`.

rizar:
Okay, just create an issue for that and I will take a look. `FakeAttentionRecurrent` is not the nicest thing in Blocks indeed, but it requires a bit of thinking about a better design of sequence generator framework.

PR #466, opened 2015-03-13
Fix docstring

fvisin:


blocks/bricks/conv.py
```diff
@@ -185,7 +185,7 @@ class ConvolutionalActivation(Sequence, Initializable):
         The application method to apply after convolution (i.e.
         the nonlinear activation function)
 
-    See :class:`Convolutional` for explanation of other parameters.
+    See: :class:`Convolutional`
```
bartvm: Not sure if this is much of an improvement, considering it's now a bit unclear why you should look at `Convolutional`. The proper NumPy docstring syntax by the way is:
```
See Also
--------
:class:`Convolutional`
```

fvisin: The problem is twofold:

1. "See:" is interpreted as a keyword , leading to print in the documentation `See(class:Convolutional` 
2. "parameters" is interpreted as a keyword too, leading to the weird text:
    * See (class:Convolutional and MaxPooling for explanations of) –
    * parameters. (other) – 

As you can see [here](http://blocks.readthedocs.org/en/latest/api/bricks.html#blocks.bricks.conv.ConvolutionalLayer]).

I was not sure how to fix the second problem. If you suggest me a way to solve it, I will gladly update the PR.

bartvm: It's interpreted like that because there is no new section, so it thinks it's a parameter. Try the one I mentioned, it should work:
```
See Also
--------
:class:`Convolutional` for the other parameters
```

PR #468, opened 2015-03-13
Added the test for VariableFilter.

Thrandis:
Hey,
Here is my ccw for this week. I couldn't find a better solution for testing the filtering by name and by application. Let me know if you have a better idea.
Cesar

### DIFF
--- a/tests/test_VariableFilter.py
+++ b/tests/test_VariableFilter.py
@@ -1,0 +1,53 @@
+from blocks.bricks import Bias, Linear, Sigmoid
+from blocks.filter import VariableFilter
+from blocks.graph import ComputationGraph
+from blocks.roles import BIASES, FILTERS, PARAMETER
+
+from theano import tensor
+
+
+def test_VariableFilter():
+    # Creating computation graph
+    brick1 = Linear(input_dim=2, output_dim=2, name='linear1')
+    brick2 = Bias(2, name='bias1')
+    activation = Sigmoid(name='sigm')
+
+    x = tensor.vector()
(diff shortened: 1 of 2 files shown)


tests/test_VariableFilter.py
```diff
@@ -0,0 +1,53 @@
+from blocks.bricks import Bias, Linear, Sigmoid
+from blocks.filter import VariableFilter
+from blocks.graph import ComputationGraph
+from blocks.roles import BIASES, FILTERS, PARAMETER
+
+from theano import tensor
+
+
+def test_VariableFilter():
```
bartvm: The naming convention would be `test_variable_filter`, CamelCase is only for classes.

Thrandis:
Why does my test decrease the coverage? Should it not increase it?

bartvm:
I think it's because Coveralls compares your current branch merged into the current master against the last known coverage result in your branch's history (which would be the commit in `master` that you branched from). The problem is that that particular commit is by now outdated. From a quick look at the changes, it looks as if you branched out before we removed the `equizip` function from `utils.py` (which was fully tested). This means that the master you branched from had `(covered + N) / (all + N)` lines in `utils.py` covered, but after merging this branch into the *current* `master` this function disappears and you only cover `covered / all` lines (which is lower). Hence Coveralls reports that the coverage since you branched from `master` compared to what the current `master` would look like after merging this branch has decreased.

To get rid of the error, you'll probably want to fetch the latest master to the latest one, and then rebase on top of that. My guess is that that should make Coveralls report correctly on the changed statistics.

PR #469, opened 2015-03-13
Max norm regularization StepRule

dwf:


dwf:
@vdumoulin This may interest you.

dwf:
So... isn't coveralls invoked by Travis? Why is coveralls reporting a coverage decrease before Travis is finished?

bartvm:
Because Travis submits multiple coverage reports (one for each job) which are then combined by Coveralls. However, Coveralls doesn't know how many to expect, so it start spitting out statistics as soon as the first job finishes, and then updates as the other ones finish too.

dwf:
I see. That's mildly irritating. :)

dwf:
Coverage decrease was from two untested lines (raising exceptions). I've added tests for them which should make all checks pass. So, if someone wants to give this a look it should be ready for human consumption now that it has been deemed fit by the machines.

dwf:
@rizar @bartvm Any thoughts on this one? Just rebased.

blocks/algorithms/__init__.py
```diff
@@ -575,6 +577,64 @@ def compute_steps(self, previous_steps):
         return steps, []
 
 
+class VariableClipping(StepRule):
+    """Clip the maximum norm of individual variables along certain axes.
+
+    This :class:`StepRule` can be used to implement L2 norm constraints on
+    individual hidden units, groups of hidden units, convolutional
```
rizar: I am not sure I understand how it can applied to anything different from entire weight tensors.

blocks/algorithms/__init__.py
```diff
@@ -575,6 +577,64 @@ def compute_steps(self, previous_steps):
...
+            raise ValueError("Invalid axes {} for {}, ndim={}".format(
+                self.axes, param, previous_step.ndim))
+        squares = tensor.sqr(previous_step)
+        if len(self.axes) == 0:
+            norms = l2_norm([previous_step])
+        else:
+            norms = tensor.sqrt(
+                reduce(lambda t, a: t.sum(axis=a, keepdims=True),
+                       sorted(self.axes), squares))
+        return (previous_step * tensor.switch(norms > self.threshold,
```
rizar: It seems like you are computing `L2` norm of the step, not of the parameter after update. Shouldn't it be 
`l2_norm([param - previous_step])`?

blocks/algorithms/__init__.py
dwf: If you had convolutional filters of the form `(n_in, n_out, filter_height, filter_width)`, for example, passing axes=(2, 3) would restrict the norm of each `(filter_height, filter_width)` element.

blocks/algorithms/__init__.py
dwf: You were right, of course. I was thinking of the return OrderedDict as a list of Theano updates, when it is in fact not that. Fixed in  83700c8, PTAL.

vdumoulin:
If I understand correctly, this implements gradient rescaling on a per-parameter (parameter being e.g. a weights matrix or a bias vector) basis?

Do you think it's worth adding a way to select which parameters this rescaling applies to?

Regarding the name, you're rescaling, not clipping, so I would reflect that in the class name.

dwf:
Actually, it implements *parameter* rescaling, in a kind of funny way. 

`max_col_norm` and `max_row_norm` in pylearn2 essentially rescales the magnitude of every column of a weight matrix if they exceed a threshold after gradient update. This kind of regularization works a little like weight decay but has the nice property that if your weights are small it doesn't do anything. It also lets you use a quite aggressive learning rate with e.g. dropout and not have everything blow up. The way that `StepRules` work is that they return a delta on the current parameter value rather than an updated value for that parameter, so what this does is compute what the updated parameter value would have been, scales it, and then subtracts the parameter (and negates the sign) to give you an equivalent *step* to apply. See [this line](https://github.com/bartvm/blocks/pull/469/files#diff-be5f60fd147704b628ba4b7e36cbb224R636).

This makes it suitable for the last stage in a `CompositeRule` (or any stage, if you're doing something more exotic). As for how to apply only to certain parameters, we already have a way: [Restrict](http://blocks.readthedocs.org/en/latest/api/algorithms.html#blocks.algorithms.Restrict). So something like `Restrict(VariableClipping(5, axes=(1,)), [my_weight_matrix, my_other_weight_matrix]))` should get the job done.

dwf:
As for rescaling vs. clipping, it is clipping in the sense that it only rescales if the norm grows beyond a certain threshold.

vdumoulin:
You have me convinced.

One last thing: there's a hardcoded assumption that the user is doing gradient descent. Even though that's the only thing implemented at the moment, I think we should document it.

dwf:
@vdumoulin Addressed in b8ef916 (I think).

vdumoulin:
It does. LGTM.

rizar:
@dwf, merge when you want!

PR #471, opened 2015-03-13
Change references to equizip to picklable_itertools.extras implementation

dwf:
Reduce code duplication between repositories. Reduce total line count in Blocks. :)


dwf:
dwf/picklable_itertools@62697ea fixes the one test failure I observed locally.

PR #472, opened 2015-03-13
Sequence generator cost refactoring

### LINKED ISSUE #434: SequenceGenerator's cost method returns cost matrix
Just wondering if we should make it behave similar to the cost bricks instead. That is, the current `cost` method should really be called `cost_matrix` and `cost` should instead be the average cost per sequence element
```python
def cost(...):
    cost = self.cost_matrix(...) / mask.sum()
    cost = add_role(cost, COST)
    return cost
```

orhanf:
Issue #434 CCW ticket

Fixes #434.

### DIFF
--- a/blocks/bricks/sequence_generators.py
+++ b/blocks/bricks/sequence_generators.py
@@ -137,8 +137,10 @@ class BaseSequenceGenerator(Initializable):
 
     @application
     def cost(self, application_call, outputs, mask=None, **kwargs):
-        """Returns the average cost of the sum of per token costs in a sequence
-        over the minibatch.
+        """Returns the average cost of over the minibatch.
+
+        The cost is computed by averaging the sum of per token costs for each
+        sequence over the minibatch.
 
         .. warning::
             Note that, the computed cost can be problematic when batches
@@ -165,7 +167,7 @@ class BaseSequenceGenerator(Initializable):
 
         Adds average cost per sequence element `AUXILIARY`
         variable to the computational graph with name
-        `sequencegenerator_cost_per_sequence_element`
+        `~sequencegenerator_cost_per_sequence_element`
 
         """
         # Compute the sum of costs
--- a/blocks/bricks/sequence_generators.py
+++ b/blocks/bricks/sequence_generators.py
@@ -11,6 +11,7 @@ from blocks.bricks.lookup import LookupTable
 from blocks.bricks.recurrent import recurrent
 from blocks.bricks.attention import (
     AbstractAttentionRecurrent, AttentionRecurrent)
+from blocks.roles import add_role, COST
 from blocks.utils import dict_union, dict_subset
 
 
@@ -136,6 +137,34 @@ class BaseSequenceGenerator(Initializable):
 
     @application
     def cost(self, application_call, outputs, mask=None, **kwargs):
+        """Returns sum of generation costs for output sequences.
+
+        Parameters
+        ----------
+        outputs : :class:`~tensor.TensorVariable`
+            The 3(2) dimensional tensor containing output sequences.
+            The dimension 0 must stand for time, the dimension 1 for the
+            position on the batch.
+        mask : :class:`~tensor.TensorVariable`
+            The binary matrix identifying fake outputs.
+
+        Notes
+        -----
+        The contexts are expected as keyword arguments.
+
+        """
+        # Compute the sum of costs
+        cost = self.cost_matrix(outputs, mask=mask, **kwargs).sum()
+        add_role(cost, COST)
+
+        # Add auxiliary variable for per sequence element cost
+        application_call.add_auxiliary_variable(
+            (cost / mask.sum()) if mask is not None else cost,
+            name='per_sequence_element')
+        return cost
+
+    @application
+    def cost_matrix(self, application_call, outputs, mask=None, **kwargs):
         """Returns generation costs for output sequences.
 
         Parameters
--- a/blocks/bricks/sequence_generators.py
+++ b/blocks/bricks/sequence_generators.py
@@ -139,8 +139,8 @@ class BaseSequenceGenerator(Initializable):
     def cost(self, application_call, outputs, mask=None, **kwargs):
         """Returns the average cost of over the minibatch.
 
-        The cost is computed by averaging the sum of per token costs for each
-        sequence over the minibatch.
+        The cost is computed by averaging the sum of per token costs for
+        each sequence over the minibatch.
 
         .. warning::
             Note that, the computed cost can be problematic when batches
--- a/blocks/bricks/sequence_generators.py
+++ b/blocks/bricks/sequence_generators.py
@@ -137,7 +137,7 @@ class BaseSequenceGenerator(Initializable):
 
     @application
     def cost(self, application_call, outputs, mask=None, **kwargs):
-        """Returns the average cost of over the minibatch.
+        """Returns the average cost over the minibatch.
 
         The cost is computed by averaging the sum of per token costs for
         each sequence over the minibatch.
@@ -165,20 +165,18 @@ class BaseSequenceGenerator(Initializable):
         -----
         The contexts are expected as keyword arguments.
 
-        Adds average cost per sequence element `AUXILIARY`
-        variable to the computational graph with name
-        `~sequencegenerator_cost_per_sequence_element`
+        Adds average cost per sequence element `AUXILIARY` variable to
+        the computational graph with name `~per_sequence_element`
 
         """
         # Compute the sum of costs
-        cost = tensor.mean(self.cost_matrix(outputs,
-                                            mask=mask,
-                                            **kwargs).sum(axis=0))
+        costs = self.cost_matrix(outputs, mask=mask, **kwargs)
+        cost = tensor.mean(costs.sum(axis=0))
         add_role(cost, COST)
 
         # Add auxiliary variable for per sequence element cost
         application_call.add_auxiliary_variable(
-            (cost / mask.sum()) if mask is not None else cost,
+            (costs / mask.sum()) if mask is not None else costs.sum(),
             name='per_sequence_element')
         return cost
 
--- a/blocks/bricks/sequence_generators.py
+++ b/blocks/bricks/sequence_generators.py
@@ -137,24 +137,41 @@ class BaseSequenceGenerator(Initializable):
 
     @application
     def cost(self, application_call, outputs, mask=None, **kwargs):
-        """Returns sum of generation costs for output sequences.
+        """Returns the average cost of the sum of per token costs in a sequence
+        over the minibatch.
+
+        .. warning::
+            Note that, the computed cost can be problematic when batches
+            consist of vastly different sequence lengths.
 
         Parameters
         ----------
         outputs : :class:`~tensor.TensorVariable`
             The 3(2) dimensional tensor containing output sequences.
-            The dimension 0 must stand for time, the dimension 1 for the
-            position on the batch.
+            The axis 0 must stand for time, the axis 1 for the
+            position in the batch.
         mask : :class:`~tensor.TensorVariable`
             The binary matrix identifying fake outputs.
+
+        Returns
+        -------
+        cost : :class:`~tensor.Variable`
+            Theano variable for cost, computed by summing over timesteps
+            and then averaging over the minibatch.
 
         Notes
         -----
         The contexts are expected as keyword arguments.
 
+        Adds average cost per sequence element `AUXILIARY`
+        variable to the computational graph with name
+        `sequencegenerator_cost_per_sequence_element`
+
         """
         # Compute the sum of costs
-        cost = self.cost_matrix(outputs, mask=mask, **kwargs).sum()
+        cost = tensor.mean(self.cost_matrix(outputs,
+                                            mask=mask,
+                                            **kwargs).sum(axis=0))
         add_role(cost, COST)
 
         # Add auxiliary variable for per sequence element cost
@@ -167,18 +184,9 @@ class BaseSequenceGenerator(Initializable):
     def cost_matrix(self, application_call, outputs, mask=None, **kwargs):
         """Returns generation costs for output sequences.
 
-        Parameters
-        ----------
-        outputs : :class:`~tensor.TensorVariable`
-            The 3(2) dimensional tensor containing output sequences.
-            The dimension 0 must stand for time, the dimension 1 for the
-            position on the batch.
-        mask : :class:`~tensor.TensorVariable`
-            The binary matrix identifying fake outputs.
-
-        Notes
-        -----
-        The contexts are expected as keyword arguments.
+        See Also
+        --------
+        :meth:`cost` documentation for parameters.
 
         """
         # We assume the data has axes (time, batch, features, ...)
--- a/blocks/bricks/sequence_generators.py
+++ b/blocks/bricks/sequence_generators.py
@@ -176,7 +176,7 @@ class BaseSequenceGenerator(Initializable):
 
         # Add auxiliary variable for per sequence element cost
         application_call.add_auxiliary_variable(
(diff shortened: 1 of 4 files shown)


blocks/bricks/sequence_generators.py
```diff
@@ -136,6 +137,34 @@ def _push_allocation_config(self):
 
     @application
     def cost(self, application_call, outputs, mask=None, **kwargs):
+        """Returns sum of generation costs for output sequences.
+
+        Parameters
+        ----------
+        outputs : :class:`~tensor.TensorVariable`
+            The 3(2) dimensional tensor containing output sequences.
+            The dimension 0 must stand for time, the dimension 1 for the
```
dwf: I'd replace "The dimension" (twice) on this line with "axis", as that's what NumPy calls them.

blocks/bricks/sequence_generators.py
```diff
@@ -136,6 +137,34 @@ def _push_allocation_config(self):
...
+            position on the batch.
```
dwf: on -> in

blocks/bricks/sequence_generators.py
```diff
@@ -136,6 +137,34 @@ def _push_allocation_config(self):
...
+        mask : :class:`~tensor.TensorVariable`
+            The binary matrix identifying fake outputs.
```
dwf: Maybe specify that a 0 indicates padding.

blocks/bricks/sequence_generators.py
```diff
@@ -136,6 +137,34 @@ def _push_allocation_config(self):
...
+
+        Notes
+        -----
+        The contexts are expected as keyword arguments.
```
dwf: Can you elaborate a bit here?

I don't really know what this code does, but that makes it a good smoke test for the docstring. :)

blocks/bricks/sequence_generators.py
```diff
@@ -136,6 +137,34 @@ def _push_allocation_config(self):
...
+
+        Parameters
+        ----------
+        outputs : :class:`~tensor.TensorVariable`
+            The 3(2) dimensional tensor containing output sequences.
+            The dimension 0 must stand for time, the dimension 1 for the
+            position on the batch.
+        mask : :class:`~tensor.TensorVariable`
+            The binary matrix identifying fake outputs.
+
```
dwf: Missing a `Returns` section.

blocks/bricks/sequence_generators.py
```diff
@@ -136,6 +137,34 @@ def _push_allocation_config(self):
...
+        """
+        # Compute the sum of costs
+        cost = self.cost_matrix(outputs, mask=mask, **kwargs).sum()
+        add_role(cost, COST)
+
+        # Add auxiliary variable for per sequence element cost
+        application_call.add_auxiliary_variable(
+            (cost / mask.sum()) if mask is not None else cost,
+            name='per_sequence_element')
```
dwf: Maybe `per_sequence_element_cost`?  @bartvm @rizar You guys might be more opinionated on naming here.

blocks/bricks/sequence_generators.py
```diff
@@ -136,6 +137,34 @@ def _push_allocation_config(self):
...
+            position on the batch.
+        mask : :class:`~tensor.TensorVariable`
+            The binary matrix identifying fake outputs.
+
+        Notes
+        -----
+        The contexts are expected as keyword arguments.
+
+        """
+        # Compute the sum of costs
```
dwf: I'd ditch this comment and the one below, they are pretty uninformative.

blocks/bricks/sequence_generators.py
dwf: I just noticed that this is a direct copy of the `cost_matrix` docstring where these deficiencies originate. You can fix them if you like, but even if not, for now, I'd rather that we didn't duplicate identical documentation for two methods.

@bartvm I looked at the metadocumentation and couldn't find a final word on this. Should we handle this in `cost_matrix` with a 

```
Notes
-----
See :meth:`cost` documentation for parameters.
```
or something like that?

blocks/bricks/sequence_generators.py
dwf: [See above](https://github.com/bartvm/blocks/pull/472/files#r26462484).

blocks/bricks/sequence_generators.py
dwf: [See above.](https://github.com/bartvm/blocks/pull/472/files#r26462484)

blocks/bricks/sequence_generators.py
dwf: [See above.](https://github.com/bartvm/blocks/pull/472/files#r26462484)

blocks/bricks/sequence_generators.py
dwf: [See above.](https://github.com/bartvm/blocks/pull/472/files#r26462484)

dwf:
Nice work! The only thing I'd suggest is:

* Delete the two uninformative comments I mentioned
* Wait for @bartvm to respond to [this query](https://github.com/bartvm/blocks/pull/472/files#r26462484) about duplicated docstrings and implement his recommendation
* (Optionally) do some or all of the documentation fixes I suggested. If not, we can open another ticket.

orhanf:
OK, I'm on it. 
Another thing that I realized was this; there does not exist any information about the `AUXILIARY` variables that are added to cg by a method (i couldn't find any examples tho, if there is, then just discard the following and please inform me:) ). Actually it would be nice to also document which `AUXILIARY` variables are added to cg on the corresponding method's docstring, apart from parameters and returns. Is this unnecessary or already done? or am i missing a point?

blocks/bricks/sequence_generators.py
rizar: @dwf , it would not be fair to make @orhanf to fix my fault of barely documenting `sequence_generators.py` :)

blocks/bricks/sequence_generators.py
rizar: Agreed!

rizar:
@dwf there is already a huge ticket #200 which is unfortunately progressing very slowly, my bad.

rizar:
@orhanf I like your idea that auxiliary variable added to the CG in a method should be mentioned in the docstring of this method. You can show us a good example by doing that your code!

blocks/bricks/sequence_generators.py
```diff
@@ -136,6 +137,34 @@ def _push_allocation_config(self):
...
+        cost = self.cost_matrix(outputs, mask=mask, **kwargs).sum()
```
rizar: What if I have batches of different size? Shouldn't we divide over the number of examples in a batch here?

bartvm: No, because then you would give more weight to batches with shorter sequences. See https://groups.google.com/forum/#!topic/blocks-users/Vfl92MkXAE0

blocks/bricks/sequence_generators.py
bartvm: @orhanf Might be relying on the fact that the parameter name in the end will be `brick_application_name`, so it will probably be `sequence_generator_cost_per_sequence_element` in the end.

blocks/bricks/sequence_generators.py
bartvm: I would use
```
See Also
--------
:meth:`cost` documentation for parameters.
```

blocks/bricks/sequence_generators.py
rizar: There you discussing dividing by the mask sum. I suggest divide over the batch size, those are quite different things.

bartvm: Same argument holds for that. If your batches are of different sizes, dividing by batch size will give greater weight to smaller batches. That can be problematic if you are e.g. forcing batches to only contain sequences of equal length. Outliers (with potentially only one example per batch) will weigh very heavily then.

rizar: But if you do not normalize at all, then you have to decrease the learning rate when you increase the batch size. 

At least it has to be done uniformly for all `COST` variables in Blocks. `CategoricalCrossEntropy` and `CostMatrix` normalize costs.

blocks/bricks/sequence_generators.py
orhanf: @bartvm as you pointed out, the resulting name becomes `sequence_generator_cost_per_sequence_element`

blocks/bricks/sequence_generators.py
bartvm: That's true... Maybe normalize for batch size, but give all cost applications a keyword argument `batch_normalize=True` that allows people to turn it off?

dwf: Hmm. I didn't catch this issue but probably should've. Unfortunately I don't know what's standard with respect to RNNs. @rizar does Bart's proposal sound good to you?

rizar: It sounds good. I would be more in favour of adding unnormalized cost as an auxiliary variable than adding a keyword argument though. For sake of uniformity again, since the cost normalized per character is an auxiliary variable.

bartvm: @rizar's proposal sounds good. I guess the case of widely variable batch sizes isn't common enough to warrant the pain of having to change your learning rate every time you change batch size. @kyunghyuncho, do you normalize the cost by dividing by the batch size in GroundHog too?

kyunghyuncho: On Mon, Mar 16, 2015 at 1:53 PM, Bart van Merriënboer <
notifications@github.com> wrote:

> In blocks/bricks/sequence_generators.py
> <https://github.com/bartvm/blocks/pull/472#discussion_r26510001>:
>
> > +        Parameters
> > +        ----------
> > +        outputs : :class:`~tensor.TensorVariable`
> > +            The 3(2) dimensional tensor containing output sequences.
> > +            The dimension 0 must stand for time, the dimension 1 for the
> > +            position on the batch.
> > +        mask : :class:`~tensor.TensorVariable`
> > +            The binary matrix identifying fake outputs.
> > +
> > +        Notes
> > +        -----
> > +        The contexts are expected as keyword arguments.
> > +
> > +        """
> > +        # Compute the sum of costs
> > +        cost = self.cost_matrix(outputs, mask=mask, **kwargs).sum()
>
> @rizar <https://github.com/rizar>'s proposal sounds good. I guess the
> case of widely variable batch sizes isn't common enough to warrant the pain
> of having to change your learning rate every time you change batch size.
> @kyunghyuncho <https://github.com/kyunghyuncho>, do you normalize the
> cost by dividing by the batch size in GroundHog too?
>
Yes. The cost should be an average over a minibatch but a sum over
timesteps.

bartvm: Let's do that then, and maybe keep the `batch_normalize` keyword argument for a later PR. Let's put a warning somewhere to remind people though that this is problematic if your batches are of vastly different sizes.

dwf:
@orhanf So here's your checklist:

* [ ] [Naming fix](https://github.com/bartvm/blocks/pull/472#discussion_r26501956)
* [ ] [Normalization issue](https://github.com/bartvm/blocks/pull/472#discussion_r26483784)
* [ ] [Auxiliary variable doc](https://github.com/bartvm/blocks/pull/472#issuecomment-81558272)
* [ ] [Remove redundant comments](https://github.com/bartvm/blocks/pull/472#discussion_r26462262)

dwf:
(Be sure to comment when you've pushed new commits so I can take another look, as GitHub doesn't send out notifications for pushed commits.)

orhanf:
@dwf , I made the following changes, please see my comments below
- [ ] Naming fix
    * This was already done, name is prepended by `brick_application_name` so the resulting name is `sequencegenerator_cost_per_sequence_element`
- [x] Normalization issue
    * Taking the average cost over the minibatch, where each cost is the of sum of per token costs for a sequence.
- [x] Auxiliary variable doc
    * I think adding another section will make the docstring hard to follow so I've added few lines to the Notes section for now. I'll think on it tho, and may be another ticket covering all the `AUXILIARY` documentations can be beneficial.
- [x] Remove redundant comments

blocks/bricks/sequence_generators.py
```diff
@@ -136,21 +137,59 @@ def _push_allocation_config(self):
...
+        """
+        # Compute the sum of costs
+        cost = tensor.mean(self.cost_matrix(outputs,
+                                            mask=mask,
+                                            **kwargs).sum(axis=0))
+        add_role(cost, COST)
+
+        # Add auxiliary variable for per sequence element cost
+        application_call.add_auxiliary_variable(
+            (cost / mask.sum()) if mask is not None else cost,
```
rizar: This seems wrong, because it's assumed that `cost` is a sum, but it's already a mean.

blocks/bricks/sequence_generators.py
```diff
@@ -136,21 +137,59 @@ def _push_allocation_config(self):
...
+            Theano variable for cost, computed by summing over timesteps
+            and then averaging over the minibatch.
+
         Notes
         -----
         The contexts are expected as keyword arguments.
 
+        Adds average cost per sequence element `AUXILIARY`
+        variable to the computational graph with name
+        `~sequencegenerator_cost_per_sequence_element`
```
rizar: In Blocks variables have two names: `x.name` and `x.tag.name`. In fact the second is more important, the first is assembled from various bits and piece for debugging purposes, for instance instead of `sequencegenerator` there can be the name of a particular brick. So please write simply `per_sequence_element` here.

orhanf:
Fixed according to the comments of @rizar

blocks/bricks/sequence_generators.py
```diff
@@ -136,21 +137,57 @@ def _push_allocation_config(self):
...
+
+        """
+        # Compute the sum of costs
+        costs = self.cost_matrix(outputs, mask=mask, **kwargs)
+        cost = tensor.mean(costs.sum(axis=0))
+        add_role(cost, COST)
+
+        # Add auxiliary variable for per sequence element cost
+        application_call.add_auxiliary_variable(
+            (costs / mask.sum()) if mask is not None else costs.sum(),
```
rizar: I am afraid this is wrong again. `costs / mask.sum()` is a matrix, not scalar.

This is what happens when you do not add test all changes in your PR!

orhanf: Sorry for the sloppiness, fixed `auxiliary_variable` computation and added the corresponding tests for that as well.

tests/bricks/test_sequence_generators.py
```diff
@@ -101,18 +120,35 @@ def test_integer_sequence_generator():
...
+    assert cost.ndim == 0
+    cost_val = theano.function([y, mask], [cost])(y_test, m_test)
+    assert_allclose(cost_val, 16.0942, rtol=1e-5)
+
+    # Test 'AUXILIARY' variable 'per_sequence_element' in 'cost' method
+    cg = ComputationGraph([cost])
+    var_filter = VariableFilter(roles=[AUXILIARY])
+    aux_var_name = '_'.join([generator.name, generator.cost.name,
+                             'per_sequence_element'])
+    cost_per_el = [el for el in var_filter(cg.variables)
```
rizar: We can leave it, but just a hint: there is a class `VariableFilter` that can help your to find a variable with a given `x.tag.name`, i.e. `per_sequence_element` in this case.

rizar:
Looks good! Can you please rebase this PR (see http://blocks.readthedocs.org/en/latest/development/index.html#sending-a-pull-request for an instruction on how to do it)?

orhanf:
Okay done, any other conventions that i should follow and/or pay attention for the future ones, like stashing, merging etc (that are not [here](http://blocks.readthedocs.org/en/latest/development/pull_request.html) ).

rizar:
Nothing special except the link you gave. It is just convenient to have nice commit history.

Congats, we are done with this!

PR #473, opened 2015-03-14
Improve tutorials

fvisin:


docs/documentation.rst
```diff
@@ -0,0 +1,46 @@
...
+
+.. code-block:: bash
+
+   $ pip install sphinx sphinx-autobuild sphinxcontrib-napoleon
+
+.. tip::
+   If you don't have sudo privileges, you can use the ``-- user`` flag to 
+   install the required packages locally.
+
+Enter in the docs directory and rename the existing conf.py
```
bartvm: Why do you do this? I've always been able to build the documentation without doing any of this.

fvisin: This has to be done to generate the Makefile (and the `make.bat` file for Windows users).
An alternative is to commit them, I thought you didn't do it in the first place for a reason.

bartvm: I just double checked, and most of these instructions are actually in the current documentation already. Please have a look at https://github.com/bartvm/blocks/blob/master/docs/development/docs.rst, it should show up as a subsection "Writing and building documentation" on the "Development" page. You'll notice the use of `[docs]` (which means that we can update `setup.py` instead of the documentation) and the use of the `sphinx-build` command (which avoids having to create this `Makefile`, which is just a convenience feature of Sphinx anyway, it's not needed).

I think you could just remove this new file. Sorry!

fvisin: My bad, I looked for instructions on how to build the documentation a while ago when I first began working on the docstrings and I couldn't find it. I didn't check again yesterday before writing the new file. Luckily it didn't take me long! I removed it.

docs/tutorial.rst
```diff
@@ -52,6 +52,10 @@ means we will minimize the sum of
 
 Building the model
 ------------------
+Blocks uses "bricks" to build models. Bricks are **parametrized Theano 
+operations**. You can read more about it in the 
+:ref:`building with bricks <bricks>` tutorial.
```
bartvm: Does this link work? I think it should be `<bricks_overview>`

bartvm:
Great, LGTM except for that link that I'm not sure is correct. I'll merge after that.

docs/tutorial.rst
fvisin: Yes, it points to the anchor in https://github.com/bartvm/blocks/pull/473/files#diff-091662f3fd06a0802f6e2d6e9a3edfe0R11
It would be probably better to put the anchor at line 8 though, so that you see the subsection title. I will modify that in a minute.

fvisin: Done.

bartvm: Ah okay, missed that! LGTM now.

Issue #441, opened 2015-03-10
Code Version Extension


rizar:
We should have an extensions that saves the theano, blocks, fuel, picklable-itertools and current project revisions into status. Additionally we could save diffs, `sys,argv`,  theano, blocks and fuel configurations.

dwf:
Probably also worth keeping around the versions of `numpy` and maybe `scipy` (if it's used at all in Blocks so far), BLAS info, etc.

rizar:
Just adding to the list: CUDA version.

rizar:
Just to serve as an inspiration for the one who dares to write this extension:

```python
class CodeVersion(SimpleExtension):

    def __init__(self, packages, **kwargs):
        self.packages = packages
        kwargs.setdefault('before_training', True)
        super(CodeVersion, self).__init__(**kwargs)

    def do(self, *args, **kwargs):
        package_paths = {name: loader.path
                         for loader, name, _ in pkgutil.iter_modules()}
        for package in self.packages:
            path = package_paths[package]
            last_commit_record = "_{}_last_commit".format(package)
            git_diff_record = "_{}_git_diff".format(package)
            self.main_loop.log.status[last_commit_record] = (
                subprocess.check_output("git --no-pager log -1",
                                        cwd=path, shell=True))
            self.main_loop.log.status[git_diff_record] = (
                subprocess.check_output("git diff",
                                        cwd=path, shell=True))
```

PR #476, opened 2015-03-15
WIP: Database backend for log

### LINKED ISSUE #470: Implement SQLite3 backend for log
This idea has been floating around for a while. It was recently brought up again as a way to speed up pickling in https://github.com/bartvm/blocks/issues/442#issuecomment-78056128, but after inspecting the pickle files the overhead of the current approach seems negligible (with 80,000 entries it loads in 0.12 seconds and takes around 3MB of space in an 10+GB pickle file).

The main advantage of using SQLite, for me at least, would be the ability to analyze and plot during training. I can't unpickle any of the checkpoints, because they need the GPU (which is being used for training), and the checkpoints take up to a minute to unpickle anway.

SQLite databases should scale okay for our needs I think. It would be nice if a user could use the same database for multiple experiments (perhaps in different tables)? So that you can collect the results from a series of experiments in a single file, and plot them together easily.

bartvm:
Fixes #470.

I started to implement a log with MongoDB as a backend (as suggested by @dwf). MongoDB has a nice Python library--I'm using version 3.0b0 of PyMongo which isn't on PyPI yet if you want to try it--and its document structure fits the log quite easily. I think that the ability to host the database on an external server could be very beneficial later on e.g. when running experiments on clusters (sending results back to your computer) and for analysis (a web server or Bokeh could query the database server to read and present results, at LISA we could even run a dedicated MongoDB server for this, like we have with PostgreSQL).

I started with a rewrite of the `TrainingLog` class though because there were *a lot* of abstractions in there, which seemed to just complicate implementing a new backend. Forcing rows to be a separate object seems a bit much, and the possibility to set default values seems completely unused. I would suggest to switching to simplest data structure possible: A nested, ordered dictionary.

Considering that the log gets created by the `MainLoop` class it might be easier to delegate from a single `TrainingLog` class to different log handlers (`DefaultOrderedDict` and `MongoLog`) instead of them being their own proper classes, so that's what I did.

Performance-wise a local MongoDB database seems to do pretty well with inserting keys (approx. 300 µs). If logging to an external server turns out to be too slow, it seems we can always switch to asynchronous calls with Gevent easily.

- [x] Add docstrings
- [ ] Tests
- [ ] Support for multiple logs in one database
- [x] Change to new syntax in other files

### DIFF
--- a/blocks/log.py
+++ b/blocks/log.py
@@ -4,7 +4,11 @@ from numbers import Integral
 from operator import methodcaller
 
 from picklable_itertools import imap
-from pymongo import ASCENDING, MongoClient
+try:
+    from pymongo import ASCENDING, MongoClient
+    PYMONGO_AVAILABLE = True
+except ImportError:
+    PYMONGO_AVAILABLE = False
 
 
 class DefaultOrderedDict(OrderedDict):
@@ -141,6 +145,8 @@ class TrainingStatus(Mapping):
 class MongoTrainingLog(Mapping):
     """A training log stored in a MongoDB database."""
     def __init__(self):
+        if not PYMONGO_AVAILABLE:
+            raise ImportError('pymongo not installed')
         self.client = MongoClient()
         self.db = self.client.blocks_log
         self.entries = self.db.entries
--- a/blocks/log.py
+++ b/blocks/log.py
@@ -3,11 +3,24 @@ from collections import Mapping, OrderedDict
 from numbers import Integral
 from operator import methodcaller
 
+from picklable_itertools import imap
 from pymongo import ASCENDING, MongoClient
-from six.moves import map
 
 
 class DefaultOrderedDict(OrderedDict):
+    r"""A ordered dictionary that supports default values.
+
+    Parameters
+    ----------
+    default_factor : callable, optional
+        A callable that returns the default value when a key is not found.
+        If not given, a KeyError will be raised instead.
+    \*args
+        Interpreted as ``dict``.
+    \*\*kwargs
+        Interpreted as ``dict``.
+
+    """
     def __init__(self, default_factory=None, *args, **kwargs):
         if default_factory is not None and not callable(default_factory):
             raise TypeError('first argument must be callable')
@@ -26,6 +39,33 @@ class DefaultOrderedDict(OrderedDict):
 
 
 class TrainingLog(Mapping):
+    """The log of training progress.
+
+    A training log stores the training timeline, statistics and other
+    auxiliary information. Information is represented as nested dictionary
+    of iteration, key values.
+
+    In addition to the set of records of the training progress, a training
+    log has a status object whose attributes describe the current state of
+    training e.g. whether training has started, whether termination has
+    been requested, etc.
+
+    Parameters
+    ----------
+    backend : {'default', 'mongo', 'sqlite'}
+        The log can be stored in a variety of backends. The `default`
+        backend stores the values in a nested dictionary (a
+        :class:`DefaultOrderedDict` to be precise). This is a simple and
+        fast option, but the log is difficult to access for analysis during
+        training, and can only be saved as a pickled Python object.
+
+        The `mongo` option uses a MongoDB database server to store the
+        logs. The `sqlite` value will use a local SQLite file to store the
+        results. Both of these backends will allow you to analyze results
+        easily even during training, and will allow you to save multiple
+        experiments to the same database.
+
+    """
     def __init__(self, backend='default', **kwargs):
         self.status = TrainingStatus()
         if backend == 'default':
@@ -56,6 +96,25 @@ class TrainingLog(Mapping):
 
 
 class TrainingStatus(Mapping):
+    """The status of the training process.
+
+    By default this contains two keys: `iterations_done` and `epochs_done`.
+
+    Parameters
+    ----------
+    exclude : list, optional
+        A list of status keys (strings) that should be excluded when
+        iterating over the status items. This can be useful when you want
+        to store a particular tidbit of information that should be
+        accessible to e.g. extensions, but don't want this to show up in
+        the log at every time step.
+
+    Attributes
+    ----------
+    exclude : list
+        See the parameters section.
+
+    """
     def __init__(self, exclude=None):
         self._status = {'iterations_done': 0, 'epochs_done': 0}
         self.exclude = [] if exclude is None else exclude
@@ -74,6 +133,7 @@ class TrainingStatus(Mapping):
 
 
 class MongoTrainingLog(Mapping):
+    """A training log stored in a MongoDB database."""
     def __init__(self):
         self.client = MongoClient()
         self.db = self.client.blocks_log
@@ -83,15 +143,31 @@ class MongoTrainingLog(Mapping):
         return MongoEntry(self, key)
 
     def __iter__(self):
-        return map(methodcaller('pop', '_id'),
-                   self.entries.find(projection=['_id'],
-                                     sort=[('_id', ASCENDING)]))
+        return imap(methodcaller('pop', '_id'),
+                    self.entries.find(projection=['_id'],
+                                      sort=[('_id', ASCENDING)]))
 
     def __len__(self):
         return self.entries.count()
 
 
 class MongoEntry(Mapping):
+    """A single entry in the MongoDB-based log.
+
+    Parameters
+    ----------
+    log : :class:`MongoTrainingLog`
+        The log this entry belongs to.
+    key : int
+        The key (``iterations_done``) that this entry belongs to.
+
+    Notes
+    -----
+    Entries are their own objects so that we can read lazily from the
+    MongoDB database, so if the user just wants to write new values, it
+    won't retrieve the old values.
+
+    """
     def __init__(self, log, key):
         self.log = log
         self.key = key
--- a/blocks/log.py
+++ b/blocks/log.py
@@ -39,7 +39,7 @@ class DefaultOrderedDict(OrderedDict):
 
 
 class TrainingLog(Mapping):
-    """The log of training progress.
+    r"""The log of training progress.
 
     A training log stores the training timeline, statistics and other
     auxiliary information. Information is represented as nested dictionary
@@ -64,10 +64,16 @@ class TrainingLog(Mapping):
         results. Both of these backends will allow you to analyze results
         easily even during training, and will allow you to save multiple
         experiments to the same database.
+    status_exclude : list, optional
+        A list of status keys that should be excluded from iteration (and
+        hence, printing)
+    \*\*kwargs
+        Additional keyword arguments are passed to the log's backend e.g.
+        database configuration. See the relevant classes for details.
 
     """
-    def __init__(self, backend='default', **kwargs):
-        self.status = TrainingStatus()
+    def __init__(self, backend='default', status_exclude=None, **kwargs):
+        self.status = TrainingStatus(exclude=status_exclude)
         if backend == 'default':
             self._log = DefaultOrderedDict(dict)
         elif backend == 'mongo':
--- a/blocks/log.py
+++ b/blocks/log.py
@@ -1,15 +1,19 @@
 """The event-based main loop of Blocks."""
+import binascii
+import datetime
+import os
 import sqlite3
 from abc import ABCMeta, abstractproperty
 from collections import Mapping, OrderedDict
 from numbers import Integral
-from operator import methodcaller, itemgetter
+from operator import itemgetter
 
 import numpy
 import six
 from picklable_itertools import imap, groupby
 try:
-    from pymongo import ASCENDING, MongoClient
+    from bson.objectid import ObjectId
+    from pymongo import MongoClient
     PYMONGO_AVAILABLE = True
 except ImportError:
     PYMONGO_AVAILABLE = False
@@ -60,7 +64,11 @@ class TrainingLog(Mapping):
 
     Parameters
     ----------
-    backend : {'default', 'mongo', 'sqlite'}
+    hash_ : str, optional
+        A hash that identifies this experiment. If given and a database
+        backend is used, it will try to reload the log. If not, a new hash
+        will be created for this log.
+    backend : {'default', 'mongodb', 'sqlite'}, optional
         The log can be stored in a variety of backends. The `default`
         backend stores the values in a nested dictionary (a
         :class:`DefaultOrderedDict` to be precise). This is a simple and
@@ -72,20 +80,46 @@ class TrainingLog(Mapping):
         results. Both of these backends will allow you to analyze results
         easily even during training, and will allow you to save multiple
         experiments to the same database.
-    status_exclude : list, optional
-        A list of status keys that should be excluded from iteration (and
-        hence, printing)
     \*\*kwargs
         Additional keyword arguments are passed to the log's backend e.g.
         database configuration. See the relevant classes for details.
 
+    Attributes
+    ----------
+    info : dict
+        A dictionary with information (metadata) about this experiment,
+        such as a description of the model, the package versions used, etc.
+        It also contains the unique hash that identifies experiments (e.g.
+        in the database).
+    status : :class:`TrainingStatus`
+        This dictionary contains information on the current status of the
+        training e.g. how many iterations/epochs have been completed,
+        whether training has finished, if errors occurred, etc.
+
+    Notes
+    -----
+    Currently logs assume that entries are added in order.
+
     """
-    def __init__(self, backend='mongodb', status_exclude=None, **kwargs):
-        self.status = TrainingStatus(exclude=status_exclude)
+    def __init__(self, hash_=None, backend='mongodb', **kwargs):
+        self.status = TrainingStatus()
+        if hash_ is None:
+            hash_ = binascii.hexlify(os.urandom(12)).decode()
+        else:
+            is_valid_hash = True
+            try:
+                hash_bytes = binascii.unhexlify(hash_)
+            except TypeError:
+                is_valid_hash = False
+            if len(hash_bytes) != 12:
+                is_valid_hash = False
+            if not is_valid_hash:
+                raise ValueError('invalid hash')
+        self.info = {'hash': hash_}
         if backend == 'default':
             self._log = DefaultOrderedDict(dict)
         elif backend == 'mongodb':
-            self._log = MongoTrainingLog(**kwargs)
+            self._log = MongoTrainingLog(log=self, **kwargs)
         elif backend == 'sqlite':
             self._log = SQLiteLog(**kwargs)
         else:
@@ -115,6 +149,8 @@ class TrainingStatus(Mapping):
     """The status of the training process.
 
     By default this contains two keys: `iterations_done` and `epochs_done`.
+    This object behaves like a dictionary, except that the keys in `exclude`
+    are ignored in many cases.
 
     Parameters
     ----------
@@ -146,29 +182,59 @@ class TrainingStatus(Mapping):
         return (key for key in self._status if key not in self.exclude)
 
     def __len__(self):
-        return len(self._status)
+        return len(iter(self))
 
 
-class MongoTrainingLog(Mapping):
+class TrainingLogBackend(Mapping):
+    def __init__(self, log):
+        self.log = log
+
+
+class MongoTrainingLog(TrainingLogBackend):
     """A training log stored in a MongoDB database."""
-    def __init__(self):
+    def __init__(self, database='blocks', host='localhost', port=27017,
+                 *args, **kwargs):
         if not PYMONGO_AVAILABLE:
             raise ImportError('pymongo not installed')
-        self.client = MongoClient()
-        self.client.drop_database('blocks_log')
-        self.db = self.client.blocks_log
-        self.entries = self.db.entries
+        super(MongoTrainingLog, self).__init__(*args, **kwargs)
+
+        self.host = host
+        self.port = port
+        self.database = database
+        self.mongo_id = ObjectId(self.log.info['hash'])
+        self._connect()
+
+    def _connect(self):
+        self.client = MongoClient(host=self.host, port=self.port)
+        self.db = self.client[self.database]
+        self.experiments = self.db['experiments']
+        self.entries = self.db['entries']
+        # Create the experiment if it does not exist already
+        self.experiments.update_one(
+            {'_id': self.mongo_id},
+            {'$setOnInsert': {'created': datetime.datetime.utcnow()}},
+            upsert=True)
+
+    def __getstate__(self):
+        state = self.__dict__.copy()
+        for attr in ['client', 'db', 'entries', 'experiments']:
+            del state[attr]
+        return state
+
+    def __setstate__(self, state):
+        self.__dict__.update(state)
+        self._connect()
 
     def __getitem__(self, key):
         return MongoEntry(self, key)
 
     def __iter__(self):
-        return imap(methodcaller('pop', '_id'),
-                    self.entries.find(projection=['_id'],
-                                      sort=[('_id', ASCENDING)]))
+        return imap(itemgetter('iteration'),
+                    self.entries.find({'experiment': self.mongo_id},
+                                      projection=['iteration']))
 
     def __len__(self):
-        return self.entries.count()
+        return self.entries.count({'experiment': self.mongo_id})
 
 
 @six.add_metaclass(ABCMeta)
@@ -223,17 +289,20 @@ class MongoEntry(DatabaseEntry):
     @property
     def entry(self):
         if not hasattr(self, '_entry'):
-            self._entry = self.log.entries.find_one({'_id': self.key},
-                                                    projection={'_id': False})
-        if self._entry is None:
-            raise KeyError(self.key)
+            self._entry = self.log.entries.find_one(
+                {'experiment': self.log.mongo_id, 'iteration': self.key},
+                projection={'_id': False, 'experiment': False,
+                            'iteration': False})
+            if self._entry is None:
+                self._entry = {}
         return self._entry
 
     def __setitem__(self, key, value):
         if isinstance(value, numpy.ndarray):
             value = value.tolist()
-        self.log.entries.update_one({'_id': self.key}, {'$set': {key: value}},
-                                    upsert=True)
+        self.log.entries.update_one({'experiment': self.log.mongo_id,
+                                     'iteration': self.key},
+                                    {'$set': {key: value}}, upsert=True)
 
 
 def _grouper_to_dict(grouper):
--- a/blocks/log.py
+++ b/blocks/log.py
@@ -1,9 +1,12 @@
 """The event-based main loop of Blocks."""
+import sqlite3
+from abc import ABCMeta, abstractproperty
 from collections import Mapping, OrderedDict
 from numbers import Integral
-from operator import methodcaller
+from operator import methodcaller, itemgetter
 
-from picklable_itertools import imap
+import six
+from picklable_itertools import imap, groupby
 try:
     from pymongo import ASCENDING, MongoClient
     PYMONGO_AVAILABLE = True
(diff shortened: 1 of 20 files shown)


bartvm:
@dwf @rizar Any initial comments before I continue with this?

rizar:
Will give you feedback tomorrow morning!

blocks/log.py
```diff
@@ -1,274 +1,117 @@
...
-    def __getitem__(self, time):
-        self._check_time(time)
-        return TrainingLogRow(self, time)
 
-    @property
-    def current_row(self):
-        return self[self.status.iterations_done]
+class TrainingStatus(Mapping):
+    def __init__(self, exclude=None):
+        self._status = {'iterations_done': 0, 'epochs_done': 0}
```
rizar: There used to be also an `_epoch_ends` attribute, and it is quite useful for analysis. Do you want to get rid of it or just make the main loop responsible for initializing it?

blocks/log.py
```diff
@@ -1,274 +1,117 @@
...
-                yield attr, value
-
-
-class TrainingLog(AbstractTrainingLog):
-    """A simple training log storing information in main memory."""
-    def __init__(self):
-        self._storage = defaultdict(dict)
-        self._default_values = {}
-        self._status = TrainingStatus()
+        return map(methodcaller('pop', '_id'),
```
rizar: Hmm, wasn't it you how insisted to deprecate `map` in Blocks?

blocks/log.py
```diff
@@ -1,274 +1,117 @@
...
 
-    def set_default_value(self, key, value):
-        self._default_values[key] = value
 
-    def _add_record(self, time, key, value):
-        self._storage[time][key] = value
+class MongoEntry(Mapping):
+    def __init__(self, log, key):
+        self.log = log
+        self.key = key
```
rizar: It's confusing that you call both time and name keys, when `self.key` is time and `key` is name.

rizar:
In general it looks good. I would like to discuss the following points:

- there used to be attribute-based access to status and log rows, that now became log entries. Are you sure we want to drop it?

- a typical use-case for a stored log is to retrieve a "column", that is values of a certain quantity from all time steps. In SQL I would just do `SELECT cost FROM records`. Would MongoDB efficient at such requests?

- what about refactoring log first in one PR and adding database backend in another one? That would make it a bit easier to work with.

blocks/log.py
bartvm: It wasn't immediately clear to me what the use was, so I just kept it out of my initial draft. Not really opinionated on re-introducing it.

blocks/log.py
bartvm: Yeah, I just need to check whether using a generator here is picklable. A list comprehension won't work, because this is supposed to be lazy.

blocks/log.py
bartvm: I'll change that. I don't like `time` much though, maybe `iteration` is better, considering that's the actual time scale we use.

bartvm:
About the attribute-access: Yeah, I would drop it. Attributes in my mind are supposed to be more static. I know that Python blurs the lines between attributes and keys sometimes, but in this case I think keys are the right option. For example, Scrutinizer actually makes it an error if an attribute is assigned outside of `__init__`, and letting other objects randomly add attributes is one step up from that.

Those kind of queries should be fine for MongoDB. For very large logs it might be slightly slower than SQL-databases, but querying all the keys for a specific time step are faster in MongoDB again, so it's a bit of give and take. Either way, speed-wise both should be more than fast enough for the amount of data we have.

I can split them up eventually, but let me develop a bit further with both at the time first, so that I can make sure the interfaces match.

dwf:
About to have a look at this myself.

blocks/log.py
rizar: It's used in `blocks-plot`. I suggest to keep it so far.

bartvm: It seems like there would be more standard ways of storing and retrieving this information though. Either by saving `epochs_done` as part of the log for each time step, or by just adding a `new_epoch_started = True` to the log whenever a new epoch begins. I'm not sure why `_epoch_ends` should be treated differently if we said that everything should go into the log.

blocks/log.py
```diff
@@ -1,274 +1,117 @@
...
-
-        The iterator should yield (attribute name, attribute value) pairs.
-
-        """
-        pass
+from collections import Mapping, OrderedDict
+from numbers import Integral
+from operator import methodcaller
+
+from pymongo import ASCENDING, MongoClient
```
dwf: How difficult would it be to retain pymongo as an optional dependency? The less the better.

blocks/log.py
dwf: Shouldn't you be using `picklable_itertools.imap` in that case?

bartvm: I was scared about the overhead, giving that this could be used to request 100,000s of log entries in a query. I'll just replace it with `imap` from `picklable_itertools` for now and if it turns out to be slow I'll look at it again.

blocks/log.py
rizar: `epoch_finished=True` + a method for the log to fetch a column as a dataframe would do the job:

```
epoch_ends = log.get_column('epoch_finished').dropna().index
```

But it orthogonal to two other tasks this PR tries to solve and I think can be postponed. Small PRs are a way easier to review.

bartvm:
Still WIP, but now with basic SQLite support and docstrings.

Also updated the rest of Blocks to use the key instead of attribute interface, and to no longer rely on `None` being returned as a default which seemed dangerous to me e.g. how do you know whether the value stored was `None` or whether it wasn't present, and it can lead to obscure errors if you think you're reading a value but you're really just saving `None`. Better to make those calls explicit with `get`.

bartvm:
Working MongoDB support! Need to fix SQLite now, because I think I broke it in the process. Some things I ran into:

* MongoDB maps tuples automatically to lists. In practice I don't think is is ever much of a problem, but it broke some tests. Need to add this strange quirk to the docstring probably.
* To support storing multiple experiments I create a unique 12-byte hash for each log/experiment. These can then correspond directly to the hashes that MongoDB uses to identify documents.
* Insertion into MongoDB arrays grows linearly with their size, so they're not a very good idea for our log entries. Instead entries are now documents of their own, which should make for constant insertion time. Reading time is slightly worse this way, but insertion is definitely the bigger bottleneck in our case.

Beyond `log.status` I want to add a `log.info` dictionary, which can contain the hash, as well as the creation date, a description, etc. This should make it easy for the user to describe the experiment (e.g. the hyperparameters, the model, etc.) for future reference.

bartvm:
@rizar @dwf Just wondering if you have ideas or clean solutions about/for an issue I ran into with database backends:

* Train for 5 iterations
* Save the model to A
* Train for 5 iterations
* Load model A
* Model A will think it has 10 iterations done, because it is reading from the same database that was written to after resumption.

Two options I see:

* Copy on resumption. When training is resumed, the existing log is copied in its entirety. Relatively easy, but clearly very inefficient space wise. I guess that there are cases where this could be problematic e.g. if a model is extensively pre-trained before a hyperparameter search is performed on fine-tuning.
* On training resumption a new experiment hash is assigned, the `status` and `info` sections are copied, but the first log entry contains a `resumption_of` entry with the hash of the previous run. This allows for a complicated tree structure where you can reconstruct the log of any series of resumptions by following these links. Being able to read the log from start to finish won't be possible with a single query anymore though.

bartvm:
I guess another option is effectively creating linked lists e.g. each log entry has a parent. Could make for slow reading, but writes will still be fast, and at least we don't need a special case to handle root nodes.

dwf:
I'm missing something, where in the list of things that happened is the
resumption taking place?

bartvm:
At "Load model A" I'm resuming from the pickled file that was saved at the "Save the model A" step. To be a bit more clear:

* Train for 5 iterations
* Save the model to `A.pkl`
* Train for 5 iterations; crash happens/training gets interrupted
* Resume training by loading `A.pkl`, which has parameters trained for 5 iterations
* Model A will think it has 10 iterations done, because it is reading from the same database that was written to after resumption.

rizar:
This is an interesting issue you hit @bartvm , but I believe there is a simple solution: the `status` part has to be dumped in the pickle file. Then when you resume all new log rows generated before the crash are simply overwritten. 

I am also not big fan of having both `log.status` and `log.info`. My original idea was that `log.status` are simply those records for which iteration number is not defined. Put it differently, all the information collected during training is represented as key -> value and time, key -> value mappings. Now you want to have two `key -> value` mapping which will immediately trigger questions of "what should go where" spirit.

blocks/log.py
```diff
@@ -1,274 +1,418 @@
...
+    info : dict
+        A dictionary with information (metadata) about this experiment,
+        such as a description of the model, the package versions used, etc.
+        It also contains the unique hash that identifies experiments (e.g.
+        in the database).
+    status : :class:`TrainingStatus`
+        This dictionary contains information on the current status of the
+        training e.g. how many iterations/epochs have been completed,
+        whether training has finished, if errors occurred, etc.
+    hidden : dict
```
rizar: Hmm.. I used to think that we should go this way, but now I don't. In a good design every component has its clear purpose, and when a component has some data that by design is used only by another component, it's a bad sign. I guess this is what you call "feature envy".

In a long-term perspective it would be nice to store meta-information about the log records, but (a) similarly to roles in the computation graph their semantics should be completely transparent to the log (b) I do not think we really need it now.

blocks/log.py
```diff
@@ -1,274 +1,418 @@
...
-        The numbers of the epochs last iterations.
+    In addition to the set of records of the training progress, a training
+    log has a status object whose attributes describe the current state of
+    training e.g. whether training has started, whether termination has
+    been requested, etc.
 
-    .. todo::
+    Parameters
+    ----------
+    hash_ : str, optional
```
rizar: Why do think it should be done by the log? It can be just yet another record that the extensions from #441 can do.

rizar:
I think this is very important PR and we should finish it before very controversial #528.

bartvm:
You don't want to overwrite log rows though, since you're making the assumption here that there is always only one training run i.e. this should work too:

* Train 10 iterations, save as `A.pkl`
* Finetune `A.pkl` using algorithm #1, save as `B.pkl`
* Finetune `A.pkl` using algorithm #2, save as `C.pkl`

If you overwrite the rows as you suggest, your log of `B.pkl` is now going to be the one of `C.pkl`. It's also dangerous; if I do `blocks-continue` on the wrong file I could accidentally wipe out part of the log of a very long-running experiment.

I don't think that keeping `.status` in the pickle file is a good idea. For live monitoring it is important to know whether the training has finished or is currently in progress, else we wouldn't know when to stop querying the database for new entries for example.

I can merge `.status` and `.info`, that's not much of a problem. The "what will go where"-answer for me is: Anything that changes (can change) during a run goes to `.status`, everything that doesn't (e.g. the title of the experiment, the unique hash the database uses to identify the experiment) goes in `.info`.

blocks/log.py
bartvm: The idea is--although not entirely worked out yet--that in order to load a log stored in the database, you can instantiate it by passing the correct hash.

blocks/log.py
bartvm: Not sure if it's feature envy. Feature envy is the design pattern where one class keeps on using the methods of another class, while not having it as a subclass/handler. I guess this could count as "inappropriate intimacy"?

I'd very much like to get rid of it. I just kept it because the `Printing` module depends on some method of being able to determine what to print or not. I'd happily move that responsibility to the `Printing` extension itself somehow, because I agree that it has no place here.

rizar:
> If you overwrite the rows as you suggest, your log of B.pkl is now going to be the one of C.pkl. It's also dangerous; if I do blocks-continue on the wrong file I could accidentally wipe out part of the log of a very long-running experiment.

But this is exactly what we currently have with logs dumped to the disk and I think we should stick to this scheme for sake of simplicity. If one wants to continue an experiment, he should make a copy of the log. I read your proposals of how to do it differently and I felt that they were too complicated.

Put it differently, I don't think we should support anything new with the database backend that is impossible with in-memory backend.

> I don't think that keeping .status in the pickle file is a good idea. For live monitoring it is important to know whether the training has finished or is currently in progress, else we wouldn't know when to stop querying the database for new entries for example.

I mean not to keep `.status` in the pickle, but to dump status when checkpointing and overwrite it when resuming. I guess that would be compatible with line monitoring.

blocks/log.py
rizar: "data envy"? 

You mean what you want go get rid of leading underscores?

bartvm:
How do we currently have that? If you dump to `B.pkl` and `C.pkl` as in my example, you would have `B_log.pkl` and `C_log.pkl`, both of which contain the complete log from iteration 0 till the end. C_log.pkl` would not suddenly start overwriting the log that is stored in `B_log.pkl`, which is what would happen in the current database implementation.

blocks/log.py
bartvm: I'd have less of a problem with leading underscores now that they're keys instead of attributes (accessing "private" attributes everywhere was really bad), but in general, yes: Whether or not to print log entries is not something for the log to decide. You could imagine multiple printing/monitoring extensions (normal printing, debug printing, ncurses interface, graphing interface) all of which might have different opinions on which keys are important and not.

rizar: Okay, we can add a light-weight meta information for the log and status records. They can have roles similarly as variable currently do. One of the roles can be `PRIVATE`, this stands for the records that a particular extension (or the main loop) uses to store its state and that should be of little interest to the everything/everybody else. The `Printing` extension can ignore those by default.

rizar:
If you mean the following protocol

```
cp A.pkl B.pkl
blocks-continue B.pkl
```

then the funny thing is, it will actually overwrite `A.pkl` because the paths have not been changed. That is `blocks-continue` is not about forking, for forking some additional action needs to be taken.

blocks/log.py
bartvm: What is currently considered private e.g. whether an exception occurred or when an interrupt was intercepted, can actually be very useful if you're writing some sort of monitoring interface which wants to inform the user of the fact that an error occurred.

My point is really that these leading underscores are basically there to determine what `Printing` should print or not. That works okay now, but if/when we have many types of monitoring interfaces, it becomes problematic.

bartvm:
This additional action is very straightforward (I do it all the time). You do
```
cp A.pkl B.pkl
blocks-continue A.pkl
cp A.pkl C.pkl
blocks-continue A.pkl
```
Compared to my first example, `C.pkl` is now `A.pkl`, `A.pkl` is `B.pkl` and `B.pkl` is `C.pkl`, but whatever, the point is that forking is easy to do. And as I mentioned in https://github.com/bartvm/blocks/pull/476#issuecomment-84617961, it has important use cases too e.g. hyperparameter search for fine-tuning a pretrained model.

> Put it differently, I don't think we should support anything new with the database backend that is impossible with in-memory backend.

Although nice in principle, that doesn't really work here. Supporting the "forking" use case might require an entirely different database schema, so it would be just be creating extra work if we first do it one way and then switch entirely again in a later PR.

blocks/log.py
rizar: > What is currently considered private e.g. whether an exception occurred or when an interrupt was intercepted, can actually be very useful if you're writing some sort of monitoring interface which wants to inform the user of the fact that an error occurred.

That's very true, but one can just configure his favourite monitoring front-end to stop ignoring those. That is we should split annotation and its interpretation in exactly the same way we did in `blocks/graph.py`.

In fact (and in this I am basically citing @janchorowski) there is already a good example we can take inspiration from: Python logging packages with its levels, DEBUG, INFO, etc.. 

A random crazy idea: roles of variables monitored by `DataStreamMonitoring` and `TrainingDataMonitoring` can be added as meta-information to the log, which would allow the user to e.g. display only costs in his favourite ncurses based training progress monitor.

bartvm: Annotating log entries with a certain level makes sense, since that separates the annotation from the interpretation as you mentioned.

Implementing this requires a complete overhaul of the interface though. You can't just do `log[0].cost = 4` or `log[0]['cost'] = 4` anymore, because you need a way to annotate it somehow. The `logging` module-way of doing it would be `log.info(0, 'cost', 4)` I guess, which we could consider. I do think that it takes away from the simplicity of just having a nested dictionary as the interface though. If I log this way, it's not directly clear at all how I would go about extracting data from the log. The `logging` module doesn't have this problem, because it just writes everything to a file or screen, but for us this accessibility is very important i.e. not being able to do `[t, v['cost'] for t, v in log.items() if 'cost' in v]` is a big downside.

rizar: I do not say that we should consider `logging` interface as an example, e.g. `log.info` and and etc. We are quite different from logging because we now that a record has always the same priority level. That is a have something like the following interface in mind:

```python
class TrainingLog(object):
    ...
    def add_annotation(self, record_name, annotation):
         """Annotates a log record with `annotation`. 

If `annotation` is already there,

         pass

    def has_annotation(self, record_name, annotation):
        passs
```

rizar: Sorry, don't read the one above yet!

janchorowski: Well, log[0].cost is (log[0]).cost

Why not use the fact that setitem can take multiple parameters? Then you 
couild write log[0,priority].cost =4 where behind the scenes 
log[0,priority] would return you a view of log, linked to the log, whose 
setattr sets the priority to 4. However, you cannot put something into 
the log with multiple priorities, e.g the following must fail:

log[0,log.INFO].cost =10
log[0,log.DEBUG].cost = 11 #exception raised

also, log[0] is equivalent to log[0,some_default_level]

Since we disallow having the same value with different priorities, 
getattr is easy - it just doesn't look at the priorities, hence the 
following might be the syntax:

log[0,log.INFO].cost =10

# log[0,log.DEBUG].cost = 11 #exception raised

assert log[0, log.DEBUG].cost == 10

#and also
  assert log[0].cost == 10

rizar: Done, you can read it!

bartvm: @janchorowski I'm very much trying to get rid of attribute access because it's very non-idiomatic IMHO, but I like the idea of passing tuples as keys that can be intelligently interpreted by the `__getitem__` and `__setitem__` methods, since it allows us to keep a pretty standard interface when querying the log. So
```python
log[0]['cost', log.INFO]
```
or like @rizar suggested, we could even do something like
```python
log[0]['cost', COST]
```
where `COST` is the variable role.

rizar: @janchorowski , why should we specify the level every time we log altogether? This is redundant and we should rather do it once by a separate call like `add_annotation` from above. Also we might find it useful to annotate variables not only with the level but with something else, e.g. by the extension that produced them, so that one could later filter only values produced by a given extension.

bartvm: That said, for simplicity's sake--because storing this well in 3 different backends can quickly turn into a nightmare--I think it's worth considering keeping the annotations to string tags e.g.
```python
log[0]['cost', 'COST'] = 5
log[0]['saved_to', 'INFO'] = '/home/model.pkl'
```

bartvm: All of that said, I think the important question to ask is whether all of this added complexity justifies the increase in functionality. Saving all of these annotations in 3 different backends, making each extension declare the level and nature of their log entries, etc. is a lot of complexity for what is basically a glorified filter, so I don't think it does. If variables are half-decently named, doing something along the lines of `grep cost` should be good enough for 99% of the use cases. Let's keep the codebase small and avoid feature bloat.

janchorowski: +1

rizar: I would put it differently: this would be definitely pleasant to have in some later version, but it's a way too early for such things.

I agree that for now we can stick to name mangling of log records.

bartvm:
So a bit more thought on the linked-list idea: I think we can make the data structure consistent across backends using weak references. We'd basically end up with something like:

```python
@six.add_metaclass(ABCMeta)
class LogEntry(MutableMapping):
    pass

class InMemoryLogEntry(dict, LogEntry):
    pass

log[0] = InMemoryLogEntry({'cost': 5, 'prev_entry': None})
log[1] = InMemoryLogEntry({'cost': 4, 'prev_entry': weakref.ref(log[0])})
log[2] = InMemoryLogEntry({'cost': 3, 'prev_entry': weakref.ref(log[1])})

entry = log[2]
costs = []
while entry['prev_entry']:
    costs.append(entry['cost'])
    entry = entry['prev_entry']()
costs.append(entry['cost'])
costs = costs[::-1]
```
In the case of databases, the weak references are replaced by objects which read from the database when called e.g.
```python
class MongoDBLogEntry(LogEntry):
    def __getitem__(self, key):
        if key == 'prev_entry':
            return lambda: MongoDBLogEntry(key=self.get_prev_entry_key())
        # Return other requests by reading from the database
```

This leaves the question of how/where to store statuses in databases. I see two options:

* On resumption, copy the entire `status` dictionary and edit the new copy from thereon. That way we each pickled version has a corresponding `status` object somewhere.
* Keep a full history of the `status` instead i.e. whenever a change is made, copy the entire status object. (We'll probably want to exclude `iterations_done` and `epochs_done` in that case.)

rizar:
That would probably the most memory efficient solution in case of lots of branching experiments, but I am not yet ready to say goodbye to random access.

rizar:
> Although nice in principle, that doesn't really work here. Supporting the "forking" use case might require an entirely different database schema, so it would be just be creating extra work if we first do it one way and then switch entirely again in a later PR.

No sure what you are talking about. Seems like we do not have a fixed scheme neither in SQLite nor in MongoDB approaches.

That said, I should say that I lost the point what the main new issue with the database backends is.

bartvm:
> That would probably the most memory efficient solution in case of lots of branching experiments, but I am not yet ready to say goodbye to random access.

The in-memory solution would have random access either way. In the database random access doesn't make much sense, because all experiments/branches are saved in the same database, so doing `log[10]` could have results in several different continuations of the same experiment.

> No sure what you are talking about. Seems like we do not have a fixed scheme neither in SQLite nor in MongoDB approaches.

SQLite has schemes, they're just not strongly typed. MongoDB doesn't have fixed schemes, but nonetheless you need to decide on one, right? If I save everything as a simple array of entries now, but then decide to support multiple experiments, I'll have to rewrite everything, because it changes the way entries are read and written entirely.

> That said, I should say that I lost the point what the main new issue with the database backends is.

Main issue is this: With pickling every possible branch of your experiment (i.e. if you resume from the same pickle file twice) has its entire log stored in self-contained pickle file. In a database, different continuations of the same experiment need to save their data in the same database however, without interfering with each other. This means we can do two things: Copy all the data each time we resume an experiment, or do a more intelligent (and less memory-wasteful) kind of "branching". I'm proposing doing the latter.

bartvm:
I was hoping to spend some CCW work hours on this next week, but I'd like to hear from some people first whether they think is a useful addition, because it could make the logging framework quite a bit more complicated than it is now. A quick summary:

We would like to log directly to an SQLite or MongoDB database, allowing for

  * Continuous access to the log during training (not just whenever `Checkpoint` is called)
  * Accessing results without the need to unpickle (faster)
  * Stops extensions from having to interface with the training process directly (indirectly fixing https://github.com/bartvm/blocks/issues/496)
  * Easy comparison of logs from different experiments/you can store all the logs of a series of experiments in one centralized place

The problem with this is that resumption becomes less obvious:

* I train for 10 epochs and pickle.
* I train for 5 epochs more, and I decide to stop my experiment.
* I resume my experiment from epoch 10

Since the resumed experiment is writing to the same database, what do we do? The easy solution is that we assume that every experiment progresses linearly; we simply start overwriting the log from iteration 10 onwards. The obvious downsides are:

* You resume an experiment from the wrong pickle file, and could start overwriting your log results from days of training unknowingly.
* I might want to pre-train and then fine-tune with a variety of hyperparameters. I might not want to start overwriting results each time I resume training from the pre-trained model.
* Very unexpected behavior i.e. I save model `A`, resume from `A` to train `B`, then I resume from `A` to train `C`. Now when I load `B`, the log records are actually those from `C`!

The alternative is that the database log should allow for a branching/tree structure. Every time the user resumes from a particular point in time, the log is branched. The result is a tree structure with a different branch for each resumption. The downsides here are:

* Naively implemented as a (doubly) linked list this gives O(n) access time (i.e. O(n) database queries), which could become problematic for long-running experiments.
  * There are more intelligent data structures possible that could make this a lot faster. For example, we could store the log as an array/hash table where the first entry links to the previous log. You could easily reconstruct the full log in *O(m)* then (where *m* is the number of branches).
* An alternative approach is simply to copy the entire log whenever training is resumed. Simple, but very inefficient of course (you could be copying over a million log entries every time you resume.)

Anyway, it's a complexity vs. functionality trade-off, and I'd just like to know whether people think this functionality is important i.e. do you care about the results from each separate experiment resumption? @dwf @jbornschein @vdumoulin @dmitriy-serdyuk @pbrakel etc.

jbornschein:
I did not yet read this thread from it's beginning -- so I might be missing some important details here.

But being able to branch-off experiments from an earlier one seems to be a very useful feature that should be supported in a sane way. In academia we are trained to always "start from scratch" to make our experiments reproducible. Yet,  it's quite insightful to see what happens when you e.g. decrease your learning rate after you thought it had converged already.

So I think continuations/branches should never overwrite their ancestors. They should have a unique id and they should somehow refer to their ancestor so that their origin can be traced back.

I'm *not* convinced that each (continued) experiment has to have it's full history (including their ancestors log) homogeneously stored in the database backend. We will probably have support code/classes that will read out TrainingLogs from the database. Maybe the code that stitches together the full history could be in there.

I can imagine one situation where this approach seems inappropriate: When somebody conducts a single long-running experiment from many small segements. E.g. when trying to fill a cluster with many few-hour allocations. But I guess somebody doing this, would also be fine running some kind of 'training-log-defrag' tool that takes all the segments and stitches them together as a new single-segment experiment. Just in case that person would be interested in making it really look like it was one long-running experiment.

dwf:
I feel like the conceptually clean way to address this might be to treat
each training launch as its own separate run, whether it is resumed from a
checkpoint or not. Each run gets a UUID and jobs that were not initialized
from scratch get a non-null resumed_from tag that explains their provenance
via the previous run's UUID. It is up to log-examination client to piece
these logs together upon display, and this may include using the same seed
job when displaying multiple child jobs. Within-job logs are stored as
efficiently as possible but digging back through pre-resume history
requires a separate lookup. I can't imagine these chains growing longer
than a few dozen at most, so asymptotics aren't really worth worrying too
much about.

bartvm:
@jbornschein Completely agree.

@dwf Yes, I think that's the right way to go, and it's similar to what I had in mind when I mentioned "more intelligent data structures".

There are a few nitty gritty details, but they should be easy to address. For example, if `A` was saved at `t = 5` but continued until `t = 10` afterwards, and then `B` is resumed from `A`'s checkpoint, then `B.resumed_from = A`, but when reconstructing the full history, we only care about the entries from `A` for `t <= 5` (i.e. we have to ignore the log entries that were written to `A` after its main loop was checkpointed). That's easily solved by checkpointing not only a reference to the log, but also a reference to the current entry in the log, so that we can do `B.resumed_from = A` and `B.resumed_at = iterations_done`.

I think it's probably a good idea to try and make this work in MongoDB and SQLite simultaneously, because hiding this behind a clean interface with 3 different backends might still get a bit messy.

rizar:
Guys, can we separate resumption with a new UUID and resumption with the old one? I would not like to have a fragmented log when I stop and resume for some purely technical reason. This new resumption that you want could be called `blocks-fork` as opposed to the current name `blocks-continue`. A safety check could be done in `blocks-continue` to ensure that it does not overwrite old records but only creates new ones.

bartvm:
I think the forking should be handled at a much lower level than inside the `blocks-continue` or `blocks-fork` script, since people very often resume experiments without using those kind of scripts (e.g. because they want to edit some of the extensions, change the algorithm, etc.)

I guess it would be possible to check on resumption whether the current log has been continued after pickling, and only fork if that's the case. I wonder if it's necessary though. The forked data structure should be almost entirely transparent to the user either way, so I don't think having a fragmented log should be much of a problem if your experiment crashes a few times. The advantage is that the behaviour of always forking on resumption is more predictable and less error-prone. It's also more consistent e.g. with the pre-train and hyperparameter search example, it's strange if 1 experiment is considered a continuation, while all the others are considered forks. It might make more sense for all of them to be considered forks.

rizar:
Apparently obsolete since #710  is already in.

dwf:
Doing a branches cleanup -- @bartvm is this safe to delete?

bartvm:
Yeah, it is/was

PR #477, opened 2015-03-15
Make default bekeh server configurable via .blocksrc

jbornschein:
I think it makes sense to have thins configurable -- being forced to add an explicit ``server_url`` to the Plot extensions in all main scripts makes these scripts clearly less portable between systems.


dwf:
Looks good to me!

PR #481, opened 2015-03-16
Add parents attribute to bricks

bartvm:
This introduces a `parents` attribute to bricks which is automatically kept in sync with the children attribute. It is very useful if you have a particular variable and you want to know where it comes from. Often `get_brick` doesn't tell you more than `<blocks.bricks.Linear object at 0x7fe24f9cbb10: name=linear>`, which is utterly useless. This way you can do `get_brick(var).parents`, etc.

Right now `Children` and `Parameters` share almost everything, so I will probably merge them into the same class before merging this.

Also, maybe we should make `self.parents` explicitly read-only.

### DIFF
--- a/blocks/bricks/base.py
+++ b/blocks/bricks/base.py
@@ -30,38 +30,64 @@ def create_unbound_method(func, cls):
 property_ = property
 
 
-class Parameters(MutableSequence):
-    """Behaves exactly like a list, but annotates the variables."""
-    def __init__(self, brick, params):
+@add_metaclass(ABCMeta)
+class BrickList(MutableSequence):
+    """Like a list, but performs operations on insert/delete."""
+    def __init__(self, brick, items):
         self.brick = brick
-        self._params = []
-        for param in params:
-            self.append(param)
+        self._items = []
+        for item in items:
+            self.append(item)
 
     def __repr__(self):
-        return repr(self._params)
+        return repr(self._items)
+
+    def __eq__(self, other):
+        return self._items == other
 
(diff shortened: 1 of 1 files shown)


blocks/bricks/base.py
```diff
@@ -556,6 +582,7 @@ def __init__(self, name=None):
         self.name = name
 
         self.children = []
```
rizar: Why do we have `children` both here and as a property?

bartvm: It's to initialize the list, otherwise `self.children` would result in an `AttributeError` instead of returning an empty list. There's a dozen way to do this though, I could also do a `if not hasattr(self, '_children'): self.children = []` in the getter, but this was one line shorter :p

rizar: Not sure I get it. You use `_children` it the `children` property, and here you create another `children` attribute. And also you want `children` property always return `Children` object. Seems like there is some mismatch here.

Maybe you want `self._chidren = Children(self, [])`?

bartvm: I don't create an attribute. This calls the setter, so `self.children = []` results in `self._children = Children(self, [])` being called by the property setter. Then the next call to `self.children` will return the `self_children` attribute which is now set to a `Children` object. If I don't call the setter here, the first call to `self.children` will result in an `AttributeError`. It's better to call the setter than it is to set private attributes directly in the constructor IMHO.

rizar: Oh, now I see, sorry.

PR #482, opened 2015-03-16
Renames roles to be singular. Fixes #363.

### LINKED ISSUE #363: Rename roles to all be singular
Currently there are ```WEIGHTS```, ```BIASES```, ```FILTERS``` but on the other hand there are ```PARAMETER```, ```INPUT```, ```OUTPUT```. Is there a reason for mixing singulars and plurals?

Also ```WeightsRole``` inherits ```ParameterRole```, so it applies to class naming too.

mducoffe:


mducoffe:
@dwf done as you say

dwf:
Unfortunately between when you fetched and branched and when you sent the pull request, we merged the VariableFilter tests which use the old names.  I will quickly fix it up.

dwf:
Reopened as #483.

PR #483, opened 2015-03-16
Rename roles to singular. Fixes #363.

### LINKED ISSUE #363: Rename roles to all be singular

dwf:
Just putting the finishing touches on this.


dwf:
@bartvm Merged this without further review of my tweaks as it a straight search-and-replace and all checks passed. Hope that's kosher.

PR #485, opened 2015-03-16
Fix bug in RNG found by @vdumoulin

bartvm:


vdumoulin:
Test are passing, but coveralls seems to have stalled.

bartvm:
It does that... Ah well.

PR #486, opened 2015-03-16
WIP: Strict option for recurrent

dmitriy-serdyuk:
I started work on implementing `strict` option in `recurrent` which [is going to be introduced in theano for scan](https://github.com/Theano/Theano/pull/2622). 

Currently I'm a bit stuck, because I need to provide all the shared variables to the scan function. Seems, that it can be easily done in `recurrent` decorator. But I don't understand if it is possible to extract information about brick parameters and substitute it with the parameters from scan.

@rizar @bartvm 
Thank you for your help!

### DIFF
--- a/blocks/algorithms/__init__.py
+++ b/blocks/algorithms/__init__.py
@@ -194,8 +194,7 @@ class GradientDescent(DifferentiableCostMinimizer):
             logger.info("Taking the cost gradient")
             self.gradients = dict(
                 equizip(self.params, tensor.grad(self.cost, self.params,
-                                                 known_grads=known_grads,
-                                                 add_names=True)))
+                                                 known_grads=known_grads)))
             logger.info("The cost gradient computation graph is built")
         else:
             if known_grads:
--- a/blocks/algorithms/__init__.py
+++ b/blocks/algorithms/__init__.py
@@ -3,13 +3,16 @@ import logging
 import itertools
 from abc import ABCMeta, abstractmethod
 from collections import OrderedDict
+from six.moves import reduce
+
+from picklable_itertools.extras import equizip
 
 import theano
 from six import add_metaclass
 from theano import tensor
 
 from blocks.graph import ComputationGraph
-from blocks.utils import dict_subset, named_copy, shared_floatx, equizip
+from blocks.utils import dict_subset, named_copy, shared_floatx
 from blocks.theano_expressions import l2_norm
 
 logger = logging.getLogger(__name__)
@@ -572,6 +575,80 @@ class StepClipping(StepRule):
             (param, step * multiplier)
             for param, step in previous_steps.items())
         return steps, []
+
+
+class VariableClipping(StepRule):
+    """Clip the maximum norm of individual variables along certain axes.
+
+    This :class:`StepRule` can be used to implement L2 norm constraints on
+    e.g. the weight vectors of individual hidden units, convolutional
+    filters or entire weight tensors. Combine with :class:`Restrict`
+    (and possibly :class:`CompositeRule`), to apply such constraints only
+    to certain variables and/or apply different norm constraints to
+    different variables.
+
+    Parameters
+    ----------
+    threshold : float
+        Maximum norm for a given (portion of a) tensor.
+    axes : iterable, optional
+        An iterable collection of integer axes over which to
+        sum in order to calculate the L2 norm. If `None`
+        (the default), the norm is computed over all elements
+        of the tensor.
+
+    Notes
+    -----
+    Because of the way the :class:`StepRule` API works, this particular
+    rule implements norm clipping of the value *after* update in the
+    following way: it computes ``param - previous_step``, scales it
+    to have (possibly axes-wise) norm(s) of at most `threshold`,
+    then subtracts *that* value from `param` to yield an 'equivalent
+    step' that respects the desired norm constraints. This procedure
+    implicitly assumes one is doing simple (stochastic) gradient descent,
+    and so steps computed by this step rule may not make sense for use
+    in other contexts.
+
+    Investigations into max-norm regularization date from [Srebro2005]_.
+    The first appearance of this technique as a regularization method
+    for the weight vectors of individual hidden units in feed-forward
+    neural networks may be [Hinton2012]_.
+
+    .. [Srebro2005] Nathan Srebro and Adi Shraibman.
+       "Rank, Trace-Norm and Max-Norm". *18th Annual Conference
+       on Learning Theory (COLT)*, June 2005.
+
+    .. [Hinton2012] Geoffrey E. Hinton, Nitish Srivastava,
+       Alex Krizhevsky, Ilya Sutskever, Ruslan R. Salakhutdinov.
+       "Improving neural networks by preventing co-adaptation of
+       feature detectors". arXiv:1207.0580.
+
+    """
+    def __init__(self, threshold, axes=None):
+        axes = axes if axes is not None else ()
+        self.axes = set(axes)
+        self.threshold = shared_floatx(threshold)
+        if len(axes) != len(self.axes):
+            raise ValueError("axes must be unique")
+
+    def compute_step(self, param, previous_step):
+        if any(axis >= previous_step.ndim for axis in self.axes):
+            raise ValueError("Invalid axes {} for {}, ndim={}".format(
+                self.axes, param, previous_step.ndim))
+        if len(self.axes) == 0:
+            norms = l2_norm([param - previous_step])
+        else:
+            squares = tensor.sqr(param - previous_step)
+            norms = tensor.sqrt(
+                reduce(lambda t, a: t.sum(axis=a, keepdims=True),
+                       sorted(self.axes), squares))
+        # We want a step s* that is the same as scaling (param - previous_step)
+        # by threshold / norm when threshold < norm.
+        shrinking_step = (param -
+                          (self.threshold / norms) * (param - previous_step))
+        return tensor.switch(norms > self.threshold,
+                             shrinking_step,
+                             previous_step), ()
 
 
 class Adam(StepRule):
--- a/blocks/algorithms/__init__.py
+++ b/blocks/algorithms/__init__.py
@@ -194,7 +194,8 @@ class GradientDescent(DifferentiableCostMinimizer):
             logger.info("Taking the cost gradient")
             self.gradients = dict(
                 equizip(self.params, tensor.grad(self.cost, self.params,
-                                                 known_grads=known_grads)))
+                                                 known_grads=known_grads,
+                                                 add_names=True)))
             logger.info("The cost gradient computation graph is built")
         else:
             if known_grads:
--- a/blocks/algorithms/__init__.py
+++ b/blocks/algorithms/__init__.py
@@ -9,7 +9,7 @@ from six import add_metaclass
 from theano import tensor
 
 from blocks.graph import ComputationGraph
-from blocks.utils import dict_subset, named_copy, shared_floatx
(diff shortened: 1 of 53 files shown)


rizar:
Interesting, so it turns out that putting shared variables in `non_sequence` list does make sense, hmm, I thought it doesn't..

So far it seems to me that a proper solution would be to just fix that in Theano, but looks like it can take some time.

Temporarily we could do the following in Blocks:

- assume that only shared variables that are arguments of a recurrent application method or parameters of the brick or its children will be used in the computation graph

- with this assumption we can collect them using `Selector.get_params`, see `blocks/select.py`

- in `scan_function` we can substitute them with placeholders given by `theano.scan` using `ComputationGraph.replace`

I would be curious if it actually reduce `tensor.grad` and `theano.function` times.

bartvm:
I don't understand the placeholder and replace part. Can't we just get the parameters, and then pass them all as `non_sequences`?

It's very model dependent, but apparently it massively reduces `tensor.grad` as well as computation time in cases where you have multiple scans following each other (e.g. 4-5+ scans dependent on each other in your model). For most normal models (e.g. 1-2 scans) the advantage is muss less noticeable.

blocks/algorithms/__init__.py
```diff
@@ -194,7 +194,8 @@ def __init__(self, step_rule=None, gradients=None, known_grads=None,
             logger.info("Taking the cost gradient")
             self.gradients = dict(
                 equizip(self.params, tensor.grad(self.cost, self.params,
-                                                 known_grads=known_grads)))
+                                                 known_grads=known_grads,
+                                                 add_names=True)))
```
bartvm: `add_names` is `True` by default, no need to pass it.

rizar:
> I don't understand the placeholder and replace part. Can't we just get the parameters, and then pass them all as non_sequences?

I am not sure here about what actually `non_sequences` parameter means. In definitely adds new arguments to the scan function, but do you mean that they are the same as `non_sequences` elements?

bartvm:
`non_sequences` just takes variables that are used by your `scan` function, but that aren't looped over i.e. they aren't changed. From the Theano documentation:

> The list of non_sequences can also contain shared variables used in the function, though scan is able to figure those out on its own so they can be skipped. For the clarity of the code we recommend though to provide them to scan.

Now it turns out that this isn't just clarity of code. When there are many scans in a graph, Theano has a lot of difficulty figuring out which shared variables it needs, and it actually ends up re-computing some things that depend on shared variables. The solution to this is to pass all shared variables that are/could be used in the scan function as `non_sequences`, and pass `strict=True` to make sure Theano doesn't try to figure it out itself.

So I think something like this in the `recurrent` wrapper should be enough, no?

```python
# Find all shared variables
params = list(Selector(brick).get_params().values())
# Add them to non_sequences
non_sequences = list(contexts_given.values()) + params
# Just make `scan_function` discard them before calling `application_function`
```

rizar:
Okay, I know all that, but I also thought, possibly erroneously, that the arguments of the scan function should be used instead of `non_sequences`, those mentioned in the docs as 

```   
* all other arguments (the list given as `non_sequences` to
  scan)
```

Otherwise the scheme you proposed looks good.

bartvm:
I'm not sure I understand what you are saying. The documentation about giving variables as arguments alone, and not passing them as `non_sequences` says:

> To some extend [sic] scan can also figure out other non sequences (not shared) even if not passed to scan (but used by fn).

> One can opt to exclude variable used in fn from this list as long as they are part of the computational graph, though for clarity we encourage not to do so.

Basically, it's possible, but not recommended for clarity, and now it turns out it's also not recommended for performance.

dmitriy-serdyuk:
Great, seems, that discarding the shared variables works nice. I'll try to run my experiments and see if it improves memory consumption/speed in practice.

dmitriy-serdyuk:
I just talked to @lamblin and he said, that the scan may create a different non-sequence variable inside scan function. I'll try to investigate it.

dmitriy-serdyuk:
Well, seems, that a simple test shows, that the variables are the same:
```python
w = shared(np.ones((5, 5)))
def foo(x_in, w_in):
    assert w is w_in
    return w.dot(x_i)

x = tensor.matrix('x')
scan(fn=foo, sequences=[x], non_sequences=[w])
```

lamblin:
Just some clarification, after discussion with @kyunghyuncho: for performance, what is important is to pass as non-sequences *the immediate inputs of the loop*, if they depend only on shared variables, not the shared variables themselves.
For instance, if you have

    w = th.shared(some_matrix)
    x = some_complex_graph(w.T)
    y = tt.vector()
    z = th.scan(lambda y_i: x * y_i, sequences=[y])

you will not get any speed up by doing:

    z_bad = th.scan(lambda y_i, w: x * y_i, sequences=[y], non_sequences=[w])

what will help is doing:

    z_good = th.scan(lambda y_i, x: x * y_i, sequences=[y], non_sequences=[x])

I don't know how it translates to the current case, but trying to scrape the shared variables at the leaf nodes will probably be less helpful than passing the result of computation depending on them.

rizar:
@lamblin , thanks for clarification!

@dmitriy-serdyuk , we should already have what @lamblin suggests, there is even a warning [here](https://github.com/bartvm/blocks/blob/master/blocks/bricks/recurrent.py#L157) raised when one forgets to mention some of the contexts he uses. You can try to also add all shared variables and tell us if there is a performance difference, but I doubt that there will be any.

@bartvm , see @dmitriy-serdyuk code snippet above, I though that `assert w is w_in` not always holds, and I still don't know whether it's true. @lamblin , are the elements of `non_sequences` and the resulting additional arguments to the scan function the same variables or can they differ?

lamblin:
@rizar I am actually not sure. In general there is no guarantee (obviously the slices from `sequences` and the original sequences are not the same variable), it really depends on the current implementation of the `theano.scan` wrapper function. In particular, if the shared variable has a default update rule, I think it gets cloned or mapped to a non-shared variable.

lamblin:
Actually, [`safe_new`](https://github.com/Theano/Theano/blob/master/theano/scan_module/scan_utils.py#L37), which create a new variable, seems to be called:
- for shared variables with an update rule: https://github.com/Theano/Theano/blob/master/theano/scan_module/scan.py#L848
- for shared variables without an update rule: https://github.com/Theano/Theano/blob/master/theano/scan_module/scan.py#L917
- for non-sequences: https://github.com/Theano/Theano/blob/master/theano/scan_module/scan.py#L908

I guess that when `scan_fn(...)` gets called, the shared input variable is the same as in the output graph, but it will actually get cloned and replaced when building the graph of the inner theano function.

rizar:
@lamblin , to be honest I got even more confused from your answer :)

Let me formulate it simple: if add a single variable `a` in my `non_sequences` list, do I have to ensure that not `a` used in the scan function, but the corresponding argument for the scan function, that is the last argument of the scan function?

kyunghyuncho:
@rizar

Adding unused variables in non_sequences is find. theano.scan "pseudo-"
evaluates the inner function to grab all possible input variables, and
"only" those will be considered. With strict=True, any of these variables
found by the pseudo-evaluation must be present in any of the arguments to
theano.scan (sequences or non_sequences).

bartvm:
I think @rizar's question is slightly different. If I understand correctly, the question is whether this is allowed:

```python
a = theano.shared(numpy.random.rand(10))
def fn(b):
    return a + 1
theano.scan(fn=fn, non_sequences=[a])
```

That is, is the instance `a` outside of `scan` different from the instance `b` inside of `scan`? And does it matter which one of the two we use? The reason we want to do that is because bricks normally save their parameters as instance attributes, so they don't use the argument that was passed to the `scan` function.

bartvm:
Talking about this offline with @kyunghyuncho, he said he thinks it's fine to not use the arguments to scan, but another reference to the same object.

We need to keep in mind that RNGs are also shared variables, so when using `strict=True` I believe you strictly speaking need to pass those as `non_sequences` too. However, it is often much faster to create the noise outside of `scan` and pass it as a sequence, so I would be fine with not supporting RNGs in `strict=True` mode with a mention of this in the documentation.

blocks/bricks/recurrent.py
```diff
@@ -185,7 +187,7 @@ def scan_function(*args):
                 args = list(args)
                 arg_names = (list(sequences_given) + list(states_given) +
                              list(contexts_given))
-                kwargs = dict(equizip(arg_names, args))
+                kwargs = dict(zip(arg_names, args))
```
rizar: This is not cool to make such changes without any explanations.

rizar: Oh I got it, you want to get rid to trailing arguments. But this is not the nicest way to do it, replacing a safe zip with a completely unsafe one. You should rather do `equizip(arg_names, args[-len(shared_vars):])`.

blocks/bricks/recurrent.py
```diff
@@ -198,12 +200,22 @@ def scan_function(*args):
             outputs_info = (list(states_given.values()) +
                             [None] * (len(application.outputs) -
                                       len(application.states)))
+            if config.scan_strict:
+                # Find all shared variables
+                shared_vars = list(Selector(brick).get_params().values())
```
rizar: We can always add `shared_vars` to `non_sequences`, regardless whether strict is `True` or false.

That I suggest that Blocks always operates in strict compatible mode, the user just has to declare explicitly that he wants to take a risk and pass `strict` to Theano.

By the way, we could make `strict` a keyword argument of `recurrent_apply`, similarly to `iterate` and `return_initial_states`. That would give nice fine-grained control to the user.

rizar:
@bartvm , just a note that `SequenceGenerator.generate` has random generation inside the scan.

But if we made `strict` an argument of `recurrent_apply`, the user can always not give `strict=True` when working with this application method.

blocks/bricks/recurrent.py
bartvm: Agreed that `strict = kwargs.pop('strict', config.scan_strict)` would be nice. Don't forget to add it to the docstring in that case.

dmitriy-serdyuk: Right, it's better than introducing a new global configuration parameter.

bartvm: I think you should leave the global configuration, but use it like I
suggested in my code snippet. It'd be nice if a user can speed things up
easily by just setting one flag, while maintaining fine-grained control
when needed.

blocks/bricks/recurrent.py
```diff
@@ -180,12 +182,21 @@ def recurrent_apply(brick, application, application_call,
             for name, state in states_given.items():
                 states_given[name] = tensor.unbroadcast(state,
                                                         *range(state.ndim))
+            if config.scan_strict:
+                # Find all shared variables
+                shared_vars = list(Selector(brick).get_params().values())
+                # Just make `scan_function` discard them before calling
+                # `application_function`
+                strict = {'strict': True}
```
rizar: I do not see `strict` as a new keyword argument. Do you plan to add it?

Also there is something wrong with your branch that does not let the tests run, can you check why it can not be merged into `master`?

dmitriy-serdyuk: Yes, I do. I still haven't had a chance to check if this PR gets any improvement.

rizar:
What is the status of this PR?

dmitriy-serdyuk:
I carefully measured the speed and I didn't get any improvement. I guess, the main part of this trick is to preaggregate variables which can be aggregated outside of the scan. But we are doing it already (for linear transformation of inputs, for example). 

So, I'm closing this PR. But if you have any ideas why it doesn't work and how to fix it, I'm ready to continue my work.

PR #489, opened 2015-03-16
CCW 391: VariableFilter Interface Improvements 

### LINKED ISSUE #391: Various VariableFilter Inteface Improvements
Should only be done after #165 

- accept a single brick, not only list

- accept a list of bound applications, not only a single one

- rename `name` argument to `name_re` and introduce a new `name` which corresponds to exact matching instead of matching with a regular expressions

- add `one` keyword argument. If `one=True`, either return the found variable, or `None`, or raise an exception if several were found

mducoffe:
Fixes #391.


mducoffe:
@dwf Hi, I did not check tests for now. I wanted to know how to generate automatic documentation for the keywords I added and if the code was okay for now.

dwf:
Hey, so this is actually started from the branch you previously created and contains an old version of the #363 commit. Here's how to fix it:

```
git checkout master
git fetch upstream
git checkout -B ccw_391 upstream/master
git cherry-pick a17092c
git push --force origin ccw_391:ccw_391
```

@bartvm I think we need to document the pull request workflow somewhere.

dwf:
Great, I see you did it. Unfortunately all the checks are still failing. :( So I'd check out what's up with those by clicking on the details links at least on Travis and Scrutinizer.

dwf:
For how to build the documentation see http://blocks.readthedocs.org/en/latest/development/docs.html

blocks/filter.py
```diff
@@ -44,7 +44,8 @@ class VariableFilter(object):
     ----------
     roles : list of :class:`.VariableRole` instances, optional
         Matches any variable which has one of the roles given.
-    bricks : list of :class:`.Brick` classes or instances, optional
+    bricks : list of :class:`.Brick` classes or instances, optionnal
```
dwf: optionnal -> optional

blocks/filter.py
```diff
@@ -44,7 +44,8 @@ class VariableFilter(object):
...
+        or instance of :class:`.Brick` classes or instances
```
rizar: > or instance of :class:`.Brick` classes or instances

or :class:`Brick` subclass or :class:`Brick` instance?

blocks/filter.py
```diff
@@ -55,6 +56,7 @@ class VariableFilter(object):
         A regular expression for the variable name. The Blocks name (i.e.
         `x.tag.name`) is used.
     application : :class:`.Application` instance
+        or list of :class:`.Application` instance
```
rizar: instance -> instances

blocks/filter.py
```diff
@@ -55,6 +56,7 @@ class VariableFilter(object):
         A regular expression for the variable name. The Blocks name (i.e.
         `x.tag.name`) is used.
     application : :class:`.Application` instance
```
rizar: Could you please rename it to `application` as it supports multiple arguments now? You should also add documentation for new options, such as `name_re`, `one` and change the `name` docstring.

mducoffe:
@dwf Coverage fails so I wanted to check the coverage of blocks/filter.py but I received the following error message 'The file "blocks/filter.py" isn't available on GitHub. Either it's been removed, or the repo root directory needs to be updated' from the coverage tools. '

Is it a merging problem ?

dwf:
That is quite strange... I don't know why coveralls would be doing that.

From the root "blocks" directory, you can run  `rm -f .coverage && nosetests --with-coverage --cover-package=blocks.filter`

to run the tests and see a report. (Technically we use nose2 in order to run the doctests but I don't know how to get nose2 to do the coverage report.)

dwf:
I restarted the Travis build, maybe Coveralls will fix itself...

tests/test_variable_filter.py
```diff
@@ -43,11 +43,24 @@ def test_variable_filter():
...
     # Testing filtering by application
-    appli_filter = VariableFilter(application=brick1.apply)
+    appli_filter = VariableFilter(applications=brick1.apply)
+    variables = [cg.variables[1], cg.variables[8]]
+    assert variables == appli_filter(cg.variables)
+
+    # Testing filtering by applications
+    appli_filter = VariableFilter(applications=[brick1.apply, brick2.apply])
     variables = [cg.variables[1], cg.variables[8]]
     assert variables == appli_filter(cg.variables)
```
rizar: I've just noticed that the `one` option is not tested. Could you please test it?

Hint: `assert_raises` from `numpy.testing` will help to check that an exception was raised.

mducoffe:
@dwf, @rizar  About testing name_re and a list of applications, I proposed tests in https://github.com/mducoffe/blocks/blob/ccw_391/tests/test_variable_filter.py#L55
and https://github.com/mducoffe/blocks/blob/ccw_391/tests/test_variable_filter.py#L64
but in the coverage, self.applications as a list is never tested and no variable corresponding to the regex are found.

Any idea why ?
I will test the one keyword argument.

blocks/filter.py
```diff
@@ -44,7 +44,8 @@ class VariableFilter(object):
     ----------
     roles : list of :class:`.VariableRole` instances, optional
         Matches any variable which has one of the roles given.
-    bricks : list of :class:`.Brick` classes or instances, optional
+    bricks : list of :class:`.Brick` classes or instances,
+        or instance of :class:`.Brick` classes or instances, optional
```
bartvm: Line breaks in type descriptions don't work I think. Just stick with something like
```
bricks : list
    List of ... bla bla
    and more bla bla
```

blocks/filter.py
```diff
@@ -52,10 +53,17 @@ class VariableFilter(object):
...
         returned. ``False`` by default.
     name : str, optional
+        The variable name. The Blocks name (i.e.
+        `x.tag.name`) is used.
+    name_re : str, optional
         A regular expression for the variable name. The Blocks name (i.e.
         `x.tag.name`) is used.
-    application : :class:`.Application` instance
+    applications : :class:`.Application` instance
+        or list of :class:`.Application` instance, optional
```
bartvm: Same here.

blocks/filter.py
```diff
@@ -52,10 +53,17 @@ class VariableFilter(object):
...
         Matches a variable that was produced by the application given.
+    one : boolean, optional
```
bartvm: Maybe something like `match_exactly_one`, `match_one` or `exactly_one` would make more sense. `one` is very cryptic if you don't know the docstring.

blocks/filter.py
```diff
@@ -88,12 +96,20 @@ class VariableFilter(object):
 
     """
     def __init__(self, roles=None, bricks=None, each_role=False, name=None,
-                 application=None):
+                 name_re=None, applications=None, one=True):
         self.roles = roles
-        self.bricks = bricks
+        if not isinstance(bricks, list):
```
bartvm: The type checking should be the other way around (and can be one line shorter):
```python
if isinstance(bricks, Brick):
    bricks = [bricks]
self.bricks = bricks
```

bartvm: Just to clarify: This is for duck typing reasons. Right now if someone passes e.g. a `tuple` the code breaks.

blocks/filter.py
```diff
@@ -88,12 +96,20 @@ class VariableFilter(object):
...
+            self.bricks = [bricks]
+        else:
+            self.bricks = bricks
         self.each_role = each_role
         self.name = name
-        self.application = application
+        self.name_re = name_re
+        self.one = one
+        if not isinstance(applications, list):
```
bartvm: Same here.

blocks/filter.py
```diff
@@ -124,10 +140,17 @@ def __call__(self, variables):
...
             variables = [var for var in variables
                          if get_application_call(var) and
-                         get_application_call(var).application ==
-                         self.application]
-        return variables
+                         get_application_call(var).application in
+                         self.applications]
+        if self.one:
+            if len(variables) > 1:
+                raise ValueError("Variable Filter only return one variable")
```
bartvm: This error message isn't very grammatical, and a bit unclear. How about "variable filter matched multiple variables"?

bartvm:
A bunch of tests are failing because other parts of the code are passing `application`, which should now be applications`, or `name`, which should now be `name_re` I guess. You'll have to find these instances and fix them. This could also explain the lack in test coverage.

mducoffe:
Yeah I have corrected the tests : did have some issue with blocks and
picklable_itertools version on my computer so I could not run the tests.
I have still a problem with the regex but maybe it is because of my lack of
understanding :
when I am checking re.match with the variable name var.tag.name and let's
say W_nor* instead of W_norm it is working. But if my symbolic expression
is inside the string like W_no*m then there is no match.

Also I was curious reguarding the point of returning only one variable
because it is quite restrictive and many tests in test_variable_filter.py
returns several variable at once.

Thanks for your time

Mélanie

2015-03-19 16:41 GMT-04:00 Bart van Merriënboer <notifications@github.com>:

bartvm:
I'm not sure actually, and maybe we shouldn't use it. A user can always
just do

var, = VariableFilter(...)(vars)

which will give an error if there are multiple matches. No need to write
extra code for it IMHO.

dwf:
I don't think just a * is what you want. That's a shell wildcard but that's not the same as a regex.

W_no.*m should work. Or W_no.?m. or W_no[a-z]+m.

dwf:
@bartvm @mducoffe How do you feel about "name_regex"? I feel like the "gex" makes the meaning an awful lot more obvious.

blocks/filter.py
dwf: @bartvm There is always `collections.abc.Iterable` but I guess the inverse works better.

dwf:
@mducoffe Thanks for working so diligently on this. As far as I'm concerned, you've put in more than your fair share of time this week (CCW tickets are supposed to be 0.5 days/week), so don't worry too much if you're not quite done by tomorrow.

rizar:
@bartvm , with the `one` option it is the same as with `dict_subset`. It's okay to write like this one, two, three times, but finally this inconvenience and the fact that you have to write one more line has negative effect on both code readability and your performance. Why does `dict` have `__getitem__`, `get`, `pop` and `setdefault`? To save you the precious state, and for this very reason I think `one` option makes sense.

rizar:
Agreed with @dwf that it was a large ticket and there is no rush.

rizar:
Sorry, pressed a wrong button!

rizar:
@mducoffe , the idea was not to always return one variable, but make it possible to activate a check for unqiueness of the result by passing a flag `one=True` to `VariableFilter`.

rizar:
A hint: in the beam search implementation `blocks/search.py` the fact that `name` stood for regex matching was heavily used, that is `name` should be always replaced by `name_regex` in this while.

bartvm:
@rizar Call me clean freak, but I really dislike all those little helper funtions that do completely trivial things. They just cause code bloat, require users to learn new interfaces for actions that take less time to program than to read and understand the docstring, and most people will just ignore them, causing the library to end up with a mix of use and non-use of these keyword arguments.
```python
var, = VariableFilter(...)(vars)
f([var])
```
Is standard, more explicit and far cleaner than
```python
f(VariableFilter(..., exactly_one=True)(vars))
```
I still think we should get rid of `dict_subset` as well, I really don't see the point.

rizar:
@bartvm , from your kind permission, you are clean freak! Please email Guido van Gossum and ask him to remove `pop` from the `dict` interface and make everybody write 

```python
value = d["key"]
del d["key"]
```

It's an additional line for every single `VariableFilter` usage whose can be lots of all when you define your monitoring, it's awfully inconvenient when debugging because I can not write

```
snapshot[VariableFilter(..., one=True)]
```

and be sure that there was exactly one result returned, I have to use two lines again and again. I think people will simply always write `VariableFilter(...)(cg)[0]`, because (a) not everybody knows this unpacking idiom, I've learnt about it only when developing blocks (b) for the sake of simplicity, because an additional line is code bloat, and additional variable name is code bloat. The `one=True` are 7 characters which much are easier to put instead if `[0]` then to cut the `VariableFilter` call, put in on separate line, come up with a name for the new intermediate variable and reindent the code after all these changes.

bartvm:
Also, `get` isn't just to save keystrokes. It exists for a performance reasons: `k = d[k] if k in d else None` requires two hashing operations, whereas `get` only requires one, same for `pop` and `setdefault`.

rizar:
I am sure that it's not only for performance reason. It's because common things should be easy to do.

Maybe you will find it cleaner to add a `find` method to the `VaribleFilter` interface and act as `__call__(..., one=True)` instead of the `one` flag. That's perhaps a bit clearer interface-wise.

bartvm:
Common things are the ones that are in the standard library, and we can expect people to know these functions. You're trying to rewrite half the standard library, like dictionary comprehensions and variable unpacking, into new obscure functions and cryptic keyword arguments that will only be understandable by a handful of developers who know Blocks intimately. In the process you are making the code less readable for newcomers--your argument of not knowing variable unpacking is even more true for people not knowing the `one` keyword argument--less explicit and more prone to errors. What does `one` mean anyway? Return only the first one? Raise an error when there is more than one? Only except one input? Maybe we should change the name then?
```python
snapshot[VariableFilter(..., match_exactly_one=True)(vars)]
```
```python
var, = VariableFilter(...)(vars)
snapshot[[var]]
```
But now the second option is actually 19% shorter! Even if you stick with the cryptic `one`, you've saved 6%, at the cost of legibility. Not worth it IMHO.

(On the other hand `v = d.get(k, None)` compared to the full `if-else` block saves 57% in keystrokes, and provides a potential speedup of 2x. That's a very different story.)

rizar:
See my proposal to add `find` instead. Maybe this will allow to save time for both of us.

bartvm:
I prefer not needlessly complicating the interface. If you expect one result, you unpack it. You do the same when you filter a list, get results from an SQLite database, etc. I don't see why we should come up with special method for this, Blocks is getting bloated at high enough speed as it is IMHO.

mducoffe:
@bartvm, @rizar Two questions 
Do I change name_re to name_regex ?
Do I have to check the ValueError ( which should be a NoneError) when I try to return the first element of variable when no variable has been found ?

rizar:
It depends on how often you make such things. Fetching one particular variable from a CG is a super frequent thing you do when you specify what's monitored. And your complexity analysis is flawed, because when I fetch ten variables from a graph I have to give them reasonable names, and these reasonable names are typically at least ten characters and have to be typed twice. 

Another argument: list comprehensions. The trick you refer to can't be used in them. How do you suggest to rewrite https://github.com/bartvm/blocks/blob/master/blocks/search.py#L77 to keep it expressive and make it save? 

And finally: I understand when you stand your ground in questions like "what is Model", which are pivotal for further development of Blocks. I even see your point speaking about `dict_subset` though I am not convinced. But not here. Blocks is a collaborative effort and its development consists of compromises and attempts to serve 
broad audience. If every small change of interface which makes life of a particular part of this audience easier and does not in any way affect others will face such opposition, we will have very hard time. Adding a `find` method is a question of this addition to the code:

```python
def find(self, *args, **kwargs):
    results = self(*args, **kwargs)
    if len(results) != 1:
        raise ValueError
    return results[0]
```

It is unreasonable to struggle so hard against it as you do here, in my humble opinion.

@mducoffe , sorry that you we asked to implement something so controversial. You may remove all that relates to the `one` option, I hope I will convince @bartvm  and will add a `find` method later.

rizar:
Answering your question @mducoffe , `name_regex` is indeed better, and as I said before, you can just forget about all `one`-option related things because it has become an object of another hot discussion.

bartvm:
> And your complexity analysis is flawed, because when I fetch ten variables from a graph I have to give them reasonable names, and these reasonable names are typically at least ten characters and have to be typed twice.

That's what a `for` loop is for.

```python
matched_vars = []
for variable_name in variable_names:
    var, = VariableFilter(name=[variable_name])(vars)
    matched_vars.append(var)
```
You'll probably tell me that you prefer the 7% shorter
```python
matched_vars = [VariableFilter(name=[variable_name], one=True)(vars)
                for variable_name in variable_names]
```
but I've already given all my arguments: The principle of least astonishment, there should be only one obvious way to do it, etc.

> Blocks is a collaborative effort and its development consists of compromises and attempts to serve broad audience.

But you can turn this entirely around. If any person can choose to add a keyword whenever he wants to save 1 or 2 keyword strokes or in order to suit some particular coding style, we'll end up with a codebase that actually serves a far smaller audience.

> It is unreasonable to struggle so hard against it as you do here, in my humble opinion.

The reason I'm arguing against it is because this is a typical example of a slippery slope. It is hard to strongly disagree with seems like a single innocuous keyword argument, or a small function like `dict_subset`. However, they are just manifestations of a larger debate: At what point do we start replacing the standard library constructs with custom functions and patterns.

In this we simply disagree. I believe a strong case must be made in order to justify a new helper function or keyword argument, because of all the reasons I mentioned. This means that it needs to save decent amounts of repetitive code, improve performance measurably, or clarify things significantly. The `one` keyword argument and `dict_subset` function don't improve performance, they are minimally shorter, and they are less readable simply by virtue of being custom replacements for constructs that already largely exists.

rizar:
> You'll probably tell me that you prefer the 7% shorter

It's not only about percents, it's about clarity. List comprehension is a pure syntax sugar, the only benefit is not to have an additional buffer. You can even use "one obvious way" to fight it, but it's there because syntax sugar is very important. Email Guido and ask him to remove it, because people that people coming from C++ don't know how to read this and prefer `for` loops. 

> If any person can choose to add a keyword whenever he wants to save 1 or 2 keyword strokes or in order to suit some particular coding style, we'll end up with a codebase that actually serves a far smaller audience.

@bartvm , general statements like this are false. Every case should be considered on individual basis. There are `str.index` and `str.find`, there is `as_dict` argument that we added to `Application`, people do add sometimes convenience methods. Case by case the decision has to be taken if the feature is needed to sufficiently many people to integrate it into the main code. I think that almost everybody using Blocks comes across `VariableFilter`, and I think that in at least a half of cases a single variable is retrieved, which needs some sort of check, and I find it perfectly okay to let them have their check  done by choosing the method they call. Please note, I do not offer a keyword argument any more, I offer another method `find`, which you ignored in your previous post, I don't know deliberately or not. 

> The reason I'm arguing against it is because this is a typical example of a slippery slope. It is hard to strongly disagree with seems like a single innocuous keyword argument, or a small function like dict_subset. However, they are just manifestations of a larger debate: At what point do we start replacing the standard library constructs with custom functions and patterns.

“Today he is playing jazz and tomorrow he’ll betray the Motherland.” - Soviet era saying.

As a person that grew up in a post-soviet country I bear an allergy against any ideology, and this was an ideological statement. We are not going to replace anything from the standard library, but we should make interfaces of our classes convenient. Forget about `dict_subset`, that was a bad idea to refer to it protect this change, which is times, times more innocent. 

> In this we simply disagree. I believe a strong case must be made in order to justify a new helper function or keyword argument, because of all the reasons I mentioned. This means that it needs to save decent amounts of repetitive code, improve performance measurably, or clarify things significantly. The one keyword argument and dict_subset function don't improve performance, they are minimally shorter, and they are less readable simply by virtue of being custom replacements for constructs that already largely exists.

It safes decent amount of repetitive code, because one line is often much more desirable than two. It will help to make `BeamSearch` implementation cleaner. 

If you are still not fine with adding the method `find` to the `VariableFilter` class in the way proposed above, the only option I see is that I prepare a PR and we make a poll with noticeable Blocks contributors as voters, 
in which they express their opinion if the feature I propose is generic enough to be part of Blocks.

bartvm:
> this was an ideological statement

I actually isn't, at all. The entire argument I'm making is that on a case-by-case basis there should be a cost-benefit analysis (performance, code repetition, clarity). That is the opposite of an ideological statement.

> It safes decent amount of repetitive code, because one line is often much more desirable than two.

I do not see why. In Python the full `if-else` block is often preferred over the "unpythonic" ternary operator because it's considered clearer, even though it's 4 lines instead of one.

> There are str.index and str.find, there is as_dict argument that we added to Application

Yes, and `as_dict` is an argument that I would never have introduced myself, but I can see its benefit, and I am happy to compromise, as you suggested.

> The only option I see is that I prepare a PR and we make a poll with noticeable Blocks contributors as voters

Sure, but we should do this anyway. It's not just the two of us, and I'm sure there are many other opinions out there.

rizar:
Fine with me. Anyway at some point we will need a clear mechanism of taking democratic decisions, so this will serve as a nice reason to invent it.

dwf:
@rizar @bartvm Any idea what the heck happened to the history here? I think we broke GitHub.

bartvm:
No idea... This branch has been merged into itself and somehow picked up commits from 10 other people along the way, poor GitHub is clearly confused. Looks like `master` got merged without being fast-forwarded into this branch? If that's the case resetting your `master` and rebasing should work though, @mducoffe did you try that?

mducoffe:
@bartvm @dwf  I applied a rebase, are there still things that need to be modified ?

dwf:
There are failing tests :(

dwf:
Also a rather significant decrease in test coverage %...

mducoffe:
Yeah I know. Just because some programs are using name as name_regex.

2015-03-25 15:14 GMT-04:00 David Warde-Farley <notifications@github.com>:

mducoffe:
Hi tests are fixed but the corevage has decreased without any causes being
related to my pull request. Any idea ?

2015-03-25 15:15 GMT-04:00 Mélanie Ducoffe <melanieducoffe@gmail.com>:

bartvm:
https://github.com/bartvm/blocks/pull/468#issuecomment-81171808

dwf:
Replaced by #562.

PR #491, opened 2015-03-16
Add expression for Hessian times vector

bartvm:


bartvm:
Wondering whether anyone thinks this is useful enough to add. I always end up looking at [the Theano documentation](http://deeplearning.net/software/theano/tutorial/gradients.html#hessian-times-a-vector) when I need this, so I figured a helper function could be handy, but I don't know how many people use Hessians for 2nd order stuff?

@dwf @rizar @dmitriy-serdyuk

rizar:
I am in favour of merging it.

PR #494, opened 2015-03-17
Add unit test for PR #485 (Random brick bug)

vdumoulin:
I figured this would make a useful unit test.


vdumoulin:
@bartvm This should be very quick to review.

PR #495, opened 2015-03-17
Fix ConvolutionalLayer batch size bug

vdumoulin:
The batch size was not pushed to the `ConvolutionalLayer`'s convolution and pooling bricks.


vdumoulin:
@rizar Would you mind reviewing this? It should be relatively straightforward.

vdumoulin:
Thanks!

rizar:
I am not a guru in convolutions at all, but it looked pretty clear what you wanted to fix and how you did it. You're welcome!

Issue #462, opened 2015-03-12
LSTM and GatedRecurrent interface harmonization


bartvm:
Would be nice if they take `dim` and `activation` in the same order so that code can use either of them transparently.

dwf:
@orhanf This one's for you this week!

vikkamath:
Isn't this issue closed? (Refer PR #497)

rizar:
It should be, thanks!

PR #497, opened 2015-03-17
CCW ticket issue #462 changes

orhanf:
LSTM and GatedRecurrent `dim` and `activation` parameter passing harmonized.
For GatedRecurrent default value for `activation` is also set to `Tanh`.
Unit tests changed respectively.

### DIFF
--- a/blocks/bricks/recurrent.py
+++ b/blocks/bricks/recurrent.py
@@ -439,14 +439,14 @@ class GatedRecurrent(BaseRecurrent, Initializable):
 
     Parameters
     ----------
+    dim : int
+        The dimension of the hidden state.
     activation : :class:`.Brick` or None
         The brick to apply as activation. If ``None`` a
         :class:`.Tanh` brick is used.
     gate_activation : :class:`.Brick` or None
         The brick to apply as activation for gates. If ``None`` a
         :class:`.Sigmoid` brick is used.
-    dim : int
-        The dimension of the hidden state.
     use_upgate_gate : bool
         If True the update gates are used.
     use_reset_gate : bool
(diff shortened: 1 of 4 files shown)


blocks/bricks/recurrent.py
```diff
@@ -439,9 +439,10 @@ class GatedRecurrent(BaseRecurrent, Initializable):
 
     Parameters
     ----------
-    activation : :class:`.Brick`
-        The brick to apply as activation.
-    gated_activation : :class:`.Brick` or None
+    activation : :class:`.Brick` or None
```
rizar: Can you please permute the entries of `Parameters` section in the same order as argument of the `__init__` method?

tests/bricks/test_recurrent.py
```diff
@@ -143,14 +143,17 @@ def sigmoid(x):
...
+        self.gated = GatedRecurrent(dim=3,
+                                    activation=Tanh(),
+                                    gate_activation=Tanh(),
+                                    weights_init=Constant(2))
         self.gated.initialize()
-        self.reset_only = GatedRecurrent(
-            dim=3, weights_init=IsotropicGaussian(),
-            activation=Tanh(), gate_activation=Tanh(),
-            use_update_gate=False, seed=1)
+        self.reset_only = GatedRecurrent(dim=3,
```
rizar: Although technically PEP8 allows that, please do not use this Groundhog style indentation, that is one line per one keyword argument. It takes too much space and we in Blocks sort of don't do it. Sorry for being picky!

rizar:
I have only 2 comments and then we can merge!

tests/bricks/test_recurrent.py
orhanf: It's okay, good to know the convention. Fixed it.

orhanf:
fixed according to @rizar comments

rizar:
Very nice, thanks!

PR #498, opened 2015-03-18
Fixing recurrent

laurent-dinh:
Just an alternative to fixing the documentation.

### DIFF
--- a/blocks/bricks/recurrent.py
+++ b/blocks/bricks/recurrent.py
@@ -184,8 +184,10 @@ def recurrent(*args, **kwargs):
 
             def scan_function(*args):
                 args = list(args)
-                arg_names = (list(sequences_given) + list(states_given) +
-                             list(contexts_given))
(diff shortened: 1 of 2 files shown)


rizar:
In general I like the idea, but code is convoluted. Seems like the code above is equivalent

```python
outputs_info = [
    states_given[name] in name in application.states
    else None 
    for name in application.outputs]
```

laurent-dinh:
Yes, I just replaced it. Also changing the (`scan_function`)[https://github.com/bartvm/blocks/blob/master/blocks/bricks/recurrent.py#L185] so as to take into account the order.

rizar:
Okay. Can you please turn your gist into a test?

rizar:
Ping! You have a test, just add it to `tests/bricks/test_recurrent.py` and that would be it.

laurent-dinh:
Done.

rizar:
Thanks!

PR #504, opened 2015-03-18
add __repr__ method in initialization.py

yingzha:
This method prints the object like this : instance_name(attribute_name=attribute_value), I'm not sure if my code here is compact enough, if not, please correct me. :)

### DIFF
--- a/blocks/initialization.py
+++ b/blocks/initialization.py
@@ -46,6 +46,14 @@ class NdarrayInitialization(object):
             shape = var.get_value(borrow=True, return_internal_type=True).shape
         var.set_value(self.generate(rng, shape))
 
+    def __repr__(self):
+        """Printable representation of the object.
+
+        """
+        attribute_items = ['{}={}'.format(item[0], item[1])
+                           for item in self.__dict__.items()]
+        return "{}({})".format(type(self).__name__, ','.join(attribute_items))
+
 
 class Constant(NdarrayInitialization):
     """Initialize parameters to a constant.


bartvm:
Please have a look at https://github.com/bartvm/blocks/blob/master/blocks/utils.py#L373-L406

You probably don't want to print everything blindly, but this will allow you do something like `repr_attrs(self, ['std', 'mean'])`, etc.

yingzha:
I'm thinking is it proper to name the method as '__repr__' in initialization.py? Because it selectively print the attribute you enter, if that's the case, why not use repr_attrs directly from utils.py?

bartvm:
I'm not sure what you mean? My suggestion is this:

```python
def __repr__(self):
    repr_attrs(self, 'std', 'mean')
```
The reason we want to use this is because when you do
```python
init = IsotropicGaussian()
print(init)
```
it calls `__repr__` in the background.

yingzha:
OK, I just call the method like this in the  class NdarrayInitialization
     def __repr__(self):
         """Printable representation of the object with attributes.
 
         """
         return '{}'.format(repr_attrs(self, *self.__dict__.keys()))
it's the easiest way to do so, but the printing doesn't look good due to the prefix of the attribute, I'm not sure if I need to add the method to each class, what's your opinion?

bartvm:
Ah, I see what you mean. Sorry, I didn't understand the problem at first. I guess we could update `repr_attrs` to take an optional second argument, which contains alternative names for the attributes we want to print e.g. `repr_attrs(['_mean', '_std'], ['mean', 'std'])`. However, it seems like there are some tricky cases like `Constant` (which is going to print a NumPy array, which will look really messy). Mm... Let me think about it!

rizar:
I would rather go this way: in addition to accepting attribute names, `repr_attrs` can also accept (key, value) tuples of additional attributes to be rendered. Then repr's can be written like that:

```python
# for IsotropicGaussian
def __repr__(self):
    return repr_attrs(self, ('mean', self._mean), ('std', self._std))

# for Constant
def __repr__(self):
    contant = self._constant.item() if self._constant.ndim == 0 else self._constant
    return repr_attrs(self, ('constant', constant))
```

bartvm:
You'll need something slightly more robust than that though. If for
whatever reason the initialisation fails, the error message will try to
represent the object, but then crash again because those attributes in
__repr__ aren't available.

rizar:
`_mean`, `_std`, `_constant` are all set in the constructor which should not fail under any circumstances, because it only contains trivial assignments.

bartvm:
Doing things as simple as setting a breakline and checking `!self` in `pdb` at the start of `__init__` would still break it, and so would subclassing it and overwriting some attributes, etc. The reason these bugs are nasty is because they will go completely unnoticed until you do `print(obj)` at one point, and *then* suddenly things crash. The reason I wrote `repr_attrs` to be more robust is exactly for that reason: For a long time bricks would actually crash with obscure errors when you tried to print them during debugging.

rizar:
I see your point. Then we can go the original way you proposed and tolerate the fact that `repr(Constant(0))` will contain `numpy.array(0)`. Or even have `_mean` and `_std` in the repr string, I would not care too much because still our repr strings are not interpretable by `eval`. And also I do not think those should not be private attributes in the first place. It seems this PR can be done by simply making all attributes of initialization schemes public and using `repr_attrs`.

rizar:
Why did you close this one?

bartvm:
Because it was 4 lines of code that seemed unlikely to be used, haven't been touched in 2 months, and there's a corresponding ticket open (https://github.com/mila-udem/blocks/issues/463).

rizar:
Okay, I guess I will finish later, since I need this feature.

Issue #502, opened 2015-03-18
Initial feedback should be (optionally) zero


bartvm:
CC @sebastien-j

Right now the code in `BaseSequenceGenerator.cost()` sets the initial feedback as a function of `readout.initial_outputs()`. When using a `SoftmaxEmitter` this means that the initial feedback is basically the word embedding for the token with index 0. There are two problems with this:

* For the encoder-decoder translation model we we want the actual feedback to be zero i.e. we don't want to do a lookup at all. This should be supported.
* If we are hardcoding 0 as a beginning-of-sentence token, we need to make this very clear in the documentation. In all of my experiments I am using 0 as the unknown-token index, so the initial feedback the decoder gets is `UNK`, which I imagine can be quite counterproductive.

rizar:
It's very easy to fix in the `cost` method, but slightly harder in `generate`. The only 100% solution I can think is to track the step number and not use feedback when it is zero. 

Or another way: it seems reasonable for me to say that `initial_outputs` of a `SoftmaxEmitter` should return -1. Then the `LookupFeedback` can return zeros when its input is negative. This behavior is a bit shady, but can be activated with a flag, in which case it seems okay to me.

I vote for the second option, let me know if you have another ideas.

kyunghyuncho:
What I usually do when generating from these types of models is
```
    # if it's the first word, emb should be all zero
    emb = tensor.switch(y[:,None] < 0, tensor.alloc(0., 1, tparams['Wemb_dec'].shape[1]), 
                        tparams['Wemb_dec'][y])
```
As @rizar said, this should be on a lookup module which should anyways work only with a non-negative integer.

bartvm:
But these solutions effectively hardcode the assumption that we're outputting labels. What's wrong with just replacing [these lines](https://github.com/bartvm/blocks/blob/master/blocks/bricks/sequence_generators.py#L181-L182) with something like `zeros_like(...)`?

kyunghyuncho:
How about a kind of strict mode at ```LookupFeedback```? 

After all, if the input to ```LookupFeedback``` is a negative integer (and it is defined to be an integer), there will be an error raised. Instead, if ```LookupFeedback``` instance is initialized with, say, ```strict=False```, it simply returns ```tensor.alloc(0., outputs_shape[0], self.num_outputs)```.

bartvm:
[...]

PR #511, opened 2015-03-19
Configurable initial outputs for SoftmaxEmitter

rizar:


blocks/bricks/sequence_generators.py
```diff
@@ -440,7 +440,15 @@ class SoftmaxEmitter(AbstractEmitter, Initializable, Random):
...
     Interprets readout elements as energies corresponding to their indices.
 
+    Parameters
+    ----------
+    initial_output : int or a scalar :class:`~theano.Variable`
+        The initial output.
+
     """
+    def __init__(self, initial_output):
+        self.initial_output = output
```
bartvm: Forgot to pass `**kwargs`

blocks/bricks/sequence_generators.py
```diff
@@ -440,7 +440,15 @@ class SoftmaxEmitter(AbstractEmitter, Initializable, Random):
 
     Interprets readout elements as energies corresponding to their indices.
 
+    Parameters
+    ----------
+    initial_output : int or a scalar :class:`~theano.Variable`
+        The initial output.
```
bartvm: This is not very easy to understand without reading the code. I would say something like "The label (token) to output for the first time step of the :class:`.SequenceGenerator`. This corresponds to the 'beginning-of-sentence' token."

bartvm:
I think this is a reasonable workaround for now. It makes sampling from the old GroundHog models possible. Training them won't be possible, but we can leave that till we have a proper solution as discussed in #502

rizar:
I think there is an easy way to implement `zero_init_feedback`, please see my post in the issue. If you like, I would close this PR.

kyunghyuncho:
This will be useful.

bartvm:
@rizar Even with `zero_init_feedback`, I still think this is a useful PR. In some cases you might want to learn a beginning-of-sentence token, but don't want to be forced to use 0 for this e.g. because the data you are training one already uses this number for another token.

rizar:
Okay, then I will finish it!

blocks/bricks/sequence_generators.py
rizar: Although I understand that documentation is insufficient here, the proper place for it would be the `AbstractEmitter` documentation. It also will have to play nicely with all the rest of documentation, which in turn will change significantly after #522. Taking it all into account, I would not touch it here for now.f

PR #512, opened 2015-03-19
Fix using super in an application that overrides a recurrent one

rizar:


bartvm:
LGTM, makes the code clearer too.

PR #513, opened 2015-03-19
WIP: Batch normalization

### LINKED ISSUE #509: Batch normalization
As presented in [this paper](http://arxiv.org/pdf/1502.03167v3.pdf).

It would make a great feature for fully-connected as well as convolutional nets.

I think it would make sense to implement it as a wrapper class around `Activation` subclasses.

I intend on using it pretty soon, so I'd volunteer to implement it. Any thoughts, @dwf @bartvm @rizar ?

vdumoulin:
Fixes #509.


blocks/graph.py
```diff
@@ -520,3 +521,78 @@ def apply_dropout(computation_graph, variables, drop_prob, rng=None,
...
+    >>> cg = ComputationGraph(y)
+    >>> inputs = VariableFilter(
+    ...     roles=[INPUT],
+    ...     bricks=linear.linear_transformations)(cg.variables)
+    >>> cg_bn = apply_batch_normalization(
+    ...     cg, inputs, [gamma_1, gamma_2], [beta_1, beta_2])
+    >>> fprop = function(cg.inputs, cg.outputs[0])
+    >>> bn_fprop = function(cg_bn.inputs, cg_bn.outputs[0])
+    >>> linear.initialize()
+    >>> print fprop(numpy.ones((3, 2), dtype=theano.config.floatX))
```
bartvm: Docstrings are Python 3, so use `print()`

blocks/graph.py
```diff
@@ -520,3 +521,78 @@ def apply_dropout(computation_graph, variables, drop_prob, rng=None,
...
+    -----
+    For more information, see [BN]_.
+
+    .. [BN] Sergey Ioffe and Christian Szegedy. *Batch Normalization:
+       Accelerating Deep Network Training by Reducing Internal Covariate
+       Shift*, arXiv:1502.03167.
+
+    Examples
+    --------
+    >>> import numpy
```
bartvm: I think the imports from the file are available in the docstring namespace, so you can make it a bit less verbose by leaving out e.g. `import numpy` and `import theano`

vdumoulin: Comment addressed (as well as your `print()` comment).

blocks/graph.py
```diff
@@ -4,15 +4,16 @@
...
-from theano import Variable
+from theano import Variable, tensor
 from theano.gof import graph
 from theano.sandbox.rng_mrg import MRG_RandomStreams
 from theano.scan_module.scan_op import Scan
 from toolz import unique
 
 from blocks import config
-from blocks.roles import add_role, has_roles, AUXILIARY, PARAMETER, DROPOUT
+from blocks.roles import add_role, has_roles, AUXILIARY, PARAMETER, DROPOUT, BN
```
dwf: `BN` is a little cryptic...

dwf: How about `BATCH_NORMALIZED`?

blocks/graph.py
```diff
@@ -520,3 +521,82 @@ def apply_dropout(computation_graph, variables, drop_prob, rng=None,
...
+    -----
+    For more information, see [BN]_.
+
+    .. [BN] Sergey Ioffe and Christian Szegedy. *Batch Normalization:
+       Accelerating Deep Network Training by Reducing Internal Covariate
+       Shift*, arXiv:1502.03167.
+
+    Examples
+    --------
+    >>> from theano import function
```
dwf: Some commentary in between chunks of this example code would make them more useful.

blocks/graph.py
```diff
@@ -520,3 +521,82 @@ def apply_dropout(computation_graph, variables, drop_prob, rng=None,
...
+
+def apply_batch_normalization(computation_graph, variables, gammas,
+                              betas, axis=0, epsilon=1e-7):
+    """Returns a graph to variables in a computational graph.
+
+    Parameters
+    ----------
+    computation_graph : instance of :class:`ComputationGraph`
+        The computation graph.
+    variables : list of :class:`~tensor.TensorVariable`
```
dwf: I guess my first question is, is there a simple way to infer which variables in a graph need to be batch-normalized? Even if we don't do it automagically, some guidance as to this point might be worth putting in the docs.

blocks/graph.py
vdumoulin: Sounds good.

blocks/graph.py
dwf: Re: my comment about "how to use batch normalization" above, in between lines of example code is the perfect place for it.

blocks/graph.py
vdumoulin: Reading the paper, I don't know.

For fully-connected networks what I understand is that they batch-normalize post-nonlinearity activations (this is what I read as "activations", but I might be wrong).

For convnets the authors explicitly say that they batch-normalize the post-convolution, pre-nonlinearity activations.

I could put general recommendations in the documentation (after validating my understanding) but I wouldn't go as far as making assumptions for the user.

blocks/graph.py
```diff
@@ -520,3 +521,82 @@ def apply_dropout(computation_graph, variables, drop_prob, rng=None,
...
+     [ 2.  2.]
+     [ 2.  2.]]...
+
+    """
+    epsilon = numpy.cast[theano.config.floatX](epsilon)
+
+    means = [var.mean(axis=axis, keepdims=True) for var in variables]
+    variances = [var.var(axis=axis, keepdims=True) for var in variables]
+    replacements = [
+        (var, gamma * (var - mu) / tensor.sqrt(sigma_sqr + epsilon) + beta)
```
dwf: Putting epsilon inside of the sqrt magnifies its effect (as in, the square root of a small constant less than one is a larger constant). Is this what you intended?

blocks/graph.py
dwf: Yeah, I wouldn't go so far as making assumptions either. I will read the paper myself after finishing my ICML reviewing stack (argh).

blocks/graph.py
vdumoulin: That's what's in the paper (see Algorithm 1). Do you think it's a typo? From my point of view it looks like it is.

blocks/graph.py
vdumoulin: I'll address that when the interface converges to something. Right now I'm not even sure it'll stay as it is.

vdumoulin:
One of the things about which I'd like to know your opinion: is it alright to expect that the user provides the `gamma` and `beta` parameters?

I chose to go that route because it offers more flexibility: the user can choose how to initialize these parameters, if they're to be learned, etc.

Plus I could add some `normalize_batch` flag (defaulting to `True`) which, when set to `False`, just multiplies by `gamma` and adds `beta`. That would allow the user to re-use `apply_batch_normalization` at test time with the population statistics rather than the batch statistics.

blocks/graph.py
dwf: Well if the typical case is that sigma squared is rather large then it doesn't really matter. I'd just go with what the paper does.

dwf:
Leaving this open in a tab so I have a look at it first thing in the morning.

dwf:
Apparently my tab strategy did not pan out. Looking now.

blocks/graph.py
fvisin: > For fully-connected networks what I understand is that they batch-normalize post-nonlinearity activations (this is what I read as "activations", but I might be wrong).

I agree that it's not clear what they mean by that in the paper, but I reached your same conclusions basing on:
* "The scaled and shifted values _y_ are passed to other network layers." (page 4, top left, where _y_ is the output of the BN) 
* "Any layer that previously received _x_ as the input, now receives BN(_x_)" (page 4, bottom left)
* Point 4 of algorithm 2

Nonetheless, the argument they make to justify the use of the transformation before the non-linearity in the convolutional networks section seems to be valid in the fully-connected case too (since applying BN to the post-nonlinearity activation would the same as applying it to the input of the next layer, if I am not confused):
> We add the BN transform immediately before the nonlinearity, by normalizing x = W u + b. We could have also normalized the layer inputs u, but since u is likely the output of another nonlinearity, the shape of its distribution is likely to change during training, and constraining its first and second moments would not eliminate the covariate shift.

blocks/graph.py
fvisin: From the paper:
> epsilon is a constant added to the mini-batch variance for numerical stability

I guess the only thing that matters here is not to divide by zero, so both putting the epsilon inside or outside the sqrt will do.

blocks/graph.py
fvisin: In the example you should also import tensor and ComputationGraph

bartvm: Doctests are executed in the same namespace as the file I think, so `tensor` is already available. I prefer not importing everything again in the docstring, because it turns docstrings into 50% import statements otherwise.

blocks/graph.py
```diff
@@ -526,3 +529,96 @@ def apply_dropout(computation_graph, variables, drop_prob, rng=None,
...
+    ...                dtype=theano.config.floatX))) # doctest:+ELLIPSIS
+    [[ 2.  2.]
+     [ 2.  2.]
+     [ 2.  2.]]...
+
+    """
+    epsilon = numpy.cast[theano.config.floatX](epsilon)
+
+    # Broadcast gamma and beta properly
+    axis = axis if isinstance(axis, (list, tuple)) else (axis,)
```
fvisin: Since it is a list of axis, I would name it `axes` throughout the code.

blocks/graph.py
```diff
@@ -526,3 +529,96 @@ def apply_dropout(computation_graph, variables, drop_prob, rng=None,
...
+    mappings = [
+        dict([(d, i) for i, d in
```
fvisin: For the sake of readability I would rename both `d` and `a` to `axis`. (see also comment above)

blocks/graph.py
dwf: +1. Also I think you can just use `blocks.utils.pack` here.

vdumoulin:
@dwf @fvisin Your comments have been addressed, thanks!

vdumoulin:
I'm getting a syntax check error because I'm redefining the `axis` argument within list comprehension.

I really want to keep that name for the argument, as it's consistent with `theano` and `numpy`, so I'll change `axis` to `ax` within the code.

bartvm:
I guess that a (very) loose interpretation of PEP8 would suggest you use `axis_` instead of `ax`.

vdumoulin:
Close enough, I'll do that.

vdumoulin:
I think I could also write a short tutorial in the documentation to show how to use batch normalization with a deep network on MNIST.

I have a working example where an MLP with 4 sigmoid layers followed by a softmax layer fails to learn by itself but has no problem training when using batch normalization.

dwf:
Hmm, VariableClipping uses axes, I did that since it accepts multiple axes.
Should I change it for consistency?

vdumoulin:
I think we should stay consistent with theano and numpy when appropriate. I'd change `axes` for `axis` in `VariableClipping`.

dwf:
Does Theano accept multiple axes for its axis arguments though? I'm pretty
sure numpy doesn't.

vdumoulin:
It seems like it does:

```python
numpy.mean(numpy.random.uniform(size=(2, 2, 2)), axis=(1, 2))
```

although it raises an error if `axis` is a list instead of a tuple.

The analogous theano construct works on both lists and tuples.

vdumoulin:
I'd like to draw your attention to the _convnet_ part of the tutorial, in which batch normalization allows us to get down to 1.1% test error on MNIST with a deep sigmoidal convnet.

The kicker: the network is trained for 10 epochs using a fixed 0.01 learning rate and no momentum! With a GTX 480 GPU, it takes roughly 5 minutes to complete.

dwf:
1.1% is worse than you can get with a fully-connected ReLU net, but doing it with sigmoids in 10 epochs is still impressive.

fvisin:
Really impressive indeed and very nice tutorial btw! 
I can't wait to apply it on VAE, if I am not too tired I will work on that tonight. I will keep you posted on the results.

blocks/roles.py
```diff
@@ -129,3 +129,10 @@ class DropoutRole(VariableRole):
 
 #: Inputs with applied dropout
 DROPOUT = DropoutRole()
+
+
+class BatchNormalizationRole(VariableRole):
```
dwf: Just realized, since `Brick.repr(...)` will render this as `BATCH_NORMALIZATION`, we should probably change this to `BatchNormalizedRole()` (or change the constant to have name `BATCH_NORMALIZATION`, but I think `-ED` is better).

blocks/graph.py
```diff
@@ -528,3 +531,107 @@ def apply_dropout(computation_graph, variables, drop_prob, rng=None,
         replacement.tag.replacement_of = variable
 
     return computation_graph.replace(replacements)
+
+
+def apply_batch_normalization(computation_graph, variables, gammas,
```
fvisin: Specifying gammas and betas every time is somewhat annoying. I'd like them to be optional, set to a default value if an explicit value is not provided.

I guess the following behavior might not be optimal in some cases, but it should mostly work:
``` python
def apply_batch_normalization(computation_graph, variables, gammas=None, betas=None, axis=0,
                              epsilon=1e-7, use_population=False):
    if not betas:
        betas = []
        for i, v in enumerate(variables):
            betas.append(shared_floatx(numpy.zeros(v.shape[-1]), name='betas_{}'.format(i+1)))
    if not gammas:
        gammas = []
        for i, v in enumerate(variables):
            gammas.append(shared_floatx(numpy.ones(v.shape[-1]), name='gammas_{}'.format(i+1)))
```

vdumoulin: I just remembered why it's not as straightforward as one may think: you need the actual size of `v` to initialize the shared variable. In your example, `v` is a symbolic variable, so `v.shape[-1]` is a symbolic variable as well.

At the very least, the user would need to provide the dimensionality of the non-normalized axes.

docs/batch_normalization.rst
```diff
@@ -0,0 +1,397 @@
...
+procedure introduces a :math:`\beta` vector that will act as a bias vector.
+
+Let's now retrieve a reference to the symbolic variables that need to be
+normalized:
+
+>>> from blocks.filter import VariableFilter
+>>> from blocks.roles import OUTPUT
+>>> cg = ComputationGraph([cost, error_rate])
+>>> variables = VariableFilter(
+...     bricks=mlp.linear_transformations, roles=[OUTPUT])(cg.variables)
```
fvisin: I guess this line should be 
```python
bricks=mlp.linear_transformations[0], roles=[OUTPUT])(cg)
```

dwf: That would just get the first one, no?

fvisin: You are perfectly right...sorry. I looked into `MLP` and my comment doesn't make any sense! I was fooled by the fact that applying the filter with `bricks=[encoding_mlp.linear_transformations]` didn't return any result. It was on a slightly different setting though, I will check if the same issue arises with the code of the example and report it if so.

jimsotelo:
@vdumoulin I was thinking of extending your code for inference. Do you have any suggestions on how I should implement this?

vdumoulin:
@jimsotelo By inference, you mean test-time prediction?

I intend on putting something about that in the tutorial, but I haven't figured out a clean way to do it either.

You need to have a reference to the variables you replaced lying around. Alternatively, you can get them by filtering your graph for `BATCH_NORMALIZED` variables and access their `tag.replacement_of` attribute to get the replaced variables. Once you have that, you can compile a function whose output is the value of those variables and compute their mean and standard deviation across the training set (see Algorithm 2, line 10 of the paper).

Then you compute all "population" gammas and betas (Algorithm 2, line 11 in the paper) and call `apply_batch_normalization` on the non-replaced graph using these values and setting `use_population=True`.

jimsotelo:
Thanks for the advice! A problem that I was thinking might arise is that you kind of have to do it iteratively. This is what I mean:

Let's say you have 2 batch normalized layers. If you compile the function with both layers' output and then get the mean and std dev over the training set, it could be that the 2nd layer is computed given the unstandardized 1st layer as input. This could be different from the case when you first standardize the 1st layer, then compute the 2nd layer and calculate the statistics.

What do you think about this?

vdumoulin:
Reading the algorithm a bit more carefully, I think you're right. That makes things awkward: in the middle of the procedure you need to have the lower half of your network use the population statistics, while the upper half uses the batch statistics.

At the moment, `apply_batch_normalization` is either using population statistics for all variables, or batch statistics for all variables.

I guess `use_population` could instead be a dict mapping variables to booleans that indicates whether this particular variable needs to use population statistics or batch statistics. What do you think?

vdumoulin:
@jimsotelo The latest commit should make it work for you.

jimsotelo:
I think I don't fully understand the logic behind this code:
```python
    if use_population:
        replacements = [(var, gamma * var + beta) for var, gamma, beta in
                        zip(variables, gammas, betas)]
    else:
        replacements = [
            (var, gamma * (var - mu) / tensor.sqrt(sigma_sqr + epsilon) + beta)
            for var, mu, sigma_sqr, gamma, beta in
            zip(variables, means, variances, gammas, betas)]
```
I think that when you use the ```use_population``` you are just applying a linear transformation without standardizing at all. Maybe the dictionary is not needed, instead maybe passing the mu and var values that should be applied.

Are you in the lab? Maybe it would be better to talk about this in person.

vdumoulin:
Not today, but tomorrow I'll be there.

The idea was simply to have a common function to call for training and test time. In both cases you're applying a linear transformation and tagging variables, which is useful to encapsulate, but in one case that transformation is data-dependent whereas in the other case it's not.

In the `use_population=True` case, `gamma` is considered to be `gamma / sqrt(Var[x] + epsilon)` and `beta` is considered to be `beta - (gamma . E[x]) / sqrt(Var[x] + epsilon)`.

dwf:
I've been thinking about accumulating population statistics, and I think
the right way to attack it is to write a function that takes a data stream
and a computation graph, and then accumulates sum(X) and sum(X**2) for
every variable with a given tag. From these it is straightforward to
compute the mean (divide sum(X) by total number of cases) and variance
(using the formula X_squared_sum / num_cases - (X_sum / num_cases)**2,
which also happens to be the numerically stable way of computing the
variance.

That function then spits out a dict of {variable: (mean, variance)} that
can then be used to build the final test time graph.

jimsotelo:
I think I understand now. So the difference is that for training you do something like this:
```python
cg = apply_batch_normalization(
    cg, variables, gammas, betas, axis=[0, 2, 3], epsilon=1e-5)
```
whereas for testing you have to first modify the gammas and betas to the new data-dependent values and then do:
```python
cg_test = apply_batch_normalization(
    cg, variables, gammas_mod, betas_mod, axis=[0, 2, 3], epsilon=1e-5, use_population=True)
```
Finally, to deal with the issue I raised before, this would be done inside a loop over the batch normalized variables.

I've just read @dwf comment and it seems to be a great way of computing the mean and variance for each variable. I mean that it would be inside the loop.

jimsotelo:
I also want to state that I'm just planning to evaluate the model after it has been fully trained. For the monitoring, it would probably be better to use a running mean. I think they also suggest this in the paper.

What do you think?

jimsotelo:
In order to advance the work for testing time, I've worked on the tensor support for aggregators:

PR: https://github.com/bartvm/blocks/pull/595

This should make possible to implement @dwf idea in a simpler way.

vdumoulin:
@jimsotelo Very nice! I'll have a look at that.

I'm starting to think that we should tag `replacement` with references to the `gamma` and `delta` parameters as well. Otherwise retrieving them is awkward when loading a saved model.

EderSantana:
@jimsotelo and @vdumoulin did you get to test the batch normalization yet (I mean the final model on the test set)? I just tried to do it this way: https://gist.github.com/EderSantana/435e67c79800f50aa5f5
But I observed a little degrade in the accuracy. Also, I was a little confused about how to use the the modified computation graph, I'm not really sure if the solution above is 100% right...  

This also rises another question. Why is it better to do batch normalization by modifying an existing computation graph instead of using a layer that could save moving averages as states, for example? I'm sure this is something based on Blocks core design choices... I would just really appreciate to have some clarification on that.

dwf:
So far I don't think we have a standard piece of code that computes
statistics for the entire training set. @jimsotelo worked on some stuff for
aggregators but it seems that you need to do multiple passes over the
training set to do it properly. The reason is that the correct population
statistics for e.g. layer 3 depend on the layer 2 outputs being computed
with the population-normalized statistics, which themselves depend on the
layer 1 outputs being computed with the layer 1 population statistics.

vdumoulin:
Regarding implementation choices, I was afraid that going the "layer" way would potentially paint us in the corner further down the road by forcing people to buy into the whole `MLP` or `Convolutional` set of bricks to use this single feature.

Working on the computational graph itself is more in line with the Blocks philosophy, which is to be a framework that facilitates building computational graphs. In theory, someone could write pure Theano code and use Blocks just for the batch normalization part.

dwf:
Yeah. I think that's the right way to do it, operating on a graph. If we
can avoid it, we also don't want to constrain its use to things that are
just chains. e.g. if we have a multi-modal net that takes images and text
using two separate pathways we want to be able to apply it to the whole
thing, which necessitates some sort of breadth-first search backward from
the output(s).

But re-reading the paper, it seems like they just record moving averages of
the activations in a single pass (see algorithm 2). Since that's what the
original proposers use, it's probably what we should do first, even if
someone eventually wants to complicate their life and experiment with the
"proper" way. So @jimsotelo's work is not in vain. :)

An implementation detail: for the variance, what you probably want to do is
track the sum (or mean) of the square of the activations, and then at the
_end_ subtract square-of-mean activations from mean-of-squared-activations.
That minimizes the amount of numerical error you end up with by doing less
subtractions.

I'm a little confused by the m/(m - 1) term in the paper for the population
statistics. Normalizing by (m - 1) is an unbiased estimator of the standard
deviation from a sample; it would make more sense to me to use 1/(m - 1) in
Algorithm 1 and (m - 1)/m in Algorithm 2 (since normalizing by m for "the
population", i.e. the entire training set, is more reasonable). Note that
it's also necessary to do some gymnastics if you have a dangling minibatch
of size < m.

dwf:
Yoshua just convinced me that the m/(m-1) thing is correct for what
they're doing, though it's still puzzling why they don't just divide
by m-1 at training time, since that would reduce the bias of their
per-batch variance estimate.

 Oh well; I'd say we implement exactly what the paper prescribes to
start, which has the advantage of being pretty simple computationally
speaking.

sotelo:
Sorry for the SUPER LATE reply. I've written a mean and variance aggregator for multivariate (more than scalar) theano tensors. I haven't made a pull request, but it can be found [here.] (https://gist.github.com/bd70927c2ce04e227e49) I'm currently using the biased estimate, although the unbiased estimate should be really easy to compute from there. (Changing line 53).

Either way, I will post an example of the correct way of computing test performance. As @dwf pointed out, this involves having a loop over the different layers where BN is applied. Unfortunately, multiple passes over the training set are unavoidable unless you're willing to approximate the answer in some way. I was thinking of using the aggregation code, changing the variables that will get aggregated and using the new values for later computations. A proof of concept of what I mean can be found in this [gist] (https://gist.github.com/sotelo/c5544f798bc817d48e81) (changing lines 42-46 with the aggregation code).

As they mention in the paper, this is probably something that you don't want to do during training but it might be useful for evaluating the model at the end. In my (limited) experience, increasing the number of examples considered for the statistics tends to improve classification at the end.

dwf:
@jmsotelo Upon more careful reading, the guys who wrote the paper actually
just do one pass and average their per minibatch estimates, doing a
correction at the end. That's probably what we should support in Blocks
before anything else.

sotelo:
I guess you refer to algorithm 2. Will you be at the lab tomorrow? I would be willing to write the code but I would appreciate some guidance for planning what to do.

What do you mean by correcting at the end?

sotelo:
@dwf I'm probably confused, but I think they actually do several passes for testing. In line 8 of algorithm 2, they have a for loop over variables. This means they first loop over variables (or BN-ed layers) and then they say that they process multiple mini-batches (in line 10). Of course, the extreme would be to process ALL the mini-batches in line 10, in this case, they would process all the data once per each BN-ed layer. I don't think they did that, but I think this is closer to the 'correct' way of doing it.

I'll try to post a working version of this later today. I'm almost there :)

dwf:
I just checked with Christian Szegedy. He said the way they use it pretty
much exclusively is to learn a mean and variance during training, basically
by regressing the parameter with squared error. This has the added benefit
that you can monitor the performance of the "test time" network during
training.

rizar:
I am curious why this highly desired feature hanging as a PR for 3 months...

dwf:
Basically because we haven't agreed upon the correct way to produce a test
time network. I think we ought to create appropriate tested shared
variables and return an OrderedDict of additional updates that implement
the moving average strategy. (I can't remember off the top of my head if
there is a facility for making GradientDescent apply some updates verbatim,
but maybe there should be...)

dwf:
You're very right, that is an annoying detail (this is where the pylearn2
"output space" concept would be handy). @rizar, thoughts?

bartvm:
For aggregators we solve this using a flag and `ifelse`. Not pretty, but it works: https://github.com/mila-udem/blocks/blob/master/blocks/monitoring/aggregation.py#L111-L121

rizar:
I am confused: it seems like @cooijmanstim's pull request to @vdumoulin branch was merged, but somehow the code here is not updated. @vdumoulin , do you know why?

rizar:
Closed in favour of #851.

Issue #514, opened 2015-03-19
Computation graph replacements and aggregation schemes


rizar:
When I apply dropout to a computation graph  in which certain variables have aggregation schemes, this makes no changes to aggregation schemes. I find it very unfortunate and very dangerous.

I think that we could teach `ComputationGraph.replace` to make replacements in aggregation scheme inputs.

dwf:
Can you explain a little more? As far as I understand it aggregation schemes are something used by monitoring. You almost always want to be monitoring the non-dropped-out version of the model.

rizar:
An addition: same problem for auxiliary variables, they also must be changed.

@dwf, I am sort of used to always monitor the training cost on every batch in addition to the validation cost. Monitoring both dropped out cost and clean cost is very computationally expensive, that's why I want to monitor 
clean validation cost but dropped out training cost.

Aggregation schemes (`blocks/monitoring/aggregation.py`) are a way to define what is the right way to average your variable over a few batches. For instance if you monitor the cost per sequence element and you want to get average over ten batches, you have to sum ten total costs, sum all masks and divide the former by the latter. The aggregation schemes are handled in `AggregationBuffer` and `DatasetEvaluator`, which are in `blocks/monitoring/evaluators.py`.

dmitriy-serdyuk:
@rizar , are you applying aggregations after replacements? Seems, that it should work.

rizar:
@dmitriy-serdyuk , I am applying replacements after I attached aggregation schemes to some variables, and this does not work. I want to monitor similar expressions on training and vaidation data, with the difference that the former are monitored with regulazarion, and the latter without.

rizar:
Thinking about it a bit more, the problem is that annotations are not affected by `ComputationGraph.replace`. That involves their auxiliary variables and updates and might have very nasty consequences, when updates added to the computation graph and invalidated by a transformation performed on it. Aggregation schemes in fact should inherit from `Annotation`, it's just that they were historically earlier introduced that they do not. 

I am thinking in the direction that the `Annotation` interface should be expanded with a `replace` method:

```python
class Annotation(object):

    def replace(self, updates, auxiliary_variables):
         """Create a clone of this annotation with new updates and auxiliary variables."""
         return Annotation(updates, auxiliary_variables)
```

For annotations such as `Brick`, it should simply `return self`, but for `ApplicationCall` and `AggregationScheme` it should produce a clone with the variables given.

dwf:
[...]

PR #516, opened 2015-03-19
Fixed replacement of outputs

### LINKED ISSUE #515: ComputationGraph.replace doesn't work on graph outputs
The following example fails:
```python
>>> from theano import tensor
>>> from blocks.graph import ComputationGraph
>>> x = tensor.scalar()
>>> y = x + 1
>>> cg = ComputationGraph([y])
>>> doubled_cg = cg.replace([(y, 2 * y)])
>>> doubled_cg.outputs[0].eval({x: 2})
3.0
```

dmitriy-serdyuk:
Fixes #515.

### DIFF
--- a/blocks/graph.py
+++ b/blocks/graph.py
@@ -204,17 +204,15 @@ class ComputationGraph(object):
         # Sort `replacements` in topological order
         for node in apply_nodes_sorted:
             for input_ in node.inputs:
-                if input_ not in replacements:
-                    continue
-                replacement_keys_cur.append(input_)
-                replacement_vals_cur.append(replacements[input_])
+                if input_ in replacements:
+                    replacement_keys_cur.append(input_)
+                    replacement_vals_cur.append(replacements[input_])
 
         # Add outputs of the computation graph
         for output in self.outputs:
-            if output not in replacements:
-                continue
-            replacement_keys_cur.append(output)
-            replacement_vals_cur.append(replacements[output])
+            if output in replacements:
+                replacement_keys_cur.append(output)
+                replacement_vals_cur.append(replacements[output])
 
         # Replace step-by-step in topological order
         while replacement_keys_cur:
--- a/blocks/graph.py
+++ b/blocks/graph.py
@@ -209,6 +209,13 @@ class ComputationGraph(object):
                 replacement_keys_cur.append(input_)
                 replacement_vals_cur.append(replacements[input_])
 
+        # Add outputs of the computation graph
+        for output in self.outputs:
+            if output not in replacements:
+                continue
+            replacement_keys_cur.append(output)
+            replacement_vals_cur.append(replacements[output])
+
         # Replace step-by-step in topological order
         while replacement_keys_cur:
             replace_what = replacement_keys_cur[0]
--- a/tests/test_graph.py
+++ b/tests/test_graph.py
@@ -71,6 +71,16 @@ def test_computation_graph():
     assert all(v in cg6.scan_variables for v in scan.inputs + scan.outputs)
 
 
+def test_replace():
+    # Test if replace works with outputs
+    x = tensor.scalar()
+    y = x + 1
+    cg = ComputationGraph([y])
+    doubled_cg = cg.replace([(y, 2 * y)])
+    out_val = doubled_cg.outputs[0].eval({x: 2})
+    assert out_val == 6.0
+
+
 def test_apply_noise():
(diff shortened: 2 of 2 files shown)


blocks/graph.py
```diff
@@ -209,6 +209,13 @@ def replace(self, replacements):
                 replacement_keys_cur.append(input_)
                 replacement_vals_cur.append(replacements[input_])
 
+        # Add outputs of the computation graph
+        for output in self.outputs:
+            if output not in replacements:
```
vdumoulin: The following would be more straightforward:
```python
if output in replacements:
    replacement_keys_cur.append(output)
    replacement_vals_cur.append(replacements[output])
```

dmitriy-serdyuk: Is it a bad practice to use `continue`? I tried to decrease the number of indents.

bartvm: I don't think `continue` is discouraged per se, and it can often be the right choice. In this case I agree with @vdumoulin though; `if .. in ..` just reads a bit more natural.

dwf:
LGTM.

blocks/graph.py
```diff
@@ -204,10 +204,15 @@ def replace(self, replacements):
         # Sort `replacements` in topological order
         for node in apply_nodes_sorted:
             for input_ in node.inputs:
-                if input_ not in replacements:
-                    continue
-                replacement_keys_cur.append(input_)
-                replacement_vals_cur.append(replacements[input_])
+                if input_ in replacements:
```
rizar: There is a bug here, because a variable can be an input of multiple nodes.

vdumoulin: Can you give an example of where this would fail?

dwf: A failing test case, even, would be good.

rizar: Sure, I will. I got it when if was monitoring mean activation of a dropped-out layer, that is the values from this layers were used for monitoring and cost computations.

It took me a while to find this bug and this is the main reason why I am lagging behind in a few important discussions here.

vdumoulin: I get the cause, but what impact does this bug have on the correctness of computation?

rizar: It made me panic for two days, because after I started to use the aforementioned monitoring applying dropout _caused_  severe overfitting. This sort of makes sense, because division by `1 - dropout_prob` happened twice and the training and validation objective were no longer similar, but quite different.

I will fix it tomorrow, if nobody does it before me.

vdumoulin: OH! I get it, I think. So you're saying that if a node to be replaced is an input of multiple nodes in the graph, its replacement will happen more than once which will affect the correctness of the computation.

Here's a minimal example that reproduces the issue:
```python
from theano import tensor
from theano.printing import debugprint

x = tensor.scalar('x')
y = 2 * x
z = x + 1

cg = ComputationGraph([y, z])
debugprint(cg.outputs[0])
cg = cg.replace({x: 0.5 * x})
print('----------')
debugprint(cg.outputs[0])
print('----------')
print cg.outputs[0].eval({x: 1.0}), cg.outputs[1].eval({x: 1.0})
```

The expected output for `y` and `z` is 1 and 1.5, respectively, but the output produced by the code snippet above is 0.5 and 1.25, which corresponds to the replacement being applied twice on `x`, as shown in the `debugprint` output.

@rizar I'll make a pull request for that right away.

rizar: Exactly! Sorry for not explaining it as clearly as you did.

> @rizar I'll make a pull request for that right away.

Cool, thanks!

rizar: It seems to me that the whole iteration could be actually replaced by `for variable in self.variables`. They are also sorted topologically, exactly in the way as done here (correct me if I am wrong), but the duplicates are removed, which is just what we need here.

vdumoulin: @rizar see my PR in #525

vdumoulin: I think you're right, `self.variables` is topologically sorted.

vdumoulin: Although it also contains auxiliary variables. Is it an issue?

rizar: Needs a bit of thought. So far it is definitely not because replacements of auxiliary variables is not supported, see #514. If we support it later we will have to be able to traverse all variables including auxiliaries in topological order, but again, the right tool for it would be `self.variables`.

So I would say it's not, maybe we should mention it in a comment in `replace`.

PR #519, opened 2015-03-20
Fix unique name check

rizar:


rizar:
Right, removed square brackets. Set comprehensions... well, it has not  been yet officially said that we don't support Python 2.7.

dwf:
They are available in Python 2.7:
https://docs.python.org/2/tutorial/datastructures.html#sets

rizar:
Thas was a typo, I wanted to say 2.6, sorry.

bartvm:
Isn't it? We use `OrderedDict`, dictionary comprehensions, multiple context managers in a single `with` statement and `importlib`, all of which are Python 2.7 only. I have no desire whatsoever to support 2.6.

rizar:
Okay, I did not remember for all these things the version in which they were introduced. Changed to set comprehension.

bartvm:
Actually, looks like `variable_names` is zipped at [line 137](https://github.com/rizar/blocks/blob/fix_dataset_evaluator_check/blocks/monitoring/evaluators.py#L137). So it doesn't make sense if you turn it into a set, since those are unordered.

rizar:
Right, thanks for pointing this out, trying again.

PR #522, opened 2015-03-23
WIP: Write a bit of docs and make sequence generator more standard

rizar:
My work in progress on fixing quirks of sequence generator. It seems like having `outputs_t` generated from `states_t`, not from `states_{t-1}` is the current standard and makes code easier.

On my way I am changing docs a bit, though this is not the main point of this PR.

### DIFF
--- a/blocks/bricks/sequence_generators.py
+++ b/blocks/bricks/sequence_generators.py
@@ -76,15 +76,15 @@ class BaseSequenceGenerator(Initializable):
        readout, initial states and glimpses given by the `initial_state`
        method of the transition.
 
-    2. The next states are computed from the current states, current
-       glimpses, feedback from the current outputs. This happens in in the
-       transition's `apply` method. The `fork` brick is used to
-       compute the transition's inputs from the feedback.
-
-    3. Given the contexts, the next states and the current glimpses
+    2. Given the contexts, the current states and glimpses
        the attention mechanism hidden in the transition
        produces the next glimpses. This happens in the `take_glimpses`
        method of the transition.
+
+    3. The next states are computed from the current states, next
+       glimpses, feedback from the current outputs. This happens in in the
+       transition's `apply` method. The `fork` brick is used to
+       compute the transition's inputs from the feedback.
 
     4. Using the contexts, the feedback from the current outputs, the
        next states and glimpses, the readout brick is used to generate
@@ -263,19 +263,19 @@ class BaseSequenceGenerator(Initializable):
         contexts = dict_subset(kwargs, self._context_names)
         glimpses = dict_subset(kwargs, self._glimpse_names)
 
+        next_glimpses = self.transition.take_glimpses(
+            as_dict=True, **dict_union(states, glimpses, contexts))
         feedback = self.readout.feedback(outputs)
         inputs = self.fork.apply(feedback, as_dict=True)
         next_states = self.transition.compute_states(
-            as_list=True,
-            **dict_union(inputs, states, glimpses, contexts))
-        next_glimpses = self.transition.take_glimpses(
-            as_dict=True, **dict_union(next_states, glimpses, contexts))
+            as_dict=True,
+            **dict_union(inputs, states, next_glimpses, contexts))
         next_readouts = self.readout.readout(
             feedback=feedback,
             **dict_union(next_states, next_glimpses, contexts))
         next_outputs = self.readout.emit(next_readouts)
         next_costs = self.readout.cost(next_readouts, next_outputs)
-        return (next_states + [next_outputs] +
+        return (list(next_states.values())+ [next_outputs] +
                 list(next_glimpses.values()) + [next_costs])
 
     @generate.delegate
--- a/blocks/bricks/sequence_generators.py
+++ b/blocks/bricks/sequence_generators.py
@@ -8,8 +8,8 @@ element is fed back into the recurrent network state. Sometimes
 also an attention mechanism is used to condition sequence generation
 on some structured input like another sequence or an image.
 
-This module provides :class:`SequenceGenerator` that builds a sequence generating
-network from three main components:
+This module provides :class:`SequenceGenerator` that builds a sequence
+generating network from three main components:
 
 * a core recurrent transition, e.g. :class:`~blocks.bricks.recurrent.LSTM`
   or :class:`~blocks.bricks.recurrent.GRU`
@@ -275,7 +275,7 @@ class BaseSequenceGenerator(Initializable):
             **dict_union(next_states, next_glimpses, contexts))
         next_outputs = self.readout.emit(next_readouts)
         next_costs = self.readout.cost(next_readouts, next_outputs)
-        return (list(next_states.values())+ [next_outputs] +
+        return (list(next_states.values()) + [next_outputs] +
                 list(next_glimpses.values()) + [next_costs])
 
     @generate.delegate
--- a/blocks/bricks/sequence_generators.py
+++ b/blocks/bricks/sequence_generators.py
@@ -1,4 +1,33 @@
-"""Sequence generation framework."""
+"""Sequence generation framework.
+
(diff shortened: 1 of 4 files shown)


rizar:
Okay, the most difficult part is done, `reverse_words` is working again.

I should say that these changes have quite positive impact on the code, and it seems that it will be possible to make `SequenceGenerator` a _way_ more simple by hiding the attention from it altogether.

I am just enumerating people who might be interested in this PR: @sebastian-j, @kyunghyuncho, @pbrakel, @dmitriy-serdyuk. Note, when merged it will break whatever you have!

blocks/bricks/sequence_generators.py
```diff
@@ -47,38 +76,29 @@ class BaseSequenceGenerator(Initializable):
...
-
     **Notes:**
 
     * For machine translation we would have only one glimpse: the weighted
       average of the annotations.
 
-    * For speech recognition we would have three: the weighted average,
-      the alignment and the monotonicity penalty.
+    * For speech recognition we would have two: the weighted average,
+      the alignment and the alignment.
```
bartvm: The alignment and the alignment?

rizar: Sorry, this is very much WIP :)

rizar:
Okay, this PR is mostly ready and I am waiting for code comments and feedback from the machine translation team. It opens door for lots of refactoring, but I would prefer to have a separate later PR for it.

bartvm:
Isn't @dmitriy-serdyuk in a better position to review changes to the beam search?

The changes to `BaseSequenceGenerator` look okay to me.

rizar:
As mentioned in the NMT repo, applying this commit make convergence of my experiments significantly slower. So far I have no idea why.

dmitriy-serdyuk:
It looks much cleaner. I suppose, that the [picture](http://blocks.readthedocs.org/en/latest/_images/sequence_generator_scheme.png) should also be modified.

dmitriy-serdyuk:
No, I'm wrong, it is the same.

blocks/search.py
```diff
@@ -104,35 +107,25 @@ def _compile_initial_state_computer(self):
...
-        self.next_state_computer = function(
-            self.contexts + self.input_states + next_outputs, next_states)
-
-    def _compile_logprobs_computer(self):
+                       for name in self.state_names_no_output]
         # This filtering should return identical variables
         # (in terms of computations) variables, and we do not care
         # which to use.
-        probs = VariableFilter(
+        logprobs = -tensor.log(VariableFilter(
```
dmitriy-serdyuk: I realized that `logprobs` may be misleading. Should we rename it to `nlprobs` for negative log probs?

rizar: Technically you are right, but I still would prefer to keep it like this. Too long or too abbreviated names sometimes kill clarity and I feel it's one of these cases.

dwf: `neglogprobs`?

rizar: Okay, there are too many of you to simply ignore your opinion :) I will change it to `neglogprobs`.

rizar:
Lol guys, we are all sooooo stupid. 

Briefly: the change from this PR is a nightmare because the glimpse is chosen without using the last generated symbol.

Verbose version: there are basically three things an attention-equipped RNN does: it attends the input sequence, it generates a new symbol, it generates  states. Let's call those "look", "generate" and "update" actions.

In both Groundhog and the original version of the sequence generation framework the order of these three actions was the following: "look", "generate", "update". This PR changes it to "look", "update", "generate". And the state used at the "look" stage does not contain the last generated symbol! In other words "generate" must take place between "look" and "update" which leaves us the following choice:

- "look", "generate", "update". This is how we do in Blocks master, but implementation wise this is not very easy
   and is the main reason why we now have the ugly `FakeAttentionRecurrent`.

- "update", "look", "generate".
 
- "generate", "update", "look".

Unlike the option 1 options 2 and 3 always require carrying over information from "look" at the step _t_ to "update" at the step _t + 1_. Option 1 allow to have glimpses both produced and consumed in the same time step.

Having all that said, I am inclined to freeze sequence generators as they currently are (i.e. the option 1) for Blocks 0.1 to make speech and machine translation experiments possible right away. Switching to either of the options 2 or 3 would probably require significant refactoring and would inevitably introduce bugs.

I would be interested though to heat which of the three methods you find a better candidate for happy future of Blocks.

@kyunghyuncho, @sebastien-j, I believe the text above is very important.

kyunghyuncho:
This is true, and unfortunately, so far all our experiments (on MT, caption
generations, etc.) did not take into account the previously generated
symbol. Obviously taking into account the previously generated symbol is
important but may not be too much so, because the previous hidden state
should contain the information about the conditional distribution of the
previous symbol. Of course, this ignores the fact that the conditional
distribution of the previous symbol had one dominant mode, and if it has
many equally likely modes, taking into account the previously generated
symbol (deciding which mode out of many) become more important.

dmitriy-serdyuk:
@kyunghyuncho , do you mean, that GroundHog is different from blocks and doesn't take into account the previous symbol?

kyunghyuncho:
If you look at A.2 in (http://arxiv.org/pdf/1409.0473v5.pdf), it's clear
that we have not used the previously generated symbol when computing the
alignment weights. This can be easily fixed (or modified) by including
y_{i-1} in computing e_{ij}.

sebastien-j:
In the short term, keeping SequenceGenerator as it is now could be helpful as it allows loading Groundhog models .

I don't like the third option very much. I believe we should `look` before `generating` the first symbol.

@dmitriy-serdyuk , if I understand correctly, the attention mechanism in Blocks explicitely uses the previous generated word, rather than implicitely through the hidden state as in GroundHog. Is this correct?

rizar:
@kyunghyuncho , I feel I failed to explain what I mean. Let's consider your  scheme from https://github.com/kyunghyuncho/NMT/pull/3#issuecomment-86162489:

```
state_t -> context_{t+1} 
state_t, context_{t+1}, output_t -> state_{t+1} 
state_{t+1}, context_{t+1}, output_t -> output_{t+1}
```

By `the glimpse is chosen without using the last generated symbol` I mean the following issue:

```
context_{t+1} <- state_t <- state_{t-1}, context_{t}, output_{t-1} <- ... output_{1 ... t-1}
```

This is quite bad because during both training and test stage `output_{t}` is not used to compute `context_{t}`, e.g. to find the location in the sequence relevant for generating `output_{t+1}`. 

We do not have this issue neither in Grounhog master nor in Blocks master. It's also different in the paper

```
h_i <- h_{i-1}, y_{i-1}
c_i <- h_i
y_i <- h_i, c_i, y_i
```

and in the correct description of what happens in Groundhog by @sebastien-j https://github.com/kyunghyuncho/NMT/pull/3#issuecomment-86162489

That is the current Blocks implementation is quite nice and I suggest that it enters Blocks 0.1. After the NIPS deadline we can thing about changing the code in a way that does not require `FakeAttentionRecurrent`, that is use one of the options 2 and 3 from my earller post.

rizar:
> if I understand correctly, the attention mechanism in Blocks explicitely uses the previous generated word, rather than implicitely through the hidden state as in GroundHog. Is this correct?

No, it does not! But the "cleaner" version we considered would fail to use the previous generated word even through the previous hidden states, see formulas above.

> In the short term, keeping SequenceGenerator as it is now could be helpful as it allows loading Groundhog models .

By all means, yes, if the current Blocks implementation is indeed compatible with Groundhog, let's use it for the transition period. In fact I tried hard to make it compatible when I wrote in November. Let me know when you try to match Groundhog cost using the attention mechanism, if we are able to make them exactly equal that would be perfect!

We can explore alternatives later, and that would hopefully allow to significantly simplify the code.

rizar:
I am closing this PR as the change I tried to here turned out to be a bad plan. I will open another PR with the little documentation written here.

PR #523, opened 2015-03-23
Move annotating list to new utils.containers

bartvm:


bartvm:
The `DefaultOrderedDict` from #476 could go here as well eventually.

blocks/utils/containers.py
```diff
@@ -0,0 +1,57 @@
...
+    def __eq__(self, other):
+        return self._items == other
+
+    def __ne__(self, other):
+        return self._items != other
+
+    def __getitem__(self, key):
+        return self._items[key]
+
+    def _setitem(self, key, value):
```
dwf: `@abstractmethod`?

blocks/utils/containers.py
```diff
@@ -0,0 +1,57 @@
...
+        """The operation to perform when an item is inserted/appended."""
+        pass
+
+    def _delitem(self, key):
```
dwf: ditto, `@abstractmethod`?

dwf: Actually I can see this one perhaps not being annotated as `@abstractmethod`, but `_setitem` seems fundamental.

bartvm: Yeah, don't think this one should, sometimes you only want to perform an action on insert. I figured there theoretically could be a case where you only want to do something on deletion, which is why I didn't annotate the `_setitem` method, but I can't actually think of an example, so I guess I might as well :)

bartvm:
@dwf Updated and rebased

dwf:
It seems like we're losing coverage by a thousand papercuts, but as long as it's still above 91% I'm not going to worry too much about it.

PR #525, opened 2015-03-23
Fix bug in ComputationGraph.replace

vdumoulin:
If a variable to be replaced is an input to multiple nodes, the replacement is made multiple times, which is incorrect.

### DIFF
--- a/blocks/graph.py
+++ b/blocks/graph.py
@@ -205,8 +205,9 @@ class ComputationGraph(object):
         for node in apply_nodes_sorted:
             for input_ in node.inputs:
                 if input_ in replacements:
-                    replacement_keys_cur.append(input_)
-                    replacement_vals_cur.append(replacements[input_])
+                    if input_ not in replacement_keys_cur:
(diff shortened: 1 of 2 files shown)


blocks/graph.py
```diff
@@ -205,8 +205,9 @@ def replace(self, replacements):
         for node in apply_nodes_sorted:
```
rizar: If you also find this identical to `for variable in self.variables`, then maybe we should we switch to it, because it's a way easier.

vdumoulin: I think `self.variables` also contains auxiliary variables.

vdumoulin: Thinking of it, it's not, unless the user happens to try to replace an auxiliary variable.

vdumoulin:
Should we merge now and refactor later if need be?

blocks/graph.py
rizar: I guess we could even log a warning the user tries to do such nasty things.

Issue #337, opened 2015-02-23
Bricks initialization: discussion and possible improvements


fvisin:
I'd like to discuss two potential issues of the current bricks initialization:

1. I find somewhat redundant and confusing that a brick should be initialized even when initialization arguments are given in the constructor. I would prefer the following behaviour: 
    - __all the initialization argument passed to the constructor__: the brick object quietly calles `initialize()` right before the first time it is "used". This could be achieved with an \_initialized flag set in the object.
    - __some initialization arguments are not passed to the constructor__: the brick object should be either  explicitly initialized later calling `initialize(args_dict)`, where `args_dict` is a dictionary of parameters and values, or Block will try to automatically initialize it by exploiting the nested bricks hierarchy.
2. In my opinion setting the arguments to 0 by default and initialize them only when `initialize()` is called, is potentially very harmful because it can lead to silent failures, i.e. to the model running without raising errors but with wrong arguments. In my opinion it would be preferable to fail when the brick has not been either explicitly or automatically (nested bricks scenario) initialized.

bartvm:
I agree in principal with these issues, but the decision was definitely made keeping these things in mind. One way we hope to alleviate "silent" failing is through https://github.com/bartvm/blocks/issues/244, using `NaN` instead of zero should make it clearer that you forgot to initialize, because everything (including the cost) will have `NaN` as a value before training even started.

The main problem is this: You strictly speaking *don't know* whether you can initialize or not. You don't know what arguments are needed; right now, it's almost always just `weights_init` and `biases_init`, but in principle the initialization process could rely on a range of other settings.

So the only way to initialize "by default" would be to use a `try` and `except` block. That's messy: You could be sampling from the RNG, so initialization could become irreproducible. 

I'm not sure what you mean by "exploiting the nested bricks hierarchy".

A more comprehensive solution to this would be to implement what is discussed in https://github.com/bartvm/blocks/issues/31. If we actually define in the signature which values are needed for which part of the life cycle, it would become possible to check if all necessary arguments for  allocation/initialization were provided.

bartvm:
Another issue that comes to mind with your approach: It can be confusing that `initialize` needs to be called in some cases, and not in others. (Especially if people are programmatically passing arguments.) The question then becomes what a second call to `initialize` should result in: If you re-initialize, the user might be overriding the old initialization unknowingly (causing problems with reproducibility). If you don't re-initialize, it becomes impossible for the user to change the initialization settings and re-initialize.

[...]

Issue #39, opened 2014-11-21
Switch to using decorators for _initialize, _allocate, etc.?


bartvm:
We currently have the situation where we want a set of methods of all the derived classes of `Brick` to be wrapped by methods defined in the `Brick` base class e.g. `_initialize`, `_allocate`, `_push_allocation_config`, etc. are all wrapped by the methods `initialize`, `allocate`, ....

There are a few ways of doing this. The way we do it right now is by calling the public method `allocate` on the base class, which internally calls the non-public method `_initialize` defined on the child classes. This works, but personally I am not a fan of non-public methods being the main features of bricks. It might also be confusing to people (and I think ugly) that they need to define `_initialize` and then call `initialize`, nothing will work. The current method also doesn't allow us to document the `initialize` method of each brick separately.

I was wondering if it would not be better to switch to using decorators here. We use them quite heavily already, but I think this is exactly what they are made for: wrapping methods simply, and explicitly. So instead we would have

``` python
class SomeBrick(Brick):
    @lazy
    def __init__(self, dim):
        self.dim = dim

    @application
    def apply(self, x, y):
        """Applying this brick does a, b and c."""
        ...

    @initialization
    def initialize(self):
        """Initialize the 3 weight matrices of this block."""
        ...

    @allocation
    def allocate(self):
        """Allocate the parameters. Note that `dim` must be set before calling this."""
        ...
```

PR #528, opened 2015-03-23
WIP: New bricks

bartvm:
WIP, but this is basically a rewrite of `blocks/bricks/base.py`. It implements the `lazy` decorator as discussed in https://github.com/bartvm/blocks/issues/337#issuecomment-75884106 with @rizar and @fvisin. It also ditches metaclasses in favor of class decorators; I need to confirm later on that these actually pickle well, but my initial tests seem to show they do.

Currently it's still very messy, but it cuts 20% of the lines `base.py`. There are a few functionalities which took acrobatics to support, which I'm not sure are actually used, so I would propose removing them if at all possible. Notably, the ability to set the attributes of applications i.e. to do `brick.apply.inputs = ['x']`. This is annoying because it prevents us from actually turning `brick.apply.inputs` in a bound method (it's impossible to set an attribute of an `instancemethod`).

This also switches to using decorators for `allocation`, `initialization`, etc. as suggested in https://github.com/bartvm/blocks/issues/39. Expecting a user to override private methods is pretty bad form I think, so we need to either change the name of the method to something without a leading underscore, or use decorators. I went with decorators for now, because otherwise it's confusing for users to implement one method, but call another.

### DIFF
--- a/blocks/bricks/__init__.py
+++ b/blocks/bricks/__init__.py
@@ -1,5 +1,6 @@
 """The interface of bricks and some simple implementations."""
 import logging
+from abc import ABCMeta
 
 import numpy
 from six import add_metaclass
@@ -9,7 +10,7 @@ from toolz import interleave
 from picklable_itertools.extras import equizip
 
 from blocks import config
-from blocks.bricks.base import application, _Brick, Brick, lazy
+from blocks.bricks.base import application, Brick, lazy
 from blocks.roles import add_role, WEIGHT, BIAS
 from blocks.utils import pack, shared_floatx_nans
 
@@ -428,7 +429,7 @@ class LinearMaxout(Initializable, Feedforward):
         return output
 
 
-class ActivationDocumentation(_Brick):
+class ActivationDocumentation(ABCMeta):
     """Dynamically adds documentation to activations.
 
     Notes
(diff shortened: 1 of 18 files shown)


bartvm:
@fvisin If you want to help make your idea work, please help me update all the bricks to use the new `lazy` decorator :) They should all look like:

```python
@lazy(allocation=['args', 'needed', 'for', 'allocation'],
      initialization=['likewise'])
def __init__(self, always_needed, args, needed, for allocation,
             likewise, allocation_var_with_default=None):
    pass
```

fvisin:
Nice job Bart!
I will look into it as soon as I have some spare time and I will work on the decorators too. Probably not before Thursday though.

rizar:
I really respect you Bart for your ability to work on several huge PRs at the same time, but this exceeds my reviewing capabilities. Can we finish with log refactoring first?

bartvm:
There's not much point in reviewing this right now anyway, it's too much WIP. Currently running into the same pickling issues, but trying to come up with a cleaner solution this time. I'll let people know if/when I get it into a working state.

The database backend is indeed important, but the problem I ran into is quite fundamental, and needs a bit of thinking and/or some good suggestions.

PR #529, opened 2015-03-23
Refactor Checkpoint slightly to make it more testable.

dwf:
Follow-up to #527.


dwf:
Okay, this finally passes all tests.

dwf:
No I didn't!

PR #530, opened 2015-03-23
Refactor Checkpoint slightly to make it more testable (again).

dwf:
#529 got screwed up by an errant push on my part.

This follows up on #527 by actually testing the code in question.


rizar:
It's nice to have that, but in future I think `Checkpoint` backend (file pathes formation and etc.) should be separated from the frontend (that is the extension class).

PR #534, opened 2015-03-24
Fix bug causing ComputationGraph.variables duplicates

### LINKED ISSUE #533: Possible duplicates in ComputationGraph.variables attribute
This piece of code

```python
from theano import tensor


x = tensor.matrix('x')
y = tensor.matrix('y')
w = x + y
w.name = 'w'
z = tensor.exp(w)
z.name = 'z'

cg = ComputationGraph([z, w])
print(cg.variables)
print(cg.variables[2] is cg.variables[-1])
```

produces the following output:

```
[x, y, w, z, w]
True
```

Is it desireable behaviour?

vdumoulin:
Fixes #533.

### DIFF
--- a/blocks/graph.py
+++ b/blocks/graph.py
@@ -128,8 +128,8 @@ class ComputationGraph(object):
             main_vars = (
                 [var for var in list(chain(
                     *[apply_node.inputs for apply_node in sorted_apply_nodes]))
-                 if not (var in seen or seen.add(var))]
-                + [var for var in self.outputs if var not in seen])
+                 if not (var in seen or seen.add(var))] +
+                [var for var in self.outputs if var not in seen])
 
             # While preserving order add auxiliary variables, and collect
             # updates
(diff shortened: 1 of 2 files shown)


blocks/graph.py
```diff
@@ -125,9 +125,11 @@ def _get_variables(self):
...
 
             seen = set()
-            main_vars = [var for var in list(chain(
-                *[apply_node.inputs for apply_node in sorted_apply_nodes]))
-                if not (var in seen or seen.add(var))] + self.outputs
+            main_vars = (
+                [var for var in list(chain(
+                    *[apply_node.inputs for apply_node in sorted_apply_nodes]))
+                 if not (var in seen or seen.add(var))]
+                + [var for var in self.outputs if var not in seen])
```
bartvm: I don't understand why Scrutinizer doesn't complain, but it does on my computer: PEP8 says that binary operators should go before the line break i.e.
```python
if not ...] +
[var for var ...
```

vdumoulin: I'll change that.

Issue #384, opened 2015-03-01
Initialization discussion


adbrebs:
Hi all,

First, well done for your hard work! I have just read the tutorial and I have a few remarks/questions. They may sound critical but don't get me wrong, I'm sure that your design choices are well thought out and I just want to understand them.

1. Why do we need to call explicitly brick.initialize? Can't the initialization be done when apply is called? Or during the allocation?

2. When is lazy initialization (isn't the term "lazy configuration" more suitable?) necessary/convenient? The example given in the tutorial (connecting two networks with a linear layer) could be done as easily without the lazy initialization. If it is not necessary/convenient, it may be dangerous to allow too much flexibility especially if Blocks wants to stay quite low-level. Having different ways to do the same thing can be quite confusing for a user and tedious for us to develop (the smaller the library the better).

3. "Many neural network models, especially more complex ones, can be considered hierarchical structures. Even a simple multi-layer perceptron consists of layers, which in turn consist of a linear transformation followed by a non-linear transformation.
As such, bricks can have children. "
The terminology (children, parent) might be slightly confusing. The hierarchy you mention is actually a sequence of bricks in which each brick has one child (the previous brick) and one parent (the next brick). MLP is a superBrick composed of subBricks.

4. Why do we need to call mlp.push_initialization_config when we override the initialization schemes of the children? Would it be possible to remove this step?

5. "By default, all our parameters are set to 0."
It seems dangerous to initialize the parameters to zero when they are allocated. Users might forget to call the initialize method.

I will do a PR with the few typos I have found in the tutorial.

rizar:
Thanks for your feedback!

I hope it is okay that I will answer your question one (or a few) at a time, cause I hate writing too long entries :)

1. In the most general case we might want not to initialize at all. Initialization of a very big model can take quite some time and is not necessary when one loads parameters into the bricks from the disk. We acknowledge, that in most cases it is OK do it automatically and this is a subject of active discussion, see for instance #337. After we switched to initializing parameters with NaNs, (see #378),
one can not simply forget to call `initialize` any more: his cost will be NaN from the beginning. So currently one has to always call `initialize`, which is not very convenient, but neither makes the code much more difficult.

rizar:
5. You are absolutely right, but we do not do it any more! #378 fixes that.

rizar:
[...]

Issue #31, opened 2014-11-19
Lazy Initialization: Discussion


bartvm:
I noticed that the `MLP` class actually didn't work with lazy initialisation (because `len(activations)` gave an error when `activations = None`). We should probably add a unit test which tries to initialise each `Brick` subclass without arguments to see if they are lazy-loading compatible.

PR #536, opened 2015-03-24
New lazy decorator

bartvm:
@fvisin @rizar (and I believe @adbrebs touched on this in #384)

This is a stricter version of lazy initialization (spin-off of #528). The lazy decorator is now expected to say which values are needed at which stage of the brick life cycle. I also made laziness non-configurable, because of an important point that @fvisin brought up: It makes code annoying to port, because whether things work or not might depend on someone's configuration. I've also never seen someone explicitly set `Brick.lazy = False`.

```python
@lazy(allocation=['args', 'needed', 'for', 'allocation'],
      initialization=['likewise'])
def __init__(self, always_needed, args, needed, for allocation,
             likewise, allocation_var_with_default=None):
    pass
```

The main advantages:

* Arguments that are needed for brick construction are now actually compulsory again; in the current situation arguments needed in `__init__` would just be replaced with `None` and lead to weird errors.
* No more semi-allocated or semi-initialized bricks with cryptic errors in cases where the configuration wasn't set. We can now check in advance if needed configurations for allocation and initialization are missing.
* Simplification because users don't have to choose between lazy and non-lazy anymore. Bricks are just always lazy.

### DIFF
--- a/blocks/bricks/base.py
+++ b/blocks/bricks/base.py
@@ -8,7 +8,7 @@ import six
 from six import add_metaclass
 from theano import tensor
 from theano.gof import Variable
-from toolz import merge
+from toolz import first, merge_with
 
 from blocks.graph import add_annotation, Annotation
 from blocks.roles import add_role, PARAMETER, INPUT, OUTPUT
@@ -265,8 +265,6 @@ class Application(object):
         # Allocate before applying, and optionally initialize
         if not brick.allocated:
             brick.allocate()
-        if not brick.initialized and not brick.lazy:
-            brick.initialize()
 
         # Annotate all the input variables which are Theano variables
         def copy_and_tag(variable, role, name):
@@ -458,13 +456,6 @@ class Brick(Annotation):
     ----------
     name : str
         The name of this brick.
-    lazy : bool
-        ``True`` by default. When bricks are lazy, not all configuration
-        needs to be provided to the constructor, allowing it to be set in
-        another way after construction. Many parts of the library rely on
-        this behavior. However, it does require a separate call to
-        :meth:`initialize`. If set to ``False`` on the other hand, bricks
-        will be ready to run after construction.
     print_shapes : bool
         ``False`` by default. If ``True`` it logs the shapes of all the
         input and output variables, which can be useful for debugging.
@@ -515,7 +506,7 @@ class Brick(Annotation):
 
     Examples
     --------
-    By default, bricks have lazy initialization enabled.
+    Most bricks have lazy initialization enabled.
 
     >>> import theano
     >>> from blocks.initialization import IsotropicGaussian, Constant
@@ -528,23 +519,7 @@ class Brick(Annotation):
     linear_apply_output
     >>> linear.initialize()  # Initializes the weight matrix
 
-    In simple cases, eager bricks are easier to deal with.
-
-    >>> from blocks.initialization import IsotropicGaussian, Constant
-    >>> Brick.lazy = False
-    >>> linear = Linear(5, 3, weights_init=IsotropicGaussian(),
-    ...                 biases_init=Constant(0))
-    >>> linear.apply(x)
-    linear_apply_output
-
-    .. doctest::
-       :hide:
-
-       >>> Brick.lazy = True  # Reset for other doctests
-
     """
-    #: See :attr:`Brick.lazy`
-    lazy = True
     #: See :attr:`Brick.print_shapes`
     print_shapes = False
 
@@ -819,11 +794,17 @@ def lazy(allocation=None, initialization=None):
         initialization = []
 
     def lazy_wrapper(init):
+        def strict_merge(values):
+            if len(values) > 1:
+                raise ValueError
+            return first(values)
+
         def lazy_init(*args, **kwargs):
             self = args[0]
             self.allocation_args = allocation
             self.initialization_args = initialization
-            kwargs = merge(args_to_kwargs(args, init), kwargs)
+            kwargs = merge_with(strict_merge, args_to_kwargs(args, init),
+                                kwargs)
             for allocation_arg in allocation:
                 kwargs.setdefault(allocation_arg, NoneAllocation)
             for initialization_arg in initialization:
--- a/blocks/bricks/base.py
+++ b/blocks/bricks/base.py
@@ -609,20 +609,15 @@ class Brick(Annotation):
         reset the parameters.
 
         """
+        if any(getattr(self, arg) is NoneAllocation
+               for arg in self.allocation_args):
+            raise ValueError('allocation config not set')
         if not self.allocation_config_pushed:
             self.push_allocation_config()
         for child in self.children:
             child.allocate()
         self.params = []
-        try:
-            self._allocate()
-        except Exception:
-            if self.lazy:
-                reraise_as("Lazy initialization is enabled, so please make "
-                           "sure you have set all the required configuration "
-                           "for this method call.")
-            else:
-                raise
+        self._allocate()
         self.allocated = True
 
     def _allocate(self):
@@ -649,21 +644,16 @@ class Brick(Annotation):
         call the :meth:`allocate` method in order to do so.
 
         """
+        if any(getattr(self, arg) is NoneInitialization
+               for arg in self.initialization_args):
+            raise ValueError('initialization config not set')
         if not self.allocated:
             self.allocate()
         if not self.initialization_config_pushed:
             self.push_initialization_config()
         for child in self.children:
             child.initialize()
-        try:
-            self._initialize()
-        except Exception:
-            if self.lazy:
-                reraise_as("Lazy initialization is enabled, so please make "
-                           "sure you have set all the required configuration "
-                           "for this method call.")
-            else:
-                raise
+        self._initialize()
         self.initialized = True
 
     def _initialize(self):
--- a/blocks/bricks/base.py
+++ b/blocks/bricks/base.py
@@ -8,11 +8,10 @@ import six
 from six import add_metaclass
 from theano import tensor
 from theano.gof import Variable
-from toolz import first, merge_with
 
 from blocks.graph import add_annotation, Annotation
 from blocks.roles import add_role, PARAMETER, INPUT, OUTPUT
-from blocks.utils import pack, repr_attrs, reraise_as, unpack
+from blocks.utils import dict_union, pack, repr_attrs, reraise_as, unpack
 from blocks.utils.containers import AnnotatingList
 
 
@@ -797,19 +796,13 @@ def lazy(allocation=None, initialization=None):
         initialization = []
 
     def lazy_wrapper(init):
-        def strict_merge(values):
-            if len(values) > 1:
-                raise ValueError
-            return first(values)
-
         def lazy_init(*args, **kwargs):
             self = args[0]
             self.allocation_args = (getattr(self, 'allocation_args',
                                             []) + allocation)
             self.initialization_args = (getattr(self, 'initialization_args',
                                                 []) + initialization)
-            kwargs = merge_with(strict_merge, args_to_kwargs(args, init),
-                                kwargs)
+            kwargs = dict_union(args_to_kwargs(args, init), kwargs)
             for allocation_arg in allocation:
                 kwargs.setdefault(allocation_arg, NoneAllocation)
             for initialization_arg in initialization:
--- a/blocks/bricks/base.py
+++ b/blocks/bricks/base.py
@@ -8,6 +8,7 @@ import six
 from six import add_metaclass
 from theano import tensor
 from theano.gof import Variable
+from toolz import merge
 
 from blocks.graph import add_annotation, Annotation
 from blocks.roles import add_role, PARAMETER, INPUT, OUTPUT
@@ -555,6 +556,9 @@ class Brick(Annotation):
         self.children = []
         self.parents = []
 
+        self.allocation_args = []
+        self.initialization_args = []
+
         self.allocated = False
         self.allocation_config_pushed = False
         self.initialized = False
@@ -768,64 +772,76 @@ class Brick(Annotation):
         return [self.get_dim(name) for name in names]
 
 
-def lazy(func):
+def args_to_kwargs(args, f):
+    arg_names, vararg_names, _, _ = inspect.getargspec(f)
+    return dict((arg_name, arg) for arg_name, arg
+                in zip(arg_names + [vararg_names], args))
+
+
+class LazyNone(object):
+    def __init__(self, name):
+        self.name = name
+
+    def __repr__(self):
+        return self.name
+
+    def __bool__(self):
+        return False
+
+    __nonzero__ = __bool__
+
+NoneAllocation = LazyNone('NoneAllocation')
+NoneInitialization = LazyNone('NoneInitialization')
+
+
+def lazy(allocation=None, initialization=None):
     """Makes the initialization lazy.
 
-    Any positional argument not given will be set to ``None``. Positional
-    arguments can also be given as keyword arguments.
+    This decorator allows the user to define positional arguments which
+    will not be needed until the allocation or initialization stage of the
+    brick. If these arguments are not passed, it will automatically replace
+    them with a custom ``None`` object. It is assumed that the missing
+    arguments can be set after initialization by setting attributes with
+    the same name.
 
     Parameters
     ----------
-    func : method
-        The __init__ method to make lazy.
+    allocation : list
+        A list of argument names that are needed for allocation.
+    initialization : list
+        A list of argument names that are needed for initialization.
 
     Examples
     --------
     >>> class SomeBrick(Brick):
-    ...     @lazy
+    ...     @lazy(allocation=['a'], initialization=['b'])
     ...     def __init__(self, a, b, c='c', d=None):
     ...         print(a, b, c, d)
     >>> brick = SomeBrick('a')
-    a None c None
+    a NoneInitialization c None
     >>> brick = SomeBrick(d='d', b='b')
-    None b c d
-    >>> Brick.lazy = False
-    >>> brick = SomeBrick('a')
-    Traceback (most recent call last):
-      ...
-    TypeError: __init__() missing 1 required positional argument: 'b'
-
-    .. doctest::
-       :hide:
-
-       >>> Brick.lazy = True  # Reset for other doctests
+    NoneAllocation b c d
 
     """
-    arg_spec = inspect.getargspec(func)
-    arg_names = arg_spec.args[1:]
-    defaults = arg_spec.defaults
-    if defaults is None:
-        defaults = []
+    if not allocation:
+        allocation = []
+    if not initialization:
+        initialization = []
 
-    def init(self, *args, **kwargs):
-        if not self.lazy:
-            return func(self, *args, **kwargs)
-        # Fill any missing positional arguments with None
-        args = args + (None,) * (len(arg_names) - len(defaults) -
-                                 len(args))
-
-        # Check if positional arguments were passed as keyword arguments
-        args = list(args)
-        for i, arg_name in enumerate(arg_names[:len(arg_names) -
-                                               len(defaults)]):
-            if arg_name in kwargs:
-                if args[i] is not None:
-                    raise ValueError
-                args[i] = kwargs.pop(arg_name)
-
-        return func(self, *args, **kwargs)
-    wraps(func)(init)
-    return init
+    def lazy_wrapper(init):
+        def lazy_init(*args, **kwargs):
+            self = args[0]
+            self.allocation_args = allocation
+            self.initialization_args = initialization
+            kwargs = merge(args_to_kwargs(args, init), kwargs)
+            for allocation_arg in allocation:
+                kwargs.setdefault(allocation_arg, NoneAllocation)
+            for initialization_arg in initialization:
+                kwargs.setdefault(initialization_arg, NoneInitialization)
+            return init(**kwargs)
+        wraps(init)(lazy_init)
+        return lazy_init
+    return lazy_wrapper
 
 
 class ApplicationCall(Annotation):
--- a/blocks/bricks/base.py
+++ b/blocks/bricks/base.py
@@ -531,9 +531,6 @@ class Brick(Annotation):
         self.children = []
         self.parents = []
 
-        self.allocation_args = []
-        self.initialization_args = []
-
         self.allocated = False
         self.allocation_config_pushed = False
         self.initialized = False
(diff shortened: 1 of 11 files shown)


dwf:
:+1: Like, very much.

bartvm:
So there was a small bug in there that causes bricks that used `super().__init__` to lose their `allocation_args`, so I fixed that. I also made it "optional" in the sense that you can write a brick without using the `@lazy` decorator, and it shouldn't crash.

Lastly, `Initializable` should technically not have `weights_init` as a compulsory argument. `Initializable` bricks pass their initialization configuration on *if they have it*. I ran into this case with `ConvolutionalSequence`: You can either configure it on the sequence, and pass it on to the components, or you can just set it on the children directly. In the second case, the sequence won't have `weights_init` set, but everything is fully configured anyway...

We should either add `weights_init` to the `@lazy` decorator of the bricks that actually use it in their `initialize` method, or we should split `Initializable` into two: Those that pass on weights configuration on, and those that use weights configuration.

fvisin:
Very nice indeed, thank you Bart! 
I am sorry I couldn't help you updating the @lazy annotations, I didn't understand you needed it done so soon.

I would imagine that most of the bricks will  have `weights_init` as a compulsory argument, but also that there will be other cases similar to `ConvolutionalSequence`. In this case maybe the best solution is to be flexible and split `Initializable` in two. Having a superclass of `Initializable` with less restrictive assumptions doesn't seems like a bad idea to me.

bartvm:
Ugh, second annoying corner case:

`Parallel` requires `input_dims` in order to allocate. Its child `Fork` requires `input_dim` and uses it to set `input_dims` in `push_allocation_config`. Currently this crashes: `Fork` refuses to call `push_allocation_config` because it thinks it needs `input_dims`, but it really doesn't.

One naive way of solving this would be to first call `push_allocation_config`, and *then* check whether everything is configured before calling `allocate`. However, this can still lead to strange errors if `push_allocation_config` tries to iterate over `NoneAllocation` objects for example...

Another naive way would be to not inherit necessary configuration definitions from the parent brick in the class hierarchy, but this makes lazy initialization useless in most cases.

The last solution is simple and my suggestion: If a brick sets configurations required by its parent in `push_allocation_config`, it must set the configuration to `None` explicitly after calling `super`.

fvisin:
Although it is a simple solution, I don't like the idea of having as a rule a "workaround" that has to be read of in the docstring of the lazy decorator or in the documentation. I can easily imagine users failing to create a new brick and also failing to find the _intended way_ of dealing with the situation. 

I like better the idea of having two optional parameters `ignore_inherited_allocation` and `ignore_inherited_initialization` (maybe something less verbose) that specify which inherited initialization/allocation will be handled in a custom way. I am not saying I don't like your idea at all, but I think that the Zen of Python applies here (Explicit is better than implicit).

bartvm:
But also "Simple is better than complex." The Zen of Python is awesome because you can justify anything with it :)

On a more serious note: This is a relatively minor corner case, and there's an argument here which I tend to dislike: "The user won't read the documentation, so let's foresee and handle all possible mistakes the user can make." Although that argument has its merit, in many cases, especially the more advanced ones, I think we should stop taking responsibility for a user not reading the documentation carefully.

If this really does turn into a recurring problem, with users complaining about it, we can always reconsider. For now, let's not overengineer things. The point of this PR was mostly to remove lines, not to write new ones.

fvisin:
Ahahah! You beat me at my own game! What can I say? :)

More seriously, in general I agree that we should take the user accountable if he doesn't read the documentation. In this case though I think that being the parameters optional they wouldn't make the general case more verbose, but it would be clear what to do in the corner case. Importantly, all of this could be achieved with few lines of code. That said, given that you beat me at the Zen of python game, if you still think it's better the other way I won't insist. ;)

bartvm:
Practicaly beats purity ;) So I'd prefer to defer the discussion, maybe make an issue out of it so that we don't forget.

Meanwhile, ready for review by whoever wants to have a closer look!

rizar:
I should say that I like the new interface, and maybe we actually should have something like this.

I am a not fan of forcing they class attributes to have the same names as constructor arguments. This violates encapsulation and is a pretty shady restriction. I would prefer something like that instead:

- the same lazy decorator replaces all not given lazy arguments with instances of `MissingAllocationOption` or `MissingInitializationOption`. Different from the current `AllocationNone` those would have the constructor argument name and the brick object as attributes. Whenever these objects are used in any context an exception like this would be thrown: 

```
AllocationError: missing the "input_dim" argument to the "fork" brick.

You get this error because some of lazy constructor arguments was not given and the corresponding
configuration actions were not taken. Blocks can not allocate parameters without you telling it how to do it!
```

If don't miss anything, that would also solve all the issues you have here. Those `MissingAllocationOption` that were never used will not trigger an error, `ConvolitionalSequence` is happy. Those that will be overridden by the subclass brick ('input_dims`) will not cause any issues as well.

bartvm:
Mm, not a bad idea, but a few concerns:

* We're back to semi-allocated and semi-initialized bricks, because we won't know whether we have the needed configuration until we try and fail.
* Writing these `MissingAllocationOption` objects is messy. You want them to raise an error whatever the code tries to do with them, but that could be accessing attributes, accessing keys, iterating over, testing boolean values, getting length, comparing to another object, used in arithmetic operations, used in NumPy/Theano, etc. You would need to intercept any possible call made to any of these actions to make sure you can raise this sensible error. Is there an easy way to do this? If you miss one, there's a chance again that your code silently does the wrong thing.

I've actually always found the fact that lazy arguments need to be set as attributes later on a feature instead of a restriction. "Either pass `foo` to the constructor, or otherwise set `bar` later on" becomes overly complicated. Lazy construction is already a bit dodgy, I like keeping it simple by really just forcing it to be deferred construction: Either pass `foo` now or set it later. I don't see the link with encapsulation by the way?

I thought about the problem with `Fork` that I ran into a bit more, but I actually don't think the current solution is a "workaround" as much as @fvisin suggested: If a parent brick takes responsibility for configuring a particular setting of one of its children, it needs to signal this in some way. The way to do so is by explicitly setting the configuration to something different from `NoneAllocation`.

rizar:
> We're back to semi-allocated and semi-initialized bricks, because we won't know whether we have the needed configuration until we try and fail.

Here I am not sure I follow. Isn't is the same in the current code of this PR? It seemed to me that you check all the attributes in `allocate`, which is exactly the point where my `MissingOption`'s would start panic.

> Writing these MissingAllocationOption objects is messy.

Indeed that's not super easy. Here is what I came up with:

```
class MissingOption(object):

    def __getattr__(self, name):
        self.panic()

    def panic(self, *args, **kwargs):
        raise Exception("I am missing option")

# The full list can be taken from
# https://docs.python.org/2/reference/datamodel.html#special-method-names
for name in ['__len__', '__add__', '__eq__']:
    setattr(MissingOption, name, MissingOption.panic)

mo = MissingOption()

# All of these work as expected
mo == 1
1 == mo
len(mo)
mo + 1

# Found no way to make this crash.
hasattr(mo, 'a')
```

That is if one has his own checks of allocation parameters validity, those will prevail, which I find tolerable.

> Either pass foo now or set it later. I don't see the link with encapsulation by the way?

What I meant is that it forces the interface (the constructor arguments) and the implementation (the objects attributes) to be the same. In a way the attributes are also a part of interface, so that was not the most successful argument.

However I still find it too restrictive and don't like it. Having two `Initializable` bricks is only one of sacrifices that have to be done to make this work, and it alone looks rather confusing. I will try come up with more concrete examples.

bartvm:
> Here I am not sure I follow. Isn't is the same in the current code of this PR? 

If `MissingOption` appears somewhere at the end of the `_allocate` function, everything before that will be executed. By checking explicitly, the `_allocate` function will never be called if one of the configurations isn't set.

> Having two Initializable bricks is only one of sacrifices that have to be done to make this work

I actually think that this is a bug either way: It doesn't make sense to have `weights_init` as a positional argument if it's in fact optional. It's just that this PR brings this problem more to the foreground.

> However I still find it too restrictive and don't like it.

I believe we understand lazy bricks differently then. This is how I see them: We want to avoid the situation where we need to pass the entire configuration for construction/allocation/initialization in one method call, because we sometimes have complex interdependencies of configuration between bricks. We address this is by saying: "Look, we'll take the arguments from `Brick.allocate` and move them to `Brick.__init__`. You can either pass them there, or you can set them anytime between calling `Brick.__init__` and calling `Brick.allocate`." By that reasoning, it makes perfect sense that attributes need to have exactly the same name as the arguments (has in my mind they've always had to).

blocks/bricks/base.py
```diff
@@ -512,7 +505,7 @@ class Brick(Annotation):
 
     Examples
     --------
-    By default, bricks have lazy initialization enabled.
+    Most bricks have lazy initialization enabled.
```
fvisin: Shouldn't this be _all_ bricks now?

fvisin:
I'd like to share an idea. The premise is that it just popped into my mind and it has probably some serious limitation or drawbacks that I am not taking into account. Still, I think it might be worth discussing so....here it is.

We could use the decorator to remove from the object the attributes that have not been set in the constructor (i.e., instead of setting them to `NoneAllocation` and `NoneInitialization`, we remove them). We would then keep two lists of the "suppressed attributes", one for allocation and one for initialization. We can then override __setattr__ and re-add the attribute to the object when the user sets its value (if in the list). This would allow us to call `push_allocation_config`, and _then_ to check if everything has been configured before calling `allocate`. And this check would be as simple as checking if the list is empty. What do you think about it?

rizar:
@bartvm , I will sidestep the complex ideological question for now since I have another example in mind: quite often I do something like this:

```
attention = AttentionMechanism(

```

rizar:
@bartvm , I will sidestep complex ideological questions for now since I have another example in mind: quite often I do something like this:

```python
seq_gen = SequenceGenerator(..., Readout(merge=MLP([None, 500, 500, None])))
```

Normally these `None`'s are to be filled by parents, but once in a while something goes wrong and one gets an exception from `shared_floatx_nan`. Your security checks do not work in such case, whereas the concept of `MissingAllocationOption()` is still applicable:

```python
seq_gen = SequenceGenerator(..., Readout(merge=MLP([
    MissingAllocationOption(), 500, 500, MissingAllocationOption()])))
# This can even remember the stack trace when `MissingAllocationOption` was created.
```

That looks quite bulky, but a syntax sugar to make shorter can be always invented. The main point is that "delayed exception" is a much more flexible paradigm of detecting missing configuration.

blocks/bricks/base.py
bartvm: Nope, a user can technically write his own brick without the `@lazy` decorator. I think this should be allowed (and I actually edited the code specifically to do so) because it can make for easier development when quickly hacking a brick together e.g. you can just do
```python
StupidBrick(Brick):
    def __init__(self, in, out):
        self.parameters.append(theano.shared(np.random.rand((in, out))))

    @application
    def apply(self, x):
        return x + self.parameters[0]
```

bartvm:
If the user wants to pass a complicated object that raises an error when used, he can do that anyway, this PR isn't stopping the user from doing that, nor does your suggestion solve the problem you point out without requiring some sort of user intervention.

"Delayed exception" is more flexible, but it's also significantly more complicated. My paradigm is: Missing arguments can be set as attributes before calling `allocate`/`initialize`. From the signature of `__init__` and the `@lazy` decorator the user can see *what* they need to set and *when* they need to set it. It's clean and simple. Your solution seems far more complex. If a user doesn't pass a particular argument, it is replaced with a special object, or the user needs to pass this special object manually (introducing more custom concepts the user needs to be familiar with). When and how the user then gets rid of this object needs to be deduced from the code and/or docstring, and could require intimate knowledge of the brick.

rizar:
> If the user wants to pass a complicated object that raises an error when used, he can do that anyway, this PR isn't stopping the user from doing that, nor does your suggestion solve the problem you point out without requiring some sort of user intervention.

Well, this some sort of user intervention is very basic: he just should use the right `None`. And it's not such a complicated object, see my implementation above. Of course this "right None" trick can be used in addition to your checks in `allocate`, but we should we have two ways of doing the same thing when one of the two is anyway required and covers all cases?

> "Delayed exception" is more flexible, but it's also significantly more complicated. 

This is very subjective. I find it very unnatural to bind object attributes and constructor arguments. 

> My paradigm is: Missing arguments can be set as attributes before calling allocate/initialize. From the signature of __init__ and the @lazy decorator the user can see what they need to set and when they need to set it. 

As I said I like your new syntax, but I also propose to use it and provide the user with the same information. What differs is the implementation method and yours has certain limitations, which I find quite important. 

> It's clean and simple. Your solution seems far more complex. If a user doesn't pass a particular argument, it is replaced with a special object, or the user needs to pass this special object manually 

Note, only in the case when your solution is helpless.

> (introducing more custom concepts the user needs to be familiar with).

Your solution introduces a non-standard constraint the user has to be familiar with: attribute names and keyword arguments must be same. Learning about either of these two takes 3 minutes of reading docs.

>  When and how the user then gets rid of this object needs to be deduced from the code and/or docstring, and could require intimate knowledge of the brick.

This is a general issue of lazy initialization though. You never know when a particular missed allocation or initialization setting actually arrives to your brick. 

Yes, one would have to read the `Attributes` section to find out how to configure the brick after allocation, similarly to what we have now. In most cases the section will repeat `Parameters` section, but this is standard Python situation. We are not developing a new DSL and you are the one who protects standard Python practices here, remember #489. 

To sum it up: your solution interferes in Python object life-cycle, whereas what I propose is just a light-weight addition on top of fully lazy bricks that tries to catch the error and explain it to the user in a meaningful way. It can be even implemented without decorator like it was originally proposed in #31.

dwf:
> I find it very unnatural to bind object attributes and constructor arguments.

In general, I agree. But note that @bartvm's solution isn't doing it in general, only for allocation and initialization arguments. This is the way it usually is done now, as far as I can tell: if the user wants to specify weights initialization after construction they set the `weights_init` attribute.

I think I may have a way to make @rizar's proposal more tolerable to you, @bartvm: what if we just checked for these special `MissingWhatevers` in the base class `allocate` and `initialize` methods? That would (as far as I can tell) obviate the need for any real use of them in user code.

bartvm:
> This is the way it usually is done now, as far as I can tell: if the user wants to specify weights initialization after construction they set the weights_init attribute.

That's exactly the behavior I think we should follow. Asking a user to pass `weights_init` *or* set a completely different attribute afterwards is counter-intuitive and requires the user to read through every docstring to see what attributes to set for the arguments they didn't pass.

> what if we just checked for these special `MissingWhatevers` in the base class allocate and initialize methods? 

I'm not sure I understand your proposal. If I understand @rizar's suggestion, he never wants to check for them. We just wait to see if the `_allocate` and `_initialize` method end up using an attribute that somehow contains these objects, and if they do, we hope that they get used in such a way that they raise an error (e.g. doesn't use `hasattr`).

If you are proposing to check whether particular attributes have the value `MissingWhatever` during `allocate` and `initialize` before calling `_allocate`/`_initialize`, then this is exactly what I'm doing. But to know which attribute values to check, you need to have a correspondence between the arguments and the attributes (since you can't just check all attributes, because some of them need to be set for allocation, some for initialization).

rizar:
`MissingWhatever` turned out to be a bad idea: C code from numpy is very hard to fool and all feedback you get when calling `numpy.zeros(whatever)` is `TypeError: an integer is required`. Their check types in a rather tough way [here](https://github.com/numpy/numpy/blob/master/numpy/core/src/multiarray/conversion_utils.c#L771), and have no desire whatsoever to dig that way.

On the other hand this PR does not do the job as well. Apart from `None`'s in lists that I have already mentioned, consider the case of `biases_init`. It is only required when `has_bias == True`, that is it can not be mentioned in the decorator. That is we have to have a separate sanity check for it in e.g. `Linear` if we don't want to have uninformative error messages like those we have now. 

That being said, I think the only place where reliable security checks can be placed are `_allocate` and `_initialize` methods. And the only one who knows how to do it is the `Brick` author. I think we should not waste our time for partial solution of this problem because of our "lean and mean motto" and to avoid code bloat.

However, I find it very nice that the current `@lazy` decorator (a) separates actually mandatory arguments and those that can be provided later (b)  indicates clearly when is the deadline for an option to be provided. We could have it also insert information which argument of which constructor it was in the `AllocationNone` values it creates, and those could be used to form a nice error message in `_alllocate` and `_initialize`. The `_initialize` routine from `Linear` could look like this:

```
def _initialize(self):
    if self.use_bias:
        if not isinstance(self.biases_init, InitializationScheme):
            complain(self.biases_init)
        W, b = self.params
        self.biases_init.initialize(b, self.rng)
    else:
        W, = self.params
    if not isinstance(self.weights_init, InitializationScheme):
        complain(self.weights_init)
    self.weights_init.initialize(W, self.rng)
```

where `complain` would use information from `self.biases_init` to create a nice error message. One could also explore `self.biases_init` in `pdb` to try to understand what the hell is happening.

I would be happy as well if just get rid of `lazy` decorator altogether. Lack of manpower has been mentioned hundred of times recently, but still I feel we avoid simple solutions at all costs...

bartvm:
I'm not a fan of patching the `__init__` function of each `Brick` automatically, I'd rather keep that explicit by requiring a decorator. The other alternative is for people to write something like `def __init__(self, input_dim=NoneAllocation, output_dim=NoneAllocation)`, but that will look horrendous for bricks like `ConvolutionalLayer`, so I prefer separating that information out into a decorator.

Of course the precise arguments can only be fully checked by the `Brick` itself, but there's a direct parallel here with normal functions:
```python
def foo(a, b=1, c=2):
    if b == c:
        raise ValueError("b and c cannot have the same value")
```
In thise case, the Python function pre-emptively checks whether `a` was provided, because it was compulsory, and calling `foo()` will obviously crash. If we call `foo(1, 1, 1)` it will only crash inside the function.

What my decorator does is simply mirror this precise behaviour: If an argument isn't given (either in `__init__` or later set as an attribute) it crashes pre-emptively. If the actual *value* of the argument is wrong (like `None` in a list) or if several arguments are values that are incompatible, these checks can clearly only be performed inside of the function. This is why I disagree with the statement that I'm coming up with an entire DSL: I'm simply proposing to more strictly mirror Python's default behaviour in our lazy constructor.

rizar:
> I'm not a fan of patching the __init__ function of each Brick automatically

Sorry, where have I proposed it?

bartvm:
You said you would want to get rid of the decorator entirely. In that case there are two options to do lazy initialization that I can see, which I both addressed: Patching `__init__` automatically, or expecting the signature to contain the `NoneAllocation`/`NoneInitialization` defaults.

rizar:
No, what I meant is that we just use `weights_init=None` for all optional parameters in the good old Python way. That is `Linear` would become

```
    def __init__(self, input_dim=None, output_dim=None, **kwargs):
        super(Linear, self).__init__(**kwargs)
        self.input_dim = input_dim
        self.output_dim = output_dim
```

That way we do not reinvent the wheel and do not add any new concepts to Python.
In Python there are mandatory parameters and optional parameters. Now we try to add some allocation parameters, initialization parameters, which are something in between. Should we really do that? If one can skip a parameter when calling a constructor, then it is optional and a default value must be provided in the constructor signature, period. Let's be lean and mean. 

Now that you added the `parent` links debugging missing configuration becomes much easier, one can trace the issue upwards. Also a minimalistic addition to the code above would be to create custom none objects every time that remember where they were created:

```
    def __init__(self, input_dim=NoneA(), output_dim=NoneA(), **kwargs):
        super(Linear, self).__init__(**kwargs)
        self.input_dim = input_dim
        self.output_dim = output_dim
```

bartvm:
> Also a minimalistic addition to the code above would be to create custom none objects every time that remember where they were created:

That's not too useful though, because you can only remember the line in which you were created, and that will be the same line that every other possibly missing configuration was created.

In general, I'm ambivalent about your proposal. I like the fact that it removes the need for custom decorators, but it ruins the signature of the constructor. There is now way to tell from the constructor anymore what needs to be provided and what doesn't. Each docstring will need to minutely detail this, and currently none of the docstrings do this (even though we said we would). This is actually one of the problems this new decorator addresses: Making it immediately clear what needs to be provided when. 

For `Linear` your solution this isn't too bad, but do we really want
```python
class ConvolutionalLayer(Brick):
    def __init__(self, activation, filter_size=None, num_filters=None, pooling_size=None,
                 num_channels=None, conv_step=(1, 1), pooling_step=None,
                 batch_size=None, image_size=None, border_mode='valid',
                 **kwargs):
```
Now it's impossible to tell from the signature that `num_channels` and `pooling_size` are needed, but `pooling_step` and `batch_size` are optional.

> That way we do not reinvent the wheel and do not add any new concepts to Python.

But we are anyway. Keyword arguments in Python are *optional*. In our case they aren't. Both solutions to the problem are equally ugly IMHO:

1. There can be a discrepancy between the signature (which suggests `num_filters` is required) and the functional call (not passing it works), which is what we currently do. This PR proposes making this discrepancy clearer by having a decorator which shows *which* arguments are optional until *when*.

2. Your suggest of getting rid of `@lazy` entirely introduces a discrepancy between the signature and the actual configuration of the brick. The default keyword arguments would suggest that the brick should be configured fully after constructing it, but the documentation needs to be read to see that this isn't true at all: The brick is worthless until a bunch of attributes have been set.

Both approaches mess with the way the objects are normally constructed in Python. The advantage of the `@lazy` decorator has always been that at least the signature makes sense. This new decorator adds even more information to the signature. Your suggestion would make the actual call more sensible, but it removes all information from the code that would show the user what is the necessary configuration.

bartvm:
Rebased this and wondering whether we can merge.

Regardless of what is the optimal solution, I am convinced that this is at least better than what we have right now. The only restriction it introduces is that arguments must correspond directly to attributes, which is a good thing I think, and 99% of bricks did this anyway (only `Fork` didn't do this).

dwf:
This limitation, while indeed restrictive, means there's a standard way to
set such attributes after construction that is discernable from merely
inspecting the call signature. So I'm +1 for merging.

blocks/bricks/base.py
```diff
@@ -768,64 +740,84 @@ def get_dims(self, names):
...
-        return func(self, *args, **kwargs)
-    wraps(func)(init)
-    return init
+    if not allocation:
+        allocation = []
+    if not initialization:
+        initialization = []
+
+    def lazy_wrapper(init):
+        def strict_merge(values):
```
rizar: That's not nice to use `strict_merge` here when in the rest of the library so far `dict_union` is used. When you write a replacement and try to use it you will mostly likely only take care about `dict_union` occurences and forget about this one.

rizar:
Okay, let's merge. I have mixed feelings about this PR, but we can give it a try.

blocks/bricks/base.py
bartvm: Oops, forgot about this one, just put it in there to test.

PR #537, opened 2015-03-24
Rename axes->axis for consistency.

dwf:
Also accepts singleton integers via pack().


dwf:
cc @vdumoulin

dwf:
Err, the discussion was on #513.

PR #538, opened 2015-03-25
Detect MRG RNG variables

bartvm:
Following https://github.com/Theano/Theano/pull/2670 we can now make sure that the state of MRG RNGs don't show up as shared variables. Strictly speaking the `hasattr` check should be enough, but I'll leave the `isinstance` one in there for people who haven't updated their Theano.


bartvm:
@dwf Done.

PR #541, opened 2015-03-25
Switch to items instead of attributes for log entries

bartvm:
Spin-off of #476 (and an attempt to make that PR a bit less humongous). This switches from using attributes (`log[0].cost`) for log items to keys (`log[0]['cost']`), which I personally think is more standard; keys seem to be more apt for very mutable data. It also allows us to simply subclass `defaultdict` instead of having a large number of custom methods.

I also removed the functionality of setting defaults, which wasn't used anywhere, and significantly simplified the logging interface and classes. The interface is now simply:

* `log` behaves as a `defaultdict`, accepting `i > 0` keys and returning a `dict` as a result.
* `log[i]` is a dictionary
* `log.status` is a dictionary

### DIFF
--- a/blocks/extensions/__init__.py
+++ b/blocks/extensions/__init__.py
@@ -153,9 +153,9 @@ class Predicate(object):
 
     def __call__(self, log):
         if self.condition.endswith('epochs'):
-            entry = log.status.epochs_done
+            entry = log.status['epochs_done']
         else:
-            entry = log.status.iterations_done
+            entry = log.status['iterations_done']
         if self.condition.startswith('every'):
             return entry % self.num == 0
         else:
@@ -163,7 +163,7 @@ class Predicate(object):
 
 
 def has_done_epochs(log):
-    return log.status.epochs_done == 0
+    return log.status['epochs_done'] == 0
 
 
 def always_true(log):
@@ -363,7 +363,7 @@ class FinishAfter(SimpleExtension):
         super(FinishAfter, self).__init__(**kwargs)
 
     def do(self, which_callback, *args):
-        self.main_loop.log.current_row.training_finish_requested = True
+        self.main_loop.log.current_row['training_finish_requested'] = True
 
 
 class Printing(SimpleExtension):
@@ -377,7 +377,7 @@ class Printing(SimpleExtension):
         super(Printing, self).__init__(**kwargs)
 
     def _print_attributes(self, attribute_tuples):
-        for attr, value in sorted(attribute_tuples, key=first):
+        for attr, value in sorted(attribute_tuples.items(), key=first):
             if not attr.startswith("_"):
                 print("\t", "{}:".format(attr), value)
 
@@ -387,7 +387,7 @@ class Printing(SimpleExtension):
 
         print()
         print("".join(79 * "-"))
-        if which_callback == "before_epoch" and log.status.epochs_done == 0:
+        if which_callback == "before_epoch" and log.status['epochs_done'] == 0:
             print("BEFORE FIRST EPOCH")
         elif which_callback == "on_resumption":
             print("TRAINING HAS BEEN RESUMED")
@@ -403,7 +403,7 @@ class Printing(SimpleExtension):
             print("Training status:")
             self._print_attributes(log.status)
             print("Log records from the iteration {}:".format(
-                log.status.iterations_done))
+                log.status['iterations_done']))
             self._print_attributes(log.current_row)
         print()
 
@@ -460,7 +460,7 @@ class ProgressBar(TrainingExtension):
 
         """
         iter_per_epoch = self.get_iter_per_epoch()
-        epochs_done = self.main_loop.log._status.epochs_done
+        epochs_done = self.main_loop.log.status['epochs_done']
 
         if iter_per_epoch is None:
             widgets = ["Epoch {}, step ".format(epochs_done),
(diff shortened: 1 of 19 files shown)


bartvm:
Tests pass! Should be ready for review.

blocks/extensions/__init__.py
```diff
@@ -363,7 +363,7 @@ def __init__(self, **kwargs):
         super(FinishAfter, self).__init__(**kwargs)
 
     def do(self, which_callback, *args):
-        self.main_loop.log.current_row.training_finish_requested = True
+        self.main_loop.log.current_row['training_finish_requested'] = True
```
rizar: That should be come a upper-cased constant, I guess (`TRAINING_FINISH_REQUESTED`). Something I wanted to do since long, because otherwise typos only show themselves in runtime and also in an unpredictable way.

I am less concerned about reads like `log.status['epochs_done']` because a wrong read usually quickly let's one know about itself, but doing write with plain text strings is really dangerous.

blocks/extensions/__init__.py
```diff
@@ -550,44 +550,44 @@ def log(self):
...
-            epoch_ends = self.log.status._epoch_ends
-            self.log.status._epoch_before_interrupted = (
-                self.clock_function() -
-                0 if not epoch_ends else self.log[epoch_ends[-1]].total_took)
+        self.log.status['total_before_interrupted'] = (
+            self.log.current_row['final_total_took'])
+        if self.log.status['epoch_started']:
+            epoch_ends = self.log.status['_epoch_ends']
+            self.log.status['epoch_before_interrupted'] = (
+                self.clock_function() - 0
```
rizar: What is `- 0`?

blocks/log.py
```diff
@@ -10,265 +9,61 @@
...
+        super(TrainingLog, self).__init__(dict)
+        self.status = {
+            'iterations_done': 0,
+            'epochs_done': 0,
+            '_epoch_ends': []
+        }
 
-        """
-        pass
+    def __reduce__(self):
```
rizar: Why do we need `__reduce__` here?

blocks/log.py
```diff
@@ -10,265 +9,61 @@
...
-
-    @abstractmethod
-    def get_row_iterator(self, time):
-        """Returns an iterator over key-value pairs of a row of the log."""
-        pass
-
-    def _check_time(self, time):
-        if not isinstance(time, int) or time < 0:
-            raise ValueError("time must be a positive integer")
+        return self[self.status['_epoch_ends'][-1]]
```
rizar: Just remembered another reason why I wanted to have these `_epoch_ends`: to implement early stopping. This allow to quickly fetch your validation cost from all epochs and make a conclusion, whether training should be continued.

blocks/extensions/__init__.py
bartvm: I believe we've had this discussion before, but I think that would be a bit overly defensive. It introduces unnecessary new variables which need to be imported if you want to use them in other files (which people will probably just forget anyway), so you end up with a complicated web of import statements if you don't want to start redefining constants in each file (which would defeat the purpose). You end up with a whole bunch of them, which I think is ugly anyway, but it also seems arbitrary to use constants in some places and free strings in other. If you write to the wrong key, you'll now get a `KeyError` when you try to read it (instead of the `None` you used to get), so a typo should show itself quite quickly. In the end, catching typos is just one of the reasons that we do code reviews. For callbacks I think a custom `__eq__` was reasonable, because we're comparing against a limited, defined set of callbacks, but in this case I think we don't need to be as paranoid.

blocks/extensions/__init__.py
bartvm: I don't know, you wrote this code? I just moved it from the line below :)

blocks/log.py
bartvm: Because `defaultdict` like most mappings defines a custom `__reduce__` method, which doesn't save its `__dict__` as a state (because its assumed that the state is fully defined by reading `self.items()`. If you don't overwrite this, it will wipe out `self.status` every time you pickle it.

blocks/log.py
bartvm: In the other PR I replaced it with a `self.current_entry['epoch_ended'] = True` entry in the log, and that actually works just fine. You can just do `epoch_ends = [key for key, value in self.log.items() if 'epoch_ended' in value]`. Slightly longer than `epoch_ends = self.log.status['_epoch_ends']`, but a lot more consistent, since right now we're effectively saving log entries into the status for some reason.

blocks/extensions/__init__.py
rizar: > If you write to the wrong key, you'll now get a KeyError when you try to read it

That does not always help because sometimes `get` is used to access the value. That is if one writes a custom extension that should finish training at some point and makes a typo, he will only notice it two days later when his training does not stop when it should have stopped. 

@dwf, what do you think?

blocks/log.py
rizar: Right, but this is `O(n)` where `n` is the number of iterations done, and it restricts the frequency with which such an early stopping extension can be called. Anyway, I am fine with switching to `epoch_ended` later.

blocks/extensions/__init__.py
rizar: Ooos.. okay, then just leave it, we will see later.

bartvm:
@dwf @rizar Could one of you two have a look and merge this when you've got some time? It touches 19 files, so sooner or later there are going to be rebasing issues otherwise.

@rizar's comment in https://github.com/bartvm/blocks/pull/541#discussion_r27103287 that we should switch to constants instead of string keys is a problem that currently exists as well (and is actually partly alleviated by this PR), so I'd like to defer the discussion or implementation of further changes to another PR.

rizar:
Sure, I wanted to merge it, simply forgot about it, sorry!

PR #543, opened 2015-03-25
Remove unneeded Brick argument to ApplicationCall.

### LINKED ISSUE #540: ApplicationCall arguments
`ApplicationCall` takes two arguments, an `Application` and a `Brick`. But it sounds like it does not actually need the `Brick` and could just get this from the `.brick` attribute of the `Application`. Is that the case?

dwf:
Fixes #540.


rizar:
LGTM!

PR #549, opened 2015-03-25
Made ProgressBar work when no scheme

dmitriy-serdyuk:
ProgressBar now works when the top data stream doesn't have an iteration scheme. It searches parent streams with `num_batches` or `num_examples` attribute. I suppose, that it is legitimate because usually a transformer without an iteration scheme doesn't influence these attributes.

### DIFF
--- a/blocks/extensions/__init__.py
+++ b/blocks/extensions/__init__.py
@@ -444,7 +444,12 @@ class ProgressBar(TrainingExtension):
 
     def get_iter_per_epoch(self):
         """Try to infer the number of iterations per epoch."""
-        iter_scheme = self.main_loop.data_stream.iteration_scheme
+        data_stream = self.main_loop.data_stream
+        while (not hasattr(data_stream.iteration_scheme, 'num_batches') and
+               not hasattr(data_stream.iteration_scheme, 'num_examples') and
+               hasattr(data_stream, 'data_stream')):
+            data_stream = data_stream.data_stream
+        iter_scheme = data_stream.iteration_scheme
         if hasattr(iter_scheme, 'num_batches'):
             return iter_scheme.num_batches
         elif (hasattr(iter_scheme, 'num_examples') and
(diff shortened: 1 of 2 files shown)


dmitriy-serdyuk:
@jbornschein , please take a look.

jbornschein:
Looks good to me!

Although I think it would be cleaner if DataStreams would automatically "inherit" num_examples / num_batches from their parent streams or iterators. But that would require some refactoring (and some discussions) in Fuel.

-> Good to merge from my side.

bartvm:
I'm not convinced we should merge this, because it's quite hacky. The problem is that there is no saying whether a `Transformer` actually conserves the number of examples. A `Filter` or `Mapping` transformer for example can discard examples, and there is no way in knowing in advance how many it will discard or not, so in theory your progress bar might be completely off the mark.

My suggestion would be something along these lines: By default `Dataset`s, `Transformer`s and `DataStream`s return `None` for `num_examples` and `num_batches`, but each `Transformer` can override this behaviour, instead calculating these values based on whatever they're wrapping, if possible. This way they get passed through the chain, but intelligently, and if you have a `Filter` anywhere along the way, it could just be passing `None` from there on, because you just don't know.

You would then end up with something like
```python
>>> dataset = Dataset()
>>> dataset.num_examples 
100
>>> dataset.num_batches
None
>>> data_stream = DataStream(dataset, iteration_scheme=SequentialScheme(100, 10))
>>> data_stream.num_examples
100
>>> data_stream.num_batches
10
>>> transformed_data_stream = DropHalf(data_stream)
>>> transformed_data_stream.num_examples
50
>>> transformed_data_stream.num_batches
5
```
Anyway, you get the point. Some of this logic should go into the iteration schemes I guess.

rizar:
+1 that this is not reliable enough to make it to Blocks, and that the proper direction to go is to explore pulling information through the transformer chain.

PR #552, opened 2015-03-26
New timing context manager

bartvm:
This is WIP, but I wanted to optimize something, and realised that the current timing we do isn't very useful, so it was hard to tell where the main loop was spending time at all. I don't think it's feasible to put timing completely in an extension as we tried before; the results simply aren't reliable or precise enough. E.g. if you want to figure out which extension is taking up a lot of time, you need to time each extension separately.

This PR introduces a context manager that basically keeps a nested record of timings. (Right now this nested dictionary is actually slow to update, so I need to use a different data structure, but it's just a proof of concept right now.) This allows it to print a report like:

```
  before_training: 0.00019 (0.00%, 0.00%)
    Checkpoint: 7e-06 (3.66%, 0.00%)
    DataStreamMonitoring: 1.4e-05 (7.33%, 0.00%)
    Printing: 2.7e-05 (14.14%, 0.00%)
    FinishAfter: 2.2e-05 (11.52%, 0.00%)
    TrainingDataMonitoring: 5.6e-05 (29.32%, 0.00%)
  initialization: 0.19 (1.54%, 1.54%)
  training: 1.2e+01 (97.83%, 97.83%)
    epoch: 1.2e+01 (95.19%, 93.12%)
      before_batch: 1.3 (11.28%, 10.50%)
        Checkpoint: 0.072 (5.42%, 0.57%)
        DataStreamMonitoring: 0.13 (9.87%, 1.04%)
        Printing: 0.27 (20.60%, 2.16%)
        FinishAfter: 0.092 (6.98%, 0.73%)
        TrainingDataMonitoring: 0.12 (9.39%, 0.99%)
      after_batch: 1.4 (11.58%, 10.78%)
        Checkpoint: 0.073 (5.37%, 0.58%)
        DataStreamMonitoring: 0.14 (9.98%, 1.08%)
        Printing: 0.27 (20.15%, 2.17%)
        FinishAfter: 0.11 (8.45%, 0.91%)
        TrainingDataMonitoring: 0.13 (9.32%, 1.00%)
      batch: 7.6 (65.15%, 60.67%)
    before_epoch: 0.22 (1.80%, 1.76%)
      Checkpoint: 1.6e-05 (0.01%, 0.00%)
      DataStreamMonitoring: 0.22 (99.83%, 1.76%)
      Printing: 0.00018 (0.08%, 0.00%)
      FinishAfter: 2.7e-05 (0.01%, 0.00%)
      TrainingDataMonitoring: 3e-05 (0.01%, 0.00%)
    after_epoch: 0.37 (3.01%, 2.94%)
      Checkpoint: 2.1e-05 (0.01%, 0.00%)
      DataStreamMonitoring: 0.37 (99.71%, 2.93%)
      Printing: 0.00032 (0.09%, 0.00%)
      FinishAfter: 0.00017 (0.05%, 0.00%)
      TrainingDataMonitoring: 0.00031 (0.08%, 0.00%)
  after_training: 0.079 (0.63%, 0.63%)
    Checkpoint: 0.079 (99.62%, 0.63%)
    DataStreamMonitoring: 1.3e-05 (0.02%, 0.00%)
    Printing: 0.00019 (0.24%, 0.00%)
    FinishAfter: 1e-05 (0.01%, 0.00%)
    TrainingDataMonitoring: 1.2e-05 (0.02%, 0.00%)
```
It basically prints `total time spent (% time of parent, % time of total)` right now. The times don't really add up, because it actually spends a lot of time keeping track of the timings itself, which isn't counted :)

Anyway, todo:

- [x] Speed up tracking of total times
- [x] Sort timings sensibly
- [x] Print prettier
- [x] Write extension that prints report


rizar:
> This is WIP, but I wanted to optimize something, and realised that the current timing we do isn't very useful, so it was hard to tell where the main loop was spending time at all. I don't think it's feasible to put timing completely in an extension as we tried before; the results simply aren't reliable or precise enough. E.g. if you want to figure out which extension is taking up a lot of time, you need to time each extension separately.

Reliability and precision could be achieved, but there is no way to monitor how much time every single extension consumes. Having that said seems that timing has to be hardcoded in the main loop.

rizar:
In general that would be nice to have. One thing to which we should pay attention is as usual training resumption. 

Let me know when you want it to be reviewed.

bartvm:
Faster implementation now. Still need to work on the pretty printing, and clean up the code a bit, but it's functional now. The profile of `mnist.py` is interesting though. Notice how we actually only spend 74% of the time training and reading data... Might make it worthwhile to at one point sit down and optimize things a bit.

```
initialization: 0.19 (1.64%)
after_training: 0.078 (0.66%)
  Plot: 2.4e-05 (0.00%)
  TrainingDataMonitoring: 1.3e-05 (0.00%)
  Timing: 2e-05 (0.00%)
  DataStreamMonitoring: 1.4e-05 (0.00%)
  Printing: 0.00019 (0.00%)
  Checkpoint: 0.078 (0.66%)
  FinishAfter: 9e-06 (0.00%)
  other: 3.8e-05 (0.00%)
training: 1.2e+01 (97.70%)
  epoch: 1.1e+01 (93.04%)
    batch: 7.6 (64.46%)
    read_data: 0.94 (7.95%)
    before_batch: 0.98 (8.30%)
      Printing: 0.27 (2.25%)
      Plot: 0.12 (0.99%)
      Timing: 0.027 (0.23%)
      Checkpoint: 0.066 (0.56%)
      FinishAfter: 0.085 (0.72%)
      DataStreamMonitoring: 0.12 (1.04%)
      TrainingDataMonitoring: 0.12 (1.00%)
      other: 0.18 (1.51%)
    after_batch: 1.2 (10.10%)
      DataStreamMonitoring: 0.13 (1.08%)
      Plot: 0.12 (1.00%)
      Timing: 0.19 (1.63%)
      FinishAfter: 0.1 (0.88%)
      TrainingDataMonitoring: 0.12 (1.03%)
      Printing: 0.27 (2.26%)
      Checkpoint: 0.067 (0.57%)
      other: 0.2 (1.65%)
    other: 0.26 (2.23%)
  after_epoch: 0.37 (3.14%)
    Timing: 9.4e-05 (0.00%)
    DataStreamMonitoring: 0.35 (2.99%)
    TrainingDataMonitoring: 0.00027 (0.00%)
    Checkpoint: 1.9e-05 (0.00%)
    FinishAfter: 0.00014 (0.00%)
    Printing: 0.00039 (0.00%)
    Plot: 0.017 (0.14%)
    other: 0.00013 (0.00%)
  before_epoch: 0.18 (1.52%)
    DataStreamMonitoring: 0.17 (1.43%)
    Checkpoint: 1.5e-05 (0.00%)
    Plot: 0.0095 (0.08%)
    Printing: 0.00017 (0.00%)
    Timing: 3.3e-05 (0.00%)
    FinishAfter: 2.5e-05 (0.00%)
    TrainingDataMonitoring: 2.7e-05 (0.00%)
    other: 6.3e-05 (0.00%)
  other: 0.00018 (0.00%)
before_training: 0.00018 (0.00%)
  DataStreamMonitoring: 1.4e-05 (0.00%)
  Printing: 2.6e-05 (0.00%)
  FinishAfter: 1.6e-05 (0.00%)
  TrainingDataMonitoring: 7.2e-05 (0.00%)
  Timing: 1e-05 (0.00%)
  Checkpoint: 7e-06 (0.00%)
  Plot: 1.2e-05 (0.00%)
  other: 2.6e-05 (0.00%)
```

dwf:
This is fantastic. Killer feature I didn't know I wanted.

bartvm:
Fixed endless merge conflicts with #541 and rebased. Should now be fully functional except for documentation (and maybe tests, but not really sure how to test this besides just having it run), so ready for (initial) review.

The current output format is this. This is very much a style/preference thing, so any suggestions or comments are welcome:

```
Section                                  Time     % of total
------------------------------------------------------------
Before training                          0.00          0.00%
  Timing                                 0.00          0.00%
  Finishafter                            0.00          0.00%
  Datastreammonitoring                   0.00          0.00%
  Trainingdatamonitoring                 0.00          0.00%
  Checkpoint                             0.00          0.00%
  Plot                                   0.00          0.00%
  Printing                               0.00          0.00%
  other                                  0.00          0.00%
Initialization                           0.19          1.68%
Training                                11.39         98.32%
  Before epoch                           0.19          1.62%
    Timing                               0.00          0.00%
    Finishafter                          0.00          0.00%
    Datastreammonitoring                 0.18          1.54%
    Trainingdatamonitoring               0.00          0.00%
    Checkpoint                           0.00          0.00%
    Plot                                 0.01          0.08%
    Printing                             0.00          0.00%
    other                                0.00          0.00%
  Epoch                                 10.82         93.38%
    Read data                            0.93          8.03%
    Before batch                         1.06          9.19%
      Timing                             0.09          0.77%
      Finishafter                        0.07          0.61%
      Datastreammonitoring               0.12          1.04%
      Trainingdatamonitoring             0.12          1.02%
      Checkpoint                         0.07          0.57%
      Plot                               0.12          1.01%
      Printing                           0.27          2.30%
      other                              0.22          1.88%
    Batch                                7.43         64.15%
    After batch                          1.11          9.58%
      Timing                             0.11          0.96%
      Finishafter                        0.07          0.64%
      Datastreammonitoring               0.12          1.06%
      Trainingdatamonitoring             0.12          1.03%
      Checkpoint                         0.07          0.59%
      Plot                               0.12          1.01%
      Printing                           0.27          2.30%
      other                              0.23          1.97%
    other                                0.28          2.43%
  After epoch                            0.38          3.32%
    Timing                               0.00          0.00%
    Finishafter                          0.00          0.00%
    Datastreammonitoring                 0.37          3.16%
    Trainingdatamonitoring               0.00          0.00%
    Checkpoint                           0.00          0.00%
    Plot                                 0.02          0.15%
    Printing                             0.00          0.00%
    other                                0.00          0.00%
  other                                  0.00          0.00%
After training                           0.00          0.00%
  Timing                                 0.00          0.00%
  other                                  0.00          0.00%
```

bartvm:
* Changed the capitalization of the words e.g. `DataStreamMonitoring` instead of `Datastreammonitoring`.

bartvm:
Now with tests for `OrderedSet` so that the coverage shouldn't fall as dramatically.

@dwf @rizar @vdumoulin Any of you got time to have a look?

blocks/extensions/__init__.py
```diff
@@ -497,98 +496,30 @@ def before_batch(self, batch):
...
-        Return the current time. By default `time.time` is used,
-        which means that user time is tracked.
-
-    Notes
-    -----
-    When training is interrupted this extension saves intermediate
-    time measurements to the training status, i.e. it should be robust
-    to any training interruptions.
+    By default the extension prints the report at the end of the training.
+    Not that printing a report does not reset the timer.
```
vdumoulin: `Not -> Note`?

bartvm: Fixed.

vdumoulin:
Running the `mnist.py` example, I get a negative value for `After training/Other` which has the same absolute value as `After training/Checkpoint`.

rizar:
Added to my tomorrow's todo list.

bartvm:
@vdumoulin Nice catch. It's because when the extension gets run `after_training`, the timer is actually still in the middle of timing the "after training" segment, so it's equal to 0. However, it has finished timing the specific extensions preceding `Timing` (such as `Checkpoint`)...

Not sure how to solve that. Considering that this extension just prints to standard output, I'm inclined to suggest maybe doing something like Theano. We can create a configuration value that allows you to print the report at the end of a main loop either with `BLOCKS_PROFILE` or setting `blocks.config.profile = True`.

vdumoulin:
@bartvm That would work, I think.

bartvm:
Okay, done. Running with `BLOCKS_PROFILE=true` or setting `profile: True` in your configuration file will now print a report at the end of the main loop.

dwf:
Just a thought for somewhere down the line: there is a standard-ish format
format for profiling output, that can be read by a variety of graphical
tools. Might be nice to be able to dump to it, one day, eventually. Might
be overkill though.

http://kcachegrind.sourceforge.net/html/CallgrindFormat.html

bartvm:
Ah, interesting. I actually thought about dumping it to JSON or something, but wasn't sure how to visualize it. If there's already a standard format that could be nice, could be a nice CCW ticket sometime in the future I guess.

dwf:
It will require some reading as to whether the format actually meets our
needs, it's typically for more general code profiling whereas we are sort
of doing it a few steps up the semantic hierarchy.

blocks/main_loop.py
```diff
@@ -42,6 +45,87 @@
 a model. Check documentation of your extensions."""
 
 
+class Timer(object):
```
rizar: It seems to me that this is class is generic and not specific to the main loop and should go to a separate file. Also it needs a bit of docs to explain how it's different from other classes named `Timer`, e.g. `timeit.Timer` from the standard library. If I get right, it's special and very nice feature is that it's hierarchical, maybe it should be even called `HierachicalTimer`?

blocks/main_loop.py
```diff
@@ -42,6 +45,87 @@
...
+    def exit(self, t):
+        self.total[tuple(self.current)] += t
+        self.current.pop()
+
+    def report(self):
+        """Print a report of timing information to standard output.
+
+        .. todo::
+
+           This method could accept different I/O objects (e.g. standard
```
rizar: This todo should be completed before merge, I guess.

blocks/main_loop.py
```diff
@@ -42,6 +45,87 @@
...
+    def __init__(self, name, timer):
+        self.name = name
+        self.timer = timer
+
+    def __enter__(self):
+        self.timer.enter(self.name)
+        self.start = time.clock()
+
+    def __exit__(self, *args):
+        self.timer.exit(time.clock() - self.start)
```
rizar: You are using `time.clock()`, which is CPU time. I am slightly worried if it always correctly includes time spent on waiting for a GPU operation to finish. Do you know if it's accounted?

Also I think documentation should mention this choice.

blocks/main_loop.py
```diff
@@ -94,6 +178,8 @@ def __init__(self, algorithm, data_stream,
         self.log = log
         self.extensions = extensions
 
+        self.timer = Timer()
```
rizar: I am slightly unhappy that the timing information is not a part of the log. On the other hand thinking about putting it here with all this uncertainty we have in #476 is rather hard. Thoughts?

blocks/main_loop.py
```diff
@@ -42,6 +45,90 @@
...
+class Timer(object):
+    def __init__(self):
+        self.total = defaultdict(int)
+        self.current = []
+        self.order = OrderedSet()
+
+    def enter(self, name):
+        self.current.append(name)
+        # We record the order in which sections were first called
+        self.order.add(tuple(self.current))
```
rizar: I am not sure we need `OrderedSet` in Blocks only to support this one case. You can make `self.order` an `OrderedDict` and use `self.order[tuple(self.current)] = True` instead.

rizar:
Killing feature, as @dwf said, but some word is needed.

I am worried that currently timing information is only printed when training is finished. I think it's important to have some timing information printed on the go, like it was before with `total_took`, `initailization_took` and `iteration_took`. The challenge is that all the information is commited after all extensions for a callback were finished, which makes writing extensions using information from `self.timer` rather problematic.

bartvm:
> The challenge is that all the information is commited after all extensions for a callback were finished, which makes writing extensions using information from self.timer rather problematic.

Well, the timing of the individual extensions that would precede `Timer` in the list of extensions would already have been saved. But this is a problem that is pretty much inherent to timing. Pylearn2 solved it by reporting the timing of the *previous iteration* instead of the current one. I always found it a bit confusing though, because you have all the information of the current batch mixed in with a single datapoint from the previous batch.

blocks/main_loop.py
bartvm: Main issue with `OrderedDict` is that it's nearly twice as slow as `OrderedSet`, which can become significant if the user is performing batch-monitoring on a fast model (i.e. when `enter` gets called a lot). I'd have to do some benchmarking to see if that difference is noticeable.

blocks/main_loop.py
bartvm: Is there a particular use-case in mind that you think is really essential? I figured to just leave it for a future ticket.

rizar: The default destination for Theano profiling is stderr, which makes me want the same destination for Blocks profiling.

bartvm: Then I can just make that the default.

rizar: And then I load my main loop pickle in an IPython notebook and want to print report to standard output.

I think it's basic functionality and requires roughly 3 lines to support.

blocks/main_loop.py
bartvm: Ah, the docstring of the extension had this documentation, forgot to remove it when I copied. The Python docs say

>  in any case, this is the function to use for benchmarking Python or timing algorithms.

but I'll switch to `timeit.default_timer` which seems to be the better choice.

blocks/main_loop.py
bartvm: I agree the naming is awkward, I'll try and come up with something better, although I find `TimeAccounting` and `HierarchicalTimer` very obscure.

blocks/main_loop.py
bartvm: Already done.

bartvm:
@rizar This should address all your comments, except for printing to the log during training. There are two options I see:

* Only report on the actual time spent in `Algorithm.process_batch`, this information is available before the `after_batch` and `after_epoch` extensions get called.
* Report on statistics from the previous batch, or the previous epoch, in this case information about everything, including all the extensions, is available.

Can we move that to a subsequent PR though? I've got an assignment to finish this weekend, and I would argue this is already a big improvement over the current state of timing since the blanket number the current timing extension gives is useless to me, because I don't know whether it's spent in extensions, reading data or actually training.

rizar:
>  Only report on the actual time spent in Algorithm.process_batch

That would be useful and clean information. In addition the total time so far and the time initialization took would be very helpful.

> Report on statistics from the previous batch, or the previous epoch

As you said, this is not really intuitive.

> Can we move that to a subsequent PR though? I've got an assignment to finish this weekend, and I would argue this is already a big improvement over the current state of timing since the blanket number the current timing extension gives is useless to me, because I don't know whether it's spent in extensions, reading data or actually training.

Do not get me wrong, this is indeed a huge improvement, but also a slight step backward because there is completely zero information available during training. And that can hurt many people, including me, for which this information was an essential part of workflow. Considering that there is no rush for this extension, I would wait until at least something is written to the log before merging it. I don't see any issue in waiting a couple of days for it. 

Alternatively you can just keep `Timing` extension in `master` until we have something better. Put a warning that it's deprecated and go ahead with merging.

bartvm:
> Considering that there is no rush for this extension, I would wait until at least something is written to the log before merging it. I don't see any issue in waiting a couple of days for it.

I disagree. I really dislike having open PRs lying around which just need to be rebased continuously. In general, if something brings the `master` in a better state than it currently is it should be merged IMHO.

Either way, I added the timing extension back; I believe it reports similar data as before now.

rizar:
Sure, I agree that open PRs are annoying.

With the `Timing` extension added back,  I have no objections against merging.

bartvm:
Okay, as I think both @rizar and @vdumoulin have had a look, merging now!

Issue #474, opened 2015-03-14
Merge Dump and Checkpoint into a single extension?


bartvm:
This is more intended to be a long-term goal, but the custom pickler of `joblib` made me think that we could probably do something very similar as `joblib.dump` tha combiens the functionality of `Dump` and `Checkpoint`, which would remove a lot of duplicate functionality.

Just like `joblib` does we could intercept NumPy arrays and save them to disk instead of pickling them. The only functionality we need to add on top is giving those files a slightly more sensible name than `model.23423423.npy`. The same pickling class could give sensible errors for when unpicklable objects are encountered.

There are probably some details to iron out, but overall I think that on the mid/long-term this could be a very nice and clean solution to the duplicity we have right now with `Dump`, `Checkpoint` and a range of semi-informative pickling error messages.

dwf:
+1

rizar:
In fact `Checkpoint` and `Dump` can be merged even now, just adding a flag to `Checkpoint` making save `.npz` file as well should do the job. We should get rid of `LoadFromDump` - the rationale behind it was that we have to load the log after the algorithm as initialized, because algorithm would not be initialized when resuming training with an old log. Instead we should make algorithm initialization lazy and always called.

bartvm:
That's more of a stop-gap solution though, because from what I understand you want to save everything twice? Both as a large pickle file, and then again separately as an `.npz` file. What I'm proposing is a single, more intelligent pickling solution that automatically saves any NumPy array in the object hierarchy separately.

There shouldn't be any need then for forcing parameters to have a unique path/name in the brick hierarchy. You basically save them to files with a name like `uniqueid.parameter_name.npy`. You'll only have to ensure that parameter names are unique in cases where you want to distribute the NumPy arrays separately or transfer/inspect them easily.

This actually comes back to #480, because although forcing uniqueness on the short-term makes sense, if we have this solution, I think we shouldn't enforce it. Its' a pain in the ass to make sure every variable has a unique name when you're not using bricks, so if we can handle serialization without it, we should.

rizar:
> What I'm proposing is a single, more intelligent pickling solution that automatically saves any NumPy array in the object hierarchy separately.

But parameters are not numpy arrays, they are shared variables hosted on GPU. Do you think wil be able to save them as numpy arrays and in separate files and then load back at unpickling? 

As for name uniqueness, I never was a fan of this idea. It's pain in the ass, as you said.

bartvm:
[...]

PR #556, opened 2015-03-26
WIP: Unique names for bricks for pickling

dmitriy-serdyuk:
Conserns #474.

### DIFF
--- a/blocks/bricks/base.py
+++ b/blocks/bricks/base.py
@@ -774,8 +774,6 @@ class Brick(Annotation):
         else:
             return [self]
 
-    def get_full_name(self):
-        return '/' + '/'.join([brick.name for brick in self.get_unique_path()])
 
 def lazy(func):
     """Makes the initialization lazy.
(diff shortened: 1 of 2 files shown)


blocks/bricks/base.py
```diff
@@ -765,6 +766,16 @@ def get_dims(self, names):
...
 
+    def get_unique_path(self):
+        """Returns unique path to this brick in the application graph."""
+        if self.parents:
+            parent = min(self.parents, key=operator.attrgetter('name'))
+            return parent.get_unique_path() + [self]
+        else:
+            return [self]
+
+    def get_full_name(self):
```
bartvm: I would leave this method out, and we can move it to the pickler eventually. The precise string format to use might be better decided by the pickler or whatever consumer of the function, because it could depend on many things (e.g. file names can't have `/` or `:` in it).

rizar:
This PR is apparently abandoned in favour of #615

PR #557, opened 2015-03-26
Allow multi-word role names to be rendered with underscores.

dwf:


dwf:
As we make more and more use of variable roles (e.g. #513) we're going to be using more multi-word role names, which will be  currently rendered as `SEVERALWORDSSQUASHEDTOGETHER` by `repr()` with the current code. This fixes it in the most elegant way I could think of (the standard library `re` module doesn't support matching an indefinite number of groups, unfortunately, or I'd have used that).

bartvm:
Doesn't this work?
```python
re.sub(r'(?!^)([A-Z]+)', r'_\1', camel).lower()
```

dwf:
Hmm, seems like it does. Time to shave some lines off this pull request :P

dwf:
@bartvm replaced entire pull request with a 2 line call to `re.sub`, PTAL.

PR #558, opened 2015-03-27
Update plotting.rst

vikkamath:
Removed line "sample = theano.tensor.scalar('data')" because it wasn't being used anywhere in the example code snippet


bartvm:
Thanks! The build just failed because of the ongoing DDoS attack on GitHub, has nothing to do with your changes.

vikkamath:
Yeah! I was wondering why that happened. The error log looked dubious

PR #560, opened 2015-03-29
Clean up docs a bit

bartvm:


bartvm:
@dwf Changed to capitalized sentences and added the periods back in, thanks.

PR #562, opened 2015-03-30
Rebase of #489 (CCW 391: VariableFilter Interface Improvements)

### LINKED ISSUE #391: Various VariableFilter Inteface Improvements

dwf:
Fixes #391.


dwf:
@bartvm I fixed the single failing test (8a8ecbc) and addressed two of your docstring-related comments that went unaddressed in #489 (4f47af2), if that's to your liking then merge at will.

dwf:
Scrutinizer failed because of Baidu's DDoS attack, but I'm fairly certain everything is fine since 4f47af2 passed and changes in a2e949d are minimal, so I'm merging.

PR #563, opened 2015-03-30
WIP: Improve documentation of the sequence generator

rizar:
Instead of #522 I am now writing better documentation for the sequence generator. 

I am only halfway through, but something is already there. I would be glad if you try to read it and tell me where important information is lacking.

@dmitriy-serdyuk , @sebastien-j, @bartvm 

- [x] document attributes and move common constructor arguments for `AbstractFeedback`, `AbstractEmitter` and `AbstractReadout`

- [x] `AbstractReadout` should not inherit from `AbstractFeedback` and `AbstractEmitter`


rizar:
Apart from documentation I am also trying to refactor things a bit in this PR:

- simplified `initial_outputs` signature to take only `batch_size` - done

- `AbstractReadout` should not inherit from `AbstractFeedback` and `AbstractEmitter`. This was done to save some typing but in fact is rather weird. If the are no objections, I will separate these interfaces tomorrow.

mjwillson:
The extra docs here are really helpful!

blocks/bricks/sequence_generators.py
```diff
@@ -20,49 +49,72 @@ class BaseSequenceGenerator(Initializable):
...
+      generation step from states, contexts and the previous step glimpses.
+      They are computed in the transition's `apply` method when not given
+      or by explicitly calling the transition's `take_glimpses` method. The
+      set of glimpses considered is specified in
+      `transition.glimpse_names`.
+
+    * Outputs :math:`y_i` are produced at every step and form the output
+      sequence. A generation cost :math:`c_i` is assigned to each output.
 
     **Algorithm:**
```
mjwillson: A graphviz diagram of this would be awesome... just drawing one on scrap paper as I read this

blocks/bricks/sequence_generators.py
```diff
@@ -1,4 +1,33 @@
-"""Sequence generation framework."""
+"""Sequence generation framework.
+
+Recurrent networks are often used to generate/model sequences.
+Examples include language modelling, machine translation, handwriting
+synthesis, etc.. A typical pattern in this context is that
+sequence elements are generated one ofter another, and every generated
```
dmitriy-serdyuk: Here is a typo: ofter.

blocks/bricks/sequence_generators.py
```diff
@@ -20,49 +49,72 @@ class BaseSequenceGenerator(Initializable):
...
 
-    1. The initial states are computed from the contexts. This includes
-       fake initial outputs given by the `initial_outputs` method of the
-       readout, initial states and glimpses given by the `initial_state`
-       method of the transition.
+    1. Initialization.
+
+       :math:`y_0 = readout.initial\_outputs(contexts)`
+
+       :math:`s_0, g_0 = transition.initial\_states(contexts)`
```
dmitriy-serdyuk: Does the initial state depend on contexts?

dmitriy-serdyuk: Does `BaseRecurrent.initial_state` receive contexts as keyword arguments? It is not indicated in the docs.

rizar: Actually it takes all '*args' and '**kwargs' https://github.com/bartvm/blocks/blob/master/blocks/bricks/recurrent.py#L32 that your recurrent application method takes. This way it supports hidden states computed from contexts.

rizar:
I rebased this PR and will resume working on it.

blocks/bricks/sequence_generators.py
mjwillson: Question about this:

It seems that s_0 is special-cased in that all the other s_i for i>0 depend on the (forked) feedback f_i, but s_0 doesn't depend on the f_0 which is computed from y_0. Is that the intention?

I was also wondering why in fact why we need a fixed "initial output" y_0 at all -- is the idea that this is used for a mandatory start-of-sequence marker for example? It seems like the only thing it influences is the first readout r_1. I'm guessing some of the more advanced models need it, but in a lot of cases it may be sufficient to specify the initial state s_0 to achieve desired control over the initial readout. And the fact that the y_0 is special-cased in that it doesn't influence s_0 makes it feel a bit awkward at first glance.

rizar: Right, `s_0` does not depend on anything, it is determined by the transition itself. Options included fully zero state, a trainable one, or one predicted from the contexts. 

The reason why we have `y_0` and `f_0` is that there exists a popular hack to use `y_{i-1}`, more specifically `f_{i-1}`, to predict `y_i`. By default this does not happen though, one should include 'feedback' into the `source_names` of the readout to switch it on. This is why `y_0` is needed, to be used for prediction of `y_1`.

Your edit sounds like a nice summary for why we need `y_0` :)

rizar:
I am think I wrote most part of required documentation and also did a bit of refactoring. Now I would like this to be reviewed and merged.

@mjwillson , would be nice if you can provide some feedback about these new docs.

blocks/bricks/sequence_generators.py
dmitriy-serdyuk: @rizar , can you show, where `initial_states` with `contexts` argument is called? I see, that contexts are provided, but I can't understand why. Or is it an implicit behaviour of `recurrent` decorator?

blocks/bricks/sequence_generators.py
```diff
@@ -577,14 +735,7 @@ def get_dim(self, name):
...
 class FakeAttentionRecurrent(AbstractAttentionRecurrent, Initializable):
-    """Adds fake attention interface to a transition.
-
-    Notes
-    -----
-    Currently works only with lazy initialization (can not be initialized
-    with a single constructor call).
-
-    """
+    """Adds fake attention interface to a transition."""
```
dmitriy-serdyuk: Maybe it worth mentioning, that it uses no glimpses and just feeds states and contexts to the transition.

blocks/bricks/sequence_generators.py
rizar: `initial_state` is called [here](https://github.com/bartvm/blocks/blob/master/blocks/bricks/recurrent.py#L175) in the `recurrent.py`. Moreover, this is the only place where it is used. 

Do you have suggestions on how to better document this behaviour? I though `BaseRecurrent.initial_states` docstring is explanatory enough.

blocks/bricks/sequence_generators.py
rizar: Good point, wrote docs for this class.

rizar:
I am getting the following from Scrutinizer:

```
blocks/bricks/sequence_generators.py:442 in public class `Readout`:
        D301: Use r""" if any backslashes in a docstring
blocks/bricks/sequence_generators.py:800 in public class `SequenceGenerator`:
        D301: Use r""" if any backslashes in a docstring
```

But I use backslashes to skip line breaks. With `r"""` I can not do this. @bartvm , can you help me?

bartvm:
I generally just avoid them, the rule enforced is from PEP257,

> Use r"""raw triple double quotes""" if you use any backslashes in your docstrings.

If you really, really need them I guess we could disable the rule, but then it becomes important that we make sure (manually) that all escape sequences are avoid (`\n`, `\b`, `\\`, etc.) which might not be pleasant when you're writing LaTeX.

rizar:
Well, if there is any other way to break the line in a parameter docstring (like in the line 810 of this PR) I would be glad to use it. Do you know how to do that?

blocks/bricks/sequence_generators.py
dmitriy-serdyuk: I would add here something like "Notice, that standard implementations of recurrent bricks (`SimpleRecurrent`, `LSTM`, 'GatedRecurrent`) return all zeros for the initial state regardless the value of contexts, if you need to redefine this behaviour, inherit from a recurrent brick and implement `initial_states` method".

rizar: Done, please check.

blocks/bricks/sequence_generators.py
```diff
@@ -16,53 +45,83 @@
...
+
+       .. math::
+
+           y_0 = readout.initial\_outputs(contexts)\\
+           s_0, g_0 = transition.initial\_states(contexts)\\
+           i = 1\\
+
+        The default `initial_states` method that a recurrent transition
+        inherits from :class:`.BaseRecurrent` creates an initial states
+        of all zeros. Use your own transition classes to get have custom
```
dmitriy-serdyuk: "get have" is it a typo?

rizar: Sure, fixed.

rizar:
I am struggling to see how I should typeset [here](https://github.com/rizar/blocks/blob/docs_sequence_generator/blocks/bricks/sequence_generators.py#L815) without backslashes. @bartvm , can you please disable this rule?

bartvm:
I'd prefer not disabling it because it can lead to really weird stuff when people type e.g. latex (you'd have to type `\\nabla` instead of `\nabla`, etc.)

What I've done in these cases is
```
attention : obj, optional
    An instance of :class:`~blocks.bricks.attention.AbstractAttention`.
```

rizar:
Well, people still can use `r"""`, it they like. I do not see a necessity to force everybody to use it though.

bartvm:
Because people forget all the time, and then the documentation doesn't render correctly but nothing will warn you about that. The reason for forcing people to do it that way is because it's a PEP257 guideline not to use `\` for new lines in docstrings.

rizar:
Hard to argue against PEP257, but I think that the case I refer to will be not that rare one. Any reference to a class which is not in the same module tends to be quite long, and a good documentation tends to have a lot of then. I will try to look a bit deeper into Sphinx or reformat it as you suggest.

rizar:
I fixed the Scrutinizer errors, but when I build the docs locally, I get the following:

```
/home/rizar/Dist/blocks/blocks/bricks/sequence_generators.py:docstring of blocks.bricks.sequence_generators.AbstractReadout.readout:3: WARNING: Inline strong start-string without end-string.
/home/rizar/Dist/blocks/blocks/bricks/sequence_generators.py:docstring of blocks.bricks.sequence_generators.AbstractReadout.readout:5: WARNING: Inline strong start-string without end-string.
/home/rizar/Dist/blocks/blocks/bricks/sequence_generators.py:docstring of blocks.bricks.sequence_generators.Readout:37: WARNING: Inline strong start-string without end-string.
/home/rizar/Dist/blocks/blocks/bricks/sequence_generators.py:docstring of blocks.bricks.sequence_generators.Readout:39: WARNING: Inline strong start-string without end-string.
/home/rizar/Dist/blocks/blocks/bricks/sequence_generators.py:docstring of blocks.bricks.sequence_generators.SequenceGenerator:17: WARNING: Inline strong start-string without end-string.
/home/rizar/Dist/blocks/blocks/bricks/sequence_generators.py:docstring of blocks.bricks.sequence_generators.SequenceGenerator:21: WARNING: Inline strong start-string without end-string.
```

I find it quite hard to understand even what the line numbers mean in these messages, because both lines 3 and 5 of `AbstractReadout` docs look absolutely innocent.

bartvm:
It normally means that you've documented *args or **kwargs, but * and **
are symbols for starting italic/strong in rST. The fix is to use r""" and
\* or \*\* instead.

rizar:
Hell yeah, finally it seems to be done! Thanks Bart!

Considering how desperately this part of Blocks needs to be documented, I would be willing to merge this PR quickly.

PR #564, opened 2015-03-30
Pin dependencies in req*.txt files

bartvm:


bartvm:
To make sure that

* Builds don't break when some package gets updated (which happens when `flake8` gets updated, or `sphinx`)
* Keep our dependencies a bit more organized
* Automatically track whether our dependencies are up to date using requires.io

bartvm:
@dwf @rizar Any problems with pinning dependencies? It's good practice, and will hopefully stop us from running into problems like https://github.com/Theano/Theano/pull/2697

rizar:
I am rather ignorant about such things but so far I do not see any issues.

dwf:
No objection to pinning dependencies from me. Seems like a very reasonable
idea.

PR #566, opened 2015-03-30
Add prototype to class SequenceGenerator

sebastien-j:


blocks/bricks/sequence_generators.py
```diff
@@ -641,6 +641,9 @@ class SequenceGenerator(BaseSequenceGenerator):
         If ``True``, the :class:`AttentionRecurrent` wrapping the
         `transition` will add additional contexts for the attended and
         its mask.
+    prototype : :class:`.FeedForward`, optional
+        A transformation prototype. If ``None``, the default prototype
```
rizar: I am sorry for confusion, but I have just realized that it's not the right way to go. All you need is to receive a fork from user, that is the line 660 should be changed to `kwargs.setdefault('fork', fork).

I will make a PR in a minute.

rizar:
Replaced by #567.

PR #567, opened 2015-03-30
Create default Fork in BaseSequenceGenerator

rizar:
A better way to implement functionality from #566

### DIFF
--- a/blocks/bricks/sequence_generators.py
+++ b/blocks/bricks/sequence_generators.py
@@ -88,6 +88,9 @@ class BaseSequenceGenerator(Initializable):
         The transition component of the sequence generator.
     fork : :class:`.Brick`
         The brick to compute the transition's inputs from the feedback.
+        If not given, :class:`~blocks.bricks.parallel.Fork` is used
+        and all the sequential inputs which names do not contain
+        the 'mask' substring are forked.
 
     Notes
     -----
@@ -95,8 +98,12 @@ class BaseSequenceGenerator(Initializable):
 
     """
     @lazy()
-    def __init__(self, readout, transition, fork, **kwargs):
+    def __init__(self, readout, transition, fork=None, **kwargs):
         super(BaseSequenceGenerator, self).__init__(**kwargs)
+        if not fork:
+            fork = Fork([name for name in transition.apply.sequences 
+                         if not 'mask' in name])
+        
         self.readout = readout
         self.transition = transition
         self.fork = fork
@@ -649,12 +656,7 @@ class SequenceGenerator(BaseSequenceGenerator):
 
     """
     def __init__(self, readout, transition, attention=None,
-                 fork_inputs=None, add_contexts=True, **kwargs):
-        if not fork_inputs:
-            fork_inputs = [name for name in transition.apply.sequences
-                           if name != 'mask']
-
-        fork = Fork(fork_inputs)
+                 add_contexts=True, **kwargs):
         if attention:
             distribute = Distribute(fork_inputs,
                                     attention.take_glimpses.outputs[0])
@@ -665,4 +667,4 @@ class SequenceGenerator(BaseSequenceGenerator):
             transition = FakeAttentionRecurrent(transition,
                                                 name="with_fake_attention")
         super(SequenceGenerator, self).__init__(
-            readout, transition, fork, **kwargs)
+            readout, transition, **kwargs)
--- a/blocks/bricks/sequence_generators.py
+++ b/blocks/bricks/sequence_generators.py
@@ -88,9 +88,6 @@ class BaseSequenceGenerator(Initializable):
         The transition component of the sequence generator.
     fork : :class:`.Brick`
         The brick to compute the transition's inputs from the feedback.
-        If not given, :class:`~blocks.bricks.parallel.Fork` is used
-        and all the sequential inputs which names do not contain
-        the 'mask' substring are forked.
 
     Notes
     -----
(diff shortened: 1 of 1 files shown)


rizar:
@sebastien-j with this PR you should be able to give your custom `fork` to `BaseSequenceGenerator`. Can you please try it?

sebastien-j:
Ok, will do.

Couldn't weird behaviour occur if the inputs to a custom fork differ from `[name for name in transition.apply.sequences if 'mask' not in name]`?

sebastien-j:
With minor adjustments to the other script, it works.

bartvm:
Why not just do `fork=None` instead of hiding it in `kwargs`?

rizar:
> Why not just do `fork=None` instead of hiding it in kwargs?

Because `fork` is a keyword argument of the superclass, `BaseSequenceGenerator`. One typically does not repeat keyword argument of the superclass in the constructor signature.

> Couldn't weird behaviour occur if the inputs to a custom fork differ from [name for name in transition.apply.sequences if 'mask' not in name]?

Those are actually names of outputs the fork should produce in fact.

In general sequence generator does not know which inputs should be forked. For convenience I put this `name for name in transition.apply.sequences if 'mask' not in name` combo, but one can imagine cases when it would not work.

If you are using `GatedRecurrent` and create a fork with wrong `output_names` then one of two things can happen:

- if you forget about e.g. `reset_inputs`, you will have an "argument not given" error during computation graph building

- if you add `mask` to them you can really experience very weird behavior!

bartvm:
> Because fork is a keyword argument of the superclass, BaseSequenceGenerator. One typically does not repeat keyword argument of the superclass in the constructor signature.

Why not? I don't repeat the keyword argument if it's not used in the child class (e.g. `name` for bricks), but if the keyword argument is used in the child class, and a default is set in the child class, then the keyword argument should be repeated IMHO. I find `fork=None` followed by `if fork is None` is much clearer than `kwargs.setdefault('fork', ...)`.

rizar:
I think it's a matter of coding style. All descendant of `SimpleExtension` use `setdefault`, so this practice is quite widespread in Blocks, and it's just less lines.

And something tells me that we will eventually merge `SequenceGenerator` and `BaseSequenceGenerator` so this doesn't matter much. It just needs a good coder to make attention optional while keeping code readable.

rizar:
@sebastien-j , now it's in master and you can get rid of the copy you had to create in the NMT repo.

bartvm:
But right now `fork` is not part of the signature, *and* there is no docstring... Merged a bit too soon maybe?

rizar:
`fork` is a part of `BaseSequenceGenerator` signature and there is a docstring for it there.

It should be mentioned in the `SequenceGenerator` docstring that it provides a default for `fork`, but it was not  done before this PR as well and will be addressed in #563. That is this PR improved state of the master without introducing new issues, which you as argued recently should be enough to merge.

bartvm:
If `fork` would be part of the signature, then a user could reasonably be expected to look at the docstring of the base class to find its documentation. However, since you're hiding `fork` in `kwargs`, there is no way to tell--without reading the code--that `SequenceGenerator` accepts the `fork` keyword argument at all. (A derived class doesn't necessarily pass on all of its `kwargs` to the base class, there's no way for the user to know that this is the case here.)

This means it actually leaves `master` in a worse state in that respect, because before the user could tell from the signature that he could pass `fork_inputs`. Right now, the user needs to read the code to know that `fork` is an accepted argument.

rizar:
> This means it actually leaves master in a worse state in that respect, because before the user could tell from the signature that he could pass fork_inputs. Right now, the user needs to read the code to know that fork is an accepted argument.

`fork_inputs` was not documented, therefore the only way to understand where and how fork is created was to read the code. For this reason I do not consider the new master state worse. 

> Right now, the user needs to read the code to know that `fork` is an accepted argument.

He needs to read `BaseSequenceGenerator` signature to see that `fork` is an accepted argument. Absolutely similarly to that the user has to read `AbstractDataStream` documentation to get to know about `iteration_scheme` argument when he creates a `DataStream`. 

> (A derived class doesn't necessarily pass on all of its `kwargs` to the base class, there's no way for the user to know that this is the case here.)

It's a standard practice used everywhere in Blocks and is never documented, that `kwargs` (possibly with modifications) are passed to the superclass constructor. See `DataStream` and `AbstractDataStream` for an example.

bartvm:
How is it standard practice? It happens often, of course, but it doesn't make sense to always do this. Often you have cases where the child class doesn't accept some of the keyword arguments of the base class, but you might still want to use `kwargs` for the others. For example `ServerDataStream` doesn't accept `iteration_scheme`. The fact that `DataStream` doesn't document the fact that it accepts `iteration_scheme` is simply a mistake.

In `conv.py` each brick has a `See Also` section informing the user that the brick takes the same argument as its base classes, and refers the user to their documentation. I think that's the least we should do.

Anyway, the idea that the user should "just know" that when a method accepts `**kwargs`, they should go through the class hierarchy and assume that every single keyword argument there is accepted by the derived class, is obviously very bad practice and not user-friendly, and we should avoid it and remove it from everywhere else.

rizar:
> Anyway, the idea that the user should "just know" that when a method accepts **kwargs, they should go through the class hierarchy and assume that every single keyword argument there is accepted by the derived class, is obviously very bad practice and not user-friendly, and we should avoid it and remove it from everywhere else.

Agreed!

So let's be constructive instead of just arguing and introduce the following documentation convention: any method that does `super(ThisClass, self).method(**kwargs)` must declare that in the documentation. There lots of places in Block where we don't do this, for which reason I assumed that this is the default behavior I should expect from `**kwargs` of an overriding method. I assume that this should be mentioned in http://blocks.readthedocs.org/en/latest/development/docs.html if we decide on it.

PR #568, opened 2015-03-31
Add notebook plotting extension

bartvm:
Just hacked this together, because `Printing` in an IPython notebook is a nightmare. Should probably go into `blocks-extras` as soon as we create that repo, but in the meanwhile, is this something other people would find useful?

### DIFF
--- a/blocks/extensions/plot.py
+++ b/blocks/extensions/plot.py
@@ -9,7 +9,13 @@ try:
     BOKEH_AVAILABLE = True
 except ImportError:
     BOKEH_AVAILABLE = False
-
+try:
+    from IPython import display
+    from IPython.core import pylabtools
+    IPYTHON_AVAILABLE = True
+except ImportError:
+    IPYTHON_AVAILABLE = False
+from matplotlib import pyplot
 
 from blocks import config
 from blocks.extensions import SimpleExtension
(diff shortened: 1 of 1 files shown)


rizar:
I would rather wait until `blocks-extras`. Most people I know run training from console, which means that demand for this extension is quite limited on one hand, on another hand we could at least create `blocks-contrib`, not tested warehouse for raw Blocks components and put it there.

bartvm:
I'll wait till `blocks-extras` then, but it actually is a pretty handy extension! Not for complicated models with long training times probably, but in my case I was just doing a homework assignment with some small models. I also imagine a lot more people have IPython set up than Bokeh.

EderSantana:
you know that you can always use the very same bokeh plot extension but setting `output_notebook()` instead of `output_server()`, right? This way you get both the localhost:5006 and an in cell plot. [ref](http://nbviewer.ipython.org/github/paddymul/bokeh_tutorial/blob/master/bokeh_tutorial.ipynb)

But I think I heard you were having problems with the bokeh extension, what happend?

bartvm:
It's in heavy development, so the interface changes every release, which can be frustrating. They also don't allow streaming data to the server, so for long-running experiments it starts really slowing things down (because you need to send *all* the data again each time). It's also annoying that losing connection with Bokeh for any reason shuts down your entire training progress.

The major architectural problem is that I think monitoring extensions shouldn't be as directly integrated with the training progress as they are now. Preferably they should interface with the log only. That way, we only need to be robust and efficient in one place, namely when storing data in the log. Any plotting extension that crashes, is slow, etc. won't affect the training progress then, because it's only querying data from the log. This is something we're hoping to address with #476/#470 eventually.

PR #570, opened 2015-03-31
Attempt at adapting example to be usable without Bokeh. Fixes #174

### LINKED ISSUE #174: MNIST example without Bokeh
Originally mentioned in #173, also related to #170 (run tests with Bokeh as well).

vikkamath:


examples/mnist.py
```diff
@@ -24,7 +24,7 @@
 from blocks.main_loop import MainLoop
 
 
-def main(save_to, num_epochs):
+def main(save_to, num_epochs, no_bokeh):
```
bartvm: You should probably give a default, `no_bokeh=True` here as well, but make sure that the tests use `no_bokeh=True` so that it still gets tested on Travis.

examples/mnist.py
```diff
@@ -84,5 +89,7 @@ def main(save_to, num_epochs):
     parser.add_argument("save_to", default="mnist.pkl", nargs="?",
                         help=("Destination to save the state of the training "
                               "process."))
+    parser.add_argument("--no-bokeh", type=bool, default=False,
```
bartvm: I think you can set the default to `True`, or better, change the argument to `--bokeh`. The main reason we have Bokeh in here is for testing.

bartvm:
The errors are ocurring because you've changed the signature of `main`, but the tests still call `main` with 2 instead of 3 arguments, you'll have to update those.

tests/test_examples.py
```diff
@@ -24,7 +24,15 @@ def test_sqrt():
 def test_mnist():
     skip_if_not_available(modules=['bokeh'])
     with tempfile.NamedTemporaryFile() as f:
-        mnist_test(f.name, 1)
+        mnist_test(f.name, 1, True)
```
bartvm: Probably just run this test once with Bokeh, it'll just slow down the test suite to run it twice, and we don't really gain much from it.

vikkamath: I agree. Changes made, submitted PR

PR #572, opened 2015-04-01
Strict check for not given log and extensions

rizar:


rizar:
Travis fails because of `blocks/bricks/conv.py:238:1: W293 blank line contains whitespace`, not related to this PR.

PR #574, opened 2015-04-01

bartvm:


vikkamath:
This is probably irrelevant, but the cause of the PEP8 failure (conv.py) has been fixed in PR #570

bartvm:
Thanks, looks like Conda finally updated its repos.

PR #577, opened 2015-04-03
Add missing documentation and minor typos fix

fvisin:
This PR adds every non-documented class to the autobuilt documentation and fixes a few typos.
Please check if you like the way I grouped things together.

### DIFF
--- a/docs/development/internal_api.rst
+++ b/docs/development/internal_api.rst
@@ -1,6 +1,12 @@
 Internal API
 ============
 
+* `Bricks`_
+* `Extensions`_
+* `Utils`_
+
+Bricks
+------
 .. automodule:: blocks.bricks.base
    :undoc-members:
    :members:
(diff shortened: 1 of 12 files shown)


docs/development/internal_api.rst
```diff
@@ -12,3 +20,73 @@ Internal API
...
+
+.. automodule:: blocks.monitoring.evaluators
+   :undoc-members:
+   :members:
+   :private-members:
+   :show-inheritance:
+
+Search
+------
+.. automodule:: blocks.search
```
bartvm: Some of these modules shouldn't go into the internal API. The idea is that the internal API is used to document modules and objects that a user would normally not interface with (but a developer would) e.g. `ApplicationCall`, `BoundApplication`, that sort of things. Beam search, the wrapper bricks, and some of the methods in the aggregation module should be public though.

docs/development/internal_api.rst
```diff
@@ -12,3 +20,73 @@ Internal API
...
+   :undoc-members:
+   :members:
+   :private-members:
+   :show-inheritance:
+
+Utils
+-----
+.. automodule:: blocks.config_parser
```
bartvm: This docstring is rendered at http://blocks.readthedocs.org/en/latest/configuration.html already

fvisin: I missed it, sorry!

docs/development/internal_api.rst
fvisin: I should have corrected. Let me know if it is ok before merging please and I will rebase.

fvisin:
Should be ready to merge, let me know if anything is not where it should be.

PR #578, opened 2015-04-03
Add data consistency check for VariableFilter

fvisin:
This PR checks the input consistency of `roles` and `applications`.

I know it's borderline wrt the duck typing philosophy, but currently `VariableFilter` is not failing when for instance `bricks` is a list of lists (e.g. `bricks=[MLP.children]`). I guess duck typing is great if you also catch the errors as soon as possible, because silent fails are a pain to debug. :) Let me know if there is a better way to enforce input consistency though.

### DIFF
--- a/blocks/filter.py
+++ b/blocks/filter.py
@@ -44,12 +44,10 @@ class VariableFilter(object):
     ----------
     roles : list of :class:`.VariableRole` instances, optional
         Matches any variable which has one of the roles given.
-    bricks : :class:`.Brick`, class or list, optional
-        If an instance of :class:`.Brick`, matches variables that are
-        owned by that brick. If a class object, matches variables whose
-        bricks are instances of that class. If list, perform the
-        above matching for each :class:`.Brick` instance or
-        class object contained in the list.
+    bricks : list of :class:`.Brick` classes or list of instances of
+             :class:`.Brick`, optional
+        Matches any variable that is instance of any of the given classes
+        or that is owned by any of the given brick instances.
     each_role : bool, optional
         If ``True``, the variable needs to have all given roles.  If
         ``False``, a variable matching any of the roles given will be
@@ -60,9 +58,9 @@ class VariableFilter(object):
     name_regex : str, optional
         A regular expression for the variable name. The Blocks name (i.e.
         `x.tag.name`) is used.
-    applications : :class:`.Application` instance or list, optional
-        Matches a variable that was produced by an application (or any
-        of a list of applications) given.
+    applications : list of :class:`.Application`, optional
+        Matches a variable that was produced by any of the applications
+        given.
 
     Notes
     -----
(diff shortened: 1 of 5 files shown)


blocks/filter.py
```diff
@@ -96,6 +96,16 @@ class object contained in the list.
     """
     def __init__(self, roles=None, bricks=None, each_role=False, name=None,
                  name_regex=None, applications=None):
+        if bricks is not None and not isinstance(bricks, Brick):
```
bartvm: IMHO we should just get rid of this convention that both `Brick` and a list of bricks can be passed. It doesn't make sense to call the argument `bricks` and then allow a single one.

Anyway, this is probably shorter (and likewise for `applications`):
```python
if isinstance(bricks, Brick):
   bricks = [bricks]
if not all(isinstance(brick, Brick) for brick in bricks):
    raise ValueError("`bricks` should be a list of bricks")
```
Note that you should be raising a `ValueError`, not use `assert`. Assertions are predicates that should *always* be true. They're used to catch bugs in the code, not to alert users of wrong inputs. See [the documentation](http://blocks.readthedocs.org/en/latest/development/index.html#validating-function-arguments).

fvisin: Sorry about the assert, bad habit I guess.

I agree with getting rid of the convention of accepting both `Brick` and a list of `Brick`, but the main issue here would not be solved by that: list of lists are accepted but ignored.

Let me know if you want me to get rid of the `Brick` convention.

bartvm: Yeah, I know it's an entirely orthogonal issue, but I figured that since we're changing the code we might as well change that along with it, so if you could, please do!

fvisin: No problem, I am glad to help!
I just pushed a new commit, but I think that a little more can be done. I'll comment inline.

fvisin: I guess that `isclass` can be avoided here https://github.com/bartvm/blocks/blob/67e19f33ccdd2d21482fcfce89542a60c0385bad/blocks/filter.py#L131

Also, I am not sure if something can be changed here: https://github.com/bartvm/blocks/blob/67e19f33ccdd2d21482fcfce89542a60c0385bad/blocks/filter.py#L147-L150

dwf: My understanding of that is that you can pass in a class object such as
simply "Linear", and it will give you back all instances of Linear. Are you
saying that this can be accomplished without isclass? Pretty sure you will
get a TypeError if arg 2 to isinstance is not a [sequence of] type/class
objects.

fvisin: You are right, but the input control at https://github.com/bartvm/blocks/pull/578/files#diff-b78ec68afdd58167518263ff5b423d3aR97 wouldn't allow a non-class object.

I guess `isclass` can be avoided here, while `issubclass(brick, Brick)` should be added as input check.

dwf: If that's the case then the input control should be amended, IMO :)

fvisin: Sorry I forgot to write that I updated it right away, it should be ok right now. :)

fvisin:
If I am not wrong, I didn't decrease the coverage so this PR should be ready to merge. Let me know if I missed something please.

PR #579, opened 2015-04-03
Custom picklers

bartvm:
This is basically a cleaned up version of @abergeron's custom picklers from Pynnet, but now with ~~support for `CudaNdarray` and~~ intelligent naming of arrays that belong to shared variables. So far there is no Blocks-specific functionality in here yet, which makes me wonder whether maybe this would be considered useful enough to add to Theano instead of Blocks (@lamblin, @nouiz?). Almost everyone ends up saving their model's parameters to NumPy files one way or the other, and I think this is a nicely transparent way to do it.

Illustrative use from the docstring:

```python
>>> import theano
>>> foo_1 = theano.shared(0, name='foo')
>>> foo_2 = theano.shared(1, name='foo')
>>> with open('model.zip', 'w') as f:
...     dump((foo_1, foo_2, numpy.array(2)), f)
>>> numpy.load('model.zip').keys()
['foo', 'foo_2', 'array_0', 'pkl']
>>> numpy.load('model.zip')['foo']
array(0)
>>> with open('model.zip') as f:
...     foo_1, foo_2, array = load(f)
>>> array
array(2)
```

The plan for Blocks is now to basically extend `PersistentSharedVariableID` to a `PersistentBrickParameterID`, which names the arrays not only after the shared variable, but also taking into account its position in the Brick hierarchy (and hence fixing #474).

CC @dmitriy-serdyuk

### DIFF
--- a/blocks/serialization.py
+++ b/blocks/serialization.py
@@ -1,83 +1,239 @@
-import os
+import pickle
+import tempfile
+import zipfile
+from collections import defaultdict
+from contextlib import closing
 from pickle import HIGHEST_PROTOCOL
 try:
     from pickle import DEFAULT_PROTOCOL
 except ImportError:
     DEFAULT_PROTOCOL = HIGHEST_PROTOCOL
-import shutil
-import tempfile
 
-import six
-from six.moves import cPickle
-
-from blocks.utils import reraise_as
-
-PICKLING_ERROR = """
-
-Blocks relies on the ability to pickle the entire main loop, which includes \
-all bricks, data streams, extensions, etc. All of these must be serializable \
-using Pythons pickling library. This means certain things such as nested \
-functions, lambda expressions, generators, etc. should be avoided. Please \
-see the documentation for more detail."""
-
-INSTANCEMETHOD_ERROR = """
-
-Python is unable to serialize references to class and instance methods. This \
-means that you cannot store reference to objects like `str.lower` (a class \
-method) or `Linear().allocate` (an instance method) in an attribute. If you \
-reall need to do this, it is often possible to write a wrapper function such \
-as `def lower(s): return s.lower()` and store that instead."""
-
-LAMBDA_ERROR = """
-
-Python is unable to serialize lambda functions. Make sure that your code does \
-not rely on them and use normal function definitions instead."""
-
-NESTED_FUNCTION_ERROR = """
-
-Python is unable to serialize nested functions. Make sure that all your \
-functions are available in the global namespace."""
+import numpy
+from six import StringIO
+from theano.sandbox.cuda import CudaNdarray
+from theano.compile.sharedvalue import SharedVariable
 
 
-def pickle_dump(*args, **kwargs):
-    """A wrapper around pickle's dump that provides informative errors."""
-    kwargs.setdefault('protocol', DEFAULT_PROTOCOL)
-    try:
-        cPickle.dump(*args, **kwargs)
-    except Exception as e:
-        if six.PY3 and '<lambda>' in e.args[0]:
-            reraise_as("Pickling failed to pickle a lambda function." +
-                       LAMBDA_ERROR)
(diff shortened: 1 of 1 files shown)


abergeron:
This fails because you can't always import CudaNdarray.  You could fix it by not having a special case for it since the default pickler will convert them to numpy.ndarray anyway.

Also, on the topic of having this directly in theano, I don't see a compelling advantage.  In what way would it help the users and frameworks that depend and use theano?  I also worry about making the interface general enough to be pluggable in custom saving schemes that other frameworks may have.

bartvm:
> You could fix it by not having a special case for it since the default pickler will convert them to numpy.ndarray anyway.

Right, because the custom unpickler will be used to pickle the return values of `CudaNdarray_unpickler`, makes sense. For some reason I was under the impression that the function registered with `copy_reg` wasn't called anymore when using `persistent_id`, but it is of course.

>  In what way would it help the users and frameworks that depend and use theano?

It's just a suggestion, but as I pointed out, every single user I know, regardless of the framework they use, at one point ends up wanting to save their model's parameters. No one actually does this in a Pickle file because all of its shortcomings (e.g. unpickling GPU-trained models on CPU, different Python versions or simply different installations, faster/easier access, etc.) so everyone ends up calling `get_value()` manually on each of their parameter variables in order to save them to an NPZ file with some sensible names.

This PR would just allow the user to call `dump` and `load` as if they were pickling, but knowing that regardless of their environment anyone can access the parameters easily using `numpy.load` if they want.

>  I also worry about making the interface general enough to be pluggable in custom saving schemes that other frameworks may have.

What exactly do you have in mind? I already made the persistent ID and persistent load methods a keyword argument, and people can still use `copy_reg`, so anything short of an entirely custom `Pickler` class can easily be used together with these functions, right?

abergeron:
One problem I have is that there is no guarantee that the names are unique at the theano level.  So we would need some form of collision avoidance strategy.  You currently either raise an error or overwrite the previous value, both of which are not optimal, although I prefer the version that raises an error.

The strategy used in pynnet worked because it had the cooperation of the model classes where they would extract the value of the shared variables they used on pickling and rebuild those shared variables on load (along with the theano graph). This won't be a given in a random implementation.

Finally lots of people in the lab do use simple pickle to save their models, because it's the easiest way.  In that sense maybe having a builtin theano solution would help.

--

For the generality I mean that this interface would be useless for people storing the model values in an hdf5 file for example. But I guess that we can reasonably restrict ourselves to `pickle`.

--

On the topic of test errors it seems you deleted a couple of interfaces that where in this file before and were used elsewhere.

bartvm:
> One problem I have is that there is no guarantee that the names are unique at the theano level. So we would need some form of collision avoidance strategy. You currently either raise an error or overwrite the previous value, both of which are not optimal, although I prefer the version that raises an error.

It doesn't do that (or if it does, it's not supposed to). It should do the following (see my example):

* If the shared variable has no name, its data is stored to `array_0`, `array_1`, etc.
* If the shared variable has a duplicate name, they are stored to `name`, `name_2`, `name_3`, etc.

All of this logic is here: https://github.com/bartvm/blocks/blob/custom_pickle/blocks/serialization.py#L88-L99 I do allow people to set `allow_duplicates=False` or `allow_unnamed=False`, in which case it will raise an error as you said. (It will never overwrite though.)

> On the topic of test errors it seems you deleted a couple of interfaces that where in this file before and were used elsewhere.

Yeah, there's some work to make Blocks use this, but I figured I'd make a PR already for early feedback.

rizar:
I did my homework reading pickle docs and now I like this idea. But the current state of the PR looks somewhat raw, in particular I guess we want to keep all the custom pickling error messages we had.

bartvm:
Yes, as I said in the description, this PR is by no means finished. I
wanted to know whether there was interest in moving this to Theano first.

bartvm:
Largely merged into Theano as part of #2802, and Blocks-specific implementation in #615.

PR #580, opened 2015-04-04
Check if blocksrc is empty to avoid error

bartvm:
Fix error reported in https://groups.google.com/forum/#!topic/blocks-users/W0MpxS__tAg


rizar:
LGTM

PR #581, opened 2015-04-06
Fix division by zero in attention

rizar:
I was developing my custom attention mechanism and came across an issue: `compute_weights` did not work with a mask consisting of all zeros. I did have such masks. This PR fixed the issue.

Also fixes a syntax error in `conv.py`


rizar:
The PR seems ready, waiting for an LGTM.

rizar:
Turns out `finfo.tiny` causes nans in the gradient, for which I reason I now use a slightly more involved solution.

dwf:
LGTM, merging.

PR #583, opened 2015-04-09
Added delegate to bidirectional

dmitriy-serdyuk:


bartvm:
There's a blank line too much in the test file, otherwise it looks good to me.

PR #585, opened 2015-04-11

bartvm:


rizar:
We will do such update every time a fresher version of a package is out?

bartvm:
Yeah, given that the new package is available in Anaconda and doesn't break
the master branch. It prevents everything from breaking in case a package
introduces a non-backward compatible change.

The pull requests are made automatically though, we just need to press
merge once things pass.

PR #590, opened 2015-04-15
Getting rid of redundent biases

laurent-dinh:
There should already be a linear/bias layer before the LSTM, the biases in the LSTM class may be redundant.


rizar:
This seems like the right thing to do, but let's include the code author in the discussion.

@pbrakel , would you mind this change?

pbrakel:
Not at all. I didn't really think about that when I implemented it. Actually, the implementation turns out to be very slow (probably due to the slicing) so I expect the whole implementation will be changed again eventually.

rizar:
Right, maybe you should separate inputs variables like it is done in `GatedRecurrent`.

PR #591, opened 2015-04-15
Cleaner error messages for attribute errors

dmitriy-serdyuk:


blocks/bricks/base.py
```diff
@@ -210,12 +210,14 @@ def __getattr__(self, name):
...
                 getattr(self, self.properties[name]), self.brick))
-        raise AttributeError
+        raise AttributeError("Cannot find '%s' property in brick %s" %
+                             (name, str(self)))
 
     def __setattr__(self, name, value):
         # Mimic behavior of read-only properties
         if 'properties' in self.__dict__ and name in self.properties:
-            raise AttributeError("can't set attribute")
+            raise AttributeError("Can't set attribute '%s' in brick %s" %
```
rizar: This error message does not tell the whole story, I believe. The idea here is that `brick.apply.inputs = ['blabla']` is not supported, that is that application method properties are read-only. A good error message should explain this to user.

bartvm: The current error message, `can't set attribute` is an exact copy of the error message that Python gives when trying to set a read-only property.

dmitriy-serdyuk: @bartvm , do you mean, that I shouldn't change this message? Sometimes it's really hard to guess what is the reason of a getattr/setattr error.

rizar: @bartvm , but I am not sure that information provided in this exact copy is sufficient for the user. Especially taking into account that we typically use the word 'property', not 'attribute' when referring to things like `brick.apply.inputs`.

Maybe a longer error message after a newline would do.

bartvm: You can change it if you want, I was just pointing out that this is simply mirroring Python's behaviour, so I can live with not changing it.

rizar: Yeah, but Python error message have to be very generic because no detailed information is available at the programming language level. We in Blocks could be more generous.

But okay, I also can live with this message, it's better than simply `AttributeError`.

rizar: By the way, the message "Can't set attribute '%s' in brick %s" is still misleading, because it's triggered on an attempt to set an application's attribute, not brick's.

rizar:
@dmitriy-serdyuk , would you like to finish this little PR?

rizar:
Ping!

rizar:
Closed due to in activity.

Issue #480, opened 2015-03-16
Selector and Model fail to return parameters silently


bartvm:
If parameters have the same path/name (i.e. if a sequence contains multiple `Linear` bricks without a specific name) the `Selector.get_params` (and hence `Model.get_parameters`) will simply override the first one with the second one in the parameter dictionary it returns, and hence fail to return all the parameters.

rizar:
My opinion is that we should enforce that children of a brick have different names. You `Children` class is a good place to check that.

bartvm:
Sounds good. The same problem occurs for parameters outside of the brick hierarchy though. The `Model` class simply stores them with the `param.name` key, which could easily be the same for multiple parameters, and there too it silently doesn't return some parameters.

rizar:
Right, we should as well check uniqueness in both Model and Selector.

dwf:
I saw @mducoffe run into an error that sounds like @rizar's suggestion. Has this been addressed/should this ticket be closed?

mducoffe:
@ rizar Is there any way to see which variable is concerned by the problem. I have put names to everything during the construction but I keep having the problem. What is weird is that my cost is composed of two cost : cost_1 + cost_2. When I am considering only cost_1 I have the error ( so I can't save my model) but when I am running the sum of cost_1 and cost_2 it is working so I am a bit perplexed.

rizar:
If you are talking about "top bricks have the same name", then it's not about variable and about brick names. Can you please show the code? You can use `gist.github,com` for that.

rizar:
@dwf it's only `Model` that does this check so far,  we need it to be also done in `Children` and `Selector`.

bartvm:
This is just a note that @vdumoulin's solution in #597 only applies to parameters that belong to bricks. In my case I replace the parameters in the model with Gaussians using `ComputationGraph.replace`, which introduces a whole bunch of new parameters (the variances) which don't belong to any brick. These still get silently overwritten in [this line](https://github.com/bartvm/blocks/blob/master/blocks/model.py#L171) and never get saved.

PR #597, opened 2015-04-24
Add check in Selector.get_params for path uniqueness

vdumoulin:
This concerns issue #480. It's not a complete solution, but at least `Selector` won't do the wrong thing silently anymore.

### DIFF
--- a/blocks/select.py
+++ b/blocks/select.py
@@ -10,6 +10,12 @@ from blocks.utils import dict_union
 
 logger = logging.getLogger(__name__)
 
+name_collision_error_message = """
+
+The '{}' name appears more than once. Make sure that all bricks' children \
+have different names and that user-defined shared variables have unique names.
+"""
+
 
 class Path(object):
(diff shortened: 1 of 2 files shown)


blocks/select.py
```diff
@@ -177,6 +177,12 @@ def recursion(brick):
             for child in brick.children:
                 for path, param in recursion(child).items():
                     new_path = Path([Path.BrickName(brick.name)]) + path
+                    if new_path in result:
+                        raise ValueError(
+                            "name collision while retrieving parameters (" +
```
rizar: Can you please split this error message into a concise one and a verbose one like we do it [here](https://github.com/bartvm/blocks/blob/master/blocks/main_loop.py#L181)?

rizar:
LGTM except for the error message.

blocks/select.py
vdumoulin: Done.

rizar:
Oops, forgot about it!

Issue #594, opened 2015-04-20
Trainable Initial State for Recurrent Bricks


rizar:
It is very easy to modify `SimpleRecurrent`, `GatedRecurrent` and `LSTM` to make their initial states trainable. It gives small but consistent gain in my application. In general I think that zero hidden states might be dangerous.

What you do guys think? Should we implement that and makes the new Blocks default?

bartvm:
Yes, I think that training the initial state is the sensible thing to do in most applications.

dwf:
Sounds reasonable to me.

rizar:
Nice that you like this plan guys. I will implement in the nearest future then.

dwf:
I think this was resolved by #600, @rizar please reopen if I'm incorrect.

PR #600, opened 2015-04-26
Trainable Initial State for Recurrent Bricks

rizar:
As discussed in #594

### DIFF
--- a/blocks/bricks/recurrent.py
+++ b/blocks/bricks/recurrent.py
@@ -12,8 +12,8 @@ from blocks.bricks import Initializable, Sigmoid, Tanh
 from blocks.bricks.base import Application, application, Brick, lazy
 from blocks.initialization import NdarrayInitialization
 from blocks.roles import add_role, WEIGHT
-from blocks.utils import (pack, shared_floatx_nans, dict_union, dict_subset,
-                          is_shared_variable)
+from blocks.utils import (pack, shared_floatx_nans, shared_floatx_zeros,
+                          dict_union, dict_subset, is_shared_variable)
 
 logger = logging.getLogger()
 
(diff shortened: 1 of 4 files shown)


rizar:
The Theano bug was fixed, tests pass, PR is ready for review. @bartvm , @pbrakel

rizar:
@bartvm, @pbrakel, @dmitriy-serdyuk , somebody, please review this PR.

blocks/bricks/recurrent.py
```diff
@@ -32,6 +32,10 @@ class BaseRecurrent(Brick):
     def initial_state(self, state_name, batch_size, *args, **kwargs):
         r"""Return an initial state for an application call.
 
+        Default implementation returns a zero matrix. All the standard
+        recurrent bricks override it with trainable initial states
```
bartvm: "standard recurrent bricks" is a bit vague, I guess you mean bricks derived from `SimpleRecurrent`?

bartvm:
Looks good, but it would be nice to have some tests.

rizar:
Sure, I will think of something.

rizar:
I got rid of "standard recurrent bricks" and wrote some tests (in fact I simply just check that trainable initial states are there in the computation graph).

rizar:
Waiting for a final LGTM from anybody of @bartvm , @pbrakel , @dmitriy-serdyuk.

bartvm:
LGTM, but I'd rather @pbrakel has a quick look as well, because they use these bricks a lot and I'm not sure what the consequences are for them.

pbrakel:
As far as I can tell, this doesn't seem to break any of my code and learnable initial states seem like a good thing. Just a quick question: How do I define the initialization? Should I use `biases_init`, `weights_init` or something else?

rizar:
They are initially zeros and this is not configurable, see https://github.com/rizar/blocks/blob/gru_initial_state/blocks/bricks/recurrent.py#L385

If one needs it, a special initialization scheme can be added for those later. The initial states are neither weight nor biases, so I do not think that applying `weight_init` or `biases_init` would be reasonable.

pbrakel:
I'm perfectly fine with the standard being zero. I just wanted to be sure it wouldn't be NaN :).

rizar:
Cool, I assume this was the final OK for merge :)

PR #602, opened 2015-04-28
 Change the error message in model.py : top bricks with the same name ret...

mducoffe:
I was struggling to understand what was wrong in my code as I put differents names for every parameters in my blocks model. Eventually It turns out that this error message is about the main bricks and not the parameters inside. I just thought it would help to add the first name where some occurrences happened. Maybe it is already handled in some ticket but in case there is not, it would be great to add such precision in the error message for beginners.

Have a nice day

### DIFF
--- a/blocks/model.py
+++ b/blocks/model.py
@@ -6,6 +6,7 @@ the basic :class:`AbstractModel` interface as well as its implementations
 (currently only :class:`Model`).
 
 """
+import numpy
 import logging
 from abc import ABCMeta, abstractmethod
 from collections import OrderedDict
@@ -156,7 +157,11 @@ class Model(AbstractModel, ComputationGraph):
             if brick not in children and brick not in self.top_bricks:
                 self.top_bricks.append(brick)
         if len(set(b.name for b in self.top_bricks)) < len(self.top_bricks):
-            raise ValueError("top bricks with the same name")
+            # find the first occurences and return it in the error message
+            liste_name = [b.name for b in self.top_bricks]
+            occ = liste_name[numpy.argmax([liste_name.count(liste_name[i])
+                            for i in xrange(len(liste_name))])]
+            raise ValueError("top bricks with the same name : "+occ)
 
         brick_param_names = {
             v: k for k, v in Selector(self.top_bricks).get_params().items()}
--- a/blocks/model.py
+++ b/blocks/model.py
@@ -6,10 +6,9 @@ the basic :class:`AbstractModel` interface as well as its implementations
 (currently only :class:`Model`).
 
 """
-import numpy
 import logging
 from abc import ABCMeta, abstractmethod
-from collections import OrderedDict
+from collections import OrderedDict, Counter
 from itertools import chain
 
 from six import add_metaclass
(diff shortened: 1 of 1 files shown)


rizar:
Very nice initiative! I will have some remarks about the code, otherwise I think it is a very nice idea to have a more verbose message here.

blocks/model.py
```diff
@@ -156,7 +157,11 @@ def __init__(self, outputs):
             if brick not in children and brick not in self.top_bricks:
                 self.top_bricks.append(brick)
         if len(set(b.name for b in self.top_bricks)) < len(self.top_bricks):
-            raise ValueError("top bricks with the same name")
+            # find the first occurences and return it in the error message
+            liste_name = [b.name for b in self.top_bricks]
+            occ = liste_name[numpy.argmax([liste_name.count(liste_name[i])
+                            for i in xrange(len(liste_name))])]
```
rizar: - liste_name -> list_of_names or name_list

- occ -> repeated_name

- the counting should be much easier to implement with [collections.Counter](https://docs.python.org/2/library/collections.html). You should be interested in its `most_common` method. Also it can be used instead of 
` len(set(b.name for b in self.top_bricks)) < len(self.top_bricks):` with the advantage that it gives you the name of the duplicate.

- the space that you put in the error message after the `same name` and before colon is a little bit non-standard. I think it's better to have write `same name:`

bartvm: It's pretty redundant to call it `list_of...` of `..._list` actually. Just `names` will do.

rizar: Agreed :)

blocks/model.py
```diff
@@ -155,8 +155,12 @@ def __init__(self, outputs):
         for brick in bricks:
             if brick not in children and brick not in self.top_bricks:
                 self.top_bricks.append(brick)
-        if len(set(b.name for b in self.top_bricks)) < len(self.top_bricks):
-            raise ValueError("top bricks with the same name")
+        # check there is no occurence in top bricks name
+        names = Counter([b.name for b in self.top_bricks])
```
bartvm: This might be an easier way to get all the names that are not unique, instead of just one:
```python
names = set()
repeated_names = [brick.name for brick in self.top_bricks if brick.name in names or names.add(brick.name)]
```

bartvm:
```python
names = Counter([brick.name for brick in self.top_bricks])
repeated_names = [name for name, count in names.items() if count > 1]
```
works too of course

blocks/model.py
```diff
@@ -155,8 +155,12 @@ def __init__(self, outputs):
...
+        if names.most_common(1)[0][1] > 1:
+            repeated_name = names.most_common(1)[0][0]
+            raise ValueError("top bricks with the "
+                             "same name : %s" % repeated_name)
```
bartvm: Please use the new formatting syntax: `"top bricks with the same name: {}".format(name)`

blocks/model.py
rizar: I would rather go the second option: having things done in an `or` operand reminds me of sport programming in C++ and make me shiver.

bartvm: I agree. The first one is just a bad habit, faster but ugly.

mducoffe:
By the way I have made the tests on my own code, do you need a test or something is already implemented ?

blocks/model.py
```diff
@@ -155,8 +155,12 @@ def __init__(self, outputs):
         for brick in bricks:
             if brick not in children and brick not in self.top_bricks:
                 self.top_bricks.append(brick)
-        if len(set(b.name for b in self.top_bricks)) < len(self.top_bricks):
-            raise ValueError("top bricks with the same name")
+        # check there is no occurence in top bricks name
```
bartvm: Not sure what you mean by "occurrence" here, but you can remove the comment, it's pretty clear from the code what it does.

blocks/model.py
```diff
@@ -155,8 +155,12 @@ def __init__(self, outputs):
...
+        names = Counter([brick.name for brick in self.top_bricks])
+        repeated_names = [name for name, count in names.items() if count > 1]
+        if repeated_names:
+            raise ValueError("top bricks with the "
+                             "same name : {}".format(repeated_names))
```
bartvm: I think @rizar commented on this already, but it makes more sense to do `"same name: {}"` and since we're nitpicking anyway, `"same name: {}".format(', '.join(repeated_names))` would be even nicer.

bartvm:
I don't think a test is needed. Testing error messages gets kind of messy anyway, because you need to do string matching or something. Besides my last comment it looks ready to merge IMO.

mducoffe:
Done

2015-04-28 14:35 GMT-04:00 Bart van Merriënboer <notifications@github.com>:

bartvm:
Great, but with the latest change line 162 is one character too long... (Also, it still says `same name :` instead of `same name:`)

rizar:
Travis error seems to be not related to the PR, merging.

PR #605, opened 2015-04-29
Set names for initial states of recurrent bricks

dmitriy-serdyuk:
Currently it is really hard to debug complicated recurrent structures like ones with sequence generators. This is going to help to distinguish different initial states.

### DIFF
--- a/blocks/bricks/recurrent.py
+++ b/blocks/bricks/recurrent.py
@@ -29,7 +29,8 @@ class BaseRecurrent(Brick):
     has_bias = False
 
     @application
-    def initial_state(self, state_name, batch_size, *args, **kwargs):
+    def initial_state(self, application, state_name, batch_size, *args,
+                      **kwargs):
         r"""Return an initial state for an application call.
 
         Parameters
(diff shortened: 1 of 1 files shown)


rizar:
I understand it can be helpful, but this is quite a dirty hack. Applications outputs in Blocks are typically set either in the brick's code or in the brick's constructor, and changing them in runtime would break this invariant.

Maybe tagging the resulting variable would help?

dmitriy-serdyuk:
`initial_state` function is unusual for blocks, its return name depends on the input.

dmitriy-serdyuk:
Is it better idea to rename initial states in `recurrent` decorator?

rizar:
No sure what you mean by the last entry.

I think that we should think at this point what exactly semantics of `inputs` and `outputs` is. If we decide that both are mandatory, then I would suggest that instead of an `initial_state` method we have `initial_states` method that returns all of them at the same time. If we decide that `outputs` can be optional, then we can introduce a syntax to specify the name of the returned variable.

I like the first direction so far.

dmitriy-serdyuk:
I mean, that usually `initial_state` is called from https://github.com/bartvm/blocks/blob/master/blocks/bricks/recurrent.py#L175 and it is possible to rename it there.

rizar:
That would be mean to make an exception in clear naming policy that we have in Blocks, `brick_application_name`. I am not sure we want that.

The problem is that `initial_state` breaks the general pattern by returning semantically different variables when called with different arguments, IMHO.

dmitriy-serdyuk:
Yes, I agree. So, I'll close this PR and open an issue.

PR #606, opened 2015-04-30
add cast floatX to avoid theano warning that cuda array can not cast fro...

udibr:
add cast floatX to avoid theano warning that cuda array can not cast from float64 or int64

### DIFF
--- a/blocks/bricks/cost.py
+++ b/blocks/bricks/cost.py
@@ -56,7 +56,8 @@ class SquaredError(CostMatrix):
 class CategoricalCrossEntropy(Cost):
     @application(outputs=["cost"])
     def apply(self, y, y_hat):
-        cost = tensor.nnet.categorical_crossentropy(y_hat, y).mean(acc_dtype=floatX)
+        cost = tensor.nnet.categorical_crossentropy(y_hat, y).mean(
+            acc_dtype=floatX)
         return cost
 
 
--- a/blocks/bricks/cost.py
+++ b/blocks/bricks/cost.py
@@ -56,8 +56,7 @@ class SquaredError(CostMatrix):
 class CategoricalCrossEntropy(Cost):
     @application(outputs=["cost"])
     def apply(self, y, y_hat):
-        cost = tensor.nnet.categorical_crossentropy(y_hat, y).mean(
-            acc_dtype=floatX)
+        cost = tensor.nnet.categorical_crossentropy(y_hat, y).mean()
         return cost
 
 
--- a/blocks/bricks/cost.py
+++ b/blocks/bricks/cost.py
@@ -56,7 +56,7 @@ class SquaredError(CostMatrix):
 class CategoricalCrossEntropy(Cost):
     @application(outputs=["cost"])
     def apply(self, y, y_hat):
-        cost = tensor.nnet.categorical_crossentropy(y_hat, y).mean()
+        cost = tensor.nnet.categorical_crossentropy(y_hat, y).mean(acc_dtype=floatX)
         return cost
 
 
--- a/blocks/bricks/cost.py
+++ b/blocks/bricks/cost.py
@@ -108,6 +108,8 @@ class MisclassificationRate(Cost):
 class BinaryMisclassificationRate(Cost):
     @application(outputs=["error_rate"])
     def apply(self, y, y_hat):
+        # Here we have to cast both operands to floatX explicitly.
+        # Because int64 / float32 = float64 in Theano, unfortunately.
         return (tensor.sum(tensor.neq(y, y_hat > 0.5)).astype(
             theano.config.floatX) /
                 y.shape[0].astype(theano.config.floatX))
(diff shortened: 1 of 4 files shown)


rizar:
How do I reproduce such a warning? I ran MNIST example on my GPU and everything was clean.

udibr:
Add Xavier weight initialization and BinaryMisclassificationRate which is similar to MisclassificationRate but works with a binary (scalar) target

rizar:
@udibr , as I said in my previous message, I can not reproduce the warning you report by running a MNIST demo. Can you shed some light on when it actually happens?

blocks/initialization.py
```diff
@@ -223,3 +223,23 @@ def generate(self, rng, shape):
                                                  replace=False)
             weights[i, random_indices] = values[i]
         return weights
+
+
+class Xavier(NdarrayInitialization):
+    """Initialize with Gaussian distribution with Xavier parameters.
+
+    Use the following gaussian parameters: mean=0 and var=scale/Nin
```
rizar: Apparently you forgot a square root here. Also we would need a reference to the paper here.

rizar:
Because this sort of urgent, I opened a new PR for this issue to be merged right away, #635. However, you can continue working on this one and we will be happy to merge it later, when it is ready.

blocks/initialization.py
udibr: I've add reference (and changed var to std so as to not confuse the reader)
however, the original paper uses a uniform distribution and I am using a gaussian....
so either I change my code or not call it Xavier or keep it as is (most  places use gaussian)

blocks/bricks/cost.py
```diff
@@ -58,7 +58,8 @@ def cost_matrix(self, y, y_hat):
 class CategoricalCrossEntropy(Cost):
     @application(outputs=["cost"])
     def apply(self, y, y_hat):
-        cost = tensor.nnet.categorical_crossentropy(y_hat, y).mean()
+        cost = tensor.nnet.categorical_crossentropy(y_hat, y).mean(
+            acc_dtype=floatX)
```
rizar: Why do you think we need this?

rizar:
Xavier initialization and the fix in reverse_words look good to me, but please rebase the PR against the latest master (see our http://blocks.readthedocs.org/en/latest/development/pull_request.html for instructions)

blocks/initialization.py
```diff
@@ -223,3 +223,33 @@ def generate(self, rng, shape):
                                                  replace=False)
             weights[i, random_indices] = values[i]
         return weights
+
+
+class Xavier(NdarrayInitialization):
```
dwf: "Xavier" is the inventor's first name, so it's kind of weird to call it that. "Glorot" or "GlorotBengio" might be a better name...

blocks/bricks/cost.py
udibr: It resolved, for me, the same error messages I got in MisclassificationRate regarding float64 and CUDA. I guess this happens when y is int64.

rizar: I think it will be fixed with https://github.com/Theano/Theano/issues/2913.

I also checked that given the first `float32` and the second `int64` argument it returns `float32`, which is just what we need:

```
In [5]: tensor.nnet.categorical_crossentropy(numpy.array([[0.5, 0.5], [0.5, 0.5]], dtype='f32'), numpy.array([0, 1], dtype='i64')).eval()
Out[5]: array([ 0.69314718,  0.69314718], dtype=float32)
```

blocks/bricks/cost.py
```diff
@@ -70,3 +71,12 @@ def apply(self, y, y_hat):
...
             tensor.sum(tensor.neq(y, y_hat.argmax(axis=1))).astype(floatX) /
             y.shape[0].astype(floatX))
+
+
+class BinaryMisclassificationRate(Cost):
+    @application(outputs=["error_rate"])
+    def apply(self, y, y_hat):
+        # Here we have to cast both operands to floatX explicitly.
+        # Because int64 / float32 = float64 in Theano, unfortunately.
+        return (tensor.sum(tensor.neq(y, y_hat > 0.5)).astype(floatX) /
```
rizar: This is almost a copy of `MisclassificationRate`, which is not nice. 

You can add it as the second application method to `MisclassificationRate`, e.g. `apply_binary`, with the common code they have moved to a private method.

udibr: I also thought this way but then I noticed that in the same file you have both
BinaryCrossEntropy and CategoricalCrossEntropy so I thought it would be more appropriate to have a separate misclassification rates

rizar: Good point. I do not think there is much sense in having them as separate bricks. @bartvm , @dwf , what do you think?

rizar:
It seems like your pull request contains your master branch.. Can you make a bit clearer what of this do you actually want to contribute to Blocks? And it would be better to have separate PRs for the initialization scheme, Adagrad, etc.

udibr:
You are right. I should have a separate branch for each feature and not everything on master. I am closing this pull request and will try to fix this

PR #608, opened 2015-04-30
Clear __init__.py

bartvm:
In order to make https://github.com/mila-udem/blocks-extras work as a namespace package, the `__init__.py` file needs to be entirely cleared (see https://pythonhosted.org/setuptools/setuptools.html#namespace-packages).

To make this work:

* Set `__version__` in `__init__.py` instead of loading it with `pkg_resources`
* Move `blocks.config` to `blocks.config.config`. Not as clean, but there's not other way I can think of without programmatically changing the namespace, and I find that a bit too messy.


rizar:
LGTM

dwf:
Can I ask that you put some scary comments in `__init__.py` to discourage people from submitting pull requests that modify it?

PR #607, opened 2015-04-30
Custom pickle

dmitriy-serdyuk:
I started to port this code to theano dump/load.

PR #610, opened 2015-04-30
Dump/load from theano for custom pickler.

dmitriy-serdyuk:


dmitriy-serdyuk:
Substitutes #607.

dmitriy-serdyuk:
@bartvm , is it okey now? Strange, that it still doesn't have your commits.

bartvm:
Oh, I only just realized you're trying to merge against the `custom_pickle` branch! It will only trigger if you try to merge into `master`.
```
PR #610 accepted (branch not included or excluded)
  53ef06c - Used dump/load in DumpManager
  received at: 2015-04-30 17:41:38
```

PR #612, opened 2015-05-01
Temp dump to the right partition

nouiz:
I didn't test it, so make sure to test it before merging.

### DIFF
--- a/blocks/serialization.py
+++ b/blocks/serialization.py
@@ -74,7 +74,8 @@ def secure_pickle_dump(object_, path):
 
     """
     try:
-        with tempfile.NamedTemporaryFile(delete=False) as temp:
+        d = os.path.dirname(path)
+        with tempfile.NamedTemporaryFile(delete=False, dir=d) as temp:
             pickle_dump(object_, temp)
         shutil.move(temp.name, path)
     except:
--- a/blocks/serialization.py
+++ b/blocks/serialization.py
@@ -74,8 +74,11 @@ def secure_pickle_dump(object_, path):
 
     """
     try:
-        d = os.path.dirname(path)
-        with tempfile.NamedTemporaryFile(delete=False, dir=d) as temp:
+        # Use the same destination directory, as /tmp can be too
+        # small.  This also make the move to copy if the destination
+        # wasn't on the same partition.
+        with tempfile.NamedTemporaryFile(delete=False,
+                                         dir=os.path.dirname(path)) as temp:
             pickle_dump(object_, temp)
         shutil.move(temp.name, path)
     except:


bartvm:
What's wrong with dumping to `TMPDIR`? The idea is that even though we don't delete it immediately because we pass `delete=False`, the system should still clear it eventually. This way the user ends up with a strangely named temporary file in the target directory that needs to be deleted manually, right?

If the problem is that you want to delete temporary files, can't we just add `os.remove(temp.name)` at the end?

dwf:
The system SHOULD still clear it eventually. But not at LISA.

TMPDIR is set to /Tmp which is manual cleanup only, which is basically a
technical expression of a disbelief in human nature. :P

bartvm:
*sigh* I guess we can add an `os.remove` at the end, but I don't like the idea of littering user directories with vaguely named temporary files just to deal with the LISA systems' idiosyncrasies.

dwf:
I raise this on @nouiz's pull request in the hope that we can eventually change TMPDIR to an auto-removed directory so as to conform to expectations of practically every application in the world, even if we have /Tmp available for explicit, conscious use.

dwf:
There is an argument for putting the temporary file in the same directory, even if it gets removed eventually: it can help with absolutely catastrophic failure recovery.

bartvm:
Right, I guess that if something goes wrong during the actual copying of the temporary file, it could be useful. Not a big use case though, because the file is very likely to stay in `TMPDIR` till the next reboot, so you can just retrieve it from there yourself if things do go catastrophically wrong.

nouiz:
The problem isn't what you think. At LISA, the temp file get created under
/tmp, not /Tmp. It is way too small, so this fill it up frquently. (a few
people come in the last few days before they pickle files that where taking
Gs and filled /tmp).

If people safe big size stuff, the only place where we can be sure there is
enough space, is at the destination.

rizar:
Wouldn't be better to have a configurable temporary directory for Blocks, e.g. `config.temp_path`? This way people can always set it to '.' in their config when their `/tmp` is small.

nouiz:
By default, I think we should expect that dump can be big and it is
frequent that /tmp is small on cluster.

What you don't like about my fix?

bartvm:
The current fix doesn't remove the temporary file in the end, so you end up with a large number of temporary files (one for each checkpoint) in the target directory that are never removed, right?

nouiz:
There is already a try: except that should take care of that.

But yes, there is very rare case where this can happen.

rizar:
@bartvm, we use `shutil.move`, that is if serialization does not crash, there will be no temporary file left.f

bartvm:
My bad, didn't read the code carefully enough. In that case, this looks fine by me! (Although I'd prefer a more explanatory variable name than `d`).

rizar:
Maybe no intermediate variable altogether?

nouiz:
Done and added a comment

dwf:
@nouiz On LISA `tempfile` creates under `/Tmp` for me.

```
>>> tempfile.mkstemp()
(4, '/Tmp/wardefar/tmpmynv36n_')
```

I think it keys off $TMPDIR, and mine is set to /Tmp/wardefar, and I certainly didn't set it myself.

dwf:
(Yes, just verified it is in `/opt/lisa/os/.local.bashrc`)

rizar:
However, I think it makes sense to use the local directory by default, because it surely has enough space to contain the pickle (since the user dared to run his experiment there).

I would vote for merging this PR.

PR #615, opened 2015-05-01
Custom pickler

dmitriy-serdyuk:
Via @bartvm in the comments on #685 (adding here for posterity):

[This PR] combines the previous `Checkpoint` extension (which pickled the main loop) and the `Dump` extension (which collected the values of the parameters and saved them to an NPZ file). It does so by making use of pickle's "persistent ID" support (which we already merged into Theano a while ago, https://github.com/Theano/Theano/pull/2802).

In practice, it means that during pickling, each shared variable gets intercepted by the `PersistentSharedVariableID` class and is assigned a name (if its owned by a brick, the path to the brick is used, otherwise the name of the shared variable is used). The value of the shared variable is saved as an NPY file to a ZIP file, together with the pickled main loop (which contains string references to the names of the parameters). The final result is a single ZIP file with a pickled main loop and a series of NumPy arrays containing the values of the all the shared variables contained within the main loop.

Advantages over the old method:

* No need to use both `Checkpoint` and `Dump`. This single method allows you to resume training seamlessly, while also allowing you to inspect and share parameter values across platforms; no unpickling needed.
* Parameters are no longer silently overwritten if they had the same name and/or don't belong to a brick, as `Dump` does.
* The entire serialization situation gets (a) less confusing and (b) documented (because the current docs are out of date, and still talk about using Dill).


blocks/serialization.py
```diff
@@ -1,83 +1,109 @@
...
 
-def secure_pickle_dump(object_, path):
-    """Robust serialization - does not corrupt your files when failed.
+    """
+    def __call__(self, obj):
+        if isinstance(obj, SharedVariable):
+            if obj.name:
+                if obj.name == 'pkl':
+                    ValueError("can't pickle shared variable with name `pkl`")
+                if hasattr(obj, 'tag'):
```
bartvm: Don't shared variables always have the `tag` attribute? It could be an empty dictionary though if it's not annotated.

Thinking about it, this class should probably inherit from `PersistentNdarrayID`. Right now, when `super` gets called, doesn't the parent class (`PersistentSharedVariableID`) simply override the name you've just created?

dmitriy-serdyuk: Oh.. Funny, it was working because I made a wrong super call.

dmitriy-serdyuk:
I suppose, we won't need `DumpManager` anymore, but `load_parameters` and `save_parameters` functions might be useful for some people. 

I'd like to save `LoadFromDump` extension. In order to make it work with a new serialization I'll pickle log and iteration state separately.

Do you agree?

bartvm:
We'd only need `load_parameters`, right? There's no real need for `save_parameters`, considering that it's exactly the same as `dump`, except perhaps that `save_parameters` could raise an error for non-unique names.

I would change `LoadFromDump` to something like `Load` or `Resume` then, because the reason we had "dump" was to make a distinction between "checkpointing" (pickling everything) and "dumping" (which saved the parameters, but neglected to save a lot of other things). Ideally we get rid of this distinction, so then there will just be `Checkpoint`, `Resume`, `load_parameters`, and potentially `save_parameters` as a very thin wrapper around `dump` (for symmetry's sake).

dmitriy-serdyuk:
@bartvm , the problem was completely different and I solved it now. 

But now `blocks-continue` doesn't work for the reason we discussed: objects from global name space cannot be pickled. And actually it didn't work with normal pickler. 

I propose to remove `blocks-continue` and force people to use `Load` extension.

dmitriy-serdyuk:
It is also possible to use a construction in a script like
```
if __name__ == '__main__':
    if continue:
        continue_training(path)
    ...
```

bartvm:
So `blocks-continue` always had this limitation? Or did something change with this new approach?

The `Load` extension doesn't do the same thing though. The purpose of pickling is being able to completely reconstruct the main loop, including the state of extensions, etc. So it doesn't make sense to expect the user to first construct a `MainLoop` instance, because it would be entirely replaced by whatever the `Load` extension unpickles (and performing this replacement from within the extensions would be a mess).

If this limitation has always existed, I'd argue that we just need to document it clearly. I imagine that something like this should work fine, right?

```python
# my_model.py
def some_helper_function():
   pass
```
```python
# train_my_model.py
from my_model import some_helper_function
main_loop = MainLoop(...)  # With Checkpoint, depends on some_helper_function
main_loop.run()
```
Because when unpickling, it can load the function from `my_model`.

dmitriy-serdyuk:
Yes, it always had this limitation. 

Yes, this is going to work, but we cannot provide a general script which is able to continue any kind of experiment. So I think, the best we can provide is a function `continue_training` and explain in the documentation that it should be used like
```python
# define or import objects from global namespace like you defined it before
if __name__ == '__main__':
    if continue:
        continue_training(path)
    # Construct main loop with Checkpoint extension
```

bartvm:
Right, we could just make it a function instead of a script, but we should make it clear that using it within a script isn't necessarily the only way to use it. Technically speaking the user could have just created their model in the interpreter, without any sort of script anywhere. They should still be able to fire up a new interpreter then and call `continue_training(path)` then.

Anyway, that sounds like a reasonable solution to me. The one thing that I think would be very nice--but I have no idea how difficult this is--is being able to give a warning whenever the main loop depends on functions defined in the `__main__` namespace. It could say something like:

```
WARNING: Main loop depends on the function `_array_tuple` in `__main__` namespace. 

Because of limitations to pickling, this means that you will not be able to resume
your model outside of a namespace containing this function. In other words, you
can only call `continue_training` from within this cript.
```

Looking at `pickle.py` maybe it would work by overwriting `save_global` to check if `module` is `'__main__'` or something?

dmitriy-serdyuk:
@bartvm , is there any documentation for python 3 pickle internals? [This one](https://docs.python.org/3/library/pickle.html) doesn't seem full.

dmitriy-serdyuk:
How can I test, that arrays were not pickled in a regular way? Is there a better way then to check `pkl` file size?

bartvm:
You could load the zip file using `numpy.load` and check whether the array is one of the keys there, right?

I don't know of any documentation on pickle's internals that is any better... It's pretty badly documented I'm afraid. You can look at custom picklers like `joblib`'s to get a better idea of how to write custom ones. Is there anything in particular you're running into?

dmitriy-serdyuk:
Yes, but I'd like to check that it was not saved twice. If something goes wrong in `__call__` method, it may add a numpy file into zip, but not to return an id.

Seems, there is no `save_global` function in python 3 or its signature is different.

bartvm:
I just opened up `pickle.py` from Python 3, and it seems like the class is called `_Pickler` and the method looks like `def save_global(self, obj, name=None)`

blocks/serialization.py
```diff
@@ -1,65 +1,183 @@
...
+                name = '{0}_{1}'.format(name, count + 1)
+            self.name_counter[name] += 1
+            return name
+        return super(PersistentParameterID, self)._resolve_name(obj)
+
+    def __call__(self, obj):
+        if isinstance(obj, SharedVariable):
+            if obj.name:
+                if obj.name == 'pkl':
+                    ValueError("can't pickle shared variable with name `pkl`")
```
bartvm: Forgot the `raise` statement here.

bartvm:
I think this is starting to look pretty good, and it'd be great if some other people could have a look, because it's a pretty big change (@rizar, @dwf).

It replaces the `Dump` and `Checkpoint` extensions by a custom pickler (an extension of the ones we recently merged into Theano as part of https://github.com/Theano/Theano/pull/2802). The result is a single zip archive that contains the pickled main loop (as a file called `pkl`) and all the shared variables stored as NPY files. The names of these arrays are either the path leading to them (numbered, if need be e.g. `mlp/linear/W_0`, `mlp/linear/W_1`) or the name of the shared variable (also numbered if necessary).

The advantage is that it effectively does what `Dump` does, but without needing a `Model` instance, while still pickling the entire main loop (so the state of e.g. extensions is maintained). The resulting zip archive can be loaded using NumPy's `numpy.load` function if necessary in order to inspect parameters without unpickling.

dmitriy-serdyuk:
Maybe I need to rebase it because I made a lot of stupid commits changing things back and forth.

bartvm:
Yeah, a rebase would be nice eventually. Also, I made a comment about a missing `raise` statement.

bartvm:
Oh, and we'll need a bunch of tests I guess.

bartvm:
The main test I'd like to add is to make sure that https://github.com/bartvm/blocks/issues/480 is fixed for both brick parameters as well as normal shared variables.

rizar:
First thing I will tomorrow morning will be reading your PR, I promise.

But this is quite a major change, a first of its kind since quite some time. And people at ICLR were complaining that there is no stable version of Blocks they can refer to when releasing their code. So I propose to call whatever we have in Blocks before this PR Blocks 0.01, make a branch and merge only bug fixes in it. It should be helpful for people who are actually using Blocks. Especially taking into account that NIPS is coming, and some people (at least me) are not willing to change their habits of using `blocks-dump` right away.

What do you think guys?

dmitriy-serdyuk:
Maybe it is better to call blocks 0.01 after this PR to push people to use new checkpoint system. And there are `save_parameters` and `load_parameters` which should work with old dumps.

blocks/extensions/saveload.py
```diff
@@ -49,6 +48,8 @@ class Checkpoint(SimpleExtension):
 
     """
     def __init__(self, path, save_separately=None, **kwargs):
+        if save_separately is None:
+            save_separately = ['log']
```
rizar: Are you sure we want the log to be saved separately by default?

blocks/extensions/saveload.py
```diff
@@ -119,49 +120,28 @@ class LoadFromDump(TrainingExtension):
...
-        self.manager = MainLoopDumpManager(state_path)
+        super(Load, self).__init__(**kwargs)
+        self.state_path = state_path
+
+    def load(self):
+        with open(self.state_path, "rb") as source:
+            return load(source)
+
+    def load_to(self, main_loop):
+        loaded_main_loop = self.load()
```
rizar: I should say that this new `Load` extension is less robust than `LoadFromDump`, because the main loop, including all the bricks, has to be unpicklable. It is often the case, that one simply can not do this, because they are working already with a new code version.

bartvm: We could pickle the log and iteration state to separate files in the same zip archive, using the same persistent pickling approach. In general though, the situation covered by `LoadFromDump` that isn't covered by `Load` is quite specific, namely:

* The iteration state is picklable, but some other part of the main loop isn't
* The epoch is very long (otherwise the iteration state isn't very important)
* You don't care about the state of your extensions, algorithm, etc.

Otherwise you can just load the parameters and resume at the beginning of an epoch. So, although I agree that eventually this robustness would be a good idea, I think it's a relatively specific case, and the advantages of this new approach outweigh it IMHO.

bin/blocks-continue
```diff
@@ -1,15 +0,0 @@
...
-
-if __name__ == "__main__":
-    logging.basicConfig()
-
-    parser = ArgumentParser("Continues your pickled main loop")
-    parser.add_argument(
-        "path", help="A path to a file with a pickled main loop")
-    args = parser.parse_args()
-
-    continue_training(args.path)
```
rizar: Hmm, I was really using `blocks-continue`! What's wrong with it?

bartvm: We could keep it, but the main issue is that pickle doesn't play nice with functions or objects defined in the `__main__` namespace of the training script. If a user relied on those, they won't be able to use `blocks-continue`. That's why @dmitriy-serdyuk decided it was probably more robust to ask the user to rely on `continue_training`, with better documentation that explains how to use `continue_training` from within the training script (where the global namespace is the same).

rizar: Right, this is true, but I still like `blocks-continue`. I just got a habit of having only parameter parsing in the runnable `main.py` script.

blocks/serialization.py
```diff
@@ -1,65 +1,181 @@
...
+        module = getattr(obj, "__module__", None)
+        if module == '__main__':
+            warnings.warn(
+                MAIN_MODULE_WARNING.format(kwargs.get('name', obj.__name__))
+            )
+        pickle.Pickler.save_global(self, obj, **kwargs)
+    if six.PY3:
+        dispatch_table[six.types.FunctionType] = save_global
+    else:
+        dispatch[six.types.ClassType] = save_global
```
rizar: Maybe I searched badly, but I can not find a documentation for the 'dispatch' attribute in Pickle docs. Can you please give me a link, if you have one?

Also, I am bit concerned about the effect all these additional hooks we add will have on the performance. A function call is not that cheap in Python.

rizar: Correct if I am wrong, but seems like this PR also switches us from `cPickle` to `pickle`, which is not good in terms of performance.

bartvm: The pickle documentation isn't great. This is as good as it gets regarding `dispatch`: https://docs.python.org/3/library/pickle.html#pickle.Pickler.dispatch_table

There is no way to subclass `cPickle`, so the switch to `pickle` is necessary for any custom pickling. It's done even in libraries like `joblib` which target HPC. In the end, it's probably still faster than the current situation where you have to use both `Dump` *and* `Checkpoint` if you want seamless resumption as well as robust parameter saving.

rizar: I will check the exact slowdown tomorrow. I am worried that the training log containing lots of small objects will take forever to be handled by `pickle`.

bartvm: We might be better off fixing all those empty dictionaries in the log if it
turns out to be slow, because I do think that this serialisation method is
a big improvement over what we have right now. Also, I don't think that
serialisation is too time critical (if it is, it means you're checkpointing
too often probably).

dmitriy-serdyuk: If it is an issue we can create a custom pickler to save it in a plain text format.

rizar: @bartvm , it's not the empty dicts which are the main issue in my case. I have models with >30 parameters, and ideally for every parameter I would like to track the norm, the gradient norm and the effective step length. This is a lot of numbers and logs ends up quite large, even when they are average over 10 iterations (long live `TrainingDataMonitoring`!).

I measured how long does pickling and unpickling of a log take:

```python
In [33]: %timeit log = pickle.load(open('timit_relu100250_cdp_noise75b_log.pkl'))
1 loops, best of 3: 4.93 s per loop

In [34]: %timeit log = cPickle.load(open('timit_relu100250_cdp_noise75b_log.pkl'))
1 loops, best of 3: 1 s per loop

In [35]: %timeit pickle.dump(log, open('tmp.pkl', 'w'))
1 loops, best of 3: 13.4 s per loop

In [36]: %timeit cPickle.dump(log, open('tmp.pkl', 'w'))
1 loops, best of 3: 3.46 s per loop
```

It is not very nice that unpickling becomes 5 times slower without `cPickle`. Unpickling a log is a basic operation which you have to do lots of times when analyzing your models.

But correct if I am wrong: persistent id functionality is also available in `cPickle`, and the only thing you need a custom `Pickler` for are nice warnings about `__main__` namespace. Is it so?

rizar: Oops, I just realized that one can still use `cPickle` to unpickle the log if it was saved separately. I still need more time to think through the consequences of changes you try to make here.

rizar: I apologize for writing so many posts instead of one. I checked out your branch and benchmarked your `dump` and `load`:

```python
In [11]: %timeit  dump(log, open('log.zip', 'w'))
1 loops, best of 3: 1min 38s per loop

In [12]: %timeit  cPickle.dump(log, open('log.pkl', 'w'))
1 loops, best of 3: 3.3 s per loop

In [14]: %timeit  load(open('log.zip'))
1 loops, best of 3: 2min 10s per loop

In [15]: %timeit  cPickle.load(open('log.pkl'))
1 loops, best of 3: 2.83 s per loop
```

Also, I read more carefully your code, and it seems like you use your custom dump also for `save_separately` elements. Which effectively means that if this PR is merged, loading my logs will take up to 2 minutes instead of a few seconds. 

That being said, I still like the PR. But let's figure out how to reach its goals without drastic slowdowns.

bartvm: Did you try timing it without all the usual empty dictionaries in the log?
Curious to see the impact of that.

bartvm: Oh, and since cPickle doesn't exist in Python 3 anymore, do you have
timings for that as well?

rizar: I will try to measure both, without empty dicts and in Python 3.

rizar: `pickle` in Python3 is a successor to `cPickle` though.

It turned out be fast, whereas the new `dump` and `load` functions are even slower in Python 3:

```python
In [23]: %timeit dump(log, open('log.zip', 'wb'))
^A[^A[1 loops, best of 3: 2min 35s per loop
```

My log contains 25633 rows, of which most contain 8 records. Every tenth row contains about 100 records. I think it is moderately big.

Would be nice to know if it is the persistent ids or the custom pickler that slows down this much. I can do this measurements, if you are interested.

bartvm: I guess cPickle with persistent ID could be an option if it's a lot faster,
although I feel like a lot of people will run into the mistake of pickling
things in the __main__ namespace (I had a lot of people asking about it).
Otherwise a more efficient data structure for the log could be an option
too (e.g. a single dict with tuples as keys).

rizar: Very preliminary information: the bottleneck is not in the custom pickler. I will to profile it tomorrow.

rizar: Here comes the profile

```python
         145163322 function calls (144834560 primitive calls) in 208.903 seconds

   Ordered by: cumulative time

   ncalls  tottime  percall  cumtime  percall filename:lineno(function)
    953/1    0.032    0.000  214.053  214.053 {built-in method exec}
        1    0.086    0.086  214.053  214.053 dump.py:1(<module>)
        1    0.000    0.000  212.078  212.078 serialization.py:121(dump)
 312909/1    3.413    0.000  209.698  209.698 pkl_utils.py:326(zipadd)
        1    0.000    0.000  208.948  208.948 serialization.py:170(func)
        1    1.301    1.301  208.948  208.948 {method 'dump' of '_pickle.Pickler' objects}
   882956    1.569    0.000  207.632    0.000 serialization.py:80(__call__)
   882956    3.142    0.000  205.632    0.000 pkl_utils.py:144(__call__)
   312908    0.780    0.000  102.769    0.000 pkl_utils.py:147(write_array)
   312908    5.249    0.000  101.989    0.000 format.py:507(write_array)
   312908    6.873    0.000   93.054    0.000 format.py:270(_write_array_header)
   312908    7.256    0.000   74.038    0.000 format.py:415(_filter_header)
   312909   11.094    0.000   53.897    0.000 zipfile.py:1315(write)
  5632344   20.615    0.000   49.981    0.000 tokenize.py:473(_tokenize)
   312909    1.500    0.000   27.418    0.000 tempfile.py:571(NamedTemporaryFile)
   312910    2.044    0.000   21.790    0.000 tempfile.py:330(_mkstemp_inner)
   312908    0.619    0.000   14.401    0.000 tokenize.py:302(untokenize)
  5008285   13.965    0.000   13.965    0.000 {method 'match' of '_sre.SRE_Pattern' objects}
   312908    8.048    0.000   13.382    0.000 tokenize.py:245(untokenize)
...
```

The actual reason why it is so slow is unexpected: the log contains lots of small numpy arrays! So I think you guys should only apply the persistent id magic to the shared variables.

bartvm:
Okay, I guess that makes sense. @dmitriy-serdyuk could you make the necessary changes i.e. make sure that there's a flag in the `PersistentSharedVariableID` class that makes sure it saves *only* `ndarrays` belonging to shared variables, but none of the others? I guess this might require changes in Theano as well, although I'm not sure about that.

After that, it'd be interesting to see how fast it is to use `cPickle` instead of `pickle`. This won't allow us to give a warning about `__main__`, but if it's a big difference, we could make `pickle` the default but `cPickle` with an argument to `dump`.

dmitriy-serdyuk:
For my log it decreases pickling time from 5.3s to 1.4s while dumping without a persistent id with `cPickle` takes 238ms and 946ms with `pickle`.

dmitriy-serdyuk:
How can I create a pull request to two repositories simultaneously? Tests fail for `blocks-examples` repository.

bartvm:
The way I've done it is by making a pull request to `blocks-examples`, and then changing this PR's `.travis.yml` to do `git clone` of your branch instead. Then if everything works, and things are ready to merge, we just change the `git clone` line back just before doing so.

bartvm:
Great, so two things left to do then:

- [ ] Undo deleting `blocks-continue`, if people like @rizar use it we might as well keep it, but knowing that sometimes it won't work.
- [ ] Make `dump` optionally use `cPickle` by doing something like

```python
def dump(obj, file_handler, protocol=DEFAULT_PROTOCOL,
         persistent_id=PersistentParameterID, use_cpickle=False):
    with closing(zipfile.ZipFile(file_handler, 'w', zipfile.ZIP_DEFLATED,
                                 allowZip64=True)) as zip_file:
        def func(f):
            if use_cpickle:
                p = cPickle.Pickler(f, protocol=protocol)
            else:
                p = PicklerWithWarning(f, protocol=protocol)
            p.persistent_id = persistent_id(zip_file)
            p.dump(obj)
        pkl_utils.zipadd(func, zip_file, 'pkl')
```

rizar:
On my log in Python3 the slowdown from switching from `pickle.dump` and `pickle.load` to the functions from this PR is 3 times. This is much less than before, good job, but still quite a lot of overhead. I will try it with Python2 as well when you guys support `cPickle` there.

bartvm:
Oops, my bad on the formatting errors, but at least tests are passing again. We need a bunch of tests now, because I realised that the previous PR was actually not storing a lot of things in the ZIP file (and the initial one was storing too many things...).

@rizar You might want to re-run your benchmarks, and test the support for `cPickle`.

bartvm:
Needs to be rebased on top of https://github.com/mila-udem/blocks/pull/680 when that is merged

bartvm:
#680 was merged, so could you rebase @dmitriy-serdyuk? Earlier the doctests were failing

examples/sqrt.py
```diff
@@ -61,14 +61,14 @@ def main(save_to, num_batches, continue_=False):
             step_rule=Scale(learning_rate=0.001)),
         get_data_stream(range(100)),
         model=Model(cost),
-        extensions=([LoadFromDump(save_to)] if continue_ else []) +
+        extensions=([Load(save_to)] if continue_ else []) +
```
bartvm: Oops, missed this line, needs to be deleted

setup.py
```diff
@@ -32,7 +32,7 @@
     ],
     keywords='theano machine learning neural networks deep learning',
     packages=find_packages(exclude=['examples', 'docs', 'tests']),
-    scripts=['bin/blocks-continue', 'bin/blocks-dump', 'bin/blocks-plot'],
+    scripts=['bin/blocks-plot'],
```
bartvm: Need to add `bin/blocks-continue` back

tests/test_dump.py
```diff
@@ -3,7 +3,7 @@
 import numpy
 from picklable_itertools.extras import equizip
 
-from blocks.dump import load_parameter_values, save_parameter_values
+from blocks.serialization import load_parameter_values, save_parameter_values
 
 
 def test_save_load_parameter_values():
```
bartvm: Need to delete this file now that `save_parameter_values` is gone.

blocks/serialization.py
```diff
@@ -1,65 +1,167 @@
...
-not rely on them and use normal function definitions instead."""
+    """
+    def __call__(self, obj):
+        if isinstance(obj, SharedVariable):
+            super(PersistentParameterID, self).__call__(obj)
+            if hasattr(obj.tag, 'annotations'):
+                name = '{}.{}'.format(
+                    BRICK_DELIMITER.join([brick.name for brick in
+                                          get_brick(obj).get_unique_path()]),
+                    [obj.name]
```
bartvm: @dmitriy-serdyuk The current error is my bad, this should be `obj.name` instead of `[obj.name]`

blocks/extensions/saveload.py
```diff
@@ -51,12 +50,11 @@ class Checkpoint(SimpleExtension):
...
         kwargs.setdefault("after_training", True)
         super(Checkpoint, self).__init__(**kwargs)
-
+        if not save_separately:
+            save_separately = []
         self.path = path
         self.save_separately = save_separately
 
-        if not self.save_separately:
-            self.save_separately = []
```
bartvm: This blank line needs to go

bartvm:
So there's only one thing left to do that I can think of, which is changing `Load` to use `numpy.load` instead of unpickling when loading the parameters, and making the loading of the iteration state and log optional. This will allow someone to load the parameters from a model, even if unpickling fails completely.

PR #618, opened 2015-05-04
Changes for the new MNIST fuel dataset.

Thrandis:
Changes for MNIST refactoring. (bartvm/fuel#95)

### DIFF
--- a/examples/mnist.py
+++ b/examples/mnist.py
@@ -29,7 +29,7 @@ def main(save_to, num_epochs, bokeh=False):
               weights_init=IsotropicGaussian(0.01),
               biases_init=Constant(0))
     mlp.initialize()
-    x = tensor.tensor4('features')
+    x = tensor.matrix('features')
     y = tensor.lmatrix('targets')
     probs = mlp.apply(tensor.flatten(x, outdim=2))
     cost = CategoricalCrossEntropy().apply(y.flatten(), probs)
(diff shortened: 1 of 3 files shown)


vdumoulin:
We'll need to change `.travis.yml` as well, I'm afraid. Here's what we'll need to do:
* Change lines 2-4 to
``` yml
cache
  directories:
    - $TRAVIS_BUILD_DIR/data
```
* Remove lines 22-33 (those that download the raw MNIST files)
* Change line 34 to `export FUEL_DATA_PATH=$TRAVIS_BUILD_DIR/data`
* Right after line 40, add
``` yml
  - |
      if [ ! -f $FUEL_DATA_PATH/mnist.hdf5 ]; then
        cd $FUEL_DATA_PATH
        fuel-download mnist
        fuel-convert mnist
        fuel-download --clear mnist
        cd -
      fi
```

examples/mnist.py
```diff
@@ -29,9 +29,9 @@ def main(save_to, num_epochs, bokeh=False):
               weights_init=IsotropicGaussian(0.01),
               biases_init=Constant(0))
     mlp.initialize()
-    x = tensor.matrix('features')
+    x = tensor.tensor4('features')
```
vdumoulin: There's an even simpler way to make sure inputs are flat: pass `flatten=('features',)` to the `MNIST` constructor.

vdumoulin:
To fix the failing doctest in `docs/index.rst`, just change line 89 to
``` python
>>> mnist_train = MNIST("train", flatten=('features',))
```

Thrandis:
mhh some more doc to fix!

PR #619, opened 2015-05-04
Made merge brick apply application

dmitriy-serdyuk:


dmitriy-serdyuk:
There is something wrong with `Merge` brick. This code
```
z = merge.apply(x, y)
cg = ComputationGraph(z)
VariableFilter(bricks=[merge], roles=[OUTPUT])(cg.variables)
```
returns `[merge_apply_x, merge_apply_y]`.

dmitriy-serdyuk:
I added a test for this case. Do you have any ideas why is is happening?

rizar:
I think this weird behavior is due to the fact that `Merge` inherits from `Parallel` and call the parent's apply method in its apply method. However, I would then expect three OUTPUT variables: two of `Parallel.apply` and one of `Merge.apply`.

rizar:
As for the main body of your PR, it looks all right to me.

dmitriy-serdyuk:
Yes, you are right, there were 3 outputs.

rizar:
But then it is completely expected behavior, isn't it? Should your new test be fixed to check there 3 variables are found?

dmitriy-serdyuk:
It is still returning all three variables. I don't think that it is expected. Why filtering by `applications` gives only one output, but filtering by `bricks` returns outputs of all predecessors?

rizar:
Because when you call `merge.apply` it calls `super(Merge, self).apply`, that is two different application methods of the same brick `merge` were called. That's why what you get when you filter outputs of the `merge` brick are outputs of both application methods, two of `Parallel.apply` and one of `Merge.apply`.

dmitriy-serdyuk:
Ok, I see. But it doesn't seem intuitive.

rizar:
That is an inevitable consequence of having `Merge` as a descendant of `Parallel`. Whether or not this was a good design decision can be discussed.

rizar:
Thank for this small clean-up anyway :)

PR #624, opened 2015-05-08
Fix default value of requires in MonitoredQuantity

Treora:
MonitoredQuantityBuffer assumes requires to be a list (at
monitoring/evaluators.py:45), so the default should be a list, not None.

I have not tested the problem nor the solution; just fixed it while reading
the code.

### DIFF
--- a/blocks/monitoring/aggregation.py
+++ b/blocks/monitoring/aggregation.py
@@ -191,7 +191,7 @@ class MonitoredQuantity(object):
 
     """
     def __init__(self, requires=None, name=None):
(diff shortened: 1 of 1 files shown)


bartvm:
Thanks, that does indeed look better (unless we're overlooking something, @mohammadpz?).

For future PRs, considering using a full `if-else` block instead of the ternary operator. I usually find that a bit more readable.

mohammadpz:
As far as I see, @Treora  is right.
Thanks

PR #636, opened 2015-05-15
Temporary change of tutorial link

vdumoulin:


vdumoulin:
I'm not going to bother waiting for review, this is a one-liner in the docs that concerns an URL link.

PR #638, opened 2015-05-16
blocks-examples

bartvm:


rizar:
This is cool, but please hold on just for a moment until I am done with fixing the current state with 0.0.1 release. I will finish it tomorrow morning, if there no objections.

rizar:
I guess the tests for `Dump`, which depend on examples, could be temporarily dropped, as we plan to get rid of it anyway.

bartvm:
If we are okay with temporary disabling the test for `Dump` while finishing up https://github.com/mila-udem/blocks/pull/615, then I think this PR is good to go.

rizar:
I think we can live without them for a while. Go ahead and merge, and do not forget to write an announcement to `blocks-users`.

PR #639, opened 2015-05-16
Blocks 0.0.1

rizar:
See the subject.

@bartvm , do you know what else should be changed to switch to the two-branch system?

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -5,6 +5,7 @@ cache:
 branches:
   only:
     - master
+    - stable
 language: python
 python:
   - "2.7"
--- a/CONTRIBUTING.rst
+++ b/CONTRIBUTING.rst
@@ -1,10 +1,14 @@
-Questions about using Blocks?
-=============================
+Making a pull request
+=====================
 
-Please send your questions to the `Blocks users mailing list`_. You might not
-be the first one with this question or problem, so be sure to search both the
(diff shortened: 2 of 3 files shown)


rizar:
Cool, Travis and Coveralls seems to work. Bart, can you add the "stable" branch to Scrutinizer? Seems like it has to be done from some control panel there.

bartvm:
Enabled on Scrutinizer.

bartvm:
By the way, could you make this PR to the master branch? There's not reason to make it to `stable`.

rizar:
Scrutinizer was triggered.

Do you want me to commit these changes to master and then to create a new stable branch? Or to merge them in both?

bartvm:
Either is fine by me, as long as the commits end up in `master`, else the branches will have diverged before we even started :p

rizar:
By the way, should we move to mila-udem?

bartvm:
I guess this weekend is as good as any.

vdumoulin:
+1 for the move to @mila-udem, we should also do that for Fuel. We just need to make sure that the mailing lists are notified of the change.

rizar:
I guess I will wait until the migration is done, merge it into master and create a stable branch once again.

rizar:
Closed in favour of #641

PR #640, opened 2015-05-16
Move the mila-udem

bartvm:


rizar:
Cool! I guess Scrutinizer, and maybe some other services we use have to be notified of this migration.

PR #641, opened 2015-05-17
Release 0.0.1

rizar:
A PR to master instead of #639.

Includes some additional changes in documentation. 

A question: currently we request exact versions of `numpy`, `pandas` and other well-established packages, and bleeding-edge versions of Fuel, Theano and picklable-itertools. This makes sense for `master` but not so much for `stable`, because it will quickly end up non-installable. 

We could specify exactly (by tags, I guess, one can do that, or at least by exact hashes of revisions) which versions of these packages are required for the stable, but then we should be always careful to keep this difference between `stable` and `master`. It is concentrated in one file though, `requirements.txt`.

@bartvm , @dwf , @vdumoulin , what do you think we should do?


dwf:
I should probably just put pickable-itertools 0.1 on PyPI already,
shouldn't I.

rizar:
That would be great! Still a decision has to be made for Theano and Fuel.

How was it done in PyLearn2, btw?

bartvm:
Yeah, I guess `picklable-itertools` is stable enough to just put on PyPI. Otherwise, I think Fuel and Blocks stable releases should be synced. I can make a PR similar to this one for Fuel. For Theano I think we should just specify the commit hash.

Pylearn2 just had a single master branch without versioning.

docs/index.rst
```diff
@@ -11,7 +11,8 @@ Blocks is developed in parallel with Fuel_, a dataset processing framework.
 
 .. warning::
    Blocks is a new project which is still under development. As such, certain
-   (all) parts of the framework are subject to change.
+   (all) parts of the framework are subject to change. The last stable (and
+   thus likely an outdated version) can be found in the "stable" branch).
```
bartvm: There's a closing parenthesis too much here, and I think the first closing parenthesis should be after "outdated", not "version".

rizar:
Sounds like a good plan. A similar Fuel branch would be nice, assuming there are no urgent fixes pending as PRs there.

docs/index.rst
rizar: 615064f

PR #642, opened 2015-05-17
Fix requirements for stable

bartvm:
Following https://github.com/mila-udem/blocks/pull/641

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -11,8 +11,8 @@ git+https://github.com/dwf/picklable_itertools.git#egg=picklable-itertools
 
 --allow-external theano
 --allow-unverified theano
-git+https://github.com/Theano/Theano.git#egg=theano
+git+https://github.com/Theano/Theano.git@c6d2ab3929182cdf7cf12e4b1f191a6ff99d07cc#egg=theano
 
 --allow-external fuel
 --allow-unverified fuel
-git+https://github.com/mila-udem/fuel#egg=fuel
+git+https://github.com/mila-udem/fuel@stable#egg=fuel
--- a/requirements.txt
+++ b/requirements.txt
@@ -4,10 +4,7 @@ pandas==0.16.1
 progressbar2==2.7.3
 PyYaml==3.11
 toolz==0.7.2
-
---allow-external picklable-itertools
---allow-unverified picklable-itertools
-git+https://github.com/dwf/picklable_itertools.git#egg=picklable-itertools
+picklable-itertools==0.1.0
 
 --allow-external theano
 --allow-unverified theano


bartvm:
That just leaves `picklable-itertools`

requirements.txt
```diff
@@ -11,8 +11,8 @@ git+https://github.com/dwf/picklable_itertools.git#egg=picklable-itertools
 
 --allow-external theano
 --allow-unverified theano
-git+https://github.com/Theano/Theano.git#egg=theano
+git+https://github.com/Theano/Theano.git@c6d2ab3929182cdf7cf12e4b1f191a6ff99d07cc#egg=theano
 
 --allow-external fuel
 --allow-unverified fuel
-git+https://github.com/mila-udem/fuel#egg=fuel
+git+https://github.com/mila-udem/fuel@stable#egg=fuel
```
rizar: This is the right way to go, but if we want to keep outdated versions installable (and I think we want), we will have to create revisions pointing to a tag or exact commit in Fuel.  This is in a way the minimal form of maintenance of outdated versions. We can not afford more than that, I guess, with our limited manpower.

rizar:
Should we fix the commit hash for `picklable-itertools`?

requirements.txt
bartvm: I can't create a tag until someone merges https://github.com/mila-udem/fuel/pull/121

bartvm:
If @dwf wants to put it on PyPI, I suggest we wait for that.

requirements.txt
```diff
@@ -11,8 +11,8 @@ git+https://github.com/dwf/picklable_itertools.git#egg=picklable-itertools
 
 --allow-external theano
 --allow-unverified theano
-git+https://github.com/Theano/Theano.git#egg=theano
+git+https://github.com/Theano/Theano.git@c6d2ab3929182cdf7cf12e4b1f191a6ff99d07cc#egg=theano
```
rizar: I even know what the first "hotfix" in terminology from http://nvie.com/posts/a-successful-git-branching-model/ will be: changing Theano version to the one with https://github.com/Theano/Theano/issues/2913 fixed.

bartvm: Right, except in this case the hotfix will only be applied to the `stable` branch, which will probably just give rise to merge conflicts once we perform an actual hotfix :p This two-branch thing is going to be fun!

rizar:
Okay, LGTM. However, fixing this would eliminate our dependency on how soon he does it.

requirements.txt
rizar: Well, if we have such conflicts only in requirements.txt, I think we can manage that. But sure, fun is guaranteed :)

requirements.txt
rizar: I merged it, but still, I guess, stable branches should be synchronized, and `@stable` is the right way to make a reference. But if when we have `0.0.2` or `0.1` or later versions, I think it is important to have a revision, let's say `old-release-0.0.1` that points to the right `Fuel` version and is installable.

bartvm: You mean a separate branch? Sure.

bartvm: Also: https://github.com/mila-udem/fuel/releases/tag/v0.0.1

rizar: Yeap, I meant a branch, but a one-commit-branch, because I guess that further maintenance of outdated versions is beyond the amount of manpower we have. 

Cool, I did not know that github supports releases!

rizar: https://github.com/mila-udem/blocks/releases/tag/v0.0.1

bartvm:
Ehm, shouldn't this have been merged before creating https://github.com/mila-udem/blocks/releases/tag/v0.0.1? Installing 0.0.1 will now result in the wrong Fuel and Theano dependencies being installed.

rizar:
Oops, sure. I have rights to delete releases in Fuel but in Blocks, can you extend my rights to do that?

bartvm:
Your rights should be exactly the same for Fuel and Blocks... Not sure why you can't delete the release

rizar:
Okay. Any reason not to merge this PR?

bartvm:
Waiting for `picklable-itertools`... The weather's good and it's a Sunday, so let's give @dwf a few hours :p

dwf:
Oops, it's on PyPI now. (Actually I wasn't out but recovering from a
catastrophic accident involving me, a painting, a door frame, and a
screaming damsel in distress who in fact was just having a nightmare.
Anyway, you may see me limping tomorrow.)

vdumoulin:
Looking forward to hear about that one.

bartvm:
I think this one is ready to merge then, together with https://github.com/mila-udem/blocks/pull/646

rizar:
Definitely!

PR #647, opened 2015-05-18

bartvm:


dwf:
Whoa what, the bot makes the commit but you make the PR. What dark magic is this?

bartvm:
Haha, for some reason requires.io way of doing things is by making a commit but then using my account to actually open the pull request. I guess they do it like that so that it would even work for private repos.

PR #649, opened 2015-05-20
Assert if not BaseSequenceGenerator in beam search

orhanf:
The assertion for `SequenceGenerator` in beam search can be replaced with `BaseSequenceGenerator` without any harm. Will help if user does not extend `SequenceGenerator`  but inherit directly from `BaseSequenceGenerator`.


dwf:
LGTM, will merge when Travis passes

dwf:
Coveralls continues to act up...

PR #651, opened 2015-05-20
WIP: Separated beam search test from examples

dmitriy-serdyuk:
I'm going to simplify a bit this code.


dmitriy-serdyuk:
I think it's simple enough now.

rizar:
Sure, that was the right thing to do!

PR #652, opened 2015-05-21
Remove bias from the energy computer

rizar:
This bias is redundant, it is eliminated by softmax. @dmitriy-serdyuk , can you please review this PR?

### DIFF
--- a/tests/test_search.py
+++ b/tests/test_search.py
@@ -102,7 +102,7 @@ def test_beam_search():
     results, mask, costs = search.search(
         {inputs: input_vals}, 0, 3 * length, as_arrays=True)
     # Just check sum
-    assert results.sum() == 3160
+    assert results.sum() == 1116
 
     true_costs = simple_generator.cost(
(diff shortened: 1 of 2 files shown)


rizar:
@orhanf,  @sebastienf, can you please review this code?

tests/test_search.py
```diff
@@ -102,7 +102,7 @@ def test_beam_search():
     results, mask, costs = search.search(
         {inputs: input_vals}, 0, 3 * length, as_arrays=True)
     # Just check sum
-    assert results.sum() == 3160
+    assert results.sum() == 1116
```
dmitriy-serdyuk: Why the sum is different? It should be the same if the bias is eliminated by softmax.

rizar: That's because biases are initialized first in MLP, before weights. Which means that `use_bias=False` changes the initial random weights. This can not be called a bug, but I guess the other way round would make more sense.

dmitriy-serdyuk:
Oh, I see. In this case it looks good to me.

PR #656, opened 2015-05-21
Extend MisclassificationRate with top_k.

dwf:
It's common in computer vision applications to measure not just argmax accuracy but rather e.g. top 5 accuracy, i.e. you're allowed 5 guesses, and if any of your 5 highest outputs contain the ground truth class, you are considered to have gotten it right.

### DIFF
--- a/blocks/bricks/cost.py
+++ b/blocks/bricks/cost.py
@@ -6,8 +6,6 @@ from six import add_metaclass
 
 from blocks.bricks.base import application, Brick
 
-floatX = theano.config.floatX
-
 
 @add_metaclass(ABCMeta)
 class Cost(Brick):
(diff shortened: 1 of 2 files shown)


blocks/bricks/cost.py
```diff
@@ -63,10 +61,49 @@ def apply(self, y, y_hat):
...
+        super(MisclassificationRate, self).__init__()
+
     @application(outputs=["error_rate"])
     def apply(self, y, y_hat):
+        # Support checkpoints that predate self.top_k
+        top_k = getattr(self, 'top_k', 1)
+        if top_k == 1:
+            mistakes = tensor.neq(y, y_hat.argmax(axis=1))
+        else:
+            flattened_responses = y_hat.reshape((y_hat.shape[0] *
```
bartvm: Why not `y_hat.flatten()`? Or what am I missing?

blocks/bricks/cost.py
```diff
@@ -63,10 +61,49 @@ def apply(self, y, y_hat):
...
+            # Because we used greater-than-or-equal we have to correct for
+            # counting the true label.
+            num_higher = higher_scoring.sum(axis=1) - 1
+            mistakes = tensor.ge(num_higher, top_k)
         # Here we have to cast both operands to floatX explicitly.
         # Because int64 / float32 = float64 in Theano, unfortunately.
-        return (
-            tensor.sum(tensor.neq(y, y_hat.argmax(axis=1))).astype(floatX) /
-            y.shape[0].astype(floatX))
+        return (tensor.sum(mistakes).astype(theano.config.floatX) /
```
bartvm: So why not `mistakes.mean(dtype=theano.config.floatX)` in the end?

dwf: Yeah, actually, come to think of it, that is better.

blocks/bricks/cost.py
dwf: Heh, actually wasn't sure that existed in Theano. Pushed a new version with that change and the `.mean()` change.

PR #657, opened 2015-05-21
Remove dangerous local aliasing of floatX.

dwf:
Aliasing floatX at the module level has the side effect of incorrect behavior if floatX is ever changed at runtime as @lamblin once pointed out to me (one more instance removed in #656).

cc @rizar @dmitriy-serdyuk

### DIFF
--- a/tests/bricks/test_sequence_generators.py
+++ b/tests/bricks/test_sequence_generators.py
@@ -16,8 +16,6 @@ from blocks.graph import ComputationGraph
 from blocks.initialization import Orthogonal, IsotropicGaussian, Constant
 from blocks.roles import AUXILIARY
 
-floatX = theano.config.floatX
-
 
 class TestEmitter(TrivialEmitter):
     @application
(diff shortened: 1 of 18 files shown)


tests/bricks/test_sequence_generators.py
```diff
@@ -32,6 +30,7 @@ def test_sequence_generator():
     Such sequence generators can be used to model e.g. dynamical systems.
 
     """
+    floatX = theano.config.floatX
```
vdumoulin: Why are we aliasing here?

tests/bricks/test_sequence_generators.py
```diff
@@ -100,6 +99,7 @@ def test_integer_sequence_generator():
     Such sequence generators can be used to e.g. model language.
 
     """
+    floatX = theano.config.floatX
```
vdumoulin: Same question.

dwf: Basically because it was annoying and if you are modifying it during the
run of a single function you have bigger problems. I just wanted to get rid
of it at the module level: non persistent scopes are fine, selon moi.

vdumoulin: Nevermind, I didn't realize it's in a test.

vdumoulin:
LGTM.

PR #658, opened 2015-05-22
Refactor RemoveNotFinite

rizar:
`RemoveNotFinite` was quite confusing. It's parameter called `scaler` in fact meant that `1 - scaler` scaling will be applied. It was also never clearly said in the docs that it must be applied last. Finally, `0.1` is a kind of random default, a better default being not touching the parameters when nans are encountered.

All these things should be fixed in this PR. 

@ejls, @orhanf , you might be interested.

@memimo , you are the author.

### DIFF
--- a/blocks/algorithms/__init__.py
+++ b/blocks/algorithms/__init__.py
@@ -732,8 +732,8 @@ class RemoveNotFinite(StepRule):
         self.scaler = scaler
 
     def compute_step(self, param, previous_step):
-        not_finite = tensor.or_(
-            tensor.isnan(previous_step), tensor.isinf(previous_step)).sum()
+        not_finite = (tensor.isnan(previous_step).sum() +
+                      tensor.isinf(previous_step).sum())
         step = tensor.switch(
             not_finite > 0, (1 - self.scaler) * param, previous_step)
         return step, []
--- a/blocks/algorithms/__init__.py
+++ b/blocks/algorithms/__init__.py
@@ -710,29 +710,32 @@ class RemoveNotFinite(StepRule):
 
     Replaces a step (the parameter update of a single shared variable)
     which contains non-finite elements (such as ``inf`` or ``NaN``) with a
-    scaled version of the parameters being updated instead.
+    step rescaling the parameters.
 
     Parameters
     ----------
     scaler : float, optional
         The scaling applied to the parameter in case the step contains
-        non-finite elements. Defaults to 0.1.
+        non-finite elements. Defaults to 1, which means that parameters
+        will not be changed.
 
     Notes
     -----
+    This rule should be applied last!
+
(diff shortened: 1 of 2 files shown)


bartvm:
LGTM, but I'll let the stakeholders have a look too.

blocks/algorithms/__init__.py
```diff
@@ -710,29 +710,32 @@ class RemoveNotFinite(StepRule):
...
 
     .. _GroundHog: https://github.com/lisa-groundhog/GroundHog
 
     """
-    def __init__(self, scaler=0.1):
+    def __init__(self, scaler=1):
         self.scaler = scaler
 
     def compute_step(self, param, previous_step):
         not_finite = tensor.any(tensor.or_(
```
orhanf: We realized with @kyunghyuncho that `tensor.any` here is slowing down the overall iteration with a significant amount. For instance in a large computation graph running on gpu, by using `BLOCKS_PROFILE=True`  we were getting 
with `tensor.any` 13.5 secs per iteration
without `tensor.any` 1.9 secs per iteration

dwf: Should probably file a bug on Theano for that one.

bartvm: Here's probably why:
```python
t = CudaNdarrayType((False, False), dtype='float32')
x = t()
theano.printing.debugprint(tensor.any(x))
```
```
Any [@A] ''
 |Elemwise{neq,no_inplace} [@B] ''
   |HostFromGpu [@C] ''
   | |<CudaNdarrayType(float32, matrix)> [@D]
   |DimShuffle{x,x} [@E] ''
     |TensorConstant{0} [@F]
```
It seems to use `neq` internally, which doesn't seem to have a GPU implementation. So what it literally ends up doing is moving every single parameter step from GPU to CPU, perform an element-wise comparison to `0`, and then perform a bitwise `or` on the resulting binary (`int8`) matrix.

rizar: Holy shit!

I will try to simply replace it by sum, should do the job.

kyunghyuncho: @nouiz FYI

memimo:
LGTM, and nice catch with the tensor.any

rizar:
@orhanf , can you try  that the new version is faster? I replaced `any` with `sum`

bartvm:
@rizar I still see `any`, did you forget to push, or did you change it somewhere else than this PR?

rizar:
Sorry, I forgot to push.

blocks/algorithms/__init__.py
nouiz: I made an issue: https://github.com/Theano/Theano/issues/2951

PR #661, opened 2015-05-22
removed transposes to speed up LSTM

pbrakel:
Running the @rizar his benchmark using a GTX 580 I got this as a baseline result:

| | 100|250|1000|2000 |
|---|---|---|---|---|
| SimpleRecurrent|13|16|62|85 |
| GatedRecurrent|25|34|119|232 |
| LSTM|52|97|280|929 |

And this after removing the transpose:

| | 100|250|1000|2000 |
|---|---|---|---|---|
| SimpleRecurrent|12|16|62|85 |
| GatedRecurrent|25|34|119|231 |
| LSTM|44|85|248|611 |

So for larger networks the difference is substantial while for smaller networks the difference is not that big but still nice.

### DIFF
--- a/blocks/bricks/recurrent.py
+++ b/blocks/bricks/recurrent.py
@@ -430,8 +430,10 @@ class LSTM(BaseRecurrent, Initializable):
             Next cell activations of the network.
 
         """
+
         def slice_last(x, no):
-            return x.T[no*self.dim: (no+1)*self.dim].T
+            return x[:, no*self.dim: (no+1)*self.dim]
+
         nonlinearity = self.children[0].apply
 
         activation = tensor.dot(states, self.W_state) + inputs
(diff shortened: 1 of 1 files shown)


dwf:
nice

bartvm:
Awesome. This seems like something that a Theano optimization could/should do though, no? @lamblin @nouiz

rizar:
This PR has very convincing bytes_changed/speedup_gained ratio :)

pbrakel:
@rizar It was your suggestion to try it :)

lamblin:
Yes, that's definitely something that Theano could automatically do (and it may already), but I'm puzzled as why it makes such a difference. Is it the forward pass or the gradient that gets optimized?
If the indexing operation itself is slower, then we should bring it up to speed.
If the problem is a more complex graph being generated for the gradient, then _that_ should be the target of a Theano optimization, since the gradient is taken before the graph is simplified.

nouiz:
To know the problem, can you run both case with Theano profiler and give us the information? also enable profile_memory=True to get the strides that could affect the timming.

Issue #644, opened 2015-05-17
Decrease test runtime


vdumoulin:
The Blocks test suite isn't at the slowness level of Pylearn2 yet, but tests aren't as snappy as they could be. Advantages of decreasing test runtime include:

* Less waiting time during review process ("LGTM. Let's wait for the tests to pass and then merge.").
* Better incentive for people to run the full test suite more often during development.

One low-hanging fruit would be playing with Theano compilation modes. Most of the tests we're doing are very small scale and the compile time outweights the benefits of running the code in C. As a proof of concept, here are some timings on my machine for `tests/bricks` with either `mode=FAST_RUN` or `mode=FAST_COMPILE`:

``` bash
$ THEANO_FLAGS="mode=FAST_RUN" nosetests tests/bricks/
[...]
Ran 54 tests in 18.540s
```

``` bash
$ THEANO_FLAGS="mode=FAST_COMPILE" nosetests tests/bricks/
[...]
Ran 54 tests in 4.634s
```

rizar:
I think that running tests in `FAST_COMPILE` would be sufficient. At least for the master branch. If we wish, we can use `FAST_RUN` for the stable branch.

vdumoulin:
I tried that, but some tests with blocks/examples have full training loops are better off with FAST_RUN.

However, I think we can at least set this at a per-module basis.

PR #667, opened 2015-05-23
Reduce build matrix size

bartvm:
This does the same as we did for Pylearn2 at one point: We assume that Python 2 vs. 3 and float32 vs float64 are completely complementary, so instead of running each combination, we run Python 2 with float32 and Python 3 with float64. Should speed up Travis.

Contributes to #644.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -7,14 +7,16 @@ branches:
     - master
     - stable
 language: python
-python:
-  - "2.7"
-  - "3.4"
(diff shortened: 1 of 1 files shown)


rizar:
Sounds like the right way to go!

bartvm:
It also halves the Travis runtime (because it wasn't doing more than 4 of the 8 jobs at a time I think).

PR #670, opened 2015-05-24
AdaGrad

udibr:
Add AdaGrad learning rule (similar to Adam)

### DIFF
--- a/blocks/algorithms/__init__.py
+++ b/blocks/algorithms/__init__.py
@@ -665,9 +665,10 @@ class AdaGrad(StepRule):
     -----
     For more information, see [ADAGRAD]_.
 
-    .. [ADADGRAD] "Adaptive subgradient methods for online learning and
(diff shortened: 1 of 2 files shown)


dwf:
This looks about ready, but I'd be wary of accepting it without a small test. (It's only a -0.16% drop in coverage, but Rome wasn't burnt in a day.)

dwf:
I see you rebased. Would you mind doing a `rebase -i` and squashing out the PEP8, epsilon, and reference fix commits?

@rizar Is it permissible to be using a `StepRule` with parameters with no `.name`? Is it expected to work?

rizar:
It is not strictly required for an accumulator of a step rule to have a name. No code depends on that so far. But in future we definitely want to have the step computation graph annotated, not only with names I guess, but also with e.g. reference of the respective parameters.

For now the name mangling done in this PR is sufficient.

PR #671, opened 2015-05-24
Xavier weight initialization

udibr:
The method is called Xavier after the inventor first name and not his last name Glorot.
This is the standard name used in the industry, for example, see Caffee or http://deepdish.io/2015/02/24/network-initialization/

### DIFF
--- a/blocks/initialization.py
+++ b/blocks/initialization.py
@@ -224,6 +224,7 @@ class Sparse(NdarrayInitialization):
             weights[i, random_indices] = values[i]
         return weights
 
+
 class GlorotBengio(NdarrayInitialization):
     """Initialize parameters with Glorot-Bengio method.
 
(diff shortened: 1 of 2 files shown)


dwf:
I dont think you can claim "industry" refers to it as something based on
one nobody's blog post. I can also tell you that it's referred to as
"GLOROT_BENGIO" inside the codebase of the industry's largest consumer of
deep learning technology.

I also don't think we should be influenced by Caffe's bad decisions. If we
thought they did everything right, we'd be using Caffe.

dwf:
The initialization here doesn't actually correspond to the one in the paper at all...

dmitriy-serdyuk:
You are implemented an initialization which authors used in the beginning of the paper. Later they propose "Normalized initialization", as they call it. It is U[-a, a], where a = \sqrt{6} / \sqrt{n_j + n_{j + 1}} and n_j, n_{j + 1} are number of inputs/outputs of the layer.

bartvm:
Just a quick note here to connect this PR and this ticket: https://github.com/mila-udem/blocks/issues/576

rizar:
Closed due to inactivity and due to the fact that the PR in fact proposes another initialization scheme, not `GlorotBengio`.

PR #675, opened 2015-05-26
Add Concatenate brick

dwf:
Seems like a natural companion to Merge, fork, etc.

### DIFF
--- a/blocks/bricks/__init__.py
+++ b/blocks/bricks/__init__.py
@@ -570,6 +570,44 @@ class Sequence(Brick):
         return self.application_methods[-1].outputs
 
 
+class Concatenate(Brick):
+    """Merges several variables by concatenating them.
+
+    Parameters
+    ----------
+    axis : int
+        Axis along which to concatenate inputs.
+
+    Notes
+    -----
+    This is a very straightforward wrapping of a Theano operation without
+    parameters of its own. It's mainly useful in conjunction with the
+    :class:`~blocks.filter.VariableFilter`.
+
+    Examples
+    --------
+    >>> from theano import tensor
+    >>> a = tensor.matrix('a')
+    >>> b = tensor.matrix('b')
+    >>> concat_brick = Concatenate(axis=1)
+    >>> c = concat_brick.apply(a, b)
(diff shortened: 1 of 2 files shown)


blocks/bricks/__init__.py
```diff
@@ -570,6 +570,44 @@ def apply_outputs(self):
...
+    Parameters
+    ----------
+    axis : int
+        Axis along which to concatenate inputs.
+
+    Notes
+    -----
+    This is a very straightforward wrapping of a Theano operation without
+    parameters of its own. It's mainly useful in conjunction with the
+    :class:`~blocks.filter.VariableFilter`.
```
rizar: Do you mean that you want to have such a brick in order to have inputs and outputs tagged?

dwf: Yes, precisely. Or be able to select the inputs of all Concatenate bricks, etc.

rizar: To be honest I am afraid to create a precedent. Theano has lots of ops, and if we have a wrapper in Blocks for each and every one of them, this will hugely increase the codebase. 

What about 

```python 
class Expression(Brick): # choose a better name
   """Wraps a Theano expression in a brick."""
   def __init__(self, expression, keyword_arguments, **kwargs):
      super(Expression, self).__init__(**kwargs)
      self.expression = expression
      self.keyword_arguments = keyword_arguments

      return self.expression(*args, **dict_union(kwargs, keyword_arguments))

concatenate = Expression(tensor.concatenate, axis=1)
```

?

bartvm: These two should be changed as well then: https://github.com/mila-udem/blocks/blob/master/blocks/bricks/__init__.py#L479-L488

rizar: @dwf, do you agree that smth like `Expression` would be better?

dwf: Yeah, I started this in the midst of the NIPS fray and forgot about it.

rizar:
@dwf, what's the status of this one?

rizar:
Closed due to inactivity

PR #678, opened 2015-05-27
bug fix: has_bias should be has_biases

udibr:


rizar:
Thanks! Unfortunately, Travis says no. A bunch of test now crash because of unused `biases_init`.

udibr:
OK so ignore this pull requet and let the person who created this bug fix it...

PR #679, opened 2015-05-28
Update __init__.py

bartvm:
This brick now has its own documentation, so no longer needs to inherit from `Activation`.


dwf:
Hmm, I don't quite understand.  What documentation was it meant to be inheriting before (I see no class-level docstring, for example), and what changed?

bartvm:
Ah, never mind, I overlooked the fact that it's still inheriting the documentation for the `apply` function (it's inheriting it from the metaclass of `Activation`) https://github.com/mila-udem/blocks/blob/master/blocks/bricks/__init__.py#L432-L459

PR #680, opened 2015-05-29
Fail within if-block

bartvm:
The exit code of an `if-else` block in Bash is the one of the last command run, so right now the builds pass even when the doctests fail.


bartvm:
A 3-line diff that fixes a pretty big bug in the test suite, so would be great if someone could merge it quickly :)

PR #682, opened 2015-05-29
Add a basic early stopping extension.

### LINKED ISSUE #327: Early Stopping

dwf:
Fixes #327. Works in tandem with `TrackTheBest`. More nuanced implementations are possible (i.e. to specify a certain relative improvement is necessary) but those can be added as subclasses. CC @rizar.

A quick review would be appreciated.

### DIFF
--- a/blocks/extensions/stopping.py
+++ b/blocks/extensions/stopping.py
@@ -12,13 +12,22 @@ class FinishIfNoImprovementAfter(FinishAfter):
         The name of the log record to look for which indicates a new
         best performer has been found.  Note that the value of this
         record is not inspected.
+    patience_log_record : str, optional
+        The name under which to record the number of iterations we
+        are currently willing to wait for a new best performer.
+        Defaults to `notification_name + 'patience'`.
 
     """
-    def __init__(self, iterations, notification_name, **kwargs):
+    def __init__(self, iterations, notification_name, patience_log_record=None,
+                 **kwargs):
         self.iterations = iterations
         self.notification_name = notification_name
         kwargs.setdefault('after_batch', True)
         self.last_best = None
+        if patience_log_record is None:
+            self.patience_log_record = notification_name + '_patience'
+        else:
+            self.patience_log_record = patience_log_record
         super(FinishIfNoImprovementAfter, self).__init__(**kwargs)
 
     def update_best(self):
@@ -30,6 +39,8 @@ class FinishIfNoImprovementAfter(FinishAfter):
         self.update_best()
         iters_since = (self.main_loop.log.status['iterations_done'] -
                        self.last_best)
+        patience = self.iterations - iters_since
+        self.main_loop.log.current_row[self.patience_log_record] = patience
         if iters_since >= self.iterations:
             super(FinishIfNoImprovementAfter, self).do(which_callback,
                                                        *args)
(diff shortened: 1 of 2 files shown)


blocks/extensions/stopping.py
```diff
@@ -0,0 +1,35 @@
+from . import FinishAfter
+
+
+class FinishIfNoImprovementAfter(FinishAfter):
+    """Stop after improvements have ceased for a given period.
+
+    Parameters
+    ----------
+    iterations : int
```
bartvm: Maybe not necessary for this PR, but it'd be nice if we can support epochs as well, considering that the number of iterations per epoch can be variable.

dmitriy-serdyuk:
It would be nice to wirte patience into the log.

blocks/extensions/stopping.py
dwf: Agreed, though the question is whether you want to be monitoring for the best at the batch level. In order to then do early stopping at the epoch level, you need to examine every log record for that epoch...

dwf:
@dmitriy-serdyuk That's true.

By the way, happy birthday! Sorry I couldn't make it last night.

dwf:
@dmitriy-serdyuk I think I've added a good solution for that, PTAL.

blocks/extensions/stopping.py
dwf: In fact, since monitoring is only done every epoch by default (#684), perhaps this extension (and `TrackTheBest`) ought to default to `after_epoch`, in which case waiting for either a number of iterations or a number of epochs makes perfect sense.

dwf:
@bartvm What do you think of the proposed epoch support in 53bf349? It was either two arguments and an exclusivity check or one plus a flag, I think the former is more readable.

dwf:
Makes client code more readable, that is.

bartvm:
I agree, it's easier to read `iterations=1000` than something like `1000, type='iterations'`.

dwf:
Great. I don't think there's anything left to do in this PR except maybe fix #684 (and that's not really critical).

dmitriy-serdyuk:
@dwf, thanks. 

The PR looks good to me.

dwf:
Fixed the tests so they exercised the `patience_log_record` argument in non-default, so it's 100% coverage for that file now. Test still runs in 1ms.

dwf:
Isn't that special: generator method tests apparently do not run `setUp()` on Python 2 for whatever reason.

dwf:
Actually, it is a nose2, not Python 2, issue...

dwf:
nose-devs/nose2#80. Two year old issue. Welp. I don't have time to go spelunking in nose2 so I guess I'll just work around it.

dwf:
Anyone got further comments on this one?

PR #685, opened 2015-05-31
Custom pickler #2

bartvm:
Some fixes to #615, just want to see what Travis says


bartvm:
@rizar @dmitriy-serdyuk I fixed a few things (Python 3 compatibility, mostly) in #615 and added tests. I think it's done and ready to merge (this PR has been taking a looong time already, first work started 2 months ago).

The one decision that I'm not sure about is whether we should use `cPickle` by default, or the Python pickle so that we can raise warnings about people trying to serialize objects from the `__main__` namespace? `cPickle` gives quite a speedup, but on the other hand, I've seen people fall into the `__main__` trap several times, and what good is an informative warning if it's turned off by default.

bartvm:
Hurrah, Coveralls is finally back, 1.52% increase isn't bad. @dmitriy-serdyuk If you've reviewed it, we can finally merge this thing

rizar:
@bartvm , maybe we should wait for these 4 days before NIPS? We did make a branch for that, I know, but just in case somebody is still using master.. I really saw `Dump` in some people code.

Also, I really want to give it the last look, but NIPS rush makes it hard to find time for it.

blocks/extensions/saveload.py
```diff
@@ -85,83 +84,78 @@ def do(self, callback_name, *args):
...
+        _, from_user = self.parse_args(callback_name, args)
         try:
             path = self.path
             if from_user:
                 path, = from_user
-            already_saved_to = self.main_loop.log.current_row.get(SAVED_TO, ())
-            self.main_loop.log.current_row[SAVED_TO] = (
-                already_saved_to + (path,))
-            secure_pickle_dump(self.main_loop, path)
+            secure_dump(self.main_loop, path)
```
dmitriy-serdyuk: Why don't you use `cPickle.dump` like in line 96?

bartvm:
Well, the NIPS argument can be made two ways, because I actually very much need this for NIPS (right now the `Dump` function overwrites half of my parameters because they do not belong to a brick and have the same name). Either way, people who are doing a NIPS run can just revert to the `stable` branch or the current commit.

After 2 months and having moved the CCW deadline for it 3 times, I'd prefer merging this so that I can finally cross of it of my list. The diffs will still be there for you and everyone else to inspect next week, and we can tinker and tune then. For now, I think that considering you've given it a few looks, and both @dmitriy-serdyuk and myself have reviewed each other's code, it's been reviewed and tested enough to be merged into what is, after all, a development branch.

blocks/serialization.py
```diff
@@ -1,68 +1,166 @@
...
 
+    Parameters
+    ----------
+    obj : object
+        The object to pickle.
+    file_handler : file
+        The file handle to save the object to.
+    protocol : int, optional
+        The pickling protocol to use. Unlike Python's built-in pickle, the
+        default is set to `2` insstead of 0 for Python 2. The Python 3
```
dmitriy-serdyuk: > insstead

blocks/extensions/saveload.py
bartvm: Because for dumping the main loop, I want to use the custom `dump` function we wrote (which writes everythong to a ZIP file). However, for the attributes (normally the log and the iteration state) we can just use `cPickle.dump`, which is faster, and easier for a user to unpack (because anything serialized with our `dump` function, can only be loaded with the custom `load` function).

blocks/serialization.py
bartvm: 8666e0ec51491c6aef9881ae85e4c10c65d3a59a

blocks/extensions/saveload.py
dmitriy-serdyuk: Oh, I see now. What I meant is that we need an option to use `cPickle` here.

rizar:
If you actually need this for NIPS, you can as well use your branch. A bunch of PRs I made recently were all required for my NIPS rush, but I have my personal, regularly rebased branch for that, and don't think why you should suffer less.

I did give it a few looks, and in some cases was critically unsatisfied with what I saw. So I would not rush with this PR, because for neither of us this week is a good time to fix possible bad consequences and answer lots of emails.

To sum up, if the only argument to hurry with it is that you need for NIPS, then I think it will be a bad precedent to merge it.

blocks/serialization.py
```diff
@@ -71,17 +169,73 @@ def secure_pickle_dump(object_, path):
...
+
+    Returns
+    -------
+    A dictionary of (parameter name, numpy array) pairs.
+
+    """
+    source = numpy.load(path)
+    param_values = {'/' + name.replace(BRICK_DELIMITER, '/'): value
+                    for name, value in source.items() if name != 'pkl'}
+    source.close()
```
dmitriy-serdyuk: From my point of view, `with` statement is cleaner and more robust for typos in future changes.

blocks/extensions/saveload.py
bartvm: Right, I actually started on that but wanted to keep it for later, I guess I might as well finish it. I'll commit it in a few minutes.

bartvm: 3021bd29b97d8cded8f6998a7ef8edf71f261443

bartvm:
My main reason for wanting to merge this is not NIPS; I mentioned NIPS mostly to highlight that the entire point of creating a `stable` branch for NIPS was that we could *continue* development regardless of NIPS. My main motivation is simply that is has been in the works for 2 months now, having required several rebases along the way. A slow review and merging process was one of the things that very negatively impacted Pylearn2 development, and the main reason for me to suggest a two-branch development model. The main argument being that it would allow for faster development.

dwf:
Can someone give me the high level story of what this PR does?

blocks/serialization.py
bartvm: 5fb6753

rizar:
Sure, I agree that slow review and merging process is very hindering. But in case of this PR 2 months were required not because of too strict development policies and too lazy reviewers. It's simply a complex change in the very heart of Blocks, which required significant Python knowledge to code and review, and also causes non-trivial speed issues. Moreover, it required a change in Theano. Such things, unfortunately,  are doomed to take a lot of time.

Even with two branch system I guess we should careful before major deadline in which the main developer are involved. By the way, it was your suggestion in one of discussions here, and now I think it was a good one.

bartvm:
@dwf It combines the previous `Checkpoint` extension (which pickled the main loop) and the `Dump` extension (which collected the values of the parameters and saved them to an NPZ file). It does so by making use of pickle's "persistent ID" support (which we already merged into Theano a while ago, https://github.com/Theano/Theano/pull/2802).

In practice, it means that during pickling, each shared variable gets intercepted by the `PersistentSharedVariableID` class and is assigned a name (if its owned by a brick, the path to the brick is used, otherwise the name of the shared variable is used). The value of the shared variable is saved as an NPY file to a ZIP file, together with the pickled main loop (which contains string references to the names of the parameters). The final result is a single ZIP file with a pickled main loop and a series of NumPy arrays containing the values of the all the shared variables contained within the main loop.

Advantages over the old method:

* No need to use both `Checkpoint` and `Dump`. This single method allows you to resume training seamlessly, while also allowing you to inspect and share parameter values across platforms; no unpickling needed.
* Parameters are no longer silently overwritten if they had the same name and/or don't belong to a brick, as `Dump` does.
* The entire serialization situation gets (a) less confusing and (b) documented (because the current docs are out of date, and still talk about using Dill).

bartvm:
Rebased following #690

dwf:
Thanks, that sounds very useful. I've added it to the PR description on
#615 for GitHub archaeology purposes.

bartvm:
Rebased again.

bartvm:
Now it's failing because https://github.com/mila-udem/blocks-examples/pull/4 needs to be rebased... *sigh* @dmitriy-serdyuk Could you do that?

bartvm:
@rizar Do you have time to have a look this weekend so that we can finally merge this?

blocks/extensions/saveload.py
```diff
@@ -85,83 +88,78 @@ def do(self, callback_name, *args):
...
-                                   filenames[attribute])
+                secure_dump(getattr(self.main_loop, attribute),
+                            filenames[attribute], cPickle.dump)
         except Exception:
-            self.main_loop.log.current_row[SAVED_TO] = None
+            path = None
             raise
+        finally:
+            already_saved_to = self.main_loop.log.current_row.get(SAVED_TO, ())
+            self.main_loop.log.current_row[SAVED_TO] = (already_saved_to +
```
rizar: If saving failed, `SAVED_TO` record will be `(None)` instead of `None` as it used to be.

bartvm: Yes, that's on purpose. If multiple saves are performed in one iteration, it's very feasible that one failed, but not the other, so why should it be `None` instead of `(None, '/succesful/model')`?

blocks/extensions/saveload.py
```diff
@@ -85,83 +88,78 @@ def do(self, callback_name, *args):
...
+        training history. Defaults to `False`.
 
     Notes
     -----
-    Requires the model to be a Brick or a list of Bricks.
+    Requires the model to be created entirely using bricks, with a unique
+    path/name for each brick, so that the parameters can be matched to
+    their values.
+
+    In order to load the iteration state and the log, the saved model needs
```
rizar: For future work I would suggest to pickle log separately. We should be able to do it even without using  persistent id. Simply deleting the log attribute right before dumping and adding it back when loading should do the job.

But so far I am fine with merging the current variant.

blocks/extensions/saveload.py
rizar: True, that's reasonable.

dmitriy-serdyuk: I think, this is better in case if there are two saving extensions (e.g. an early stopping and a checkpoint after every epoch) and one of them crashes.

docs/serialization.rst
```diff
@@ -19,13 +20,13 @@ Pickling the training loop
...
-   of your pickled main loop before resuming training.
+   loop, (un)pickling will sometimes fail because it exceeds the default maximum
+   recursion depth set in Python. Increasing the limit should fix the problem.
 
-The first approach used is to pickle the entire :class:`main loop
-<.MainLoop>`, effectively serializing the exact state of the model as well
-as training. Techncially there are some difficulties with this approach:
+When checkpointing, Blocks pickles the entire :class:`main loop <.MainLoop>`,
+effectively serializing the exact state of the model as well as the training
+state (iteration state, extensions, etc.). Techncially there are some
```
rizar: _Technically_

rizar:
Go ahead and merge, great PR! The old `Dump` was quite quirky..

I would go even further in future, pickling as many as many main loop components as possible separately and adding them to the zip. Because from my experience the computation graph becomes unpicklable much faster than other components, and thus it should be as isolated as possible.

bartvm:
Great, fingers crossed then!

I agree about pickling components separately, it would make it even more robust.

dmitriy-serdyuk:
Wait! You didn't change blocks-example tests to use `mater`.

bartvm:
d10390651a40e8fd7fb0b55b36811daac6b61e0b Realised it the moment I pressed merge :)

PR #687, opened 2015-05-31
Make TrackTheBest default to after_epoch.

### LINKED ISSUE #684: TrackTheBest should default to after_epoch?
Monitoring is only done after every epoch (by default), so there's no sense in invoking TrackTheBest it after every batch (by default). Is that logic right @rizar?

dwf:
Fixes #684.

@rizar

### DIFF
--- a/blocks/extensions/training.py
+++ b/blocks/extensions/training.py
@@ -85,7 +85,7 @@ class TrackTheBest(SimpleExtension):
         self.notification_name = notification_name
         self.best_name = "best_" + record_name
         self.choose_best = choose_best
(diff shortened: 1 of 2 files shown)


rizar:
Looks sooooo good to me!

dwf:
Think I will need to fix the test :)

dwf:
Alright, @rizar, PTAL. Travis finished but Coveralls is missing in action.

rizar:
Coverage is very confused, but we know that this PR shouldn't make it less, don't we?

Issue #586, opened 2015-04-12
Bidirectional RNN : shared args and kwargs


sebastien-j:
Currently, the `Bidirectional` class uses the same positional and keywords arguments for its two children. As such, non-recurrent weights and biases will generally be shared between the two networks, which may be undesirable behaviour.

rizar:
One option would be to add another application method that would for each input `input` and context `context` expect two inputs `input_forward` and `input_backward` and two contexts `context_backward` and `context_forward` respectively. Would that do the job, @sebastien-j ?

Issue #46, opened 2014-11-28
Deep RNNs


rizar:
Stacked RNNs are so trendy nowadays that I believe somebody will need them soon. I think the right way to support it in Blocks is a `RecurrentStack` brick that takes a list of transitions and wraps them into one.

PR #688, opened 2015-05-31
stack of LSTM

udibr:
LSTMstack acts as a plug-and-play replacement for a regular LSTM

### DIFF
--- a/blocks/bricks/recurrent.py
+++ b/blocks/bricks/recurrent.py
@@ -9,12 +9,13 @@ import numpy
 import theano
 from theano import tensor, Variable
 
-from blocks.bricks import Initializable, Logistic, Tanh
+from blocks.bricks import Initializable, Logistic, Tanh, Linear
 from blocks.bricks.base import Application, application, Brick, lazy
 from blocks.initialization import NdarrayInitialization
 from blocks.roles import add_role, WEIGHT, INITIAL_STATE
 from blocks.utils import (pack, shared_floatx_nans, shared_floatx_zeros,
                           dict_union, dict_subset, is_shared_variable)
+from blocks.bricks.parallel import Fork
 
 logger = logging.getLogger()
 
@@ -633,3 +634,359 @@ class Bidirectional(Initializable):
     @apply.delegate
     def apply_delegate(self):
         return self.children[0].apply
+
+
+class RecurrentStack(BaseRecurrent, Initializable):
+    u"""Stack of recurrent networks.
+
+    Builds a stack of recurrent layers from a supplied list of
+    :class:`~blocks.bricks.recurrent.BaseRecurrent` objects.
+    Each object must have a `sequences`,
+    `contexts`, `states` and `outputs` parameters to its `apply` method,
+    such as the ones required by the recurrent decorator from
+    :mod:`blocks.bricks.recurrent`.
+
+    In Blocks in general each brick can have an apply method and this
+    method has attributes that list the names of the arguments that can be
+    passed to the method and the name of the outputs returned by the
+    method.
+    The attributes of the apply method of this class is made from
+    concatenating the attributes of the apply methods of each of the
+    transitions from which the stack is made.
+    In order to avoid conflict, the names of the arguments appearing in
+    the `states` and `outputs` attributes of the apply method of each
+    layers are renamed. The names of the bottom layer are used as-is and
+    a suffix of the form '_<n>' is added to the names from other layers,
+    where '<n>' is the number of the layer starting from 1
+    (for first layer above bottom.)
+
+    The `contexts` of all layers are merged into a single list of unique
+    names, and no suffix is added. Different layers with the same context
+    name will receive the same value.
+
+    The names that appear in `sequences` are treated in the same way as
+    the names of `states` and `outputs` if `skip_connections` is "True".
+    The only exception is the "mask" element that may appear in the
+    `sequences` attribute of all layers, no suffix is added to it and
+    all layers will receive the same mask value.
+    If you set `skip_connections` to False then only the arguments of the
+    `sequences` from the bottom layer will appear in the `sequences`
+    attribute of the apply method of this class.
+    When using this class, with `skip_connections` set to "True", you can
+    supply all inputs to all layers using a single fork which is created
+    with `output_names` set to the `apply.sequences` attribute of this
+    class. For example, :class:`~blocks.brick.SequenceGenerator` will
+    create a such a fork.
+
+    Whether or not `skip_connections` is set, each layer above the bottom
+    also receives an input (values to its `sequences` arguments) from a
+    fork of the state of the layer below it. Not to be confused with the
+    external fork discussed in the previous paragraph.
+    It is assumed that all `states` attributes have a "states" argument
+    name (this can be configured with `states_name` parameter.)
+    The output argument with this name is forked and then added to all the
+    elements appearing in the `sequences` of the next layer (except for
+    "mask".)
+    If `skip_connections` is False then this fork has a bias by default.
+    This allows direct usage of this class with input supplied only to the
+    first layer. But if you do supply inputs to all layers (by setting
+    `skip_connections` to "True") then by default there is no bias and the
+    external fork you use to supply the inputs should have its own separate
+    bias.
+
+    Parameters
+    ----------
+    transitions : list
+        List of recurrent units to use in each layer. Each derived from
+        :class:`~blocks.bricks.recurrent.BaseRecurrent`
+        Note: A suffix with layer number is added to transitions' names.
+    fork_prototype : :class:`~blocks.bricks.FeedForward`, optional
+        A prototype for the  transformation applied to states_name from
+        the states of each layer. The transformation is used when the
+        `states_name` argument from the `outputs` of one layer
+        is used as input to the `sequences` of the next layer. By default
+        it :class:`~blocks.bricks.Linear` transformation is used, with
+        bias if skip_connections is "False". If you supply your own
+        prototype you have to enable/disable bias depending on the
+        value of `skip_connections`.
+    states_name : string
+        In a stack of RNN the state of each layer is used as input to the
+        next. The `states_name` identify the argument of the `states`
+        and `outputs` attributes of
+        each layer that should be used for this task. By default the
+        argument is called "states". To be more precise, this is the name
+        of the argument in the `outputs` attribute of the apply method of
+        each transition (layer.) It is used, via fork, as the `sequences`
+        (input) of the next layer. The same element should also appear
+        in the `states` attribute of the apply method.
+    skip_connections : bool
+        By default False. When true, the `sequences` of all layers are
+        add to the `sequences` of the apply of this class. When false
+        only the `sequences` of the bottom layer appear in the `sequences`
+        of the apply of this class. In this case the default fork
+        used internally between layers has a bias (see fork_prototype.)
+
+        An external code can inspect the `sequences` attribute of the
+        apply method of this class to decide which arguments it need
+        (and in what order.) With `skip_connections` you can control
+        what is exposed to the externl code. If it is false then the
+        external code is expected to supply inputs only to the bottom
+        layer and if it is true then the external code is expected to
+        supply inputs to all layers. There is just one small problem,
+        the external inputs to the layers above the bottom layer are
+        added to a fork of the state of the layer below it. As a result
+        the output of two forks is added together and it will be
+        problematic if both will have a bias. It is assumed
+        that the external fork has a bias and therefore by default
+        the internal fork will not have a bias if `skip_connections`
+        is true.
+
+    Notes
+    -----
+    See :class:`.BaseRecurrent` for more initialization parameters.
+
+    """
+    @staticmethod
+    def suffix(name, level):
+        if name == "mask":
+            return "mask"
+        if level == 0:
+            return name
+        return name + '_' + str(level)
+
+    @staticmethod
+    def suffixes(names, level):
+        return [RecurrentStack.suffix(name, level)
+                for name in names if name != "mask"]
+
+    @staticmethod
+    def split_suffix(name):
+        # Target name with suffix to the correct layer
+        name_level = name.split('_')
+        if len(name_level) == 2:
+            name, level = name_level
+            level = int(level)
+        else:
+            # It must be from bottom layer
+            level = 0
+        return name, level
+
+    def __init__(self, transitions, fork_prototype=None, states_name="states",
+                 skip_connections=False, **kwargs):
+        super(RecurrentStack, self).__init__(**kwargs)
+
+        self.states_name = states_name
+        self.skip_connections = skip_connections
+
+        for level, transition in enumerate(transitions):
+            transition.name += '_' + str(level)
+        self.transitions = transitions
+
+        if fork_prototype is None:
+            # If we are not supplied any inputs for the layers above
+            # bottom then use bias
+            fork_prototype = Linear(use_bias=not skip_connections)
+        depth = len(transitions)
+        self.forks = [Fork(self.normal_inputs(level),
+                           name='fork_' + str(level),
+                           prototype=fork_prototype)
+                      for level in range(1, depth)]
+
+        self.children = self.transitions + self.forks
+
+        # Programmatically set the apply parameters.
+        # parameters of base level are exposed as is
+        # excpet for mask which we will put at the very end. See below.
+        for property_ in ["sequences", "states", "outputs"]:
+            setattr(self.apply,
+                    property_,
+                    self.suffixes(getattr(transitions[0].apply, property_), 0)
+                    )
+
+        # add parameters of other layers
+        if skip_connections:
+            exposed_arguments = ["sequences", "states", "outputs"]
+        else:
+            exposed_arguments = ["states", "outputs"]
+        for level in range(1, depth):
+            for property_ in exposed_arguments:
+                setattr(self.apply,
+                        property_,
+                        getattr(self.apply, property_) +
+                        self.suffixes(getattr(transitions[level].apply,
+                                              property_),
+                                      level)
+                        )
+
+        # place mask at end because it has a default value (None)
+        # and therefor should come after arguments that may come us
+        # unnamed arguments
+        if "mask" in transitions[0].apply.sequences:
+            self.apply.sequences.append("mask")
+
+        # add context
+        self.apply.contexts = list(set(
+            sum([transition.apply.contexts for transition in transitions], [])
+        ))
+
+        # sum up all the arguments we expect to see in a call to a transition
+        # apply method, anything else is a recursion control
+        self.transition_args = set(self.apply.sequences +
+                                   self.apply.states +
+                                   self.apply.contexts)
+
+        for property_ in ["sequences", "states", "contexts", "outputs"]:
+            setattr(self.low_memory_apply, property_,
+                    getattr(self.apply, property_))
+
+    def normal_inputs(self, level):
+        return [name for name in self.transitions[level].apply.sequences
+                if name != 'mask']
+
+    def _push_allocation_config(self):
+        # Configure the forks that connect the "states" element in the `states`
+        # of one layer to the elements in the `sequences` of the next layer,
+        # excluding "mask".
+        # This involves `get_dim` requests
+        # to the transitions. To make sure that it answers
+        # correctly we should finish its configuration first.
+        for transition in self.transitions:
+            transition.push_allocation_config()
+
+        for level, fork in enumerate(self.forks):
+            fork.input_dim = self.transitions[level].get_dim(self.states_name)
+            fork.output_dims = self.transitions[level + 1].get_dims(
+                fork.output_names)
+
+    def do_apply(self, *args, **kwargs):
+        """Apply the stack of transitions.
+
+        This is the undecorated implementation of the apply method.
+        A method with an @apply decoration should call this method with
+        `iterate=True` to indicate that the iteration over all steps
+        should be done internally by this method. A method with a
+        `@recurrent` method should have `iterate=False` (or unset) to
+        indicate that the iteration over all steps is done externally.
+
+        Parameters
+        ----------
+        See docstring of the class for arguments appearing in
+        self.apply.sequences, self.apply.states, self.apply.contexts
+        All arguments values are of type :class:`~tensor.TensorVariable`.
+
+        In addition the `iterate`, `reverse`, `return_initial_states` or
+        any other argument defined in `recurrent_apply` wrapper.
+
+        Returns
+        -------
+        The outputs of all transitions as defined in `self.apply.outputs`
+        All return values are of type :class:`~tensor.TensorVariable`.
+
+        """
+        nargs = len(args)
+        assert nargs <= len(self.apply.sequences)
+        kwargs.update(zip(self.apply.sequences[:nargs], args))
+
+        if kwargs.get("reverse", False):
+            raise NotImplementedError
+
+        results = []
+        last_states = None
+        for level, transition in enumerate(self.transitions):
+            normal_inputs = self.normal_inputs(level)
+            layer_kwargs = dict()
+
+            if level == 0 or self.skip_connections:
+                for name in normal_inputs:
+                    layer_kwargs[name] = kwargs.get(self.suffix(name, level))
+            if "mask" in transition.apply.sequences:
+                layer_kwargs["mask"] = kwargs.get("mask")
+
+            for name in transition.apply.states:
+                layer_kwargs[name] = kwargs.get(self.suffix(name, level))
+
+            for name in transition.apply.contexts:
+                layer_kwargs[name] = kwargs.get(name)  # contexts has no suffix
+
+            if level > 0:
+                # add the forked states of the layer below
+                inputs = self.forks[level - 1].apply(last_states, as_list=True)
+                for name, input_ in zip(normal_inputs, inputs):
+                    if layer_kwargs.get(name):
+                        layer_kwargs[name] += input_
+                    else:
+                        layer_kwargs[name] = input_
+
+            # Handle all other arguments
+            # For example, if the method is called directly
+            # (`low_memory=False`)
+            # then the arguments that recurrent
+            # expects to see such as: 'iterate', 'reverse',
+            # 'return_initial_states' may appear.
+            for k in set(kwargs.keys()) - self.transition_args:
+                layer_kwargs[k] = kwargs[k]
+
+            result = transition.apply(as_list=True, **layer_kwargs)
+            results.extend(result)
+
+            state_index = transition.apply.outputs.index(self.states_name)
+            last_states = result[state_index]
+            if kwargs.get('return_initial_states', False):
+                # Note that the following line reset the tag
+                last_states = last_states[1:]
+
+        return tuple(results)
+
+    @recurrent
+    def low_memory_apply(self, *args, **kwargs):
+        # we let the recurrent decorator handle the iteration for us
+        # so do_apply needs to do a single step.
+        kwargs['iterate'] = False
+        return self.do_apply(*args, **kwargs)
+
+    @application
+    def apply(self, *args, **kwargs):
+        """Apply the stack of transitions.
+
+        Parameters
+        ----------
+        low_memory : bool
+            Use the slow, but also memory efficient, implementation of
+            this code.
+
+        See docstring of the class for arguments appearing in
+        self.apply.sequences, self.apply.states, self.apply.contexts
+        All arguments values are of type :class:`~tensor.TensorVariable`.
+
+        In addition the `iterate`, `reverse`, `return_initial_states` or
+        any other argument defined in `recurrent_apply` wrapper.
+
+        Returns
+        -------
+        The outputs of all transitions as defined in `self.apply.outputs`
+        All return values are of type :class:`~tensor.TensorVariable`.
+
+        """
+        if kwargs.pop('low_memory', False):
+            return self.low_memory_apply(*args, **kwargs)
+        # we let the transition in self.transitions each do their iterations
+        # separatly, one layer at a time.
+        return self.do_apply(*args, **kwargs)
+
+    def get_dim(self, name):
+        # Check if we have a contexts element.
+        for transition in self.transitions:
+            if name in transition.apply.contexts:
+                # hopefully there is no conflict between layers about dim
+                return transition.get_dim(name)
+
+        name, level = self.split_suffix(name)
+        transition = self.transitions[level]
+        return transition.get_dim(name)
+
+    @application
+    def initial_state(self, state_name, batch_size, *args, **kwargs):
+        state_name, level = self.split_suffix(state_name)
+        transition = self.transitions[level]
+        return transition.initial_state(state_name, batch_size,
+                                        *args, **kwargs)
--- a/tests/bricks/test_recurrent.py
+++ b/tests/bricks/test_recurrent.py
@@ -1,6 +1,6 @@
 import itertools
 import unittest
-
+from collections import OrderedDict
 import numpy
 import theano
 from numpy.testing import assert_allclose
@@ -12,7 +12,7 @@ from blocks.bricks.base import application
 from blocks.bricks import Tanh
 from blocks.bricks.recurrent import (
     recurrent, BaseRecurrent, GatedRecurrent,
-    SimpleRecurrent, Bidirectional, LSTM)
+    SimpleRecurrent, Bidirectional, LSTM, RecurrentStack)
 from blocks.initialization import Constant, IsotropicGaussian, Orthogonal
 from blocks.filter import get_application_call, VariableFilter
 from blocks.graph import ComputationGraph
@@ -126,7 +126,7 @@ class TestLSTM(unittest.TestCase):
 
     def test_one_step(self):
         h0 = tensor.matrix('h0')
-        c0 = tensor.matrix('h0')
+        c0 = tensor.matrix('c0')
         x = tensor.matrix('x')
         h1, c1 = self.lstm.apply(x, h0, c0, iterate=False)
         next_h = theano.function(inputs=[x, h0, c0], outputs=[h1])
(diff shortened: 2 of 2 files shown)


rizar:
Thanks for sharing your code with us!

Unfortunately, this brick does not meet Blocks standard of modularity. We don't want to have `SimpleRecurrentStack`, `LSTMStack` and `GatedRecurrectStack` classes that share a lot of code. It should be possible to write a generic `RecurrentStack` brick, that accepts a prototype recurrent brick and build a stack of such bricks interleaved with `Fork`'s. There is even a ticket for that, #46 . Let me know if you are interested in implementing something like that, I would be very glad to help.

udibr:
I did implemented something like that at some point, but, every recurrent unit (LSTM, GRU, Linear) expects to see a different set of inputs (LSTM need both states and cells others just states)

As a result you cant put a generic stack in place of where you previously had an LSTM unit
(this is why I said "plug-and-play")

rizar:
I think in Blocks it is possible to make a generic stack. Given a recurrent brick, you can query its apply method what states it has and what sequential inputs it expects by simply `brick.apply.sequences` and `brick.apply.states`. By accessing those you can even write a stack that supports different components, e.g. one layer GRU and one layer LSTM.

udibr:
I have code for it and will try to check if it is working

udibr:
The code now is a general purpose stack of recurrent networks

blocks/bricks/recurrent.py
```diff
@@ -630,3 +631,140 @@ def apply(self, *args, **kwargs):
...
+            # I write it down explicitly.
+            prototype = Linear(use_bias=True)
+        depth = len(transitions)
+        self.forks = [Fork(self.normal_inputs(d), prototype=prototype)
+                      for d in range(1, depth)]
+
+        self.children = self.transitions + self.forks
+
+        # Programmatically fix the parameters of the  recurrent decoration
+        setattr(self.apply, 'sequences', transitions[0].apply.sequences)
```
rizar: Cool, this is a new technique which I never though of.. and I like it!

The standard way to have application method properties defined at runtime was so far

```
@apply.property('inputs')
def apply_inputs(self):
    return self.wrapped.inputs
```

but seems like simply setting attributes of the `BoundApplication` also does the job!

blocks/bricks/recurrent.py
```diff
@@ -630,3 +631,140 @@ def apply(self, *args, **kwargs):
...
+        # of one layer to the elements in the `sequences` of the next layer,
+        # excluding "mask".
+        # This involves `get_dim` requests
+        # to the transitions. To make sure that it answers
+        # correctly we should finish its configuration first.
+        for transition in self.transitions:
+            transition.push_allocation_config()
+
+        for d, fork in enumerate(self.forks):
+            fork.input_dim = self.transitions[d].get_dim("states")
```
rizar: Hmm, I thought that all the states should be forked, but apparently it does not match what they do in the literature... What about forking the first states, i.e. `self.transition.apply.states[0]`?

blocks/bricks/recurrent.py
```diff
@@ -630,3 +631,140 @@ def apply(self, *args, **kwargs):
...
+            else:
+                layer_kwargs = dict(zip(
+                    self.normal_inputs(d),
+                    self.forks[d-1].apply(kwargs["states_%d" % (d-1)],
+                                          as_list=True)))
+                if "mask" in sequences_names:
+                    layer_kwargs["mask"] = kwargs.get("mask")
+
+            for t in ["states", "contexts", "outputs"]:
+                suffix = '_%d' % d
```
rizar: This formatting is not available in Python 3, please use `str.format` or simply `+` operator.

rizar:
This is awesome, and we definitely want it Blocks! I left some change proposals and I will leave some more a bit later, but in general I really like it.

blocks/bricks/recurrent.py
udibr: Using "states" is ugly but this specific name does not leave any chance for error. You cant same the same on states[0]. I can keep the "states" string as a class parameter that you can override. I can also generate a warning if "states" is not located in index 0 of self.transition.apply.states

blocks/bricks/recurrent.py
```diff
@@ -630,3 +631,140 @@ def apply(self, *args, **kwargs):
...
+        for d, transition in enumerate(self.transitions):
+            sequences_names = transition.apply.sequences
+            if d == 0:
+                layer_kwargs = dict(
+                    (s, kwargs.get(s)) for s in sequences_names)
+            else:
+                layer_kwargs = dict(zip(
+                    self.normal_inputs(d),
+                    self.forks[d-1].apply(kwargs["states_%d" % (d-1)],
+                                          as_list=True)))
```
udibr: In this line I am using the states from the previous step in the sequence as the input to all of the layers (sync mode)
there is also the async option in which the output state of one layer is used as the input (through fork) of the next layer for the same step (lets call this async mode.) which one should I use?

rizar: Just stick to Alex Graves papers, I believe they set de-facto standard. Seems like it means your `async` mode.

blocks/bricks/recurrent.py
```diff
@@ -630,3 +631,168 @@ def apply(self, *args, **kwargs):
...
+    states_name : string
+        In a stack of RNN the state of each layer is used as input to the
+        next. This string identify which element in the output of each
+        layer should be used for this task. By default the element is
+        called "states". To be more precise, this is the name of the
+        element in the outputs of the apply method of each transition
+        (layer) that is used, via fork, as the sequences (input) of the
+        next layer. The same element should also appear in the states
+        parameter of the apply method.
+    sync : bool
```
rizar: I would leave only async.

blocks/bricks/recurrent.py
```diff
@@ -630,3 +631,168 @@ def apply(self, *args, **kwargs):
...
         return self.children[0].apply
+
+
+class RecurrentStack(BaseRecurrent, Initializable):
+    u"""Stack of recurrent networks.
+
+    Build a stack of recurrent layers from a supplied list of
+    BaseRecurrent objects. Each object must have a `sequences`,
+    `contexts`, `states` and `outputs` parameters to its `apply` method.
+    It is assumed that the `sequences` of all objects has an "inputs"
```
rizar: I don't see such an assumption in the code.

rizar:
This implementation could be made significantly faster by having two pathways for `iterate=True` and `iterate=False`. If `iterate=True` one can first run the lower layer until the end of sequence and then fork all its states. This can save quite some time since GPU would be used more efficiently.

Let me know if you feel like implementing it, I could give some hints.

udibr:
I'm all for speed and will look into it. I have two questions:
1) How do I detect inside apply if iterate is true/false? or should I rewrite the recurrent wrapper for his?
2) I've noticed that the memory available on the GPU depend on the number of steps (time length of the sequence.) maybe it is faster than linear dependcy. Any idea why?

rizar:
1) I would do smth like 

```python
@recurrent
def _apply(self, ...):
     """Your current apply method."""
     pass
 
@application
    if kwargs.get('iterate', True):
         # faster implementation, which can however share some code with _apply
    else:
         return self._apply(*args, **kwargs)
```

`AttentionRecurrent` from `blocks/bricks/attention.py` can be an example of doing such things.

2) Growth is normal, but it definitely should be linear. If you use attention, there might be some memory issues.

udibr:
I think current code is OK for integration. If/when I will have a faster version I will send another pull request.
P.S. I now understand why memory is linear... In backprop you need to keep all the forward results so you can latter multiply them with the backward error signal...this is the only thing that needs to grow

dwf:
On Wed, Jun 3, 2015 at 2:57 PM, Dmitry Bogdanov <notifications@github.com>
wrote:

> 2) Growth is normal, but it definitely should be linear. If you use
> attention, there might be some memory issues.
>
@rizar Superlinear memory growth could be attributable to the optimizer
giving up earlier on larger graphs and not doing all the in-place stuff
that it could.

rizar:
@udibr , regarding the memory: exactly

Would it be acceptable for you to commit it to `blocks-extras` then? We have already quite a lot of code in Blocks which has certain deficiencies and really wants to be fixed, and it would be nice to keep the ready components and the one in development separated. Since Blocks is now a namespace package, if you have both `blocks` and `blocks-extras` installed on your machine, the same import paths work for both. That is when this class is moved to Blocks, you will not have to change any code that uses it, assuming that in `blocks-extras` it will be also in `blocks.bricks.recurrent` module.

rizar:
@dwf , forgive me my poor knowledge of Theano, but I don't understand why not in-place operations can cause super-linear memory usage growth.

udibr:
not sure where to put the code in blocks-extras?
blocks/bricks/recurrent.py - this will conflict with blocks file of the same name
so to create a new python file or to put it in
blocks/extras/bricks/recurrent.py
but then I will have to modify the code to do the import from blocks.extras.bricks.recurrent

rizar:
Ooops, seems line this doesn't work!

I guess we will merge this into master then. Sorry for long delays: I have both NIPS deadline and graduation in one week.

dwf:
@rizar: my train of thought/speculation goes like this:

* One way that Theano attempts to save memory is by doing operations
in-place when destructive operations are safe. It implements this by means
of a graph optimizer.
* Theano's graph optimizers are heuristic, as doing what we'd like it to do
in full generality is NP-complete. Also a heuristic: when to stop running
further passes of the optimizer.
* If the latter heuristic gives up earlier than it could or perhaps should
on a larger graph, some optimizations may not be done, and the graph of
larger size may not be as thoroughly optimized as a smaller graph of the
same general flavour.
* If some of the optimizations that are not done are memory-saving
optimizations (most notably in-place optimizations), one may see this
manifested in superlinear growth of memory usage even with what appears to
be a linear growth in the number of mathematical operations performed.

Does that make sense?

udibr:
I've added the fast mode you asked (now it is on by default) and also a bonus distribution of inputs to all layers. Implementing fast was done with very littly codeing :-) However, it consumes more memory so I had to reduce my block_size to fit everything into the GPU and as a result the speed improvement where small (I got 30% for 3 layers)

rizar:
Great!

`also a bonus distribution of inputs to all layers.` - isn't it typically called skip connections in literature, in particular [here](http://arxiv.org/abs/1308.0850)? 

30% is a good speedup, I don't know why you call it small. Can you give an estimate of how much the memory consumption increased?

blocks/bricks/recurrent.py
```diff
@@ -630,3 +631,231 @@ def apply(self, *args, **kwargs):
...
     def apply_delegate(self):
         return self.children[0].apply
+
+
+class RecurrentStack(BaseRecurrent, Initializable):
+    u"""Stack of recurrent networks.
+
+    Build a stack of recurrent layers from a supplied list of
+    BaseRecurrent objects. Each object must have a `sequences`,
+    `contexts`, `states` and `outputs` parameters to its `apply` method.
```
rizar: These are requirements from the `recurrent` decorator, so I would add `such as the ones required by the recurrent decorator from :mod:`blocks.bricks.recurrent`.

blocks/bricks/recurrent.py
```diff
@@ -630,3 +631,231 @@ def apply(self, *args, **kwargs):
...
+    It is assumed that all `states` have a "states" element
+    (this can be configured with `states_name` parameter.)
+
+    The `sequences` (the "inputs") of the the first layer (layer 0) is
+    used as the `sequence` of the stack (the instance of this class.)
```
rizar: typo: `sequences`

blocks/bricks/recurrent.py
```diff
@@ -630,3 +631,231 @@ def apply(self, *args, **kwargs):
...
+    them. The '<n>' is the number of the layer starting from 0.
+    The same process is true for the `contexts` and `outputs` of the
+    stack.
+
+    Parameters
+    ----------
+    transitions : list
+        List of recurrent units to use in each layer.
+        Note: A suffix with layer number is added to transitions' names.
+    prototype : :class:`.FeedForward`, optional
```
rizar: Can you please rename it to `fork_prototype`? Simply `prototype` is a bit confusing.

blocks/bricks/recurrent.py
```diff
@@ -630,3 +631,231 @@ def apply(self, *args, **kwargs):
...
+        The transformation applied to each states of each layer  when it
+        is forked to the sequences of the next layer. By default a
+         :class:`.Linear` transformation with biases is used.
+    states_name : string
+        In a stack of RNN the state of each layer is used as input to the
+        next. This string identify which element in the output of each
```
rizar: `which state of each layer`, not `which element in the output`

blocks/bricks/recurrent.py
```diff
@@ -630,3 +631,231 @@ def apply(self, *args, **kwargs):
...
+        element in the outputs of the apply method of each transition
+        (layer) that is used, via fork, as the sequences (input) of the
+        next layer. The same element should also appear in the states
+        parameter of the apply method.
+    fast : bool
+        Use the fast, but also memory consuming, implementation of this
+        code. By default true.
+        WARNING: there is something strange in the
+        behaviour of the slow implementation:
+        Only outputs, used outside this class, can be found using the
```
rizar: That's probably because they are in the inner graph, which by the way can be accessed using `scan_variables` attribute of `ComputationGraph`.

blocks/bricks/recurrent.py
```diff
@@ -630,3 +631,231 @@ def apply(self, *args, **kwargs):
...
+        appear in fast mode.
+    distribute : bool
+        By default False. When true, the input (sequences) is spread to
+        all layer and not just the bottom layer. Each vector in the
+        sequences to the first layer is passed through a linear
+        transformation, without bias, to each of the sequences of the
+        other layers and is added to the
+        usually input comfing from a fork from the states of the previous
+        layer.
+        For this to work it is assumed that all layers have the same names
```
rizar: Such assumptions should be checked in the code.

blocks/bricks/recurrent.py
```diff
@@ -630,3 +631,231 @@ def apply(self, *args, **kwargs):
...
+                layer_kwargs = dict(zip(self.normal_inputs(d), inputs))
+                if "mask" in sequences_names:
+                    layer_kwargs["mask"] = kwargs.get("mask")
+
+            for t in ["states", "contexts", "outputs"]:
+                suffix = '_' + str(d)
+                for s in getattr(transition.apply, t):
+                    layer_kwargs[s] = kwargs.get(s + suffix)
+
+            for k in ['iterate', 'reverse', 'return_initial_states']:
```
rizar: I am not a big fan of hard-coding here keyword arguments `recurrent_apply` can accept.

blocks/bricks/recurrent.py
```diff
@@ -630,3 +631,231 @@ def apply(self, *args, **kwargs):
...
+
+        """
+        results = []
+        last_states = None
+        layer0_kwargs = None
+        for d, transition in enumerate(self.transitions):
+            sequences_names = transition.apply.sequences
+            if d == 0:
+                layer0_kwargs = dict(
+                    (s, kwargs.get(s)) for s in sequences_names)
```
rizar: I it is future-proof to assume that the only arguments a recurrent application method takes are sequences. You should pass all arguments, that i.s. `layer0_kwargs = kwargs`

blocks/bricks/recurrent.py
udibr: Can you please have a look at the code for recurrent decorator? Perhaps it has an error in its docstring saying inputs and not sequences  and if so this should be fixed. However, if you look at the code when handling return_initial_states case there is a reference to .owner.inputs (I guess this is OK because return_initial_states is in use in SequenceGenerator)

rizar: Thanks, this was actually a docstring error, see #702 . `.owner.inputs` should however be not confused with Blocks application method properties, these are inputs of a Theano apply node.

blocks/bricks/recurrent.py
udibr: Yes it is bad but this method (do_apply) in its fast_apply mode is mimicking the behaviour of recurrent_apply so it should be aware of its argumenrs somehow (any other idea will be welcomed.)

blocks/bricks/recurrent.py
udibr: OK maybe this is the solution ... looking at it now...

rizar: You can't handle all of them in this way though. For instance `return_initial_states=True` requires special handling because from the initial states inputs should not be forked.

udibr: There is a special if for return_initial_states few lines below this

udibr:
Took all your comments into account. As for speed, I noticed that the main limit is not from the GPU which is only ~70% utilized but from my CPU. On the CPU side only a single process (core) is active at 100%. My code does not do anything else in addition to RNN. However, it could be that I am using an external GPU on a Mac that causes the CPU to work so hard.

blocks/bricks/recurrent.py
```diff
@@ -630,3 +631,241 @@ def apply(self, *args, **kwargs):
...
+                    input_names, input_dims,
+                    self.transitions[d].get_dims(input_names),
+                    child_prefix='skip_' + str(d)))
+
+        self.children = self.transitions + self.forks + self.skip_connections
+
+        # Programmatically set the apply method
+        self.apply = self.fast_apply if fast else self.low_memory_apply
+        setattr(self.apply, 'sequences', transitions[0].apply.sequences)
+        for t in ["states", "contexts", "outputs"]:
```
rizar: This piece of code needs refactoring. We try not to use abbreviations (except very common ones) and one character identifies in Blocks. You can use `level` instead of `d`, `name` instead of `s`, `property_` instead of `t`, etc.

blocks/bricks/recurrent.py
```diff
@@ -630,3 +631,241 @@ def apply(self, *args, **kwargs):
...
+            v = [s + '_' + str(d)
+                 for d, transition in enumerate(transitions)
+                 for s in getattr(transition.apply, t)]
+            setattr(self.apply, t, v)
+
+        # sum up all the arguments we exepct to see in a call to a transition
+        # apply method, anything else is a recursion control
+        self.special_args = set(sum(map(lambda t: getattr(self.apply, t),
```
rizar: I think `set(self.apply.sequences + ... + ['mask'])` will be much simpler for the user to read.

blocks/bricks/recurrent.py
```diff
@@ -630,3 +631,241 @@ def apply(self, *args, **kwargs):
...
+
+        Returns
+        -------
+        The outputs of all transitions.
+        All return values are of type :class:`~tensor.TensorVariable`.
+
+        """
+        results = []
+        last_states = None
+        layer0_kwargs = None
```
rizar: You can already initialize `layer0_kwargs` here, there is no need to do it inside the loop

blocks/bricks/recurrent.py
```diff
@@ -630,3 +631,241 @@ def apply(self, *args, **kwargs):
...
+                    skip = self.skip_connections[d-1].apply(as_list=True,
+                                                            **layer0_kwargs)
+                    for i in range(len(inputs)):
+                        inputs[i] += skip[i]
+                layer_kwargs = dict(zip(self.normal_inputs(d), inputs))
+                if "mask" in sequences_names:
+                    layer_kwargs["mask"] = kwargs.get("mask")
+
+            # handle all arguments that should be made unique to each layer
+            for t in ["states", "contexts", "outputs"]:
```
rizar: Also here, `s` and `t` are quite bad for readability.

blocks/bricks/recurrent.py
```diff
@@ -630,3 +631,241 @@ def apply(self, *args, **kwargs):
...
+            # Handle all other arguments
+            # For example, if this method is called directly (from fast_apply)
+            # then these arguments can be the same arguments that recurrent
+            # expects to see such as: 'iterate', 'reverse',
+            # 'return_initial_states'
+            for k in set(kwargs.keys()) - self.special_args:
+                layer_kwargs[k] = kwargs[k]
+
+            result = transition.apply(as_list=True, **layer_kwargs)
+
```
rizar: I think this blank line is a bit too much, but this is up to you.

rizar:
Cool, it is getting better and better! Some work on readability is still necessary, see my comments in the code.

I would also like to suggest you a bit different design for skip connections. I think it would be better to move the transformations that yield them outside the brick. Instead, you can add additional inputs to each of the layers. That is, if `skip_connections` is `False`, only first layer gets inputs, and if `skip_connections=True`, all the layers get inputs, which are summed with the contributions from the lower states.

udibr:
The idea to put the skip inputs outside will allow for more flexibility but I see some problems with it:
1) It will complicate the default usage of this brick. Right now it is almost transparent to add it to a single layer network (you just have to decide which states_... you want to take out instead of the single states of the single layer network.)
2) If I understand you correctly, with skip_connect=True I will have multiple sequence arguments in the input with a suffix number. For examples, inputs_0, inputs_1, ... but they are not fully symetric for now because inputs_0 needs a bias while inputs_1, inputs_2,... dont need a bias because there is already a bias term coming from the fork_1, fork_2, ...
Perhaps the best solution for this would have been to move the bias to be inside the LSTM (or whatever other unit you are using.) In this case I could have used the default fork also for a single layer network.
Today when using SequenceGenerator there is a small problem that the defsult fork that feeds the input to the LSTM does not have a bias and I manually have to replace it with a fork with bias...
3) the inputs_0, inputs_1, ... are not symmetric also in respect to the fact that inputs_0 is mandaroy while _1,... are not. Perhaps the name of inputs_0 should be kept just "inputs" while all other keep their number... starting to be confusing...

rizar:
I created an issue about bad default prototype for `Fork`, #703 

Answering your concern (1): typical usage of all recurrent bricks is in a combination with `Fork`. If it were enough to create a `Fork` that would fork inputs for all levels, usage would be very similar to a usual recurrent brick. But as you rightly pointed out, the first layer inputs need biases and the others don't. That is if we implement `skip_connections` by adding more inputs, we will have a recurrent transition for which the correct strategy of forking multiple inputs from a single common one less trivial than usually.

In fact I have been thinking that in addition to these core transition bricks like `SimpleRecurrent`, `GatedRecurrent` and `LSTM` we should also have similar ones with embedded forking (I guess this corresponds to your proposal "perhaps the best solution for this would have been to move the bias to be inside the LSTM"). This would also the nasty issue #586, and in fact in my dirty fork I already have the following dirty prototype: https://gist.github.com/rizar/36451d08719833fb59d0

If we go this way, the current `LSTM` will be become e.g. `CoreLSTM` and so on. Similarly there can be two recurrent stacks: `CoreRecurrentStack`, which takes many inputs, and `RecurrentStack`, which only takes one and takes care of forking it correctly, with or without biases depending on whether its bottom-most input or a skip-connection one.

Regarding (3), inputs can be different depending on `skip_connections=True/False` since you set them programatically.

What do you think about `RecurrentStack` and `CoreRecurrentStack`? The former can inherit from the latter one, and this way no so many change in your current code would be required.

udibr:
I think the CoreRecurrentStack idea is good.
What I would do is say that when calling its apply method then all inputs are mandatory
and that all the input names have a suffix (as is already the case with states and contexts)
In addition in CoreRecurrentStack the fork between the layers will not have a bias (today it has).

The wrapper RecurrentStack will add _0 to all inputs.
The skip_connections is an argument to the wrapper (not to the core.)
If skip_connections is True then the wrapper will add a fork of the inputs
to each of the other layers but this fork will have bias (today it does not.)
If the skip_connections is False the wrapper will add a new type of brick for each of the other layer.
This new brick will be a Bias class that  does not have any input (I think you can write something like that)
Either way, the inputs will compensate for the missing bias from layer to layer inside the core.

The wrapper will be used exactly as it is today, but, the names assigned to the bias will be different
as a result you cant re-load an old model to it....

Since the interface is the same, I think we can wait in implementing the core until you decide what to do with bias in general...

udibr:
I reallized that the contexts should be the same for all layers and therefore should not have a suffix with layer number. I'm working on a fix and also trying out the separation to Core you suggested.

rizar:
Good point about contexts.

One thing I find inconsistent in your version of `CoreRecurrentStack` is that without `RecurrentStack` it will be entirely useless,  because there is no biases between layers. On the contrary the usual "core" transition are in a sense complete and can be used on their own.

I think I know how to make `CoreRecurrentStack` very similar to the basic recurrent bricks we have now. If `skip_connections=True`, `Linear(use_bias=False)` should be used as `fork_prototype` and the `apply.sequences` should contain inputs for all layers with layer indices for disambiguation. On the other hand, if `skip_connections=False`, `Linear(use_bias=True)` should be the fork and only `transitions[0].sequences` should be in `apply.sequences`. This way all inputs will need biases, just as for the other current recurrent bricks. What do you think?

udibr:
My current code has the idea that skip_connections is an init parameter only of RecurrentStack and not of CoreRecurrentStack. This is because CoreRecurrentStack is "simple" and always expect to see inputs (ie sequences) to all layers.

However, you have a point that CoreRecurrentStack should be self contained by default. So what I can do is make the proto for the fork between layers have bias by default set to True.
This will make CoreRecurrentStack work by default if you only supply it inputs to layer 0.
In other words all your input arguments should be suffixed with _0.
(I can even add code that detects input names without a suffix and renames them.)

If someone wants to use CoreRecurrentStack directly and also supply input to other layers he will have to take this into account or replace the proto type for the fork.

This also solves me from the crazy Bias brick I was talking about (very easy to implement but ugly.)

What do you think?

rizar:
But your current code does not have two separate classes currently, it only has `RecurrentStack`. 

I am not sure that `skip_connections` should be introduced at the `RecurrentStack` level. What I had mind for `XXX` and `CoreXXX` classes is that the former just adds forking to the latter one. 

Also I think it would be nicer to keep the input names without suffixes for the case when skip connections are not used. My consideration is that in a well-designed software the one not using the feature (`skip_connections`) should not pay for its existence, and without this feature it would be reasonable to simply use `transition[0].apply.sequences` as `apply.sequences`.

udibr:
After walking in circules for few days I came back to where I started - almost.
I tried to split the code to core and wrapper and it turned out at the end that the wrapper did not contributed significantly to reduce the complexity or improve the functionality.
As a result of all the code shakeup, I fixed few bugs.
I changed the fork between the layers to use bias, as a result the class can be used out-of-the box.
I also made the names that appear in the apply parameters of that are comming from the bottom layer (layer 0) to be unmodified. As a result all the input interface to the apply became much more simpler to someone who does not want to insert inputs to other layers.
From runs I am doing now on https://github.com/udibr/sketch it looks like the RNN behaves better than before so probably I did fix some important bugs.

blocks/bricks/recurrent.py
```diff
@@ -630,3 +631,281 @@ def apply(self, *args, **kwargs):
...
+                    self.transitions[level].get_dims(input_names),
+                    child_prefix='skip_' + str(level)))
+
+        self.children = self.transitions + self.forks + self.skip_connections
+
+        # Programmatically set the apply method
+        self.apply = self.fast_apply if fast else self.low_memory_apply
+        for level, transition in enumerate(transitions):
+            for property_ in ["sequences", "states", "outputs"]:
+                v = getattr(self.apply, property_, []) +\
```
rizar: Please use something a bit more verbose than `v`, e.g. `new_value`.

blocks/bricks/recurrent.py
```diff
@@ -630,3 +631,281 @@ def apply(self, *args, **kwargs):
...
+
+        self.states_name = states_name
+
+        for level, transition in enumerate(transitions):
+            transition.name += '_' + str(level)
+        self.transitions = transitions
+
+        if fork_prototype is None:
+            # By default use Linear (with bias) as a default fork.
+            fork_prototype = Linear(use_bias=True)
```
rizar: If we write here

```python
fork_prototype = Linear(use_bias=not skip_connections)
```

then I think all the `skip_connections` bricks will be not necessary anymore. Everything can be done by an external fork _with_ biases.

rizar:
The code actually has become much better, but I am still in favour of moving things out of this brick, see my comments in the code.

udibr:
I finally understood what you wanted (sorry for being so slow.)
You want the external single fork (for example the fork inside SequenceGenerator) to 
generate all inputs for all layers when skip_connections is True.

1) I fixed the code, hoepfully it is what you wanted.
2) I fixed the docstring so someone else reading it will have an easier time understanding all this. (please read)
3) I wounder if blocks in general is not missing the point if it is so easy to miss understand how to use it.

blocks/bricks/recurrent.py
```diff
@@ -630,3 +631,295 @@ def apply(self, *args, **kwargs):
     @apply.delegate
     def apply_delegate(self):
         return self.children[0].apply
+
+
+class RecurrentStack(BaseRecurrent, Initializable):
+    u"""Stack of recurrent networks.
+
+    Build a stack of recurrent layers from a supplied list of
```
rizar: _builds_

blocks/bricks/recurrent.py
```diff
@@ -630,3 +631,295 @@ def apply(self, *args, **kwargs):
...
+    BaseRecurrent objects. Each object must have a `sequences`,
```
rizar: :class:`~blocks.bricks.recurrent.BaseRecurrent`

blocks/bricks/recurrent.py
```diff
@@ -630,3 +631,295 @@ def apply(self, *args, **kwargs):
...
+    `contexts`, `states` and `outputs` parameters to its `apply` method,
+    such as the ones required by the recurrent decorator from
+    :mod:`blocks.bricks.recurrent`.
+
+    The apply parameters from all layers are concatenated.
```
rizar: "apply arguments" sounds a bit better, because the word "parameters" in Blocks already has certain semantics

blocks/bricks/recurrent.py
```diff
@@ -630,3 +631,295 @@ def apply(self, *args, **kwargs):
...
+    ----------
+    transitions : list
+        List of recurrent units to use in each layer.
+        Note: A suffix with layer number is added to transitions' names.
+    fork_prototype : :class:`.FeedForward`, optional
+        A prototype for the  transformation applied to states_name from
+        the states of each layer. The transformation is used when state
+        is forked to the sequences of the next layer. By default a
+         :class:`.Linear` transformation is used, with bias if
+         skip_connections is False.
```
rizar: ``False``

blocks/bricks/recurrent.py
```diff
@@ -630,3 +631,295 @@ def apply(self, *args, **kwargs):
...
+    states_name : string
+        In a stack of RNN the state of each layer is used as input to the
+        next. The `states_name` identify the element of the states of each
```
rizar: \'states\'

blocks/bricks/recurrent.py
```diff
@@ -630,3 +631,295 @@ def apply(self, *args, **kwargs):
...
+        layer that should be used for this task. By default the element is
+        called "states". To be more precise, this is the name of the
+        element in the outputs of the apply method of each transition
```
rizar: \'outputs\' and also below \'sequences\'

blocks/bricks/recurrent.py
```diff
@@ -630,3 +631,295 @@ def apply(self, *args, **kwargs):
...
+        (layer) that is used, via fork, as the sequences (input) of the
+        next layer. The same element should also appear in the states
+        parameter of the apply method.
+    fast : bool
+        Use the fast, but also memory consuming, implementation of this
+        code. By default true.
+    skip_connections : bool
```
rizar: A few words are missing to explain how it actually changes the model.

blocks/bricks/recurrent.py
```diff
@@ -630,3 +631,295 @@ def apply(self, *args, **kwargs):
...
+                           name='fork_' + str(level),
+                           prototype=fork_prototype)
+                      for level in range(1, depth)]
+
+        self.children = self.transitions + self.forks
+
+        # Programmatically set the apply method
+        self.apply = self.fast_apply if fast else self.low_memory_apply
+        # parameters of base level are exposed as is
+        # excpet for mask which we will put at the ver end. See below.
```
rizar: _expect_, _very_

blocks/bricks/recurrent.py
```diff
@@ -630,3 +631,295 @@ def apply(self, *args, **kwargs):
...
+        for property_ in ["sequences", "states", "outputs"]:
+            setattr(self.apply,
+                    property_,
+                    self.suffixes(getattr(transitions[0].apply, property_), 0))
+
+        # add parameters of other layers
+        if skip_connections:
+            exposed_parameters = ["sequences", "states", "outputs"]
```
rizar: "arguments" would be better, because parameters are in 99.99% shared variables

blocks/bricks/recurrent.py
```diff
@@ -630,3 +631,295 @@ def apply(self, *args, **kwargs):
...
+        # unnamed arguments
+        if "mask" in transitions[0].apply.sequences:
+            self.apply.sequences.append("mask")
+
+        # add context
+        self.apply.contexts = list(set(
+            sum([transition.apply.contexts for transition in transitions], [])
+        ))
+
+        # sum up all the arguments we exepct to see in a call to a transition
```
rizar: _expect_

blocks/bricks/recurrent.py
```diff
@@ -630,3 +631,295 @@ def apply(self, *args, **kwargs):
...
+        In addition the `iterate`, `reverse`, `return_initial_states` or
+        any other argument defined in `recurrent_apply` wrapper.
+
+        Returns
+        -------
+        The outputs of all transitions as defined in `self.apply.outputs`
+        All return values are of type :class:`~tensor.TensorVariable`.
+
+        """
+        nargs = len(args)
```
rizar: It would be cool to support positional arguments here, but then contexts should also be supported. For the first version supporting only keyword arguments would be sufficient I would say.

blocks/bricks/recurrent.py
```diff
@@ -630,3 +631,295 @@ def apply(self, *args, **kwargs):
...
+            for name in transition.apply.states:
+                layer_kwargs[name] = kwargs.get(self.suffix(name, level))
+
+            for name in transition.apply.contexts:
+                layer_kwargs[name] = kwargs.get(name)  # contexts has no suffix
+
+            if level > 0:
+                # add the forked states of the layer below
+                inputs = self.forks[level - 1].apply(last_states, as_list=True)
+                for name, input in zip(normal_inputs, inputs):
```
rizar: Please use `input_` to avoid this crazy highlighting.

tests/bricks/test_recurrent.py
```diff
@@ -205,6 +205,197 @@ def sigmoid(x):
             'initial_state', 'initial_cells'}
 
 
+class TestRecurrentStack(unittest.TestCase):
+    def setUp(self):
+        depth = 4
+        self.depth = depth
+        dim = 3  # don't change, hardwired in the code
+        tarnsitions = [LSTM(dim=dim) for _ in range(depth)]
```
rizar: _transitions_

rizar:
Perfect, this is exactly what I wanted! Hopefully it makes this new class almost interchangeable with all other recurrent bricks.

This is one more (hopefully) serious concern: `reverse=True` will break the current code because it will be applied as many times as many layers you have. Without skip connections a fix would be to pass this option only to the bottom-most layer.  With skip connections things become more involved, unfortunately. Seems like explicitly reversing the inputs would be a solution.

I am looking forward to merge this PR this week. I will be however unavailable from now until Thursday due to relocation.

Regarding understanding Blocks: point taken, we definitely need more tutorials, better documented examples, etc..  I should have time this month to sit down and at least finally compose a brick development tutorial, which explain what is a brick, why we decided to support multiple application methods, what are properties of application methods, why they are necessary and how to set them.

udibr:
Fixed everything and add more comments on usage.
I can add an assert that reverse must be False...

rizar:
> I can add an assert that reverse must be False...

OK, this will do the job for now.

rizar:
A clarification: it should be not `assert` but an exception with a sensible error message.

blocks/bricks/recurrent.py
```diff
@@ -630,3 +631,331 @@ def apply(self, *args, **kwargs):
...
+        The outputs of all transitions as defined in `self.apply.outputs`
+        All return values are of type :class:`~tensor.TensorVariable`.
+
+        """
+        nargs = len(args)
+        assert nargs <= len(self.apply.sequences)
+        kwargs.update(zip(self.apply.sequences[:nargs], args))
+
+        if kwargs.get("reverse", False):
+            raise Exception("reverse=True not implemented.")
```
rizar: That one should be `NotImplementedError` I guess.

blocks/bricks/recurrent.py
rizar: This has not yet been addressed.

blocks/bricks/recurrent.py
```diff
@@ -630,3 +631,331 @@ def apply(self, *args, **kwargs):
...
+            state_index = transition.apply.outputs.index(self.states_name)
+            last_states = result[state_index]
+            if kwargs.get('return_initial_states', False):
+                # Note that the following line reset the tag
+                last_states = last_states[1:]
+
+        return tuple(results)
+
+    @recurrent
+    def low_memory_apply(self, *args, **kwargs):
```
rizar: I am worried that the way you provide the choice between the fast and the memory efficient one is rather quirky and that the `apply` method ends up having no docs at all. Alternatively one can go this way:

```
def do_apply(self, *args, **kwargs):
    pass

@recurrent
def low_memory_apply(self, *args, **kwargs):
    return self.do_apply(*args, **kwargs)

@application
    # all docs go here!
    if kwargs.get('low_memory', False):
        return self.low_memory_apply(*args, **kwargs)
    return self.do_apply(*args, **kwargs)
```

As a consequence the `fast` flag would be not required.

rizar:
There are only a few issues left, after they are fixed this PR can be merged.

EloiZ:
Hi,
I am really interested in this PR. I have already started using it, however, I cannot figure out why the `fast_apply` method is faster than `low_memory_apply`. I thought that one scan wrapping the whole thing (`low_memory_apply`) would be faster than one scan for each layer (`fast_apply`)

udibr:
I've added the @apply as you suggested

rizar:
All right, thanks! There is still one unaddressed comment in the code, but I will merge. Great job!

rizar:
Created #724

PR #690, opened 2015-06-01
Remove remaining examples

bartvm:


bartvm:
For some reason these were missed from https://github.com/mila-udem/blocks/pull/638. All examples are now in `blocks-examples`.

PR #691, opened 2015-06-01
Rename Sigmoid to Logistic

### LINKED ISSUE #617: Sigmoid -> Logistic
The `Sigmoid` brick should really be named `Logistic`.

"Sigmoid" simply means "S-shaped" ([dictionary](http://dictionary.reference.com/browse/sigmoid?s=t), [Wikipedia](https://en.wikipedia.org/wiki/Sigmoid_function)).

bartvm:
Fixes #617.

### DIFF
--- a/blocks/bricks/__init__.py
+++ b/blocks/bricks/__init__.py
@@ -482,7 +482,7 @@ class Tanh(Activation):
         return tensor.tanh(input_)
 
 
-class Sigmoid(Activation):
+class Logistic(Activation):
     @application(inputs=['input_'], outputs=['output'])
     def apply(self, input_):
         return tensor.nnet.sigmoid(input_)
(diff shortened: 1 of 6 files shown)


dwf:
Do we want to pull the trigger on this before NIPS?

bartvm:
Well, that debate is being had right here, https://github.com/mila-udem/blocks/pull/685#issuecomment-107662178, with @rizar. My personal view is that we went through the trouble of creating a `stable` branch exactly so that we could continue development in the `master` branch without having to be afraid of breaking people's code.

I am developing with the expectation that people doing a `git pull` on `master` right before the deadline, know how to revert to an old commit. If they don't, they shouldn't be checking out a bleeding-edge version days before the deadline.

dwf:
+1

rizar:
For this one, I don't mind, because renaming things in your code is easier than adapting to a brand new serialization policy.

PR #693, opened 2015-06-02
Enable namespace packages for blocks-extras

bartvm:
The namespace package seems to throw `coverage.py` off, it runs the installed files of Blocks but measures the Git copy, resulting in 0% coverage. Installing in editable mode fixes that.


dwf:
LGTM. Although I expect that suddenly some non-negligible fraction of blocks-users traffic will start to be about namespace package fuckups...

bartvm:
I hope not... What's the alternative though? Creating a separate `blocks_users` package? I guess the change can be made if it turns out to be an issue. Otherwise we can always drop Python 2 support, PEP420 solves most issues :o)

PR #695, opened 2015-06-03
Allow child_prefix keyword in Fork

dwf:
Gets rid of potential conflict between **kwargs and the hardcoded `child_prefix=`.

@rizar


rizar:
What a stupid mistake on my side!

PR #700, opened 2015-06-05
Workaround to avoid runtime error in ifelse

lamblin:
This prevents the insertion in the graph of an expression that
cannot be computed, even if it is only in the non-necessary branch
of ifelse.
In particular, it makes it possible to use that code in a mode that
does not execute the ifelse lazily.


bartvm:
Great, thanks!

@mducoffe

PR #702, opened 2015-06-06
Fix recurrent docs (inputs -> sequences)

rizar:
@udibr , anything else?

### DIFF
--- a/blocks/bricks/recurrent.py
+++ b/blocks/bricks/recurrent.py
@@ -59,26 +59,35 @@ class BaseRecurrent(Brick):
 def recurrent(*args, **kwargs):
     """Wraps an apply method to allow its iterative application.
 
-    This decorator allows you to use implementation of an RNN transition to
-    process sequences without writing the iteration-related code again and
-    again. In the most general form information flow of a recurrent network
-    can be described as follows: depending on the context variables and
-    driven by input sequences the RNN updates its states and produces
-    output sequences.  Thus the input variables of your transition function
-    play one of three roles: an element of an input sequence, a context or
-    a state.  These roles should be specified in the method's signature to
-    make iteration possible.
+    This decorator allows you to implement only one step of a recurrent
+    network and enjoy applying it to sequences for free. The idea behind is
+    that its most general form information flow of an RNN can be described
+    as follows: depending on the context and driven by input sequences the
+    RNN updates its states and produces output sequences.
+
+    Given a method describing one step of an RNN and a specification
+    which of its inputs are the elements of the input sequence,
+    which are the states and which are the contexts, this decorator
+    returns an application method which implements the whole RNN loop.
+    The returned application method also has additional parameters,
+    see documentation of the `recurrent_apply` inner function below.
 
     Parameters
     ----------
     sequences : list of strs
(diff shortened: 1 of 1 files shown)


blocks/bricks/recurrent.py
```diff
@@ -59,27 +59,24 @@ def initial_state(self, state_name, batch_size, *args, **kwargs):
...
-    This decorator allows you to use implementation of an RNN
-    transition to process sequences without writing the
-    iteration-related code again and again. In the most general form
-    information flow of a recurrent network can be described as
-    follows: depending on the context variables and driven by input
-    sequences the RNN updates its states and produces output sequences.
-    Thus the input variables of your transition function play one of
-    three roles: an input, a context or a state. These roles should be
-    specified in the method's signature to make iteration possible.
+    This decorator allows you to use implementation of an RNN transition to
```
bartvm: I'm not following this sentence. Maybe "This decorator allows you to implement a recurrent transition to process sequences..."?

blocks/bricks/recurrent.py
```diff
@@ -59,27 +59,24 @@ def initial_state(self, state_name, batch_size, *args, **kwargs):
...
+    process sequences without writing the iteration-related code again and
+    again. In the most general form information flow of a recurrent network
```
bartvm: In *its* most general form *the* information...

blocks/bricks/recurrent.py
bartvm: Also, *the* implementation

blocks/bricks/recurrent.py
```diff
@@ -59,27 +59,24 @@ def initial_state(self, state_name, batch_size, *args, **kwargs):
...
     ----------
-    inputs : list of strs
-        Names of the arguments of the apply method that play input
-        roles.
+    sequences : list of strs
+        Specifies which of the arguments are elements of input sequences.
     states : list of strs
-        Names of the arguments of the apply method that play state
-        roles.
+        Specifies which of the arguments that are states.
```
bartvm: ~~that~~

blocks/bricks/recurrent.py
rizar: All right, although I meant only to fix a blatant argument name error in this PR,  I will also rewrite this chuck in a more readable way. Thanks for corrections!

bartvm: Oops, I didn't realise the rest was just a different wrapping of the same text.

rizar:
I rewrote the `recurrent` docs, which does not make them perfect, but I think somewhat more understandable.

rizar:
@bartvm what is the correct way to reference tutorials from docstrings?

bartvm:
Something like this should work:
```
Please read :doc:`the tutorial on RNNS <rnn>` for more information.
```

udibr:
If I remeber correctly the tutorial also needs a re-write

rizar:
@udibr , @dmitriy-serdyuk , @bartvm , can somebody of you guys give a read to these updated docs?

bartvm:
LGTM, we'll probably want examples (doctests) at one point, but this is better than it was.

PR #705, opened 2015-06-07
Remove Pandas dependency

bartvm:
So lately I've grown a bit annoyed that every time I install Blocks on e.g. a cluster it spends a long time installing and compiling Pandas, even though the only call to Pandas in the entire library is a one-liner:
```python
DataFrame.from_dict(self, orient='index')
```

I suggest we simplify the installation a bit by removing Pandas from the dependencies, and let users call `DataFrame.from_dict(self, orient='index')` manually.

### DIFF
--- a/blocks/log.py
+++ b/blocks/log.py
@@ -2,12 +2,6 @@
 from collections import defaultdict
 from numbers import Integral
 
-try:
-    from pandas import DataFrame
-    PANDAS_AVAILABLE = True
-except ImportError:
-    PANDAS_AVAILABLE = False
-
 
 class TrainingLog(defaultdict):
     """Base class for training logs.
@@ -64,10 +58,3 @@ class TrainingLog(defaultdict):
     @property
     def last_epoch_row(self):
         return self[self.status['_epoch_ends'][-1]]
-
-    def to_dataframe(self):
-        """Convert a log into a :class:`.DataFrame`."""
-        if not PANDAS_AVAILABLE:
-            raise ImportError("The pandas library is not found. You can"
-                              " install it with pip.")
-        return DataFrame.from_dict(self, orient='index')
--- a/blocks/log.py
+++ b/blocks/log.py
@@ -22,6 +22,17 @@ class TrainingLog(defaultdict):
         By default it contains ``iterations_done``, ``epochs_done`` and
         ``_epoch_ends`` (a list of time stamps when epochs ended).
 
+    Notes
+    -----
+    For analysis of the logs, it can be useful to convert the log to a
+    Pandas_ data frame:
+
+    .. code:: python
+
+       df = DataFrame.from_dict(log, orient='index')
+
+    .. _Pandas: http://pandas.pydata.org
+
     """
     def __init__(self):
         super(TrainingLog, self).__init__(dict)
--- a/docs/conf.py
+++ b/docs/conf.py
@@ -56,8 +56,7 @@ intersphinx_mapping = {
     'theano': ('http://theano.readthedocs.org/en/latest/', None),
     'numpy': ('http://docs.scipy.org/doc/numpy/', None),
     'scipy': ('http://docs.scipy.org/doc/scipy/reference/', None),
-    'python': ('http://docs.python.org/3.4', None),
-    'pandas': ('http://pandas.pydata.org/pandas-docs/stable/', None)
+    'python': ('http://docs.python.org/3.4', None)
 }
 
 
--- a/req-travis-conda.txt
+++ b/req-travis-conda.txt
@@ -2,7 +2,6 @@ coverage==3.7.1
 cython==0.22
 h5py==2.5.0
 numpy==1.9.2
-pandas==0.16.1
 pytables=3.1.1
 six==1.9.0
 toolz==0.7.2
--- a/requirements.txt
+++ b/requirements.txt
@@ -1,6 +1,5 @@
 numpy==1.9.2
 six==1.9.0
-pandas==0.16.1
 progressbar2==2.7.3
 PyYaml==3.11
 toolz==0.7.2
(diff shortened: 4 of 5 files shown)


dwf:
Merging. Also 100 points for branch name.

blocks/log.py
```diff
@@ -64,10 +69,3 @@ def previous_row(self):
     @property
     def last_epoch_row(self):
         return self[self.status['_epoch_ends'][-1]]
-
-    def to_dataframe(self):
-        """Convert a log into a :class:`.DataFrame`."""
-        if not PANDAS_AVAILABLE:
-            raise ImportError("The pandas library is not found. You can"
-                              " install it with pip.")
-        return DataFrame.from_dict(self, orient='index')
```
rizar: I don't think we should have removed that.

For one, now the user has to remember this class and typo more, and `to_dataframe` is often used in an interactive setting (and before it is done essentially nothing can be done with the log). In addition when we have more then one backend (and I believe we should), this method could hide from the user the most efficient way to convert a log to a dataframe. Also `blocks-plot` is probably broken now, as far as I remember `to_dataframe` was used there.

bartvm: Well, that's why I added the snippet to the docs. I think there's certainly a benefit to having it if we start supporting multiple backends. However, there's also a cost, and I've seen many people run into installation issues with Blocks or wait a long time for Pandas to compile (especially on clusters), while you are the only person I know of who uses this one-liner. So I'd argue that the costs outweigh the tiny, tiny benefit of being able to type `log.to_dataframe()` instead of `DataFrame.from_dict(log, orient='index')`.

I'll check if `blocks-plot` is broken. Tests passed, but perhaps it's not covered by tests.

rizar: > while you are the only person I know of who uses this one-liner

I am curious then what do the others do when they want to analyze the log. Always type `DataFrame.from_dict(...)`?

I don't find the link between long compilation times at the cluster and removal of `to_dataframe` is convincing. `pandas` can be made an optional dependency installed only on explicit request and `to_dataframe` can  gracefully fail when it is not installed.

bartvm: I generally write my own quick helper function, something like
```python
def get_values(v):
    return [(t, log[t][v]) for t in log if v in log[t]]
plt.plot(*zip(*get_values('valid_error')))
```
Not everyone uses Pandas all the time. I know it's powerful, but I personally find it pretty unwieldy, so I don't use it unless I need to.

As I mentioned in the other comment, I'm personally not a fun of bits and pieces of the library not working even after the user installed everything by the book. It makes the code messier (import-traps everywhere, custom warnings in the code) and catches users by surprise. I think it's a valid thing to do if we really need a big library that we know we won't use anywhere else, but if it's a small library I would suggest just adding it (e.g. `progressbar2`) and if it's a big library that can easily be avoided, I'd suggest getting rid of it entirely (e.g. `pandas`).

rizar: I don't agree with these arguments. Two types of dependencies should be distinguished, those on which significant amount of code depends, and those, with which Blocks supports only limited cooperation, export being an example. I think there can be as many as required dependencies of the second type, provided that they are not mandatory (i.e. not installed by default and not necessary most of the time) and that when they are missing, a meaningful error with an installation instruction is raised.

I also would say that using pandas makes much more sense than oneliners like yours, and my educated guess is that many from the Blocks audience use it. We can of course take a survey :)

Finally `SQLite` backend is coming, and having `to_dataframe` becomes more important. 

Also I think that this PR was merged too quickly, and that further PRs that remove functionality from Blocks are held a bit longer. It is reasonable to give the community at least three days to protest, e.g. I  simply didn't have time to reply. It took you and David less the 24 hours to merge it. I propose to return `to_dataframe` and `pandas` as optional dependency.

bartvm: Pandas has advantages over my one-liner, but I was trying to show that it is definitely possible to inspect logs without `to_dataframe`. Hence, IMHO it is not essental enough to justify 20 lines of code and documentation in order to bring down the amount of code the user needs to type from 41 characters to 19 characters. What ever happened to lean and mean? `blocks-extras` seems like a much better place for these kind of non-essential but handy shortcuts. We could have a `log_converter` module in there that has `to_dataframe`, `to_csv`, etc.

And "lean and mean" isn't just out of principle. Introducing these kind of "small dependencies" you argue for means more code that can break because of changes in external interface, installations that can be slow or fail, and frustration from users that commands that they rightfully expect to work end with an `ImportError`.

I also think 24 hours is a perfectly good turnaround time for a PR. If we are going to wait days so that every single person of the development team can ponder things over, development will come to a crawl. It often takes 1-2 days just for a single person to have a look, if every single one of us needs to give the LGTM we will be waiting weeks to merge 5-line PRs. If there is something you insist on reviewing, I would suggest leaving a comment saying that you have concerns and ask merging to be delayed (like you did with the new serialization method).

rizar: Your first paragraph ignores the point that `to_dataframe` for other backends can take much more than 41 character. Then you say  "installations can be slow or fail", even though I did not argue in favour of automatically installing pandas. You are right that there will be an `ImportError`, but well, if a person calls `to_dataframe` he is expected to know what pandas is and how to get it. That is frustration will be close to zero. When I tried FFT convolution in Theano it was very natural for me to install a couple of required libraries. 

Right, there is a "lean and mean" motto. But it can be used to justify whatever one wants, up to the statement that Blocks is not necessary and everything should be done in pure Theano. If 80% users convert the log to a dataframe in order to analyze it, I doubt whether the place for this functionality should be in `blocks-extras`. 

I don't agree that 24 hours is okay for the PR that drops existing functionality. It would be okay for a new feature, but when you delete something it might cause serious frustration. Especially that you did not move this functionality to extras, you simply deleted it. 


Anyway, this discussion is turning lengthy, and I just want to make clear that here I care about two things:

- I would like to call for greater caution when something is removed from Blocks

- efficient routines for exporting the Blocks logs as dataframes should be provided, and I am not sure that `blocks-extras`  is the right place for it, because I find it very essential functionality. But at the very least they should be in `blocks-extras`, and the PR that removed them without moving there should be considered an example of excessive rush with merging.

PR #707, opened 2015-06-07
Add tests for Timing extension

bartvm:
Also made the `MockMainLoop` slightly more flexible.

PR #708, opened 2015-06-07
Add tests for main loop interruption and error handling

bartvm:
Note that part of this won't count towards test coverage because `coverage.py` doesn't recognise code run by another process.

### DIFF
--- a/tests/test_main_loop.py
+++ b/tests/test_main_loop.py
@@ -24,37 +24,17 @@ class WriteBatchExtension(TrainingExtension):
 
 def test_main_loop():
 
-    class TestDataStream(object):
-
-        def __init__(self):
-            self.epochs = self._generate_data()
-
-        def _generate_data(self):
-            def wrap_in_dicts(iterable):
-                for x in iterable:
-                    yield dict(data=x)
-            yield iter(wrap_in_dicts([1, 2, 3]))
-            yield iter(wrap_in_dicts([4, 5]))
-            yield iter(wrap_in_dicts([6, 7, 8, 9]))
-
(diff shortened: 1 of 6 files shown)


bartvm:
I don't know why it was there, but for some reason `format_exc(exception)` was used, but the argument to `format_exc` should be the traceback limit (an integer)... In Python 2 this silently fails, but in Python 3 it actually crashes, which happened with the unit test. (Long live unit tests.)

bartvm:
#706, #707 and #708 are mostly just trying to cover some gaps in coverage. Could someone have a quick look and give the LGTM? @rizar @vdumoulin @dmitriy-serdyuk

tests/test_main_loop.py
```diff
@@ -77,3 +64,41 @@ def do_test(with_serialization):
...
+    p.join()
+
+
+def test_error():
+    ext = TrainingExtension()
+    ext.after_batch = MagicMock(side_effect=KeyError)
+    ext.on_error = MagicMock()
+    main_loop = MockMainLoop(extensions=[ext, FinishAfter(after_epoch=True)])
+    assert_raises(KeyError, main_loop.run)
+    ext.on_error.assert_called_once_with()
```
dmitriy-serdyuk: It's possible to check that we have a log record `got_exception` here, isn't it?

bartvm: Yeah, I can add that.

bartvm: 6ed490590d04cafc03e4aa12729061083cdac384

dmitriy-serdyuk:
Looks good, let's wait until it passes tests.

PR #710, opened 2015-06-08
WIP: Add support for SQLite

bartvm:
Todo:
- [x] Tests
- [x] Put `_epoch_ends` back now that lists are supported (or find alternative)
- [ ] Write better documentation

Future PR stuff:
- [ ] Allow comments/notes to be added to the log, to identify the correct log in the database
- [ ] Add some helper functions for logs
  - Convert to `DataFrame`
  - Something like `log[:]['valid_error']` instead of `[log[t]['valid_error'] for t in log if 'valid_error' in log[t]]`


bartvm:
I'll need to rebase and clean some stuff up, but in principle this seems functional and I invite anyone who's interested to have a first look (@dwf, @rizar). It gives each log a UUID, and whenever the log gets pickled, or the main loop restarted, the log gets a new UUID and keeps a reference to the old one, allowing us to reconstruct the entire history.

Registering a new UUID whenever the log gets pickled might give us many ancestors if you pickle regularly, but it's needed because otherwise it is impossible to keep current behaviour. An example: A user trains model A for 5 iterations, pickles, and then continues for 5 more. The user then unpickles A. It's expected that the log status at this point says 5 iterations were done. For this to be true, it must be stored in the database somewhere separately from the status of model after pickling.

Another issue is that SQLite only has support for simple datatypes: integers, floats, text, and bytes. It crashes when trying to save e.g. tuples, dicts, ndarrays, etc. I currently register an adapter for each unknown object type that comes along, which just pickles objects to bytestrings and stores that in the database. When loading, bytestrings get unpickled.

The approach is probably not too bad for small objects, but if a user ends up storing references to large objects in the database, it will quickly spin out of control. I'd almost suggest not having support for extra types, to keep things simple, but that would mean refactoring `Checkpoint` somehow, because it uses tuples. The upside is that it makes `SQLiteLog` quite transparent and more of a drop-in replacement; as long as it pickles it saves.

dwf:
Could we just set a hard upper limit on the size of the pickled
representation? Like 4kb or something before we bail...

bartvm:
Good idea. We could also just give a clear warning, telling users that this is probably not a good idea.

blocks/log.py
```diff
@@ -1,26 +1,318 @@
...
+@add_metaclass(ABCMeta)
+class _TrainingLog(object):
+    """Base class for training log.
 
     A training log stores the training timeline, statistics and other
     auxiliary information. Information is stored as a nested dictionary,
     ``log[time][key]``. An entry without stored data will return an empty
-    dictionary (like ``defaultdict(dict)``).
+    dictionary-like object that can be written to (like
+    ``defaultdict(dict)``). However, entries need to be added one by one,
```
dwf: What does "might not work" mean here? Could fail depending on the backend?

bartvm: Yeah. But I just realised that `log[t].update({'foo': 'bar'})` should work for all backends, so I'll recommend that.

dwf: Possible/worthwhile to override `__setitem__` with an informative error message?

blocks/log.py
```diff
@@ -22,6 +60,279 @@ class TrainingLog(defaultdict):
...
         ``_epoch_ends`` (a list of time stamps when epochs ended).
 
+    """
+    def __init__(self, uuid=None):
+        if uuid is None:
+            self.uuid = uuid4()
+        else:
+            self.uuid = uuid
+        if uuid is None:
+            self.status.update({
```
dwf: Deeply confused. Where does `self.status` get initially assigned?

bartvm: By the child class. Basically, the child class constructs an empty status object first, which adheres to the `MutableMapping` interface but stores its data in e.g. a database table, or a dictionary. The parent `_TrainingLog` is then responsible for adding the first default values to this `self.status` object.

bartvm:
Okay, tests pass and this is functionally complete, ready for review! Suggestions for refactoring are welcome, because it's quite a bit of code in the end (nearly 350 lines) and a bit messy here and there.

In the future we'll need a way to easily inspect experiment results from a variety of logging backends, and easily provide these results to a variety of "clients" e.g. (live) plotting (across a network), visualizing weights, etc. I think we'll want a kind of bridge, a server that interfaces with these different backends, and exposes this data easily. So I was thinking of writing a very simple REST API that could answer a set of simple requests with JSON data: All available experiments, the status and channels available for a given experiment, and the values of these channels if they are non-binary, maybe it could send the parameters as well (loading the model from the `saved_to` entry). This would be easy to integrate with e.g. Bokeh, but also with a host of other tools.

bartvm:
Thanks to Flask, here's a REST API to Blocks logs stored in SQLite databases: https://gist.github.com/bartvm/fd62fb2db16a6943d200

bartvm:
Anyone got time to review this? Starting next week I'm not sure how much time I will have to address comments.

dwf:
I have a phone interview thing at 4, I'll have a look after that.

bartvm:
Great, thanks. And bon courage with the phone interview!

blocks/log/log.py
```diff
@@ -0,0 +1,136 @@
...
+
+    In addition to the set of records displaying training dynamics, a
+    training log has a :attr:`status` attribute, which is a dictionary with
+    data that is not bound to a particular time.
+
+    .. warning::
+
+       Changes to mutable objects might not be reflected in the log,
+       depending on the backend. So don't use
+       ``log.status['key'].append(...)``, use ``log.status['key'] = ...``
```
dwf: Gross. I wish there was some syntactic sugar we could pull to make this less error-prone.

blocks/log/log.py
```diff
@@ -0,0 +1,136 @@
...
+        if uuid is None:
+            self.status.update({
+                'iterations_done': 0,
+                'epochs_done': 0,
+                '_epoch_ends': [],
+                'resumed_from': None
+            })
+
+    @property
+    def b_uuid(self):
```
dwf: Is the base class the right place to be depending on SQLite?

blocks/log/sqlite.py
```diff
@@ -0,0 +1,240 @@
...
+
+A {} object of {} bytes was stored in the SQLite database. SQLite natively \
+only supports numbers and text. Other objects will be pickled before being \
+saved. For large objects, this can be slow and degrade performance of the \
+database."""
+
+
+def adapt_obj(obj):
+    blob = sqlite3.Binary(cPickle.dumps(obj))
+    if len(blob) > 1024 * 4:
```
dwf: Maybe make this a config setting with 4096 as default?

blocks/log/sqlite.py
```diff
@@ -0,0 +1,240 @@
...
+
+LARGE_BLOB_WARNING = """
+
+A {} object of {} bytes was stored in the SQLite database. SQLite natively \
+only supports numbers and text. Other objects will be pickled before being \
+saved. For large objects, this can be slow and degrade performance of the \
+database."""
+
+
+def adapt_obj(obj):
```
dwf: Docstring

blocks/log/sqlite.py
```diff
@@ -0,0 +1,240 @@
...
+    blob = sqlite3.Binary(cPickle.dumps(obj))
+    if len(blob) > 1024 * 4:
+        warnings.warn('large objects stored in SQLite' +
+                      LARGE_BLOB_WARNING.format(type(obj), len(blob)))
+        # Prevent the warning with variable message from repeating
+        warnings.filterwarnings('ignore', 'large objects .*')
+    return blob
+
+
+def adapt_ndarray(obj):
```
dwf: Idem for docstring

blocks/log/sqlite.py
```diff
@@ -0,0 +1,240 @@
...
+    r"""Training log using SQLite as a backend.
+
+    Parameters
+    ----------
+    database : str, optional
+        The database (file) to connect to. Can also be `:memory:`. See
+        :func:`sqlite3.connect` for details. Uses `config.sqlite_database`
+        by default.
+    \*\*kwargs
+        Arguments to pass to :class:`_TrainingLog`
```
dwf: If we're going to refer to `_TrainingLog` in the docs should we give it a better name, like `TrainingLogMixin`?

dwf: Or `TrainingLogBase` or something...

blocks/log/sqlite.py
```diff
@@ -0,0 +1,240 @@
...
+
+        """
+        state = self.__dict__.copy()
+        del state['conn']
+        self.resume()
+        return state
+
+    def __setstate__(self, state):
+        self.__dict__.update(state)
+        self.conn = sqlite3.connect(self.database)
```
dwf: Will this cause the same sorts of problems that we did gymnastics to avoid with datasets? i.e. should we make this lazy, so that you can unpickle even if your database attribute is somewhere else (because you moved machines, for example, or because you're trying to use someone else's checkpoint, or whatever)?

dwf: Seems like it's exceedingly straightforward to fix, too:

```python
@property
def conn(self):
    if not hasattr(self, '_conn'):
        self._conn = sqlite3.connect(self.database)
    return self._conn
```

blocks/log/sqlite.py
```diff
@@ -0,0 +1,240 @@
...
+        with self.log.conn:
+            self.log.conn.execute(
+                "INSERT OR REPLACE INTO entries VALUES (?, ?, ?, ?)",
+                (self.log.b_uuid, self.time, key, value)
+            )
+
+    def __delitem__(self, key):
+        with self.log.conn:
+            self.log.conn.execute(
+                "DELETE FROM entries WHERE uuid = ? AND time =? AND key = ?",
```
dwf: I guess =? should be = ?, but no big deal really.

blocks/log/sqlite.py
```diff
@@ -0,0 +1,240 @@
...
+        ))
+
+
+class SQLiteEntry(MutableMapping):
+    """Store log entries in an SQLite database.
+
+    Each entry is a row with the columns `uuid`, `time` (iterations done),
+    `key` and `value`. Note that SQLite only supports numeric values,
+    strings, and bytes (e.g. the `uuid` column). This means there is no
+    support for tuples, dictionaries, NumPy arrays, etc.
```
dwf: This makes it sounds like you can't use them, whereas... it seems that you can, with all the adapter jazz.

blocks/log/sqlite.py
```diff
@@ -0,0 +1,240 @@
...
+        :func:`sqlite3.connect` for details. Uses `config.sqlite_database`
+        by default.
+    \*\*kwargs
+        Arguments to pass to :class:`_TrainingLog`
+
+    Notes
+    -----
+    .. todo::
+
+       Currently this log ignores previous logs in case of resumption.
```
dwf: Is this to be done before merge?

bartvm: Already done, forgot to update docstring.

blocks/log/sqlite.py
bartvm: Right, also forgot to update this docstring.

blocks/log/sqlite.py
```diff
@@ -0,0 +1,240 @@
...
+
+    def __len__(self):
+        return self.log.conn.execute(
+            ANCESTORS_QUERY + "SELECT COUNT(*) FROM entries "
+            "WHERE uuid IN ancestors AND time = ?",
+            (self.log.b_uuid, self.time,)
+        ).fetchone()[0]
+
+    def __iter__(self):
+        return map(itemgetter(0), self.log.conn.execute(
```
dwf: Just to verify, six.moves.map on Python 2 returns an iterator?

blocks/main_loop.py
```diff
@@ -225,7 +232,7 @@ def _run_epoch(self):
                 pass
         self.status['epoch_started'] = False
         self.status['epochs_done'] += 1
-        self.status['_epoch_ends'].append(self.status['iterations_done'])
+        self.status['_epoch_ends'] += [self.status['iterations_done']]
```
dwf: I'd leave a comment here explaining why we don't `.append()`, for posterity.

blocks/log/sqlite.py
bartvm: Yeah, it maps to `itertools.imap`, slightly confusing.

dwf:
Alrighty, finished a pass over it.  LGTM essentially. Only functional change I would recommend would be the lazy property for `self.conn`, and maybe moving the warning limit to a config option.

blocks/log/log.py
bartvm: The problem is that we need to store `resumed_from` in the status, and SQLite3 will only take it as a buffer or memoryview. The alternative is for the UUID to be a different object depending on the backend i.e. for `SQLiteLog` we would return a buffer for `log.uuid` while for `TrainingLog` we return an UUID. I'm not a big fan of that either though.

blocks/log/sqlite.py
bartvm: 7d870bb401816bb09fbc722f55c182ab09f34a32

bartvm:
I think I addressed all comments.

dwf:
Travis failure.

blocks/log/sqlite.py
```diff
@@ -0,0 +1,270 @@
...
+    def __getstate__(self):
+        """Retrieve the state for pickling.
+
+        :class:`sqlite3.Connection` objects are not picklable, so the
+        `conn` attribute is removed and the connection re-opened upon
+        unpickling.
+
+        """
+        state = self.__dict__.copy()
+        del state['conn']
```
dwf: `del state['_conn']`

dwf: Actually, you should probably gate it with an `if '_conn' In state:`.

bartvm: 4c1caf5599b721cde3ef521319851a6aceccd48f

dwf:
Travis is passing, I'll merge.

P.S. Will the snippet for pandas also work with this, given that it inherits from Mapping?

dwf:
Hrm, I just noticed "write better documentation" was left unchecked, and I might have jumped the gun a little on merging. Sorry about that. I think the docs are okay now though.

bartvm:
I have no idea whether the Pandas snippet works... I guess technically it should, assuming that Pandas doesn't do anything strange behind the scenes.

The writing better documentation was from when there was barely any, I think it's okay now, just forgot to check the box.

PR #712, opened 2015-06-09
Load extra config from blocks-extras

bartvm:


bartvm:
@jbornschein @nouiz Here's my proposal for dealing with unknown configurations. It gives a warning, but only if the unknown configuration is a likely typo. Otherwise it just silently ignores it. I prefer this over giving warnings each time a user doesn't import `blocks.extras` (regardless of whether this warning is given at imports or at exit), but it's slightly more user-friendly than being entirely silent.

nouiz:
Good direction, but how do you define typo?

bartvm:
I use [`difflib.get_close_matches`](https://docs.python.org/2/library/difflib.html#difflib.get_close_matches) with the default threshold of 0.6, which is quite sensitive (for example, `some_other_backend` is a match to `log_backend`). (Matching with edit-distance might be better, but that's not in the standard library.) For now, that seems to work well, because the few configurations we have are all quite different.

nouiz:
This seem good. I didn't know about that lib. I made a ticket in Theano about this: https://github.com/Theano/Theano/issues/3021

jbornschein:
I guess that works -- it solves the problem. And it uses a stdlib that I didn't know about.

But in general I feel a bit uncanny when it comes to "software wants to be clever about variable or configuration-option names". I think of configuration parsers as simple pieces of software that do something very deterministic and that should not surprise me. And when I ever see a message like "unknown configuration option" I would expect that I will always get such a warning. Not only when the given option is sufficiently similar to an actual option.

According to my judgement I would thus prefer to simply try to import `blocks.extra`. And rather deal with the surprises that might pop up as a result of that.

PR #716, opened 2015-06-12
WIP: Change default prototypes

### LINKED ISSUE #703: Detault prototypes in parallel.py
Currently the default prototype for `Parallel` is `Linear(use_bias=False)`, which is further inherited by `Fork` and `Merge`. I think that we should rather have no default prototype for `Parallel`, `Linear()` for `Fork` and `Linear(use_bias=False)` for `Merge`.

rizar:
Fixes #703.

@udibr , you asked for that.

### DIFF
--- a/blocks/bricks/attention.py
+++ b/blocks/bricks/attention.py
@@ -289,8 +289,8 @@ class SequenceContentAttention(GenericSequenceAttention, Initializable):
     match_dim : int
         The dimension of the match vector.
     state_transformer : :class:`.Brick`
-        A prototype for state transformations. If ``None``, the default
-        transformation from :class:`.Parallel` is used.
+        A prototype for state transformations. If ``None``,
+        a linear transformation is used.
     attended_transformer : :class:`.Feedforward`
         The transformation to be applied to the sequence. If ``None`` an
         affine transformation is used.
@@ -310,6 +310,8 @@ class SequenceContentAttention(GenericSequenceAttention, Initializable):
     def __init__(self, match_dim, state_transformer=None,
                  attended_transformer=None, energy_computer=None, **kwargs):
         super(SequenceContentAttention, self).__init__(**kwargs)
+        if not state_transformer:
+            state_transformer = Linear()
         self.match_dim = match_dim
         self.state_transformer = state_transformer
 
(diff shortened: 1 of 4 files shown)


rizar:
This PR introduces the following changes:

- `Parallel` is too generic to have a default prototype, therefore the `prototype` argument becomes mandatory for it

- for `Fork` it almost always makes  sense to have to have biases (except for the case when only one output is forked from an input which already has biases). It is the new default.

- for `Merge` it typically does not make sense, and `Linear(use_bias=False)` stays as the prototype for it

- for `Distribute` it is less obvious what the default prototype should be. So far I see a bias-less one, but perhaps there should be none?

- now the fork in `SequenceGenerator` has biases

rizar:
@udibr , any chance you can review this PR?

blocks/bricks/attention.py
```diff
@@ -290,8 +290,8 @@ class SequenceContentAttention(GenericSequenceAttention, Initializable):
     match_dim : int
         The dimension of the match vector.
     state_transformer : :class:`.Brick`
-        A prototype for state transformations. If ``None``, the default
-        transformation from :class:`.Parallel` is used.
+        A prototype for state transformations. If ``None``,
+        a linear transformation is used.
```
udibr: The issue of backward compatibility is a big issue with this PR and this is just one example.
What will happen to code outside this repository? What about blocks-extra and examples?

rizar: `blocks-extras` should not depend on this, for examples we only test that they are runnable so far, that is the test will pass.

But the main point is that backward compatibility was never a design goal for Blocks. For such an ever-changing field as deep learning it is not possible to maintain a backward compatible framework, I believe.

rizar:
@dmitriy-serdyuk , may I ask you to review this PR?

blocks/bricks/parallel.py
```diff
@@ -200,13 +208,18 @@ class Distribute(Fork):
         The names of the targets.
     source_name : str
         The name of the source.
+    target_dims : list
+        A list of target dimensions, corresponding to `target_names`.
+    source_dim : int
+        The dimension of the source input.
+    prototype : :class:`~blocks.bricks.Feedforward`, optional
+        The transformation prototype. A copy will be created for every
+        input. By default a linear transformation is used.
```
dmitriy-serdyuk: It's not linear, it's affine here, isn't it?

dmitriy-serdyuk: Oh, it's the other way around, sorry. Maybe we should rename `Linear` to `Affine` and create `Linear` which is `Linear(use_bias=False)` using current terminology.

dmitriy-serdyuk:
LGTM.

blocks/bricks/parallel.py
rizar: I like the idea to rename the current `Linear` into `Affine`! Can you please create a ticket?

PR #717, opened 2015-06-13
Disable BLOCKS_PROFILE for tests

rizar:
The test logs are on Travis are inconvenient to browse because they are mixed with profiling. @dwf , @bartvm , would you mind disabling it?


bartvm:
I agree, but can we save stderr to a file instead and cat that at the end? That way the profiling gets tested, and it can be useful anyway to have timings to see how our main loop is behaving.

rizar:
Assuming that nothing useful is outputted to stderr, we could. I will try and see that ends up there.

rizar:
Pretty much everything is written to stderr, we can't separate the profiling output this way. Is there any test for profiling other than setting it on the global level in `.travis.yaml`? If there is no, I could add one and remove it from the global testing configuration.

PR #718, opened 2015-06-14
WIP: initial_states returns all states at a time

### LINKED ISSUE #616: Initial states for recurrent bricks are not annotated with meaningful names.
See #605.

rizar:
Fixes #616, see discussion in #605.

This is one step forward towards a clearer definition of what the inputs and the outputs of an application method are. 

@dmitriy-serdyuk , you might be interested in this PR. Could you please review it?

### DIFF
--- a/blocks/bricks/attention.py
+++ b/blocks/bricks/attention.py
@@ -52,7 +52,7 @@ from blocks.bricks import (Brick, Initializable, Sequence,
 from blocks.bricks.base import lazy, application
 from blocks.bricks.parallel import Parallel, Distribute
 from blocks.bricks.recurrent import recurrent, BaseRecurrent
-from blocks.utils import dict_union, dict_subset
+from blocks.utils import dict_union, dict_subset, pack
 
 
 class AbstractAttention(Brick):
@@ -180,7 +180,7 @@ class AbstractAttention(Brick):
 
         Returns
         -------
-        initial_glimpses : dict of (str, :class:`~theano.Variable`)
+        initial_glimpses : list of :class:`~theano.Variable`
             The initial values for the requested glimpses. These might
             simply consist of zeros or be somehow extracted from
             the attended.
@@ -385,11 +385,10 @@ class SequenceContentAttention(GenericSequenceAttention, Initializable):
         return (['attended', 'preprocessed_attended', 'attended_mask'] +
                 self.state_names)
 
-    @application
+    @application(outputs=['weighted_averages', 'weights'])
     def initial_glimpses(self, batch_size, attended):
-        return {
-         "weighted_averages": tensor.zeros((batch_size, self.attended_dim)),
-         "weights": tensor.zeros((batch_size, attended.shape[0]))}
+        return [tensor.zeros((batch_size, self.attended_dim)),
+                tensor.zeros((batch_size, attended.shape[0]))]
 
     @application(inputs=['attended'], outputs=['preprocessed_attended'])
     def preprocess(self, attended):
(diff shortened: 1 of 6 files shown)


blocks/bricks/attention.py
```diff
@@ -389,12 +386,10 @@ def take_glimpses_inputs(self):
...
-    def initial_glimpses(self, name, batch_size, attended):
-        if name == "weighted_averages":
-            return tensor.zeros((batch_size, self.attended_dim))
-        elif name == "weights":
-            return tensor.zeros((batch_size, attended.shape[0]))
-        raise ValueError("Unknown glimpse name {}".format(name))
+    def initial_glimpses(self, batch_size, attended):
+        return {
+         "weighted_averages": tensor.zeros((batch_size, self.attended_dim)),
+         "weights": tensor.zeros((batch_size, attended.shape[0]))}
```
dmitriy-serdyuk: I thought, it should be done using `outputs=[...]` option of the `application` decorator, returning a dictionary looks bulky and may be a source of errors due to typos.

blocks/bricks/recurrent.py
```diff
@@ -30,18 +30,19 @@ class BaseRecurrent(Brick):
     has_bias = False
 
     @application
-    def initial_state(self, state_name, batch_size, *args, **kwargs):
-        r"""Return an initial state for an application call.
+    def initial_states(self, batch_size, *args, **kwargs):
+        r"""Return an initial states for an application call.
```
dmitriy-serdyuk: > an

blocks/bricks/attention.py
rizar: Indeed! Let's say those were cosmic rays influencing  my brain.. I will definitely do it as you suggest.

blocks/bricks/recurrent.py
```diff
@@ -179,11 +190,11 @@ def recurrent_apply(brick, application, application_call,
...
-                    # TODO init_func returns 2D-tensor, fails for iterate=False
-                    kwargs[state_name] = (
-                        brick.initial_state(state_name, batch_size,
-                                            *args, **kwargs))
-                    assert kwargs[state_name]
+                    try:
+                        kwargs[state_name] = initial_states[state_name]
+                    except KeyError:
+                        raise KeyError(
+                            "no initial state for '{}'".format(state_name))
```
dmitriy-serdyuk: May be we should add a link to a recurrent brick here. Like `No initial state for {} in brick {}`.

dmitriy-serdyuk:
Nice work!

The only comment is that you should output tuples everywhere instead of dictionaries.

blocks/bricks/recurrent.py
```diff
@@ -179,11 +190,11 @@ def recurrent_apply(brick, application, application_call,
                             kwargs[state_name](state_name, batch_size,
                                                *args, **kwargs))
                 else:
-                    # TODO init_func returns 2D-tensor, fails for iterate=False
```
dmitriy-serdyuk: Can you open an issue for it?

dmitriy-serdyuk: Oops, I meant another one TODO.

blocks/search.py
```diff
@@ -95,11 +95,10 @@ def _compile_context_computer(self):
             self.inputs, self.contexts, on_unused_input='ignore')
 
     def _compile_initial_state_computer(self):
-        initial_states = [
-            self.generator.initial_state(
-                name, self.beam_size,
+        # TODO: should be now extractable from the computation graph
```
dmitriy-serdyuk: I created #719 for it.

rizar:
Seems like I addressed all your comments. Would you mind taking one more look?

dmitriy-serdyuk:
LGTM

Is there any difference returning a tuple or a list?

rizar:
In the contexts of Blocks application methods not so much, I guess, since we anyway make it a list, see blocks/bricks/base.py:303

PR #721, opened 2015-06-15
Fix failing test after merging a PR in Fuel

vdumoulin:
I don't know why Travis showed the Blocks tests as passing in the Fuel PR, but here we are...

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -35,7 +35,7 @@ install:
   - pip install -q -r req-travis-pip.txt
 script:
   - pip install -e . -r requirements.txt # Tests setup.py
-  - curl -O https://raw.githubusercontent.com/bartvm/fuel/master/.travis-data.sh | bash -s -- mnist
+  - curl https://raw.githubusercontent.com/mila-udem/fuel/master/.travis-data.sh | bash -s -- mnist
   - # Must export environment variable so that the subprocess is aware of it
   - export THEANO_FLAGS=floatX=$FLOATX,optimizer=fast_compile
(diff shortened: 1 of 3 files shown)


vdumoulin:
Alright... the failing tests cascaded to `blocks-examples`. I'll make a PR for that as well.

bartvm:
It's because of the comment I made here: https://github.com/mila-udem/fuel/pull/156#issuecomment-110393345

Currently, the Blocks tests on the Fuel repo are actually run with the Fuel `master` branch i.e. they're not being tested at all. We need to move the Fuel installation command to *before* the Blocks installation command, otherwise pip just installs Fuel's `master` branch, and when the Fuel installation command is called later, pip just says "already installed".

vdumoulin:
I'll make a PR to change that, thanks!

vdumoulin:
@bartvm It appears MNIST isn't being properly downloaded by Blocks.

I had a look at the corresponding line in `.travis.yml` and I think there's a problem with it:

``` bash
curl -O https://raw.githubusercontent.com/bartvm/fuel/master/.travis-data.sh | bash -s -- mnist
```

should probably be

``` bash
curl https://raw.githubusercontent.com/bartvm/fuel/master/.travis-data.sh | bash -s -- mnist
```

as I think `curl -O` will write to a file without outputing anything to the stdout.

Also, should we switch from `bartvm/fuel/master` to `mila-udem/fuel/master`?

bartvm:
Yes to both :)

vdumoulin:
Tests pass, and the changes are trivial, I'll go ahead and merge.

Issue #488, opened 2015-03-16
Inconsistent usage of `params` and `parameters`


bartvm:
We have `Brick.params`, `Selector.get_params`, `Model.get_params` and `DifferentiableCostMinimizer.params`, but on the other hand we have `ComputationGraph.parameters` and `dump_parameters`.

I think we should move to `parameters` in all cases, considering the general policy of preferring full names whenever they're not extraordinarily long (which I guess is the case here). Either way, we need to be consistent.

rizar:
Sounds reasonable.

EderSantana:
There is also `role.{WEIGHT, BIAS}` but `{weights,biases}_init`

dwf:
This just got changed, and apparently we missed the
weights_init/biases_init arguments. I'd be all for changing it
eventually.

bartvm:
Yeah, @mducoffe mentioned this in https://github.com/bartvm/blocks/pull/449#issue-60671600 too, and I think we should just be consistent and change all of them then.

bouthilx:
We assigned the ticket for CCW this week. Do we all agree on using ```parameters``` rather than ```params```?

vikkamath:
Shouldn't this be closed?

bouthilx:
Lets wait for travis tests to pass before we close.

PR #725, opened 2015-06-18
Params to parameters

mohammadpz:
Hi,

Based on #488 , I replaced ```param(s)``` and ```get_param(s)``` with ```parameter(s)``` and ```get_parameter(s)```.

### DIFF
--- a/blocks/algorithms/__init__.py
+++ b/blocks/algorithms/__init__.py
@@ -60,7 +60,7 @@ class DifferentiableCostMinimizer(TrainingAlgorithm):
     ----------
     cost : :class:`~tensor.TensorVariable`
         The objective to be minimized.
-    params : list of :class:`~tensor.TensorSharedVariable`
+    parameters : list of :class:`~tensor.TensorSharedVariable`
         The parameters to be tuned.
 
     Attributes
@@ -70,7 +70,7 @@ class DifferentiableCostMinimizer(TrainingAlgorithm):
         updates are done using the old values of optimized parameters.
     cost : :class:`~tensor.TensorVariable`
         The objective to be minimized.
-    params : list of :class:`~tensor.TensorSharedVariable`
+    parameters : list of :class:`~tensor.TensorSharedVariable`
         The parameters to be tuned.
 
     Notes
@@ -90,9 +90,9 @@ class DifferentiableCostMinimizer(TrainingAlgorithm):
        currently.
 
     """
-    def __init__(self, cost, params):
+    def __init__(self, cost, parameters):
         self.cost = cost
-        self.params = params
+        self.parameters = parameters
         self._cost_computation_graph = ComputationGraph(self.cost)
         self._updates = []
 
@@ -151,9 +151,10 @@ class GradientDescent(DifferentiableCostMinimizer):
     .. code-block::  python
 
         for batch in data:
-            steps = step_rule.compute_steps(params, gradients_wr_params)
-            for param in params:
-                param -= steps[param]
+            steps = step_rule.compute_steps(parameters,
+                                            gradients_wr_parameters)
+            for parameter in parameters:
+                parameter -= steps[parameters]
 
     Note, that the step is *subtracted, not added*! This is done in order
     to make step rule chaining possible.
@@ -189,15 +190,16 @@ class GradientDescent(DifferentiableCostMinimizer):
     def __init__(self, step_rule=None, gradients=None, known_grads=None,
                  **kwargs):
         if gradients:
-            kwargs.setdefault("params", gradients.keys())
+            kwargs.setdefault("parameters", gradients.keys())
         super(GradientDescent, self).__init__(**kwargs)
 
         self.gradients = gradients
         if not self.gradients:
             logger.info("Taking the cost gradient")
             self.gradients = dict(
-                equizip(self.params, tensor.grad(self.cost, self.params,
-                                                 known_grads=known_grads)))
+                equizip(self.parameters, tensor.grad(self.cost,
+                                                     self.parameters,
+                                                     known_grads=known_grads)))
             logger.info("The cost gradient computation graph is built")
         else:
             if known_grads:
@@ -218,8 +220,8 @@ class GradientDescent(DifferentiableCostMinimizer):
         # Note: the gradients are computed in the same order in which
         # the parameters were given. Keep it like that to ensure
         # reproducibility.
-        for param in self.params:
-            all_updates.append((param, param - self.steps[param]))
+        for parameter in self.parameters:
+            all_updates.append((parameter, parameter - self.steps[parameter]))
         all_updates += self.step_rule_updates
         self._function = theano.function(self.inputs, [], updates=all_updates)
         logger.info("The training algorithm is initialized")
@@ -237,7 +239,7 @@ class GradientDescent(DifferentiableCostMinimizer):
 @add_metaclass(ABCMeta)
 class StepRule(object):
     """A rule to compute steps for a gradient descent algorithm."""
-    def compute_step(self, param, previous_step):
+    def compute_step(self, parameter, previous_step):
         """Build a Theano expression for the step for a parameter.
 
         This method is called by default implementation of
@@ -245,7 +247,7 @@ class StepRule(object):
 
         Parameters
         ----------
-        param : :class:`~tensor.TensorSharedVariable`
+        parameter : :class:`~tensor.TensorSharedVariable`
             The parameter.
         previous_step : :class:`~tensor.TensorVariable`
             Some quantity related to the gradient of the cost with respect
@@ -290,10 +292,11 @@ class StepRule(object):
             A list of tuples representing updates to be performed.
 
         """
-        parameter_wise = [self.compute_step(param, previous_steps[param])
-                          for param in previous_steps]
+        parameter_wise = [self.compute_step(parameter,
+                                            previous_steps[parameter])
+                          for parameter in previous_steps]
         steps, updates = equizip(*parameter_wise)
-        steps = OrderedDict((param, step) for param, step
+        steps = OrderedDict((parameter, step) for parameter, step
                             in equizip(previous_steps.keys(), steps))
         updates = list(itertools.chain(*updates))
         return steps, updates
@@ -342,7 +345,7 @@ class Scale(StepRule):
     def __init__(self, learning_rate=1.0):
         self.learning_rate = shared_floatx(learning_rate)
 
-    def compute_step(self, param, previous_step):
+    def compute_step(self, parameter, previous_step):
         return self.learning_rate * previous_step, []
 
 
@@ -364,8 +367,8 @@ class BasicMomentum(StepRule):
     def __init__(self, momentum=0.):
         self.momentum = shared_floatx(momentum)
 
-    def compute_step(self, param, previous_step):
-        velocity = shared_floatx(param.get_value() * 0.)
+    def compute_step(self, parameter, previous_step):
+        velocity = shared_floatx(parameter.get_value() * 0.)
         step = self.momentum * velocity + previous_step
         updates = [(velocity, step)]
         return step, updates
@@ -428,9 +431,9 @@ class AdaDelta(StepRule):
         self.decay_rate = shared_floatx(decay_rate)
         self.epsilon = shared_floatx(epsilon)
 
-    def compute_step(self, param, previous_step):
-        mean_square_step_tm1 = shared_floatx(param.get_value() * 0.)
-        mean_square_delta_x_tm1 = shared_floatx(param.get_value() * 0.)
+    def compute_step(self, parameter, previous_step):
+        mean_square_step_tm1 = shared_floatx(parameter.get_value() * 0.)
+        mean_square_delta_x_tm1 = shared_floatx(parameter.get_value() * 0.)
 
         mean_square_step_t = (
             self.decay_rate * mean_square_step_tm1 +
@@ -486,8 +489,8 @@ class BasicRMSProp(StepRule):
         self.decay_rate = shared_floatx(decay_rate)
         self.epsilon = 1. / max_scaling
 
-    def compute_step(self, param, previous_step):
-        mean_square_step_tm1 = shared_floatx(param.get_value() * 0.)
+    def compute_step(self, parameter, previous_step):
+        mean_square_step_tm1 = shared_floatx(parameter.get_value() * 0.)
         mean_square_step_t = (
             self.decay_rate * mean_square_step_tm1 +
             (1 - self.decay_rate) * tensor.sqr(previous_step))
@@ -570,8 +573,8 @@ class StepClipping(StepRule):
         multiplier = tensor.switch(norm < self.threshold,
                                    1, self.threshold / norm)
         steps = OrderedDict(
-            (param, step * multiplier)
-            for param, step in previous_steps.items())
+            (parameter, step * multiplier)
+            for parameter, step in previous_steps.items())
         return steps, []
 
 
@@ -599,9 +602,9 @@ class VariableClipping(StepRule):
     -----
     Because of the way the :class:`StepRule` API works, this particular
     rule implements norm clipping of the value *after* update in the
-    following way: it computes ``param - previous_step``, scales it
+    following way: it computes ``parameter - previous_step``, scales it
     to have (possibly axes-wise) norm(s) of at most `threshold`,
-    then subtracts *that* value from `param` to yield an 'equivalent
+    then subtracts *that* value from `parameter` to yield an 'equivalent
     step' that respects the desired norm constraints. This procedure
     implicitly assumes one is doing simple (stochastic) gradient descent,
     and so steps computed by this step rule may not make sense for use
@@ -629,21 +632,23 @@ class VariableClipping(StepRule):
         if len(axis) != len(self.axis):
             raise ValueError("axis must be unique")
 
-    def compute_step(self, param, previous_step):
+    def compute_step(self, parameter, previous_step):
         if any(ax >= previous_step.ndim for ax in self.axis):
             raise ValueError("Invalid axis {} for {}, ndim={}".format(
-                self.axis, param, previous_step.ndim))
+                self.axis, parameter, previous_step.ndim))
         if len(self.axis) == 0:
-            norms = l2_norm([param - previous_step])
+            norms = l2_norm([parameter - previous_step])
         else:
-            squares = tensor.sqr(param - previous_step)
+            squares = tensor.sqr(parameter - previous_step)
             norms = tensor.sqrt(
                 reduce(lambda t, a: t.sum(axis=a, keepdims=True),
                        sorted(self.axis), squares))
-        # We want a step s* that is the same as scaling (param - previous_step)
-        # by threshold / norm when threshold < norm.
-        shrinking_step = (param -
-                          (self.threshold / norms) * (param - previous_step))
+        # We want a step s* that is the same as scaling
+        # (parameter - previous_step) by threshold / norm
+        # when threshold < norm.
+        shrinking_step = (parameter -
+                          (self.threshold / norms) *
+                          (parameter - previous_step))
         return tensor.switch(norms > self.threshold,
                              shrinking_step,
                              previous_step), ()
@@ -675,11 +680,11 @@ class AdaGrad(StepRule):
         self.learning_rate = learning_rate
         self.epsilon = epsilon
 
-    def compute_step(self, param, previous_step):
+    def compute_step(self, parameter, previous_step):
         name = 'adagrad_sqs'
-        if param.name:
-            name += '_' + param.name
-        ssq = shared_floatx(param.get_value() * 0.,
+        if parameter.name:
+            name += '_' + parameter.name
+        ssq = shared_floatx(parameter.get_value() * 0.,
                             name=name)
 
         ssq_t = (tensor.sqr(previous_step) + ssq)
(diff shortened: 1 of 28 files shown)


blocks/algorithms/__init__.py
```diff
@@ -151,9 +151,10 @@ class GradientDescent(DifferentiableCostMinimizer):
     .. code-block::  python
 
         for batch in data:
-            steps = step_rule.compute_steps(params, gradients_wr_params)
-            for param in params:
-                param -= steps[param]
+            steps = step_rule.compute_steps(parameters,
+                                            gradients_wr_parameters)
+            for parameter in parameters:
+                parameter -= steps[parameters]
```
bouthilx: steps[parameter]

blocks/model.py
```diff
@@ -50,18 +50,18 @@ class AbstractModel(object):
 
     """
     @abstractmethod
-    def get_params(self):
+    def parameters(self):
```
bouthilx: ```parameters``` would be good for an attribute, but it's a getter method. Should be ```get_parameters```

blocks/model.py
```diff
@@ -50,18 +50,18 @@ class AbstractModel(object):
...
         -------
-        params : OrderedDict
+        parameters : OrderedDict
             Dictionary of (name, parameter) pairs.
 
         """
         pass
 
-    def get_param_values(self):
+    def parameter_values(self):
```
bouthilx: same

blocks/model.py
```diff
@@ -84,22 +85,22 @@ def set_param_values(self, param_values):
 
         Parameters
         ----------
-        param_values : OrderedDict
+        parameter_values : OrderedDict
             Dictionary of (parameter name, :class:`~numpy.ndarray`) pairs.
 
         """
-        params = self.get_params()
+        parameters = self.parameters()
```
bouthilx: Fix get_ here

blocks/model.py
```diff
@@ -161,15 +162,18 @@ def __init__(self, outputs):
...
 
-        brick_param_names = {
-            v: k for k, v in Selector(self.top_bricks).get_params().items()}
-        self.params = []
-        for param in VariableFilter(roles=[PARAMETER])(self.shared_variables):
-            if param in brick_param_names:
-                self.params.append((brick_param_names[param], param))
+        brick_parameter_names = {
+            v: k
+            for k, v in Selector(self.top_bricks).parameters().items()}
```
bouthilx: get_

blocks/model.py
```diff
@@ -161,15 +162,18 @@ def __init__(self, outputs):
...
+        self.parameters = []
```
bouthilx: You would overwrite the method here. Fix get_ and this line will be good

blocks/model.py
```diff
@@ -182,7 +186,7 @@ def get_objective(self):
             return self.outputs[0]
         raise NotImplementedError
 
-    def get_params(self):
+    def parameters(self):
```
bouthilx: get_

blocks/select.py
```diff
@@ -142,8 +143,8 @@ def select(self, path):
         current_bricks = [None]
         for node in path.nodes:
             next_bricks = []
-            if isinstance(node, Path.ParamName):
-                return list(Selector(current_bricks).get_params(node).values())
+            if isinstance(node, Path.ParameterName):
+                return list(Selector(current_bricks).parameters(node).values())
```
bouthilx: get_

blocks/select.py
```diff
@@ -155,40 +156,41 @@ def select(self, path):
             current_bricks = next_bricks
         return Selector(current_bricks)
 
-    def get_params(self, param_name=None):
+    def parameters(self, parameter_name=None):
```
bouthilx: get_

tests/bricks/test_attention.py
```diff
@@ -115,7 +115,7 @@ def generate_mask(length, batch_size):
     assert glimpses_vals.shape == (input_length, batch_size, attended_dim)
 
     assert (len(ComputationGraph(outputs).shared_variables) ==
-            len(Selector(recurrent).get_params()))
+            len(Selector(recurrent).parameters()))
```
bouthilx: get_

tests/test_model.py
```diff
@@ -18,7 +18,7 @@ def test_model():
     assert model.get_top_bricks() == [mlp1, mlp2]
     # The order of parameters returned is deterministic but
     # not sensible.
-    assert list(model.get_params().items()) == [
+    assert list(model.parameters().items()) == [
```
bouthilx: get_

tests/test_model.py
```diff
@@ -30,17 +30,19 @@ def test_model():
...
-    got_param_values = model3.get_param_values()
-    assert len(got_param_values) == len(param_values)
-    for name, value in param_values.items():
-        assert_allclose(value, got_param_values[name])
+    model3.set_parameter_values(parameter_values)
+    assert numpy.all(mlp3.linear_transformations[0].parameters[0].get_value()
+                     == 2)
+    assert numpy.all(mlp3.linear_transformations[0].parameters[1].get_value()
+                     == 3)
+    got_parameter_values = model3.parameter_values()
```
bouthilx: get_

tests/test_select.py
```diff
@@ -39,13 +39,13 @@ def test_path():
     assert hash(path4) != hash(path2)
 
 
-def test_selector_get_params_uniqueness():
+def test_selector_parameters_uniqueness():
```
bouthilx: get_

tests/test_select.py
```diff
@@ -39,13 +39,13 @@ def test_path():
...
     top = MockBrickTop(
         [MockBrickBottom(name="bottom"), MockBrickBottom(name="bottom")],
         name="top")
 
     selector = Selector([top])
-    assert_raises(ValueError, selector.get_params)
+    assert_raises(ValueError, selector.parameters)
```
bouthilx: get_

bouthilx:
There is a typo the loop at l.157 in blocks/algorithms/__init__.py.
All parameters getter method definitions and calls should be fixed. See my comments in the diff.

tests/test_select.py
mohammadpz: You're right. I'll fix it tomorrow morning.

blocks/model.py
mohammadpz: @bouthilx , Is there anything wrong here?

blocks/model.py
```diff
@@ -161,15 +162,18 @@ def __init__(self, outputs):
...
-        brick_param_names = {
-            v: k for k, v in Selector(self.top_bricks).get_params().items()}
-        self.params = []
-        for param in VariableFilter(roles=[PARAMETER])(self.shared_variables):
-            if param in brick_param_names:
-                self.params.append((brick_param_names[param], param))
+        brick_parameter_names = {
+            v: k
+            for k, v in Selector(self.top_bricks).get_parameters().items()}
+        self.parameters = []
```
bouthilx: Oh! There is a conflict with ComputationGraph.parameters (see [here](https://github.com/mila-udem/blocks/blob/master/blocks/graph.py#L92)). It seems Model previously had both params and parameters attributes. 

You can use get, set interface to get around this. Build parameters locally (without self.) and then use set_paramter_values(). 

However there is still a problem. There is a get, set interface for parameters **plus** an attribute parameters which is built differently and returns a list rather than an OrderedDict. I think ComputationGraph.parameters should be overridden. @bartvm @rizar @dwf Which one would you keep?

To override ComputationGraph.parameters we would have to define getter method with @property and hide parameters under _parameters. That means we will have 2 getter functions for _parameters. I don't like that much, why not using getter and setter from @property all over for attribute parameters? Or there is a better way to override it that I don't know?

bouthilx: @dmitriy-serdyuk What do you think?

rizar: @bouthilx , I will take a look today, sorry for delay. This is a shady area in Blocks and requires a bit of refactoring.

mohammadpz: @bouthilx , @rizar , what do you think about using ```params``` instead of ```parameters```? ```params``` is shorter while still informative enough.

rizar: Indeed, a situation when a class has `parameters` property, `get_params` method and `params` attribute is rather confusing!

The real solution would be a brutal refactoring of the `Model` class. But to finish this PR I would suggest

- rename `params` attribute to `_params`, that is simply hide it from user eyes

-  rename `get_params` into `get_parameter_dict`. As far as I remember this method is used in the `reverse_words` demo in `blocks-examples`, which means that it should be also renamed there

I think that `ComputationGraph.parameters` should be kept, since it a popular property with very clear semantics.

Does it make sense?

rizar: @mohammadpz , I agree that `params` is a quite frequent abbreviation, but still my vote goes to `parameters`. Somehow it does not feel to me as established of an abbreviations as e.g. `dim` is.

bouthilx: @rizar Why do we maintain both Model._params and ComputationGraph.parameters at the same time? Can't we build Model.get_parameter_dict on top of ComputationGraph.parameters and get rid of Model._params?

rizar: Good point, we can do that. As far as I can see the `_params` are there only for speed, but probably that was a premature optimization. I can't imagine this method being called many times.

mohammadpz:
@bouthilx , So what should I do?

blocks/model.py
```diff
@@ -69,14 +69,15 @@ def get_param_values(self):
...
-        param_values : OrderedDict
+        parameter_values : OrderedDict
             Dictionary of (parameter name, :class:`~numpy.ndarray`) pairs.
 
         """
-        return OrderedDict((name, param.get_value())
-                           for name, param in self.get_params().items())
+        return OrderedDict(
+            (name, parameter.get_value())
+            for name, parameter in self.parameters().items())
```
bouthilx: get_

blocks/model.py
bouthilx: Then we could still build get_parameter_dict on top of parameters and cache it in _parameters if necessary. As you say we can leave caching implementation for latter if we realize it's critical for speedup. Thanks for your help Dima!

bouthilx: @rizar One last thing. Is there any reason why it would be useful to build ComputationGraph.parameters as a list rather than a dictionary? We could still use ComputationGraph.parameters.values() when we only care about a list. This would simplify the common interface. Otherwise, I find names of get_parameter_dict and get_parameter_values to be quite confusing.

bouthilx:
@mohammadpz I want to be sure about new AbstractModel interface with respect to parameters before we move on. I should settle down soon with help of @rizar .

blocks/model.py
rizar: Leaving `parameters` as they currently are makes sense for interface uniformity reasons. Similar methods of the `ComputationGraph` class include `variables`, `auxiliary_variables`, `shared_variables`.

Indeed, the interface which has `parameters`, `get_parameter_dict` and `get_parameter_values` is a bit confusing. But I would like to postpone fixing it to a further PR, because in fact the whole `model.py` file requires major refactoring. I think it makes sense to accomplish the goal of this PR first, that is  to ensure that `params` abbreviation is not used.

Would it be acceptable?

bouthilx: Oh I see. 

Yep, lets keep it simple for now and plan refactoring of model.py for another ticket.

bouthilx:
@mohammadpz Here is the new plan. 

- Replace every params by parameters
- Except for Model:
    - Replace every get_params by get_parameter_dict
    - get_parameter_dict will build a dictionnary using self.parameters (which comes from ComputationGraph)

So you should only have to modify what you've done for Model.

blocks/model.py
```diff
@@ -50,18 +50,18 @@ class AbstractModel(object):
...
         """Return the model parameters.
 
         Returns
         -------
-        params : OrderedDict
+        parameters : OrderedDict
             Dictionary of (name, parameter) pairs.
 
         """
         pass
```
mohammadpz: @bouthilx ,
Should I create an ```OrderedDict``` out of ```self.parameters``` instead of ```pass```?

mohammadpz:
@bouthilx , Travis also give me errors on ```blocks-examples``` repo.

blocks/model.py
bouthilx: No. There is no self.parameters in AbstractModel. It's up to AbstractModel childs to define how they build parameter_dict. [Model](https://github.com/mila-udem/blocks/pull/725/files#diff-ff403bd4d4da2f194bd025e38a075649R122) class inherits from ComputationGraph, that's where self.parameters comes from.

blocks/model.py
```diff
@@ -190,7 +194,7 @@ def get_params(self):
         parameters that do not belong to any brick.
 
         """
-        return self.params
+        return self.parameters
```
bouthilx: self.parameters is a list

blocks/model.py
```diff
@@ -161,15 +162,18 @@ def __init__(self, outputs):
...
+        for parameter in VariableFilter(
+                roles=[PARAMETER])(self.shared_variables):
+            if parameter in brick_parameter_names:
+                self.parameters.append((brick_parameter_names[parameter],
+                                        parameter))
             else:
-                self.params.append((param.name, param))
-        self.params = OrderedDict(self.params)
+                self.parameters.append((parameter.name, parameter))
+        self.parameters = OrderedDict(self.parameters)
```
bouthilx: You should not try to set self.parameters anymore. It's ComputationGraph's property. There is no need to build it here anymore.

bouthilx:
@mohammadpz Let's finish the work here before trying to fix other blocks-

blocks/model.py
mohammadpz: I removed this part. So now ```self.parameters``` comes from ```ComputationGraph```. Right?

blocks/model.py
```diff
@@ -190,7 +181,7 @@ def get_params(self):
         parameters that do not belong to any brick.
 
         """
-        return self.params
+        return OrderedDict(self.parameters)
```
bouthilx: How should OrderedDict define the keys out of a list? ;)

mohammadpz: That's AI! :)

blocks/model.py
```diff
@@ -190,7 +179,13 @@ def get_params(self):
         parameters that do not belong to any brick.
 
         """
-        return self.params
+        if (len(set(parameter.name for parameter in self.parameters)) !=
+                len(self.parameters)):
+            raise ValueError("parameters should have different names")
```
mohammadpz: @bouthilx ,@rizar,
But the name of parameters may be the same. like weights are all ```w```'s.

bouthilx: @mohammadpz @rizar Wouldn't [l.172](https://github.com/mila-udem/blocks/pull/725/files#diff-ff403bd4d4da2f194bd025e38a075649L172) crush parameters then?

rizar: @mohammadpz , names of parameters must be different for one brick. Likewise names of children of a brick and names of top bricks must be different. This line ensures that the users and Blocks developers take care of  it...

mohammadpz: But now it fails Travis.

rizar: Can you please investigate why Travis fails? What are the duplicate names?

mohammadpz: For example in ```test_model.py```, we have:
```python
    x = tensor.matrix('x')
    mlp1 = MLP([Tanh(), Tanh()], [10, 20, 30], name="mlp1")
    mlp2 = MLP([Tanh()], [30, 40], name="mlp2")
    h1 = mlp1.apply(x)
    h2 = mlp2.apply(h1)
```
So when I print ```self.parameters``` in ```model.py```:
```python
b
b
b
W
W
W
```

rizar: Thanks, I see now what the problem is. If you read the previous version of the code, you will use that hierarchical names of the parameters were used. You can get those by using `Brick.get_unique_path`.

mohammadpz: I see. So should I extract parameters' hierarchical names and then check whether they are unique or not?
If so, should I loop over bricks in ```self.top_bricks``` and find names of parameters?

rizar: Yes, you should use hierarchical names.

No, you can get the brick a parameter belongs to by exploring its tag. Then you can use the `get_unique_name` routine of this brick.

blocks/model.py
```diff
@@ -190,7 +181,18 @@ def get_params(self):
...
+                self.top_bricks).get_parameters().items()}
+        parameter_list = []
+        for parameter in VariableFilter(roles=[PARAMETER])(
+                self.shared_variables):
+            if parameter in brick_parameter_names:
+                parameter_list.append((brick_parameter_names[parameter],
+                                       parameter))
+            else:
+                parameter_list.append((parameter.name, parameter))
+        return OrderedDict(parameter_list)
```
mohammadpz: @bouthilx , I tried to fix it similar to the previous version of the code instead of using ```brick.get_unique_path()```. What do you think?

blocks/model.py
```diff
@@ -190,7 +181,18 @@ def get_params(self):
         parameters that do not belong to any brick.
 
         """
-        return self.params
+        brick_parameter_names = {
+            v: k for k, v in Selector(
+                self.top_bricks).get_parameters().items()}
+        parameter_list = []
+        for parameter in VariableFilter(roles=[PARAMETER])(
+                self.shared_variables):
```
bouthilx: I know it's from old code, but I would use ```self.parameters``` directly rather than ```VariableFilter(roles=[PARAMETER])(self.shared_variables)```.

blocks/model.py
bouthilx: It was complicated to get hierarchical path names because you had to find the bricks for each parameter right? Then rather than building the dictionary from scratch every time, I would build it in __init__ as previously but save it under _parameter_dict.

mohammadpz: It's actually the same as to find parent of a parameter and then extract hierarchical name using Selector.

bouthilx: Well it's bottom up rather than top down. You select all bricks and return all of their parameters with hierarchical path names rather than look for parameter's brick in order to compute the hierarchical path name. But I agree the result is the same. :)

bouthilx:
OK I guess we'll have to fix it in blocks-example before we merge. :) I was not aware blocks-example code is tested here.

mohammadpz:
I think the other PR is fine. So, what's next?

bouthilx:
I don't have write access to blocks and blocks-examples. We need to wait for someone to merge it. @rizar Can you please relaunch travis once you merge the PR in blocks-examples. Thanks!

rizar:
Thanks guys, I will give the PR a final look today or tomorrow.

bouthilx:
@rizar Should we temporarily add self._parameter_dict = state.params in __setstate__ for (not that) old pickled models so we can still load them. I know there is a stable branch to avoid that, but what if model is using bleeding-edge version and cannot be loaded with stable branch. User is responsible to reset to a specific commit in order to load and resave his model? Then I believe it will be important to squash the commits and make the message clear to make it easy to find and reset.

rizar:
As far I as I understand, the old will still be possible to unpickle, just that `get_parameter_dict` will not work. If a user really cares about them, he can write a script that unpickles, adds the missing attributes and pickles back.

In general our policy is that there is no warranty at all for `master` branch and we are free to do any refactoring we find useful, even if it makes the old models non entirely reusable. With such a policy we should make frequent releases, I guess.

bouthilx:
Did you relaunch travis tests?

rizar:
To be honest, I don't know how to do it (or maybe do not have access). @dwf , can you help please?

bouthilx:
You probably do not have access then. You should see a reload button on the main page of travis for this PR.

rizar:
Actually I just had to login :)

bouthilx:
Looks like it fails because the PR is merged? Do you have buildbots on lab's machine like we have for Theano and Pylearn2? I yes I'm wondering if their tests failed after merge.

rizar:
Unfortunately we do not have a build bot. I created #742 to see if tests pass.

dribnet:
> As far I as I understand, the old will still be possible to unpickle, just that get_parameter_dict will not work. If a user really cares about them, he can write a script that unpickles, adds the missing attributes and pickles back.
> In general our policy is that there is no warranty at all for master branch and we are free to do any refactoring we find useful, even if it makes the old models non entirely reusable. With such a policy we should make frequent releases, I guess.

Disagree with this - I think blocks would be well served by at least trying not to break serialized models across releases. With some sane semantic versioning backwards compatible simple shims like @bouthilx suggests can be added for minor releases and then eventually dropped on major releases. And even if `master` is a free-for-all (and trust me - it is!) - these efforts provide model compatibility for developers moving between stable releases, and this would be a big differentiator for blocks in its ecosystem.

dwf:
I tend to view serialization as one of the few areas we should take some care about breakage. Leaving unpickled objects that came from a previous in an inconsistent state is just hugely annoying, and a complete lack of versioning means that it can be hard to detect what exactly needs to be diddled to bring the object up to date. Part of the solution to this is the extremely judicious use of state to begin with; if something can be computed in a property it usually should be, and memoized if the speed hit on repeated access is significant (with the memoization cache being thrown away at pickle-time).

rizar:
@dribnet , I see your point, but on other hand with pickle being used as the serialization mechanism preserving backward-compatibility is virtually impossible even across minor versions. For instance, if a class name is changed, the pickles containing the objects of this class become unpicklable. This is the price we pay for 100%-secure  serialization. Making exceptions from the policy that pickles can only be loaded with only the same version that was used at saving would pollute the codebase and require additional manpower, which we lack.

It does not mean, that you can not load parameters of an old model though: they are now stored as fields in a zip file, which is quite a robust format.

@dwf, I agree that we should be very careful about states, and that caching, when used, should be implemented in a way that cache is not pickled/unpickled. Also an extension that strores the Blocks version in the log has long been on the todo list, see #441.

Issue #719, opened 2015-06-15
Extract initial states from computation graph in beam search.


dmitriy-serdyuk:

Thrandis:
@rizar Why the `inner_cg` doesn't have the initial_states?

rizar:
I guess they should be in the "outer"/global/normal computation graph.

Thrandis:
All rights, I'll look into it on Monday!

PR #726, opened 2015-06-18
 Moved beam_size from constructor to search fct.

### LINKED ISSUE #500: Move `beam_size` from constructor to `search` in BeamSearch.
This should be very simple to do and would add a lot of simplicity.

Thrandis:
Fixes #500

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -52,7 +52,8 @@ script:
   - |
       if [[ $TESTS == 'blocks-examples' ]]; then
         cd $TRAVIS_BUILD_DIR
-        git clone https://github.com/mila-udem/blocks-examples.git
+        #git clone https://github.com/mila-udem/blocks-examples.git
+        git clone https://github.com/Thrandis/blocks-examples.git
         cd blocks-examples
         nose2 tests
       fi
--- a/blocks/search.py
+++ b/blocks/search.py
@@ -30,8 +30,6 @@ class BeamSearch(object):
 
     Parameters
     ----------
-    beam_size : int
-        The beam size.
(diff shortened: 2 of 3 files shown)


blocks/search.py
```diff
@@ -277,6 +276,7 @@ def search(self, input_values, eol_symbol, max_length,
             negative log-likelihood.
 
         """
+        self.beam_size = input_values[input_values.keys()[0]].shape[1]
```
dmitriy-serdyuk: It's better to provide beam size as an additional argument to `compute_initial_states`, anyway you don't use it anywhere else.

.travis.yml
```diff
@@ -52,7 +52,8 @@ script:
   - |
       if [[ $TESTS == 'blocks-examples' ]]; then
         cd $TRAVIS_BUILD_DIR
-        git clone https://github.com/mila-udem/blocks-examples.git
+        #git clone https://github.com/mila-udem/blocks-examples.git
+        git clone https://github.com/Thrandis/blocks-examples.git
```
dmitriy-serdyuk: You need to clone your branch with corresponding changes (the syntax is `git clone -b my-branch git@github.com:user/myproject.git`).

blocks/search.py
rizar: I am afraid this is not a generic solution. A sequence generator can simply have no sequential inputs, consider an encoder-decoder model.

However, after #719 is solved the beam size seems to be not necessary at all!

Thrandis:
@rizar good point! I'll move to #719 then!

PR #727, opened 2015-06-19
Stabilize attention energies

orhanf:
I've managed to accumulate enough energies to make exp overflow in some case, resulting NaN propagation all the way for machine translation. This is to stabilize as its done in regular softmax.


orhanf:
Done!

PR #729, opened 2015-06-20
RecurrentStack bug w GRU and .apply docstring+contexts to positional arguments

udibr:
fix apply's docstring of RecurrentStack and add contexts to positional arguments

### DIFF
--- a/blocks/bricks/recurrent.py
+++ b/blocks/bricks/recurrent.py
@@ -772,9 +772,9 @@ class RecurrentStack(BaseRecurrent, Initializable):
     def split_suffix(name):
         # Target name with suffix to the correct layer
         name_level = name.split('_')
-        if len(name_level) == 2:
-            name, level = name_level
-            level = int(level)
+        if name_level[-1].isdigit():
+            name = '_'.join(name_level[:-1])
+            level = int(name_level[-1])
         else:
             # It must be from bottom layer
             level = 0
--- a/blocks/bricks/recurrent.py
+++ b/blocks/bricks/recurrent.py
@@ -645,6 +645,8 @@ class Bidirectional(Initializable):
     def apply_delegate(self):
         return self.children[0].apply
 
+RECURRENTSTACK_SEPARATOR = '#'
+
 
 class RecurrentStack(BaseRecurrent, Initializable):
     u"""Stack of recurrent networks.
@@ -666,9 +668,9 @@ class RecurrentStack(BaseRecurrent, Initializable):
     In order to avoid conflict, the names of the arguments appearing in
     the `states` and `outputs` attributes of the apply method of each
     layers are renamed. The names of the bottom layer are used as-is and
-    a suffix of the form '_<n>' is added to the names from other layers,
-    where '<n>' is the number of the layer starting from 1
-    (for first layer above bottom.)
+    a suffix of the form '#<n>' is added to the names from other layers,
+    where '<n>' is the number of the layer starting from 1, used for first
+    layer above bottom.
 
     The `contexts` of all layers are merged into a single list of unique
     names, and no suffix is added. Different layers with the same context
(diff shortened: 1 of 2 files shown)


rizar:
I am sorry for the delay, this week I am really lacking time on weekdays. I promise to review this  PR on weekend though.

blocks/bricks/recurrent.py
```diff
@@ -772,9 +771,9 @@ def suffixes(names, level):
     def split_suffix(name):
         # Target name with suffix to the correct layer
         name_level = name.split('_')
-        if len(name_level) == 2:
-            name, level = name_level
-            level = int(level)
+        if name_level[-1].isdigit():
```
rizar: This will also fail if the states of the core transition have names ending with digits.

I would suggest to use a more exotic separator, e.g. `$` or `#` to indicate the name mangled states.

blocks/bricks/recurrent.py
```diff
@@ -962,18 +948,18 @@ def apply(self, *args, **kwargs):
...
-        self.apply.sequences, self.apply.states, self.apply.contexts
-        All arguments values are of type :class:`~tensor.TensorVariable`.
-
-        In addition the `iterate`, `reverse`, `return_initial_states` or
-        any other argument defined in `recurrent_apply` wrapper.
+        sequences, states, contexts : :class:`~tensor.TensorVariable`
+            See docstring of the class for arguments appearing in the
+            lists
+            self.apply.sequences, self.apply.states, self.apply.contexts
+        iterate, reverse, return_initial_states : bool
```
rizar: Please use `See Also` section for such purpose, e.g.

```
See Also
------------
:func:`~blocks.brick.recurrent.recurrent` : for all other parameters
```

Also a `Notes` section could mention that `reverse` is currently not supported.

blocks/bricks/recurrent.py
```diff
@@ -962,18 +948,18 @@ def apply(self, *args, **kwargs):
...
             Use the slow, but also memory efficient, implementation of
             this code.
-
-        See docstring of the class for arguments appearing in
-        self.apply.sequences, self.apply.states, self.apply.contexts
-        All arguments values are of type :class:`~tensor.TensorVariable`.
-
-        In addition the `iterate`, `reverse`, `return_initial_states` or
-        any other argument defined in `recurrent_apply` wrapper.
+        sequences, states, contexts : :class:`~tensor.TensorVariable`
```
rizar: The `apply` method does not have arguments `sequences`, `states`, `contexts`. It has `*args` and `**kwargs` and you can explain how those are interpreted.

blocks/bricks/recurrent.py
```diff
@@ -772,10 +773,10 @@ def suffixes(names, level):
     @staticmethod
     def split_suffix(name):
         # Target name with suffix to the correct layer
-        name_level = name.split('_')
-        if len(name_level) == 2:
-            name, level = name_level
-            level = int(level)
+        name_level = name.split(RECURRENTSTACK_SEPARATOR)
+        if name_level[-1].isdigit():
```
rizar: What if a stack element has a state with a name ending with a digit? If would rather change this condition to `if len(name_level) == 2`. You can also check that `len(name_level) < 3` and raise `ValueError` if it is not the case.

blocks/bricks/recurrent.py
```diff
@@ -963,18 +951,26 @@ def apply(self, *args, **kwargs):
...
             this code.
-
-        See docstring of the class for arguments appearing in
-        self.apply.sequences, self.apply.states, self.apply.contexts
-        All arguments values are of type :class:`~tensor.TensorVariable`.
-
-        In addition the `iterate`, `reverse`, `return_initial_states` or
-        any other argument defined in `recurrent_apply` wrapper.
+        *args : :class:`~tensor.TensorVariable`
+            Positional argumentes in the order in which they appear in
```
rizar: Please mention that they are all optional.

rizar:
Please address my last comments and I will merge this PR.

rizar:
Thanks! By the way the `initial_state` method has to be rewritten, but this is outside of this PR's scope.

PR #730, opened 2015-06-23
Use numbers.Integral to check for integer-ness.

dwf:
Reminded by mila-udem/fuel#179. You should use `numbers.Integral` when it's absolutely necessary to check types, because it catches

* `int` and `long` on Python 2, or just `int` on Python 3
* All subclasses of `numpy.integer` (`numpy.int8`, `numpy.int16`, `numpy.int32`, `numpy.int64`, `numpy.uint8`, `numpy.uint16`, `numpy.uint32`, `numpy.uint64`) that you might get when iterating over a numpy array of integers.

### DIFF
--- a/blocks/initialization.py
+++ b/blocks/initialization.py
@@ -1,8 +1,8 @@
 """Objects for encapsulating parameter initialization strategies."""
 from abc import ABCMeta, abstractmethod
+import numbers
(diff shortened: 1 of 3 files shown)


rizar:
Good to know!

PR #732, opened 2015-06-25
Extract initial states from cg graph.

### LINKED ISSUE #719: Extract initial states from computation graph in beam search.

Thrandis:
Will fix #719 and #500

### DIFF
--- a/blocks/search.py
+++ b/blocks/search.py
@@ -150,9 +150,9 @@ class BeamSearch(object):
         outputs = self.initial_state_and_context_computer(
             *[inputs[var] for var in self.inputs])
         contexts = OrderedDict((n, outputs.pop(n)) for n in self.context_names)
-        self.beam_size = outputs.pop('beam_size')
+        beam_size = outputs.pop('beam_size')
         initial_states = outputs
-        return contexts, initial_states
+        return contexts, initial_states, beam_size
 
     def compute_logprobs(self, contexts, states):
         """Compute log probabilities of all possible outputs.
@@ -265,7 +265,7 @@ class BeamSearch(object):
         if not self.compiled:
             self.compile()
 
-        contexts, states = self.compute_initial_states_and_contexts(
+        contexts, states, beam_size = self.compute_initial_states_and_contexts(
             input_values)
 
         # This array will store all generated outputs, including those from
@@ -290,7 +290,7 @@ class BeamSearch(object):
             # The `i == 0` is required because at the first step the beam
             # size is effectively only 1.
             (indexes, outputs), chosen_costs = self._smallest(
-                next_costs, self.beam_size, only_first_row=i == 0)
+                next_costs, beam_size, only_first_row=i == 0)
 
             # Rearrange everything
             for name in states:
--- a/blocks/search.py
+++ b/blocks/search.py
@@ -10,6 +10,7 @@ from blocks.bricks.sequence_generators import BaseSequenceGenerator
 from blocks.filter import VariableFilter, get_application_call, get_brick
 from blocks.graph import ComputationGraph
 from blocks.roles import INPUT, OUTPUT
+from blocks.utils import unpack
 
 
 class BeamSearch(object):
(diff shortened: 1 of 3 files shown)


dmitriy-serdyuk:
@rizar , we have a problem, that Theano cannot infer shape (which is needed for batch size) of `initial_states` using only contexts. It cannot bypass the inputs. 

There are two ways to solve this problem: have a function which computes the batch size and an additional input for `initial_state_computer` or use `givens` argument to compile the function and provide the batch size of contexts.

Second isn't not good because we might have weird contexts (like for caption generation) and we cannot rely on that dimension 1 is batch size. First one isn't very neat neither. What do you think?

blocks/search.py
```diff
@@ -74,7 +70,7 @@ def __init__(self, beam_size, samples):
         self.contexts = [
             VariableFilter(bricks=[self.generator],
                            name=name,
-                           roles=[INPUT])(self.inner_cg)[0]
+                           roles=[INPUT])(self.CG)[0]
```
rizar: `CG`?

rizar:
But why can't you provide the inputs to `initial_state_computer`?

dmitriy-serdyuk:
Oh, right. But seems like a waste of computation, especially if the input in large like in image or video caption generation (probably, I don't know much about it).

rizar:
Why is it a waste of computation? We have to somehow compute initial states anyway, don't we?

dmitriy-serdyuk:
Yes, but we could've used already precomputed contexts. Anyway, we compute it only once and all we need is a batch size.

rizar:
I am not sure I understand you. We do need initial states to do beam search. When we have them (as values), we can get the batch size, since it must be the first dimension. It is true that it is desirable to use already computed contexts to compute the initial states, but this can be achieved by computing both in one Theano function (or using `givens`).

dmitriy-serdyuk:
I'm trying to say, that theoretically initial states should depend only on contexts and not on inputs. But we don't want to rely on that batch size is second dimension of contexts.

So, I think we just create one function for initial state and contexts.

rizar:
> I'm trying to say, that theoretically initial states should depend only on contexts and not on inputs. But we don't want to rely on that batch size is second dimension of contexts.

Agreed

> So, I think we just create one function for initial state and contexts.

Sounds like a good plan.

rizar:
Sorry, pressed a wrong button.

dmitriy-serdyuk:
@Thrandis can you add a note, that the beam size is inferred from contexts?

dmitriy-serdyuk:
Oh, it's already there.

blocks/search.py
```diff
@@ -90,17 +86,19 @@ def __init__(self, beam_size, samples):
...
-            self.contexts, initial_states, on_unused_input='ignore')
+    def _compile_initial_state_and_context_computer(self):
+        initial_states = VariableFilter(
+                            applications=[self.generator.initial_states],
+                            roles=[OUTPUT])(self.cg)
+        outputs = OrderedDict([(v.tag.name, v) for v in initial_states])
+        beam_size = VariableFilter(
+                            applications=[self.generator.initial_states],
+                            name='batch_size')(self.cg)[0]
+        for i, name in enumerate(self.context_names):
```
dmitriy-serdyuk: You can iterate over `equizip(context_names, contexts)` here.

dmitriy-serdyuk:
Looks good, I think it's ready for merge. Strange that coverage decreased.

blocks/search.py
```diff
@@ -90,17 +87,19 @@ def __init__(self, beam_size, samples):
...
-                self.beam_size, as_dict=True,
-                **dict(equizip(self.context_names, self.contexts)))
-        self.initial_state_computer = function(
-            self.contexts, initial_states, on_unused_input='ignore')
+    def _compile_initial_state_and_context_computer(self):
+        initial_states = VariableFilter(
+                            applications=[self.generator.initial_states],
+                            roles=[OUTPUT])(self.cg)
+        outputs = OrderedDict([(v.tag.name, v) for v in initial_states])
+        beam_size = unpack(VariableFilter(
```
rizar: Elegant!

blocks/search.py
```diff
@@ -143,29 +141,18 @@ def compute_contexts(self, inputs):
...
+        contexts ordered like `self.context_names` and a
+        {name: :class:`numpy.ndarray`} dictionary of states ordered like
         `self.state_names`.
 
         """
-        return self.initial_state_computer(*list(contexts.values()))
+        outputs = self.initial_state_and_context_computer(
+            *[inputs[var] for var in self.inputs])
+        contexts = OrderedDict((n, outputs.pop(n)) for n in self.context_names)
+        self.beam_size = outputs.pop('beam_size')
```
rizar: Please do not use class attributes for returning results from a method. You can add `beam_size` to the return statement.

rizar:
Nice! Please change the way the beam size is returned and that should be it.

Thrandis:
@rizar done :-)

rizar:
Awesome, great job!

PR #736, opened 2015-06-26
Add consider_constant support to GradientDescent.

dwf:


dwf:
@rizar @vdumoulin @dmitriy-serdyuk I think this is fairly uncontroversial, and the drop in test coverage is negligible. Let me know if this LGTY.

Issue #526, opened 2015-03-23
Various small things in ComputationGraph.replace 


rizar:
- print a warning if the user tries to replace an auxiliary variables

- print a warning if the user tries to replace a variable that is not contained in the graph

- get rid of the topological sort and use already sorted `self.variables` instead

Makes an beginner-level CCW issue, I guess.

bouthilx:
@rizar Why not raising an error when the variable is not contained in the graph?

mohammadpz:
@bouthilx , Could you please take a look at my PR?

rizar:
@bouthilx , perhaps an error would make more sense indeed.

bouthilx:
@rizar @dmitriy-serdyuk  Can you close the issue please. :)

rizar:
Sure, thanks!

I would really like to give your merge rights, but I don't have rights to give rights, unfortunately :(

lamblin:
I added @bouthilx as an admin and @rizar as an owner.

PR #739, opened 2015-06-30
Fixing computation graph replace

mohammadpz:
Fixing #526 .

### DIFF
--- a/blocks/graph.py
+++ b/blocks/graph.py
@@ -17,6 +17,7 @@ from blocks.roles import (add_role, has_roles, AUXILIARY, PARAMETER, DROPOUT,
                           COLLECTED, COLLECTOR)
 from blocks.utils import (is_graph_input, is_shared_variable, dict_union,
                           shared_floatx_zeros, shared_like)
+import warnings
 
 logger = logging.getLogger(__name__)
 
@@ -196,8 +197,6 @@ class ComputationGraph(object):
         # Due to theano specifics we have to make one replacement in time
         replacements = OrderedDict(replacements)
 
-        apply_nodes_sorted = theano.gof.graph.io_toposort(
-            self.inputs, self.outputs)
         outputs_cur = self.outputs
 
         # `replacements` with previous replacements applied. We have to track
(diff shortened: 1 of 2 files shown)


blocks/graph.py
```diff
@@ -196,8 +197,14 @@ def replace(self, replacements):
         # Due to theano specifics we have to make one replacement in time
         replacements = OrderedDict(replacements)
 
-        apply_nodes_sorted = theano.gof.graph.io_toposort(
-            self.inputs, self.outputs)
+        for variable in replacements:
+            if variable not in self.variables:
+                warnings.warn("replace method was asked to replace a variable "
+                              "that is not a part of the computational graph.")
```
bouthilx: Would be nice to have the name of the variable if available.

blocks/graph.py
```diff
@@ -196,8 +197,14 @@ def replace(self, replacements):
...
+            if variable in self.auxiliary_variables:
```
bouthilx: has_roles(variable, [AUXILIARY]) will avoid searching in a list. Change if to elif to skip variables that are not part of the graph (we don't care if it's auxiliary, it's not in the graph).

blocks/graph.py
```diff
@@ -205,18 +212,11 @@ def replace(self, replacements):
...
-                        replacement_vals_cur.append(replacements[input_])
-
-        # Add outputs of the computation graph
-        for output in self.outputs:
-            if output in replacements:
-                replacement_keys_cur.append(output)
-                replacement_vals_cur.append(replacements[output])
+        # variables in self.variables are in topological order
+        for variable in self.variables:
+            if variable in replacements:
```
bouthilx: You could move you first loop here. Make a copy of replacements and pop items when found. If it's not empty after you print a warning for each missing variable.

blocks/graph.py
```diff
@@ -205,18 +204,24 @@ def replace(self, replacements):
...
+                replacement_keys_cur.append(variable)
+                replacement_vals_cur.append(
+                    remaining_replacements.pop(variable))
+
+        # if remaining_replacements is not empty
+        if remaining_replacements:
+            warnings.warn(
+                "replace method was asked to replace a variable(s) ({}) "
+                "that is not a part of the computational "
+                "graph.".format("".join(str(remaining_replacements.keys()))))
```
mohammadpz: @bouthilx ,
Is this the right way of doing this?

bouthilx: @mohammadpz join works on a list. If you do str(something) you get a string. 

```python
>>> "".join(str(range(10)))
'[0, 1, 2, 3, 4, 5, 6, 7, 8, 9]'
```
To help understand more, if you try to separate with comas
```python
>>> ", ".join(str(range(10)))
'[, 0, ,,  , 1, ,,  , 2, ,,  , 3, ,,  , 4, ,,  , 5, ,,  , 6, ,,  , 7, ,,  , 8, ,,  , 9, ]'
```
What you probably want is more something like
```python
>>> ", ".join(str(range(10)))
'0, 1, 2, 3, 4, 5, 6, 7, 8, 9'
```

But anyway, I think simply ```format(str(remaining_replacements.keys())))``` can be good enough.

blocks/graph.py
```diff
@@ -205,18 +204,24 @@ def replace(self, replacements):
...
-                replacement_vals_cur.append(replacements[output])
+        # variables in self.variables are in topological order
+        remaining_replacements = replacements.copy()
+        for variable in self.variables:
+            if variable in replacements:
+                if has_roles(variable, [AUXILIARY]):
+                    warnings.warn(
+                        "replace method was asked to replace a variable ({}) "
+                        "that is an auxiliary variable.".format(variable))
+                replacement_keys_cur.append(variable)
```
bouthilx: @mohammadpz maybe you could add a comment here saying that self.variables should not contain duplicates thus pop() should not fail. It will help to debug if someone breaks self.variables uniqueness property.

mohammadpz: @bouthilx , here I'm calling pop() on ```remaining_replacements```. So I should write comment about ```remaining_replacements```. Right?

bouthilx: Well, self.variables should not contain duplicates thus remaining_replacements.pop(variable) should not fail, otherwise it means the variable was already seen and popped out.

bouthilx:
@rizar @dwf @dmitriy-serdyuk  Can someone merge this PR please. I don't have write access.

dmitriy-serdyuk:
LGTM.

bouthilx:
Ah! Thank you! :)

PR #747, opened 2015-07-02
added bokeh, pandas and blocks-extras to .travis.yml #739

vikkamath:
MNIST example uses bokeh for plotting. Added bokeh, pandas and blocks-extras to .travis.yml. Also started bokeh-server. In response to https://github.com/mila-udem/blocks-examples/pull/19#issuecomment-118113170

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -33,12 +33,16 @@ install:
   # Install all Python dependencies
   - conda install -q --yes python=$TRAVIS_PYTHON_VERSION mkl --file req-travis-conda.txt
   - pip install -q -r req-travis-pip.txt
+  - pip install -q git+git://github.com/mila-udem/blocks-extras.git 
+  - conda install -q --yes bokeh
+  - conda install -q --yes pandas
 script:
   - pip install -e . -r requirements.txt # Tests setup.py
   - curl https://raw.githubusercontent.com/mila-udem/fuel/master/.travis-data.sh | bash -s -- mnist
   - # Must export environment variable so that the subprocess is aware of it
   - export THEANO_FLAGS=floatX=$FLOATX,optimizer=fast_compile
   - export FUEL_FLOATX=$FLOATX
+  - bokeh-server & #start bokeh-server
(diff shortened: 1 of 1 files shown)


rizar:
I would not call it a cyclic dependency, it's just that Blocks-extras is not a dependency for Blocks, but is a dependency of Blocks-examples, which is used to test Blocks. Can you try to install Blocks-extras after Blocks? This should do the job.

.travis.yml
```diff
@@ -33,12 +33,16 @@ install:
   # Install all Python dependencies
   - conda install -q --yes python=$TRAVIS_PYTHON_VERSION mkl --file req-travis-conda.txt
   - pip install -q -r req-travis-pip.txt
+  - pip install -q git+git://github.com/mila-udem/blocks-extras.git 
+  - conda install -q --yes bokeh
+  - conda install -q --yes pandas
```
rizar: By the way, why do you have to install Pandas here?

.travis.yml
```diff
@@ -33,12 +33,16 @@ install:
   # Install all Python dependencies
   - conda install -q --yes python=$TRAVIS_PYTHON_VERSION mkl --file req-travis-conda.txt
   - pip install -q -r req-travis-pip.txt
+  - pip install -q git+git://github.com/mila-udem/blocks-extras.git
```
rizar: There seeems to be a syntax to install extra dependencies of a package, e.g. see the command `git+git@github.com:USER/blocks.git#egg=blocks[test,docs] ` from [Blocks installation tutorial](http://blocks.readthedocs.org/en/latest/setup.html). It could be used to make sure that `bokeh` is installed automatically with `blocks-extras`.

.travis.yml
vikkamath: From what I remember from doing something similar on blocks-examples, the travis build fails without pandas. I think it has something to do with bokeh but I'll look into it again in a few hours.

.travis.yml
vikkamath: Thanks, I'll look into it.

PR #748, opened 2015-07-04
Trying to fix #739

rizar:


rizar:
Damn it, how's that even possible: when I install Blocks-extras after installing Blocks, the Python 3 test for Blocks crashes because somehow Theano gets installed wrongly!

@dwf , @vdumoulin , maybe you guys have an idea what's going now? I will continue digging tomorrow.

vikkamath:
Looks like the build is failing because of the implicit call to [Theano's 'scan'](https://github.com/Theano/Theano/blob/master/theano/scan_module/scan_op.py#L1286) module which in turn uses Python2's itertools.izip which has been made unusable (use of zip is [recommended](http://www.diveintopython3.net/porting-code-to-python-3-with-2to3.html#itertools)) instead.

rizar:
Right, but I guess it is adressed by some source code conversion run during Theano installation. And now the question is, why this conversion proceeds correctly only when I don't install Blocks-extras.

vikkamath:
So I forked Theano, [replaced the itertools.izip by theano.compat.izip](https://github.com/Theano/Theano/compare/master...vikkamath:master) and added my copy of it to the requirements file, using which the tests for blocks-examples seems to pass. I still can't figure out the source of the problem, but I'm looking into it.

dwf:
Theano was recently updated to use a single source code base for Python 2
and 3 (I started this and @abergeron graciously helped me finish it).  If
there are still Python 2 isms in master then that is a Theano bug (and one
that should have been caught by Travis or the buildbot),  please submit a
PR and a test.

vikkamath:
If these travis builds succeed, can we safely ascertain that it is indeed an upstream bug?

dwf:
I think we can, yes.

rizar:
But why have the tests been passing without `blocks-extras` then? E.g. they pass now, after I  merged https://github.com/mila-udem/blocks-examples/pull/24

vikkamath:
So it looks like the Travis tests have all passed. Ref PR #749

vikkamath:
#749 passes all checks. But I recommend that we keep this open until we can figure out the lines that caused the call of itertools.izip (as opposed to theano.compat.izip) in Theano, so we could help Theano devs to write suitable tests against it.

PR #749, opened 2015-07-05
attempt to fix #739. Ref #748

vikkamath:
UPDATE: A issue that was holding up the merging of this PR has now been fixed upstream (Theano). Unless somebody disagrees, I think it's safe to say that this PR is good to go. 

NOTE: Please do not merge until we figure out the source of the problem. If the problem is indeed upstream (Theano), please wait till the issue is resolved there first, after which I will submit another PR.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -35,7 +35,6 @@ install:
   - pip install -q -r req-travis-pip.txt
 script:
   - pip install -e . -r requirements.txt # Tests setup.py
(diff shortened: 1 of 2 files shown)


rizar:
@vikkamath , I changed the MNIST demo to use Bokeh iff blocks-extras is available. Are you find with that? If yes I will close this PR.

vikkamath:
@rizar Yes, I think that's the solution that makes most sense. Thanks :)

Issue #745, opened 2015-07-02
RecurrentStack initial_states is broken


rizar:
It was skipped in the recent refactoring #718.

Also name mangling can be improved.

@janchorowski, I assign this to you, as you asked.

rizar:
For some reason I couldn't use the "Assignee" button on the right...

PR #751, opened 2015-07-07
fix issue #745

udibr:
I know I am not assigned to it but if you like the fix then I guess its OK

### DIFF
--- a/blocks/bricks/recurrent.py
+++ b/blocks/bricks/recurrent.py
@@ -850,6 +850,8 @@ class RecurrentStack(BaseRecurrent, Initializable):
             setattr(self.low_memory_apply, property_,
                     getattr(self.apply, property_))
 
+        self.initial_states.outputs = self.apply.states
+
     def normal_inputs(self, level):
         return [name for name in self.transitions[level].apply.sequences
                 if name != 'mask']
@@ -991,8 +993,9 @@ class RecurrentStack(BaseRecurrent, Initializable):
         return transition.get_dim(name)
 
     @application
-    def initial_state(self, state_name, batch_size, *args, **kwargs):
-        state_name, level = self.split_suffix(state_name)
-        transition = self.transitions[level]
-        return transition.initial_state(state_name, batch_size,
-                                        *args, **kwargs)
+    def initial_states(self, batch_size, *args, **kwargs):
+        results = []
+        for transition in self.transitions:
+            results += transition.initial_states(batch_size, *args,
+                                                 as_list=True, **kwargs)
+        return results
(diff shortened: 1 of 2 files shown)


blocks/bricks/recurrent.py
```diff
@@ -991,8 +993,9 @@ def get_dim(self, name):
         return transition.get_dim(name)
 
     @application
-    def initial_state(self, state_name, batch_size, *args, **kwargs):
-        state_name, level = self.split_suffix(state_name)
-        transition = self.transitions[level]
-        return transition.initial_state(state_name, batch_size,
-                                        *args, **kwargs)
+    def initial_states(self, batch_size, *args, **kwargs):
+        kwargs['as_list'] = True
```
rizar: I don't think this is necessary.

blocks/bricks/recurrent.py
```diff
@@ -991,8 +993,9 @@ def get_dim(self, name):
...
+        results = []
+        for transition in self.transitions:
+            results += transition.initial_states(batch_size, *args, **kwargs)
```
rizar: You have to mangle the names, I guess.

rizar:
Your help is definitely appreciated!

blocks/bricks/recurrent.py
udibr: Very much necessary. Without it the wrapper of transition.initial_states returns a tensor which you can't add to results in the way I did

blocks/bricks/recurrent.py
udibr: You will have to detail what do you mean by mangling in this case.

rizar: Ok, I see how it works now, my bad. It would be a bit clearer to just add `as_list=True` to the function call after `batch_size`.

rizar:
It looks good, but I would also like to ask for a regression test. You can test this change in the same way I did with `Recurrent`, `GatedRecurrent` and `LSTM`: check that the initial states are indeed shared variables in the computation graph returned by the `apply` method.

udibr:
I've added the test you asked for but then I decided I am confused what init_states is used for.
The way the test works (with VariableFilter) gives you indeed the internal variables and they are indeed shared (as parameters should be) however they had the original names so in the case of a stack the same names appeared several times (when using a depth>1 of the same tranistion class.)

However, init_states returns a tensort expand (repeated to batch size) from the shared variables and as a result they are regualr tensors and not a shared variables. As such they can only be used for making predictions over a batch and not for training. However, this time because of the output names they each have the correct different name (with a #<number> suffix to differentiate the different layers.)

udibr:
P.S. I added test code so why am I seeing a decrease in coverage? (I guess this is not my fault.)

rizar:
Your test is okay, I don't know why coverage complain. Thanks for the PR!

PR #757, opened 2015-07-09
Variable filter add name

EloiZ:
Add two options for VariableFilter:
theano_name
theano_name_regex

Filtering with "name" tries to match var.tag.name with "name"
Filtering with "theano_name" tries to match var.name with "theano_name"

The behavior of the theano_name_regex is analog.

### DIFF
--- a/blocks/filter.py
+++ b/blocks/filter.py
@@ -93,7 +93,8 @@ class VariableFilter(object):
 
     """
     def __init__(self, roles=None, bricks=None, each_role=False, name=None,
-                 name_regex=None, applications=None):
+                 name_regex=None, theano_name=None, theano_name_regex=None,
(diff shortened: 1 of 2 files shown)


blocks/filter.py
```diff
@@ -93,7 +93,8 @@ class VariableFilter(object):
 
     """
     def __init__(self, roles=None, bricks=None, each_role=False, name=None,
-                 name_regex=None, applications=None):
```
rizar: Please add docstrings for new parameters.

EloiZ: Indeed, I knew I was forgetting somthing... :)
Is it OK now?

rizar:
Cool PR, thanks :)

Issue #653, opened 2015-05-21
Cost for recurrent networks


rizar:
As it was pointed out [here](https://groups.google.com/forum/#!topic/blocks-users/O-S45G6tpNY), our cost bricks can not handle masks. `SoftmaxEmitter` has an implementation of such cost, but only for one particular case.

The following sequence of changes could fix it:

- `As2D` should be able to handle multiple inputs 

- we actually switch for scalars to vector of size 1 for the integer targets

- `Softmax.categorical_crossentropy` should not compute `mean` inside. This one is the easiest to fix, I guess.

Issue #753, opened 2015-07-08
Beam search should not use `probs`


rizar:
https://github.com/mila-udem/blocks/blob/master/blocks/search.py#L119

It sucks for two reasons:

- unnecessary exponentiation causing numerical overflows

- ties BeamSearch tightly to SoftmaxReadout interface

Issue #659, opened 2015-05-22
Softmax.apply should not use theano.tensor.nnet.softmax


rizar:
...because the latter is numerically unstable, see discussion in #654

mjwillson:
Posted this in #654 but noticed that was closed, and it might be relevant here too:

I noticed in theano itself there is theano.tensor.nnet.categorical_crossentropy. For the one-hot case of this uses CrossentropyCategorical1Hot, which seems to have a grad implemented, and its docs state:

    :note: In the case that the coding distribution is the output of a
           softmax, an application of this Op will probably be optimized
           away in favour of one with a C implementation.

While I don't like the sound of "probably", is there a reason not to use this on top of theano.tensor.nnet.softmax in Softmax.categorical_crossentopy ? Is there an easy way I can find out what optimisations it's made under the hood?

It'd be good to clear this stuff up anyway, either here or in theano. I'm not enough of an expert to debug it authoritatively, but for me using blocks' Softmax.categorical_crossentopy seems to cause theano to do something suspiciously slow / inefficient (turning on profiling, around 2/3 of the time is spent in Softmax and Elemwise exp, which seems way too much...)

PR #759, opened 2015-07-10
Categorical cross entropy refactoring

rizar:
This is a refactoring of `Softmax` which I would like to do in the view of issues #753  and #653.
I thought it can be merged first as it is a pretty standalone one.


rizar:
I continued to work on this PR and now it includes some more. 

- `categorical_cross_entropy` now returns a vector instead of a scalar. I believe it makes its interface more consistent and the logic contained much more reusable. This is will be quite a noticeable interface change since many people are using `categorical_cross_entropy`, but completely safe one because old code will simply start crashing

- I refactored `As2D` (now called `WithExtraDims`) to support multiple inputs of different dimensions. This allows it to be used in `SoftmaxEmitter`, radically simplifying the code of the latter and also making it numerically stable.

Fixes #653 and #659, both being quite nasty issues.

I am kindly asking attention from @dmitriy-serdyuk (for changes in `SoftmaxEmitter`), @vdumoulin (as the author of `As2D`). Please let me know if you are able to review this in near future, guys.

rizar:
Fixes #770.

rizar:
@janchorowski , maybe you can review this PR? It is by the way quite important for speech recognition because it makes `SoftmaxEmitter` numerically stable.

blocks/bricks/__init__.py
```diff
@@ -499,35 +499,60 @@ class Softmax(Activation):
...
+            distribution.
+
+        Returns
+        -------
+        output : :class:`~theano.Variable`
+            A matrix with normalized log-probabilities for each
+            distribution from `input_`.
+
+        """
+        shifted = input_ - input_.max(axis=1).dimshuffle(0, 'x')
```
janchorowski: use keepdims=True

blocks/bricks/__init__.py
```diff
@@ -499,35 +499,60 @@ class Softmax(Activation):
...
-            The cross entropy between y and x.
+            A vector of cross-entropies between distributions y and x.
 
         """
-        x = x - x.max(axis=1).dimshuffle(0, 'x')
-        log_prob = x - tensor.log(tensor.exp(x).sum(axis=1).dimshuffle(0, 'x'))
+        x = self.log_probabilities(x)
         if y.ndim == x.ndim - 1:
-            flat_log_prob = log_prob.flatten()
+            flat_log_prob = x.flatten()
```
janchorowski: you can use 
```
log_prob[tensor.arange(y.shape[0]), y]
```

blocks/bricks/__init__.py
```diff
@@ -499,35 +499,60 @@ class Softmax(Activation):
...
         Returns
         -------
         cost : :class:`~tensor.TensorVariable`
-            The cross entropy between y and x.
+            A vector of cross-entropies between distributions y and x.
 
         """
-        x = x - x.max(axis=1).dimshuffle(0, 'x')
-        log_prob = x - tensor.log(tensor.exp(x).sum(axis=1).dimshuffle(0, 'x'))
+        x = self.log_probabilities(x)
```
janchorowski: log_probs are a useful by-product. Maybe add an auxiliary output var?

blocks/bricks/sequence_generators.py
```diff
@@ -664,14 +665,15 @@ class SoftmaxEmitter(AbstractEmitter, Initializable, Random):
 
     """
     def __init__(self, initial_output=0, **kwargs):
-        self.initial_output = initial_output
         super(SoftmaxEmitter, self).__init__(**kwargs)
+        self.initial_output = initial_output
+        self.softmax = WithExtraDims(Softmax().apply)
```
janchorowski: I'm not sure I like this. Why do we create many bricks for conceptually the same element? Maybe it is better for SoftMax to take an `axis` argument along which it works, with the last one as the default?

On a sidenote: Softmax shouldn't inherit from Activation, because it doesn't work element-wise. It acts axis-wise.

I coded this a while ago, so to  save on work:
```python
class AxisSoftmax(Brick):
    """
    Applies the Softmax function along a given axis (by default the last one)
    """
    def __init__(self, axis=-1, **kwargs):
        super(AxisSoftmax, self).__init__(**kwargs)
        self.axis = axis

    @application(inputs=['input_'], outputs=['output'])
    def apply(self, input_):
        input_ = input_ - input_.max(axis=self.axis, keepdims=True)
        ex = tensor.exp(input_)
        return ex / ex.sum(axis=self.axis, keepdins=True)

    @application(inputs=['input_'], outputs=['output'])
    def log_probs(self, input_):
        input_ = input_ - input_.max(axis=self.axis, keepdims=True)
        return input_ - tensor.log(tensor.exp(input_).sum(axis=self.axis,
                                                          keepdims=True))

    @application
    def categorical_cross_entropy(self, application_call,
                                  targets, input_,
                                  target_masks=None):
        """Return computationally stable softmax cost.

        Parameters
        ----------
        targets : :class:`~tensor.TensorVariable`
            In the case of a matrix argument, each slice along
            axis represents one distribution. In the vector case, each
            element represents the position of the '1' in a one hot-vector.
        input_ : :class:`~tensor.TensorVariable`
            Each slice along axis represents energies of a distribution,
            that is pre-softmax values.
        target_masks : None or :class:`~tensor.TensorVariable`
            If not None, the target_masks indicate which elemts to include
            in the cost
        average : bool
            Whether to compute the mean cost, or return individual costs

        Returns
        -------
        cost : :class:`~tensor.TensorVariable`
            The sum of the cross entropy between targets and Softmax(input_).

        """
        log_probs = self.log_probs(input_)

        application_call.add_auxiliary_variable(
            log_probs.copy(), name='log_probs')

        if targets.ndim == log_probs.ndim - 1:
            # move the selected axis to be the last!
            if self.axis != -1 and self.axis != targets.ndim:
                axis_order = range(log_probs.ndim)
                axnum = axis_order.pop(self.axis)
                axis_order.append(axnum)
                _log_prob = log_probs.dimshuffle(*axis_order)
            else:
                _log_prob = log_probs
            _log_prob = _log_prob.reshape((_log_prob.shape[:-1].prod(),
                                           _log_prob.shape[-1]))
            _targets = targets.ravel()
            cost = -_log_prob[tensor.arange(_targets.shape[0]),
                              _targets]
            cost = cost.reshape(targets.shape)
        elif targets.ndim == log_probs.ndim:
            cost = -(log_probs * targets).sum(axis=self.axis)
        else:
            raise TypeError('rank mismatch between x and y')
        if target_masks:
            cost = cost * target_masks
        return cost.sum()

    @application(inputs=['input_'], outputs=['output'])
    def classify(self, input_):
        return input_.argmax(axis=self.axis)

    @application
    def misclasification_count(self, application_call,
                         targets, input_,
                         target_masks=None):
        """Return computationally stable softmax cost.

        Parameters
        ----------
        targets : :class:`~tensor.TensorVariable`
           Matrix of integers encoding the proper class in a 1-hot fashion
        input_ : :class:`~tensor.TensorVariable`
            Each slice along axis represents energies of a distribution,
            that is pre-softmax values.
        target_masks : None or :class:`~tensor.TensorVariable`
            If not None, the target_masks indicate which elemts to include in
            the cost
        average : bool
            Whether to compute the mean cost, or return individual costs

        Returns
        -------
        cost : :class:`~tensor.TensorVariable`
            The sum of the cross entropy between targets and Softmax(input_).

        """
        predictions = self.classify(input_)

        application_call.add_auxiliary_variable(
            predictions.copy(), name='predictions')

        if targets.ndim == predictions.ndim:
            errs = tensor.neq(targets, predictions)
        else:
            raise TypeError('rank mismatch between x and y')
        if target_masks:
            errs = errs * target_masks
        return errs.sum().astype(theano.config.floatX)
```

blocks/bricks/wrappers.py
```diff
@@ -2,40 +2,91 @@
 from theano import tensor
 
 from blocks.bricks.base import Brick, application
+from blocks.utils import dict_subset, pack
 
 
-class As2D(Brick):
-    """Wraps a 2D application around reshapes to accomodate N-D tensors.
+class WithExtraDims(Brick):
```
janchorowski: I have mixed feelings about this brick. First, What are the other use-cases besides Softmax, which can act axis-wide with minor code changes? Second, the assumption that all input arguments will need similar flattening treatment (and that it will work) seems limiting. Third, why wrap just the application method? This forces the instantiation of many bricks, as you do with 2 softmaxes above? Maybe it is better to go the full length - walk all methods of the wrapped brick, and wrap all applications?

Brief, I would add axes to softmax and not use WithExtraDims.

janchorowski:
I like the changes to SoftMax, I would suggest adding a misclassification cost to it similarly to the cross-entropy, since this is where I would look for it.

I'd get rid of the WithExtraDims brick and make softmax genuinly n-dimensional, the code is nearly the same.

blocks/bricks/wrappers.py
rizar: First, these are not just minor changes you had to make in `Softmax` to make it handle the time dimension. Its code becomes considerably more involved, and people not needing this feature will have harder to time to understand it. Still most people work with FF nets. 

Second, there are more potential usecases for this brick than just softmax. It can wrap any operation that can act on batches of tuples, a vector and an integer scalar in the softmax case, for which writing an implementation unaware of extra axis is just simpler. But for `tensor.dot` magical ability to handle extra-axis, `Linear` and `MLP` would be the first consumers for `WithExtraDims`. Many but not all Theano expressions are as agnostic to additional axis, e.g. [fft](https://github.com/Theano/Theano/blob/master/theano/sandbox/fourier.py#L96) does not seem to be the case.

I agree that it is tempting to wrap the whole brick instead of only application method. This does not seem to be possible codewise though. We can't add application methods to a brick at instantiation time, for one. Dispatching the application method to be called by its name would be quite ugly, since I don't like application methods that do very different things depending on what their parameters are. I guess this is one of those case when we have to tolerate that one brick with multiple application method will be references as a child by a set of bricks wrapping each of these methods.

It is true that `WithExtraDims` puts certain restrictions on the inputs and outputs, but these correspond to the major case when we just have a batch of data and want to act element-wise, but the indices in the batch are 2+ dimensional.

blocks/bricks/__init__.py
rizar: That's really cool!

rizar:
Your local remarks has been addressed, on the global level I still think that keeping `Softmax` very small and help it with `WithExtraDims` is the way to go, see a  very long comment in the code.

rizar:
Regarding misclassification cost: I am not sure that `Softmax` is the place for it. Because it is not specific to softmax, if you learn to emit 1 for the right class and 0 for the wrong ones, the logic for misclassification rate would be the same.

blocks/bricks/wrappers.py
janchorowski: Hi, 

I guess we are hitting two extremes - the code I pasted has a mask argument which is not needed if you allow Softmax to return arrays. If you get rid of the masks and just let it operate on a chosen mask it will become simpler. In fact in most places you replace axis=1 by axis=self.axis. The only non-trivial change is the dimshuffle for categorical cross-entropy.

I needed the axis awareness because I implemented an FF conv net for speech.

Also, the current implementation of SoftMax as an activation is simply misleading - the documentation says that it is element-wise, which is not true and you actually have to read the code to know on which axis it operates. So the implementation is long, because I had to copy the documentation.

Then, blocks should supply the users with good implementations of common use-cases. If even we need to handle n-dimensional arrays in softmax why not simply add it?

blocks/bricks/__init__.py
bartvm: Note that this requires advanced indexing, which was never supported on GPU (maybe it has been added lately?). This means that using `tensor.arange` like this forced the tensor to be transferred to CPU. The code was written the way it was to prevent that, see also e.g. https://github.com/lisa-lab/pylearn2/blob/master/pylearn2/models/mlp.py#L1422-L1440.

janchorowski: The one you point to also relies on non-GPU arange and advanced indexing.

Then I was recently profiling networks in Blocks and the performance 
killer is CAREduce which does e.g. gradient norm monitoring. It is very 
slow in Theano and simple monitoring can take several times more than 
fprob+bprop together.

Jan

Bart van Merriënboer wrote:

blocks/bricks/wrappers.py
rizar: > I guess we are hitting two extremes - the code I pasted has a mask argument which is not needed if you allow Softmax to return arrays. If you get rid of the masks and just let it operate on a chosen axis it will become simpler. In fact in most places you replace axis=1 by axis=self.axis. The only non-trivial change is the dimshuffle for categorical cross-entropy.

True, but still you have to add reshapes before and after softmax. Why not keep the out of the Softmax brick, in another, reusable component? Isn't this modularity that we aim for in Blocks?

Similar I don't think that the axis choice should be in Softmax. Rather than having all-cases-aware monolith components like your  `AxisSoftmax` I prefer to have a few components, each with a clear simple function (Unix-style, so to say). 

The documentation of `Softmax` can indeed be further improved.

> Then, blocks should supply the users with good implementations of common use-cases. If even we need to handle n-dimensional arrays in softmax why not simply add it?

We can add a reference to `WithExtraDims` in softmax documentation.

blocks/bricks/__init__.py
rizar: I will have to look deeper, because as Jan said, it seems like the old 
version of the code also involved advanced indexing.

blocks/bricks/wrappers.py
janchorowski: My main concern is that you trade a simple change to softmax for:
1. introduction of an auxiliary class, which is rather complicated (you 
add 40 lines to do the flattening, even the documentation takes 20 
lines, that is already more than the softmax!)
2. to restore the full ability of axis softmax you need to couple it 
also with the dimshuffling brick -
3. It is very non-intuitive why instead off 1 softmax you suddenly have 
to instantiate it several times, to wrap it different apply methods. 
Then if you instantiate it once, whose children will it be?

In brief: instead of a self-contained, easy to use brick you propose to 
have 2 bricks, one barely simpler (2d softmax), one much more 
complicated (withextradims), and require the end-user to understand the 
intricacies of brick ownership. That doesn't add up for me.

Dmitry Bogdanov wrote:

rizar: Let's take a break from calling the code of each other rather complicated and consider an example. 
Suppose instead of `Softmax` I want to normalize my scores in a smoother way, e.g. by applying sigmoids instead of exponential function like we did in the TIMIT paper. With the wrapper bricks I advocate one only has to implement the required logic for the simplest case when the input is a matrix. Reshapes required to make it work with more-dimensional inputs can be added by combining the new brick `SmoothSoftmax` with `WithExtraDims` just like it is done in this PR.

On the other hand, your `AxisSoftmax` provides no reusable code for that. One has to copy and modify it in good old Groundhog style.

Please tell me if there is a flaw in this reasoning.

janchorowski: There are no flaws in your reasoning, and in principle what you suggest is good - have exactly one function do exactly one task. But in this particular situation I feel that wrapping the apply with another brick approach is not good - it is not intuitive and requires to instantiate softmax twice. Also, the usage is more difficult - instead of saying `SoftMax(axis=-1)`, i.e. take softmax along last axis you have to wrap it by saying how many dimensions to add - and separately for each of softmax applys that you want to use. And you have to double wrap it to do e.g. `SoftMax(axis=0)` on 3d tensors because you also need the dimshuffle.

Stretching the unix analogy, `sort` has the option to specify a sort key, even though you can achieve the same chaining a cut to reorder the fields, a simple sort, and another cut to reorder it back. 

An acceptable solution would be to somehow outsource the flattening-reshaping, so that you can still say `SoftMax(axis=-1)`, but you have a single implementation of the dimshuffle-flatten-reshape-dimshuffle logic. But this is a form of refactoring and can wait until we have at least two nearly-identical softmax implementations. Especially, that softmax is strange - it typically takes as inputs tensors with different number of dimensions.

blocks/bricks/__init__.py
lamblin: `log(softmax(x))[arange(y.shape[0]), y]` (where x is 2D) is one of the cases that currently has a theano optimization transforming it to `crossentropy_softmax_...`.
So if everything goes right, there should not be advanced indexing in the optimized graph.

blocks/bricks/wrappers.py
rizar: >  But in this particular situation I feel that wrapping the apply with another brick approach is not good - it is not intuitive and requires to instantiate softmax twice. 

I think I know how to refactor `WithExtraDims` and `WithAxesSwapped` to be metaclasses. This in many aspect will combine best of two worlds. 

Have you ever had to use `categorical_cross_entropy` with `axis` other than `-1`? I can imagine situation in which `softmax` has to be used with a different axis, but for categorical cross entropy, I can't.

rizar:
@janchorowski , please take a look at my new code with wrappers as metaclasses.

rizar:
Gave it yet another rewrite, now without new metaclasses. Also removed `WithAxisSwapped` because it does not seem to be used by anybody. I had a plan to use it in `blocks/bricks/attention.py`, but performance is crucial there and it makes sense to have a custom reshape- and dimshuffle- softmax. 

This still does not allow to choose `axis` like you did @janchorowski , but I believe that the new "wrappers" platform can be used for that.

blocks/bricks/__init__.py
```diff
@@ -494,45 +495,73 @@ def apply(self, input_):
...
+    @application(inputs=['y', 'x'], outputs=['output'])
+    def categorical_cross_entropy(self, application_call, y, x):
+        """Computationally stable cross-entropy for pre-softmax values.
 
         Parameters
         ----------
         y : :class:`~tensor.TensorVariable`
             In the case of a matrix argument, each slice along
-            axis represents one distribution. In the vector case, each
+            axis 0 represents one distribution. In the vector case, each
```
janchorowski: maybe use terms `rows` and `columns` since you limit this to 2d case. It is much easier to unserstand

blocks/bricks/__init__.py
```diff
@@ -494,45 +495,73 @@ def apply(self, input_):
...
             element represents the position of the '1' in a one hot-vector.
         x : :class:`~tensor.TensorVariable`
-            Each slice along axis represents energies of a distribution,
-            that is pre-softmax values.
+            Each slice along axis 0 represents unnormalized
```
janchorowski: Each row represents unnormalized probs is easier to read IMHO

blocks/bricks/__init__.py
janchorowski: Thanks

blocks/bricks/__init__.py
```diff
@@ -494,45 +495,73 @@ def apply(self, input_):
...
+            cost = -x[tensor.arange(y.shape[0]), y]
         elif y.ndim == x.ndim:
-            cost = -tensor.mean((log_prob * y).sum(axis=1))
+            cost = -(x * y).sum(axis=1)
         else:
             raise TypeError('rank mismatch between x and y')
         return cost
 
 
+class SoftmaxWithExtraDims(Softmax):
```
janchorowski: Maybe NDimensionalSoftmax? Your name has the advantage of beginning with Softmax so it will be shown by autocomplete though

blocks/bricks/wrappers.py
```diff
@@ -1,71 +1,165 @@
...
+    For documentation of the wrapped application method.
+
+"""
+
+
+@add_metaclass(ABCMeta)
+class BrickWrapper(object):
+    """Base class for wrapper metaclasses.
+
+    Sometimes one wants to add to a brick capability to handle
```
janchorowski: to extend a brick wit the capability to handle

blocks/bricks/wrappers.py
```diff
@@ -1,71 +1,165 @@
...
+    inputs different from what it was designed to handle. A typical
+    example are inputs with more dimensions that was foreseed at
```
janchorowski: more dimensions than

janchorowski:
Looks really good. Thanks!

blocks/bricks/wrappers.py
dwf: Also "foreseen"

blocks/bricks/__init__.py
rizar: I like your `NDimensionalSoftmax`. Mainly because this is not the end of the story, and we need more capabilities for this brick (e.g. to softmax along a custom axis). I believe the exact formula might change, that is another wrapper or a few other wrappers will be used to create `NDimensionalSoftmax` from `Softmax`. 

In fact we could concentrate lots of options for reshaping and reshuffling in one wrapper, that might be easier than `WithExtraDims`, `WithAxisSwapped`, `AlongSpecifiedAxis` and etc.

rizar:
I think I addressed all your remarks, @dwf  and @janchorowski .

I tried however to see what computation graph one gets with new `categorical_cross_entropy`, and below is one for the MNIST demo (after I switched to using `Softmax.categorical_cross_entropy` there). Indeed, advanced indexing happens on CPU. @lamblin , what do you think?

```
Elemwise{Composite{((-i0) / Cast{float32}(i1))}}[(0, 0)] [@A] ''   25
 |Sum{acc_dtype=float64} [@B] ''   24
 | |AdvancedSubtensor [@C] ''   23
 |   |HostFromGpu [@D] 'softmax_log_probabilities_output'   22
 |   | |GpuElemwise{Sub}[(0, 0)] [@E] ''   21
 |   |   |GpuElemwise{Sub}[(0, 0)] [@F] ''   16
 |   |   | |GpuElemwise{Add}[(0, 0)] [@G] ''   13
 |   |   | | |GpuDot22 [@H] ''   12
 |   |   | | | |GpuElemwise{Composite{tanh((i0 + i1))}}[(0, 0)] [@I] ''   11
 |   |   | | | | |GpuDot22 [@J] ''   9
 |   |   | | | | | |GpuFlatten{2} [@K] ''   7
 |   |   | | | | | | |GpuFromHost [@L] ''   5
 |   |   | | | | | |   |features [@M]
 |   |   | | | | | |W [@N]
 |   |   | | | | |GpuDimShuffle{x,0} [@O] ''   4
 |   |   | | | |   |b [@P]
 |   |   | | | |W [@Q]
 |   |   | | |GpuDimShuffle{x,0} [@R] ''   3
 |   |   | |   |b [@S]
 |   |   | |GpuDimShuffle{0,x} [@T] ''   15
 |   |   |   |GpuCAReduce{maximum}{0,1} [@U] ''   14
 |   |   |     |GpuElemwise{Add}[(0, 0)] [@G] ''   13
 |   |   |GpuElemwise{Log}[(0, 0)] [@V] ''   20
 |   |     |GpuDimShuffle{0,x} [@W] ''   19
 |   |       |GpuCAReduce{add}{0,1} [@X] ''   18
 |   |         |GpuElemwise{exp,no_inplace} [@Y] ''   17
 |   |           |GpuElemwise{Sub}[(0, 0)] [@F] ''   16
 |   |ARange [@Z] ''   10
 |   | |TensorConstant{0} [@BA]
 |   | |Prod{acc_dtype=int64} [@BB] ''   8
 |   | | |MakeVector [@BC] ''   6
 |   | |   |Shape_i{0} [@BD] ''   1
 |   | |   | |targets [@BE]
 |   | |   |Shape_i{1} [@BF] ''   0
 |   | |     |targets [@BE]
 |   | |TensorConstant{1} [@BG]
 |   |Flatten{1} [@BH] 'softmax_categorical_cross_entropy_y'   2
 |     |targets [@BE]
 |Prod{acc_dtype=int64} [@BB] ''   8
```

rizar:
To be fail, with the old `categorical_cross_entropy` everything was done on GPU:

```
HostFromGpu [@A] 'softmax_categorical_cross_entropy_output_0'   31
 |GpuElemwise{Composite{(-(i0 / i1))}}[(0, 0)] [@B] ''   30
   |GpuCAReduce{add}{1} [@C] ''   29
   | |GpuAdvancedSubtensor1 [@D] ''   28
   |   |GpuFlatten{1} [@E] ''   27
   |   | |GpuElemwise{Sub}[(0, 0)] [@F] ''   26
   |   |   |GpuElemwise{Sub}[(0, 0)] [@G] ''   21
   |   |   | |GpuElemwise{Add}[(0, 0)] [@H] ''   18
   |   |   | | |GpuDot22 [@I] ''   17
   |   |   | | | |GpuElemwise{Composite{tanh((i0 + i1))}}[(0, 0)] [@J] ''   14
   |   |   | | | | |GpuDot22 [@K] ''   11
   |   |   | | | | | |GpuFlatten{2} [@L] ''   9
   |   |   | | | | | | |GpuFromHost [@M] ''   6
   |   |   | | | | | |   |features [@N]
   |   |   | | | | | |W [@O]
   |   |   | | | | |GpuDimShuffle{x,0} [@P] ''   5
   |   |   | | | |   |b [@Q]
   |   |   | | | |W [@R]
   |   |   | | |GpuDimShuffle{x,0} [@S] ''   4
   |   |   | |   |b [@T]
   |   |   | |GpuDimShuffle{0,x} [@U] ''   20
   |   |   |   |GpuCAReduce{maximum}{0,1} [@V] ''   19
   |   |   |     |GpuElemwise{Add}[(0, 0)] [@H] ''   18
   |   |   |GpuElemwise{Log}[(0, 0)] [@W] ''   25
   |   |     |GpuDimShuffle{0,x} [@X] ''   24
   |   |       |GpuCAReduce{add}{0,1} [@Y] ''   23
   |   |         |GpuElemwise{exp,no_inplace} [@Z] ''   22
   |   |           |GpuElemwise{Sub}[(0, 0)] [@G] ''   21
   |   |Elemwise{Composite{(i0 + (i1 * i2))}}[(0, 1)] [@BA] ''   16
   |     |Flatten{1} [@BB] 'softmax_categorical_cross_entropy_y'   2
   |     | |targets [@BC]
   |     |ARange [@BD] ''   13
   |     | |TensorConstant{0} [@BE]
   |     | |Prod{acc_dtype=int64} [@BF] ''   10
   |     | | |MakeVector [@BG] ''   7
   |     | |   |Shape_i{0} [@BH] ''   1
   |     | |   | |targets [@BC]
   |     | |   |Shape_i{1} [@BI] ''   0
   |     | |     |targets [@BC]
   |     | |TensorConstant{1} [@BJ]
   |     |InplaceDimShuffle{x} [@BK] ''   8
   |       |Shape_i{1} [@BL] ''   3
   |         |W [@R]
   |GpuFromHost [@BM] ''   15
     |Elemwise{Cast{float32}} [@BN] ''   12
       |Prod{acc_dtype=int64} [@BF] ''   10
```

janchorowski:
I would profile the two versions. Probably using just the `time` command, or %timeit in IPython for the update function of the MNIST demo because Theano has some support for streams and may do the GPU-CPU memory transfer concurrently with some computations, which will get disabled with CUDA_BLOCKING=1.

rizar:
Weirdly enough, but the new version, where advanced indexing is  done on CPU, is some 20% faster.

janchorowski:
That's what I was pointing to: advanced indexing is really bad on the GPU, because you can not do coalesced memory reads, while the memory transfer is fast because it is aligned and runs in parallel to other stuff. Then indexing on the CPU is fast.

lamblin:
It looks like the optimization I was mentioning is not applied.
I guess that it got confused by the argument in the `arange`: it was looking for `targets.flatten().shape[0]`, which seems to have been optimized to `targets.shape[0] * targets.shape[1]`.
One would have to check more in depth on that example to be sure.

bartvm:
Relying on the memory transfer being fast/done in parallel is not very reliable. Whether you will see a speedup is very model-dependent, because in the case where you have a large vocabulary and batch size you could easily be copying dozens of megabytes from the GPU, which will slow things down. In Pylearn2 for word2vec we would see speedups of up to 50% for very large vocabularies and batch sizes. Taking the approach where things happen on CPU will mean that the efficiency will depend on the size of your output layer, which doesn't sound like a good idea to me.

rizar:
It seems to me that @bartvm  is right and we should make sure that everything happens on GPU by default.

janchorowski:
Then I think we should implement the basic case of advanced indexing - when all indexes are Theano tensors on the GPU. In fact we can restrict ourselves to indexing with a single Theano tensor (since computing linear indexes is an elementwise op) - this is easily achieved in CUDA. @nouiz, what do you think about that?

While it is great that Theano has stabilization and performance optimizations, it becomes difficult when it is hard to predict when will these optimizations be applied. It is much easier to properly code a numerically stable and optimal graph for a common operation, rather than code a non-optimal one hoping that the optimizer will kick in. IMHO, the comment by @lamblin suggests that even for people with in-depth Theano knowledge it is not obvious when those optimizations are applied.

rizar:
@lamblin , it seems like it should be considered a bug then.

Meanwhile, I updated this PR to use a safer version of advanced indexing and am waiting for an LGTM.

janchorowski:
Can you explain me how is the new version safer? For one, you don't check out-of-bounds conditions. Execution-wise, it will be very similar (it also happens on CPU)

rizar:
It is safer because it explicitly triggers 1D indexing, as opposed to relying on an optimization to replace 2D indexing with 1D. It is advanced 2D indexing that Theano can't do on GPU.

janchorowski:
Then all's good for me.

PR #761, opened 2015-07-13
fix minor bugs when handling masks in SequenceGenerator/Attention

udibr:
HTH

### DIFF
--- a/blocks/bricks/attention.py
+++ b/blocks/bricks/attention.py
@@ -605,8 +605,11 @@ class AttentionRecurrent(AbstractAttentionRecurrent, Initializable):
             kwargs.pop(self.preprocessed_attended_name, None),
             kwargs.pop(self.attended_mask_name, None),
             **dict_union(states, glimpses_needed))
-        if kwargs:
-            raise ValueError("extra args to take_glimpses: {}".format(kwargs))
+        # The following check is too excessive.
+        # there could be context information which is needed by the transition
+        # apply, but is also passed to this method.
+        # if kwargs:
+        #     raise ValueError("extra args to take_glimpses: {}".format(kwargs))
         return result
 
     @take_glimpses.property('outputs')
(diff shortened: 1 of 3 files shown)


blocks/bricks/attention.py
```diff
@@ -634,13 +634,15 @@ def compute_states(self, **kwargs):
             Current states computed by `self.transition`.
 
         """
-        # Masks are not mandatory, that's why 'must_have=False'
-        sequences = dict_subset(kwargs, self._sequence_names,
-                                pop=True, must_have=False)
+        # make sure we are not pooping the mask
```
janchorowski: popping

blocks/bricks/attention.py
```diff
@@ -605,8 +605,11 @@ def take_glimpses(self, **kwargs):
             kwargs.pop(self.preprocessed_attended_name, None),
             kwargs.pop(self.attended_mask_name, None),
             **dict_union(states, glimpses_needed))
-        if kwargs:
-            raise ValueError("extra args to take_glimpses: {}".format(kwargs))
+        # The following check is too excessive.
+        # there could be context information which is needed by the transition
+        # apply, but is also passed to this method.
+        # if kwargs:
+        #    raise ValueError("extra args to take_glimpses: {}".format(kwargs))
```
dmitriy-serdyuk: Can you provide an example? `take_glimpses` is called in `do_apply` where we provide only states, glimpses, and attended stuff. So it shouldn't be any variables for `transition.apply` here.

udibr: It is also called from BaseSequenceGenerator.generate which passes to it the arguments with names BaseSequenceGenerator._context_names
This is actuall a call to BaseSequenceGenerator.transition.apply.contexts which is actually a call to
AttentionRecurrent.do_apply_contexts

This returns  `attended` and `attended_mask` which are needed to take a glimpsee but it also includes context supplied by the transition (eg LSTM):
    AttentionRecurrent.transition.apply.contexts
Up to now all basic transition classes supplied in recurrent.py dont have context so the problem 
was left unnotticed. But I have wrote a wrapper for basic transition which adds context (will be published at some point) and this is how I found this bug.

dmitriy-serdyuk: I see. May I ask you to remove lines 611-612 and change the comment. It's not a good practice to leave commented out code.

And can you add a simple test which crashes if this check is not commented out. I suppose it may be a simple recurrent transition with some dummy context.

dmitriy-serdyuk:
Thank you for the fixes! I'm ready to merge the PR when you fix the comment.

I think we need a test for the sequence generator without mask but if you don't have time to add tests, we can leave it for future work.

udibr:
which comment do you want me to fix?

dmitriy-serdyuk:
I mean to remove the piece of code from the comment [here](https://github.com/udibr/blocks/blob/RecurrentStack/blocks/bricks/attention.py#L611) and change it in a way that one can understand it after, like 'kwargs may be not empty since they may contain contexts of the transition'.

dmitriy-serdyuk:
Looks good, thank you!

blocks/monitoring/evaluators.py
```diff
@@ -262,9 +262,15 @@ def _compile(self):
         if self.theano_buffer.accumulation_updates:
             updates = OrderedDict()
             updates.update(self.theano_buffer.accumulation_updates)
-            if self.updates:
-                updates.update(self.updates)
             inputs += self.theano_buffer.inputs
+        if self.updates:
```
rizar: This change should not have been merged without tests, since it would be pretty easy to write a test for it. Also it seems to me that always initializing `updates` in the line 261 to `OrderedDict()` would do the job. Am I right?

PR #764, opened 2015-07-17
compute correct per_sequence_element when no mask is given

udibr:


rizar:
Sorry that this PR was ignored for so long, it was quite a nasty bug!

PR #766, opened 2015-07-20
Fix log unpickling

### LINKED ISSUE #765: Unpickling resumed log fails
When unpickling a saved log that has previously been resumed using `resume()`, the following error is raised:

````python
In [1]: from blocks.serialization import load, dump
Using gpu device 0: NVS 4200M
In [2]: from blocks.log import TrainingLog
In [3]: log1 = TrainingLog()
In [4]: dump(log1, "log1.pkl")
In [5]: log2 = load("log1.pkl")
In [6]: dump(log2, "log2.pkl")
In [7]: load("log2.pkl")  # loading an unresumed log works
Out[7]: defaultdict(<type 'dict'>, {})
In [8]: log2.resume()
In [9]: dump(log2, "log3.pkl")
In [10]: load("log3.pkl")  # loading a resumed log does not work
<ipython-input-10-a2b8b841ca66> in <module>()
----> 1 load("log3.pkl")

/home/fabian/msc/exp/env/lib/python2.7/site-packages/theano/misc/pkl_utils.pyc in load(f, persistent_load)
    318         p = pickle.Unpickler(BytesIO(zip_file.open('pkl').read()))
    319         p.persistent_load = persistent_load(zip_file)
--> 320         return p.load()
    321 
    322 

/home/fabian/msc/exp/local/lib/python2.7/pickle.pyc in load(self)
    856             while 1:
    857                 key = read(1)
--> 858                 dispatch[key](self)
    859         except _Stop, stopinst:
    860             return stopinst.value

/home/fabian/msc/exp/local/lib/python2.7/pickle.pyc in load_newobj(self)
   1081         args = self.stack.pop()
   1082         cls = self.stack[-1]
-> 1083         obj = cls.__new__(cls, *args)
   1084         self.stack[-1] = obj
   1085     dispatch[NEWOBJ] = load_newobj

TypeError: buffer() takes at least 1 argument (0 given)
````

This may be related to [Python bug 8323](http://bugs.python.org/issue8323) which is classified as `wontfix`.

dmitriy-serdyuk:
Fixes #765 .

### DIFF
--- a/blocks/log/log.py
+++ b/blocks/log/log.py
@@ -64,7 +64,7 @@ class TrainingLogBase(object):
             })
 
     @property
-    def b_uuid(self):
+    def h_uuid(self):
         """Return a hexadecimal version of the UUID bytes.
 
         This is necessary to store ids in an SQLite database.
@@ -79,7 +79,7 @@ class TrainingLogBase(object):
         copies the status of the old log into the new log.
 
         """
-        old_uuid = self.b_uuid
+        old_uuid = self.h_uuid
         old_status = dict(self.status)
         self.uuid = uuid4()
         self.status.update(old_status)
--- a/blocks/log/log.py
+++ b/blocks/log/log.py
@@ -1,5 +1,4 @@
 """The event-based main loop of Blocks."""
-import sqlite3
 from abc import ABCMeta
 from collections import defaultdict
 from numbers import Integral
@@ -66,12 +65,12 @@ class TrainingLogBase(object):
 
     @property
     def b_uuid(self):
-        """Return a buffered version of the UUID bytes.
+        """Return a hexadecimal version of the UUID bytes.
 
-        This is necessary to store bytes in an SQLite database.
+        This is necessary to store ids in an SQLite database.
 
         """
-        return sqlite3.Binary(self.uuid.bytes)
+        return self.uuid.hex
 
     def resume(self):
         """Resume a log by setting a new random UUID.
--- a/blocks/log/sqlite.py
+++ b/blocks/log/sqlite.py
@@ -167,13 +167,13 @@ class SQLiteLog(TrainingLogBase, Mapping):
     def __iter__(self):
         return map(itemgetter(0), self.conn.execute(
             ANCESTORS_QUERY + "SELECT DISTINCT time FROM entries "
-            "WHERE uuid IN ancestors ORDER BY time ASC", (self.b_uuid,)
+            "WHERE uuid IN ancestors ORDER BY time ASC", (self.h_uuid,)
         ))
 
     def __len__(self):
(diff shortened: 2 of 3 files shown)


dmitriy-serdyuk:
Hmm, buffers are non-picklable but strings are non-sqliteable.

dwf:
@dmitriy-serdyuk ¯\_(ツ)_/¯

`__setstate__` and `__getstate__` time?

bartvm:
Just a quick note that this is what you're running into: http://bugs.python.org/issue8323 It's a bug in Python 2 that won't be fixed...

dmitriy-serdyuk:
@bartvm  we still need a workaround since lots and lots of people are still using Python2.

bartvm:
If you're going to store strings instead of binary data, don't forget to update the column preferences in the SQL commands e.g. https://github.com/mila-udem/blocks/blob/master/blocks/log/sqlite.py#L124 and https://github.com/mila-udem/blocks/blob/master/blocks/log/sqlite.py#L131

dmitriy-serdyuk:
Actually, storing ids as text makes much more sense. I've never seen that id has a binary type, I guess, more sophisticated databases have optimizations for index build by a text field.

dmitriy-serdyuk:
@dwf , can you take a look?

dwf:
@dmitriy-serdyuk Sorry, somehow I missed your request. Looking now.

dwf:
LGTM, the only thing I'm wondering is whether there are any backwards compatibility issues with existing SQLite files?

dmitriy-serdyuk:
Most probably it's not backward compatible. I'll try to figure it out, but currently I cannot run sqlite backend at all. Seems I have too old sqlite3 library.

dwf:
OK. We're pretty bold about API breakages but I'm not sure to what extent
we want to hang people's old log files out to dry. But if @bartvm gives it
his blessing (since he wrote this thing and is probably one of the few
using it) I'm happy to merge as is.

dmitriy-serdyuk:
It may be compatible if sqlite can read binary data as text. In this case uuids will be different but still unique. From [here](https://www.sqlite.org/datatype3.html):

> SQLite uses a more general dynamic type system. In SQLite, the datatype of a value is associated with the value itself, not with its container. The dynamic type system of SQLite is backwards compatible with the more common static type systems of other database engines in the sense that SQL statements that work on statically typed databases should work the same way in SQLite. However, the dynamic typing in SQLite allows it to do things which are not possible in traditional rigidly typed databases.

dwf:
@bartvm says it's no biggie if the column type changes, so let's just merge it as-is. If someone complains, we can figure out what needs to be done at that point. My guess is that this isn't actually being used very much.

PR #772, opened 2015-07-23
Tutorial on how to create a basic brick

### LINKED ISSUE #611: "Creating a Brick" tutorial
Walking through my second or so time attempting to create a custom brick, I think it would be a good idea to have a "custom brick" tutorial, in the same spirit as the Theano tutorials that walk you through the creation of a custom op (albeit more high-level).

adbrebs:
Fixes #611 @dmitriy-serdyuk 
This is still work in progress but you can already comment on the structure.

### DIFF
--- a/docs/create_your_own_brick.rst
+++ b/docs/create_your_own_brick.rst
@@ -7,7 +7,7 @@ bricks themselves) into a single operation so that you can easily reuse it.
 
 The first part of this tutorial lists the requirements and optional components
 that a brick should/can implement while the second part describes the
-construction step by step of a simple toy brick.
+construction of a simple toy brick.
 
 This tutorial assumes that you are already familiar with
 :doc:`bricks <bricks_overview>`.
@@ -38,13 +38,14 @@ from :class:`.Brick`, then you should consider overwriting the following
 methods (strictly speaking, all these methods are optional):
 
 * :meth:`.Brick.__init__`: you should pass by argument the attributes of your
-  bricks. It is also in this method that you should create the potential
-  "children bricks" that belongs to your brick. The initialiazation of the
-  attributes can be lazy as described in a further paragraph.
+  brick. It is also in this method that you should create the potential
+  "children bricks" that belongs to your brick (in that case, you have to put
+  the children bricks into ``self.children``. The initialiazation of the
+  attributes can be lazy as described later in the tutorial.
 * :meth:`you_decide_which_name`: you need to implement a method that actually
   implements the operation of the brick, taking as arguments the inputs
   of the brick and returning its outputs. It can have any name and for simple
-  bricks is often named ``apply``. You can decorate it with the
+  bricks is often named ``apply``. You should decorate it with the
   :func:`.application` decorator, as explained in the next section.
 * :meth:`.Brick._allocate`: you should implement this method if your brick
   needs to allocate its parameters.
@@ -52,22 +53,28 @@ methods (strictly speaking, all these methods are optional):
   initialize parameters of your brick.
 * :meth:`.Brick._push_allocation_config`: you should consider overwriting
   this method if you want to allocate the children bricks in a specific way.
-* :meth:`.Brick._push_initialization_config`: you should consider method if
-  you want to initialize the children bricks in a specific way.
-* :meth:`.Brick.get_dim`: this method is useful to get the dimensions of the
-  inputs and outputs of the brick.
+* :meth:`.Brick._push_initialization_config`: you should consider
+  overwriting this method if you want to initialize the children bricks in a
+  specific way. You should inherit from :class:`.Initializable` to initialize
+  the potential children bricks recursively.
+* :meth:`.Brick.get_dim`: implementing this function is useful if you want
+  to provide a simple way to get the dimensions of the inputs and outputs of
+  the brick.
 
 If you want to inherit from a specific brick, check its docstring to
-identify the particular methods to overwrite.
+identify the particular methods to overwrite and the attributes to define.
 
 you_decide_which_name method
 ~~~~~~~~~~~~~~~~~~~~~~~~~~~~
 
-The :func:`.application` decorator can be used to name variables and
-register auxiliary variables to the operation you implement. It is used as
-followed:
+The :meth:`you_decide_which_name` method described above is probably the most
+important method of your brick because it is the one that actually takes
+theano tensors as inputs, process them and return tensor outputs. You should
+decorate it with the :func:`.application` decorator, which names variables
+and register auxiliary variables of the operation you implement.
+It is used as followed (:meth:`you_decide_which_name` is named :meth:`apply`):
 
-    >>> class Foo(Brick):
+    >>> class Foo(Brick): # doctest: +SKIP
     ...     @application(inputs=['input1', 'input2'], outputs=['output'])
     ...     def apply(self, input1, input2)
     ...         ...
@@ -77,13 +84,14 @@ In the case above, it will automatically label the theano tensor variable
 input1 to ``Foo_apply_input1``, idem for input2 and the output of the method.
 
 Under the hood, the ``@application`` decorator creates an object of class
-:class:`.Application`, named ``apply``, which becomes an attribute of the brick.
+:class:`.Application`, named ``apply``, which becomes an attribute of the
+brick.
 
 In the previous examples, variables were named with strings. If you need to
 name certain variables with other variables (such as ``self .fancy_name``),
 you should define them with the apply.property decorator:
 
-    >>> class Foo(Brick):
+    >>> class Foo(Brick): # doctest: +SKIP
     ...     fancy_name = "salut_toi"
     ...     @application
     ...     def apply(self, input)
@@ -92,9 +100,10 @@ you should define them with the apply.property decorator:
     ...     def apply_inputs(self):
     ...         return self.fancy_name
 
-You can also annotate specific variables, as shown in this example:
+You can also annotate specific variables by passing ``application_call`` as
+agurment of your ``apply`` function, as shown in this example:
 
-    >>> class Foo(Brick):
+    >>> class Foo(Brick): # doctest: +SKIP
     ...     @application
     ...     def apply(self, x, application_call):
     ...         application_call.add_auxiliary_variable(x.mean())
@@ -113,11 +122,12 @@ later, after the creation of the brick. To enable this mecanism, called lazy
 initialization, you need to decorate the method :meth:`.Brick.__init__` with
 the :func:`.lazy` decorator:
 
-    >>> @lazy(allocation=['attr1', 'attr2'])
+    >>> @lazy(allocation=['attr1', 'attr2']) # doctest: +SKIP
     ... def __init__(self, attr1, attr1)
     ...     ...
 
-This allows the user to specify attr1 and attr2 after the creation of the brick.
+This allows the user to specify attr1 and attr2 after the creation of the
+brick.
 
 
 Example
@@ -134,9 +144,9 @@ not inherit from :class:`.Feedforward`, which requires a single input and a
 single output. Our brick will have to manage two shared variables
 representing the matrices to multiply the inputs with and thus, inheriting from
 :class:`.Initializable` makes perfectly sense as we will let the user decide
-which initialization scheme to use. TODO
+which initialization scheme to use.
 
-    >>> class ParallelLinear(Initializable):
+    >>> class ParallelLinear(Initializable): # doctest: +SKIP
     ...     r"""Two linear transformations without biases.
     ...
     ...     Brick which applies two linear (affine) transformations by
@@ -160,14 +170,6 @@ which initialization scheme to use. TODO
     ...         self.input_dim2 = input_dim2
     ...         self.output_dim1 = output_dim1
     ...         self.output_dim2 = output_dim2
-    ...
-    ...     @property
-    ...     def W1(self):
-    ...         return self.parameters[0]
-    ...
-    ...     @property
-    ...     def W2(self):
-    ...         return self.parameters[0]
     ...
     ...     def __allocate(self, input_dim, output_dim, number):
     ...         W = shared_floatx_nans((input_dim, output_dim),
@@ -216,3 +218,32 @@ which initialization scheme to use. TODO
     ...         if name == 'output2':
     ...             return self.output_dim2
     ...         super(ParallelLinear, self).get_dim(name)
+
+
+One can also create the brick using :class:`Linear` children bricks, which
+gives a more compact version:
+
+    >>> from blocks.bricks import Linear # doctest: +SKIP
+    >>> class ParallelLinear2(Initializable): # doctest: +SKIP
+    ...     def __init__(self, input_dim1, input_dim2, output_dim1, output_dim2,
+    ...                  **kwargs): ...
+    ...         super(ParallelLinear2, self).__init__(**kwargs)
+    ...         self.linear1 = Linear(input_dim1, output_dim1,
+    ...                               use_bias=False, **kwargs)
+    ...         self.linear2 = Linear(input_dim2, output_dim2,
+    ...                               use_bias=False, **kwargs)
+    ...         self.children = [self.linear1, self.linear2]
+    ...
+    ...     @application(inputs=['input1_', 'input2_'], outputs=['output1',
+    ...         'output2'])
+    ...     def apply(self, input1_, input2_):
+    ...         output1 = self.linear1.apply(input1_)
+    ...         output2 = self.linear2.apply(input2_)
+    ...         return output1, output2
+    ...
+    ...     def get_dim(self, name):
+    ...         if name in ['input1_', 'output1']:
+    ...             return self.linear1.get_dim(name)
+    ...         if name in ['input2_', 'output2']:
+    ...             return self.linear2.get_dim(name)
+    ...         super(ParallelLinear2, self).get_dim(name)
--- a/docs/create_your_own_brick.rst
+++ b/docs/create_your_own_brick.rst
@@ -1,10 +1,24 @@
 Create your own brick
 =====================
 
+.. doctest::
+   :hide:
+
+    >>> import numpy
+    >>>
+    >>> import theano
+    >>> from theano import tensor
+    >>>
+    >>> from blocks.bricks import Brick, Initializable, Linear, Feedforward
+    >>> from blocks.bricks.base import lazy, application
+    >>> from blocks.bricks.parallel import Parallel
+    >>> from blocks.initialization import Constant
+    >>> from blocks.roles import add_role, WEIGHT
+    >>> from blocks.utils import shared_floatx_nans
+
 This tutorial explains how to create a custom brick, which is useful if you
-want to factorize a specific several combined operations (which can be made of
-bricks themselves) into a single higher-level operation so that you can easily
-reuse it.
+want to group several specific operations (which can be bricks themselves) into
+a single one so that you can easily reuse it.
 
 The first part of this tutorial lists the requirements and optional components
 that a brick should/can implement while the second part describes the
@@ -17,9 +31,9 @@ This tutorial assumes that you are already familiar with
 Bricks ingredients and recipe
 -----------------------------
 
-All the bricks in blocks inherit directly or indirectly from the
+All the bricks in Blocks inherit directly or indirectly from the
 :class:`.Brick`. There is already a rich inheritance hierarchy of
-bricks implemented in blocks and thus, you should consider which brick level
+bricks implemented in Blocks and thus, you should consider which brick level
 to inherit from. Bear in mind that multiple inheritance is often possible and
 advocated whenever it makes sense.
 
@@ -36,9 +50,10 @@ Here are examples of possible bricks to inherit from:
   :doc:`tutorial about rnns</rnn>` for more information.
 * many more!
 
-Let's say that you want to create a brick from scracth, simply inheriting
+Let's say that you want to create a brick from scratch, simply inheriting
 from :class:`.Brick`, then you should consider overwriting the following
-methods (strictly speaking, all these methods are optional):
+methods (strictly speaking, all these methods are optional, check the docstring
+of :class:`.Brick` for a precise description of the life-cycle of a brick):
 
 * :meth:`.Brick.__init__`: you should pass by argument the attributes of your
   brick. It is also in this method that you should create the potential
@@ -54,7 +69,7 @@ methods (strictly speaking, all these methods are optional):
   :func:`.recurrent` decorator as explained in the
   :doc:`tutorial about rnns</rnn>`.
 * :meth:`.Brick._allocate`: you should implement this method to allocate the
-  shared variables (often representing parameters) of the brick. In blocks,
+  shared variables (often representing parameters) of the brick. In Blocks,
   by convention, the built-in bricks allocate their shared variables with nan
   values and we recommend you to do the same.
 * :meth:`.Brick._initialize`: you should implement this method to initialize
@@ -68,7 +83,7 @@ methods (strictly speaking, all these methods are optional):
   the children before they get initialized.
   If the children bricks need to be initialized with the same scheme, then you
   should inherit your brick from :class:`.Initializable`, which
-  automatically push the initialization schemes of your brick (provided as
+  automatically pushes the initialization schemes of your brick (provided as
   arguments ``weights_init`` and ``biases_init`` of the constructor) to the
   children bricks.
 * :meth:`.Brick.get_dim`: implementing this function is useful if you want
@@ -88,19 +103,42 @@ decorate it with the :func:`.application` decorator, which names variables
 and register auxiliary variables of the operation you implement.
 It is used as follows:
 
-    >>> class Foo(Brick): # doctest: +SKIP
+    >>> class Foo(Brick):
     ...     @application(inputs=['input1', 'input2'], outputs=['output'])
-    ...     def apply(self, input1, input2)
-    ...         ...
-    ...         return something
+    ...     def apply(self, input1, input2):
+    ...         y = input1 + input2
+    ...         return y
 
-In the case above, it will automatically label the theano tensor variable
-``input1`` to ``Foo_apply_input1``, idem for ``input2`` and the output of the
-method.
+In the case above, it will automatically rename the theano tensor variable
+``input1`` to ``Foo_apply_input1``, ``input2`` to ``Foo_apply_input2`` and the 
+output of the method to ``foo_apply_output``. It will also add roles and names 
+to the tag attributes of the variables, as shown below:
+
+    >>> foo = Foo()
+    >>> i1 = tensor.matrix('i1')
+    >>> i2 = tensor.matrix('i2')
+    >>> y = foo.apply(i1, i2)
+    >>> theano.printing.debugprint(y)
+    Elemwise{identity} [@A] 'foo_apply_output'   
+     |Elemwise{add,no_inplace} [@B] ''   
+       |Elemwise{identity} [@C] 'foo_apply_input1'   
+       | |i1 [@D]
+       |Elemwise{identity} [@E] 'foo_apply_input2'   
+         |i2 [@F]
+    >>> print(y.name)
+    foo_apply_output
+    >>> print(y.tag.name)
+    output
+    >>> print(y.tag.roles)
+    [OUTPUT]
 
 Under the hood, the ``@application`` decorator creates an object of class
 :class:`.Application`, named ``apply``, which becomes an attribute of the
-brick class (by opposition to class instances).
+brick class (by opposition to class instances):
+
+    >>> print(type(Foo.apply))
+    <class 'blocks.bricks.base.Application'>
+
 
 Application properties
 """"""""""""""""""""""
@@ -108,7 +146,7 @@ Application properties
 In the previous examples, the names of the arguments of the application methods
 were directly provided as arguments of the ``@application`` decorator because
 they were common to all instances of the classes. On the other hand, if these
-names need to be defined differently for particular instances of the class,
+names need to be defined differently for particular instances of the class, 
 you should use the ``apply.property`` decorator. Let's say that we want to
 name our attribute inputs with the string ``self.fancy_name``, then we should
 write:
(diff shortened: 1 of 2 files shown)


dmitriy-serdyuk:
The structure looks good.

We test example code, so maybe you'll need to skip some tests, see [doctest](https://docs.python.org/2/library/doctest.html) module.

docs/create_your_own_brick.rst
```diff
@@ -0,0 +1,218 @@
...
+methods (strictly speaking, all these methods are optional):
+
+* :meth:`.Brick.__init__`: you should pass by argument the attributes of your
+  bricks. It is also in this method that you should create the potential
+  "children bricks" that belongs to your brick. The initialiazation of the
+  attributes can be lazy as described in a further paragraph.
+* :meth:`you_decide_which_name`: you need to implement a method that actually
+  implements the operation of the brick, taking as arguments the inputs
+  of the brick and returning its outputs. It can have any name and for simple
+  bricks is often named ``apply``. You can decorate it with the
```
dmitriy-serdyuk: You **should**, otherwise all blocks magic won't work.

docs/create_your_own_brick.rst
```diff
@@ -0,0 +1,218 @@
...
+  :func:`.application` decorator, as explained in the next section.
+* :meth:`.Brick._allocate`: you should implement this method if your brick
+  needs to allocate its parameters.
+* :meth:`.Brick._initialize`: you should implement this method if you need to
+  initialize parameters of your brick.
+* :meth:`.Brick._push_allocation_config`: you should consider overwriting
+  this method if you want to allocate the children bricks in a specific way.
+* :meth:`.Brick._push_initialization_config`: you should consider method if
+  you want to initialize the children bricks in a specific way.
```
dmitriy-serdyuk: It worth mentioning here that usually one should just inherit from `Initializable` and it is going to initialize children recursively.

docs/create_your_own_brick.rst
```diff
@@ -0,0 +1,218 @@
...
+    ...         self.output_dim1 = output_dim1
+    ...         self.output_dim2 = output_dim2
+    ...
+    ...     @property
+    ...     def W1(self):
+    ...         return self.parameters[0]
+    ...
+    ...     @property
+    ...     def W2(self):
+    ...         return self.parameters[0]
```
dmitriy-serdyuk: 0 -> 1

docs/create_your_own_brick.rst
```diff
@@ -0,0 +1,218 @@
...
+    ...         add_role(W, WEIGHT)
+    ...         self.parameters.append(W)
+    ...         self.add_auxiliary_variable(W.norm(2), name='W'+number+'_norm')
+    ...
+    ...     def _allocate(self):
+    ...         self.__allocate(self.input_dim1, self.output_dim1, '1')
+    ...         self.__allocate(self.input_dim2, self.output_dim2, '2')
+    ...
+    ...     def _initialize(self):
+    ...         W1, W2 = self.parameters
```
dmitriy-serdyuk: You could use `self.W*` here.

dmitriy-serdyuk:
Can you also add an example on how to apply the brick to theano variables, compile a function and check the result?

dmitriy-serdyuk:
And can you add a See Also link to recurrent stuff? I think we should add a similar tutorial for recurrent brick, but it should be OK just to add a link to `BaseRecurrent` and `@recurrent` since many people use blocks for sequence generation tasks.

docs/create_your_own_brick.rst
```diff
@@ -0,0 +1,249 @@
...
+
+Let's say that you want to create a brick from scracth, simply inheriting
+from :class:`.Brick`, then you should consider overwriting the following
+methods (strictly speaking, all these methods are optional):
+
+* :meth:`.Brick.__init__`: you should pass by argument the attributes of your
+  brick. It is also in this method that you should create the potential
+  "children bricks" that belongs to your brick (in that case, you have to put
+  the children bricks into ``self.children``. The initialiazation of the
+  attributes can be lazy as described later in the tutorial.
```
dmitriy-serdyuk: I think, you need to mention that the brick can call application methods of other brick only if it is a child brick.

docs/create_your_own_brick.rst
```diff
@@ -0,0 +1,249 @@
...
+* :meth:`you_decide_which_name`: you need to implement a method that actually
+  implements the operation of the brick, taking as arguments the inputs
+  of the brick and returning its outputs. It can have any name and for simple
+  bricks is often named ``apply``. You should decorate it with the
+  :func:`.application` decorator, as explained in the next section.
+* :meth:`.Brick._allocate`: you should implement this method if your brick
+  needs to allocate its parameters.
```
dmitriy-serdyuk: We have an agreement, that when bricks are allocated, the parameters contain nans.

dmitriy-serdyuk:
Sorry for pickiness. I just realized that we should explain how to use `get_dim` and lazy constructors in `_push_allocation_config` method. I think we need a brick has two children and the first child's dimension depends on the second. Are you following me?

adbrebs:
@dmitriy-serdyuk I have modified the tutorial according to your comments except the last one (lazy), which I don't really know how to do. Could you please point out an example?

dmitriy-serdyuk:
Usually, `get_dim` is used inside `_push_allocation_config`, see for example `BaseSequenceGenerator._push_allocation_config` class: it assigns dimensions for one child depending on demensions of other. It works nicely with lazy initialization, a user may create a brick and do not specify some dimentions and they will be assigned by a parent brick.

I suggest to create one more brick `DoubleParallel` which applies two parallel-like bricks consequently like

```
class DoubleParallel:
    def __init__(transformations):
         self.transformations = transformations
    def _push_allocation_config():
        # assign output dimension of transformation 0 to the input 
        # dimension of transformation 1
    @application
    def apply(x, y):
        outs = self.transformations[0].apply(x, y)
        return self.transformations[1].apply(outs)

```

adbrebs:
Ok, but what if the two ``transformations`` already have all their attributes initialized? It is not very nice to assign them a second time (actually this might lead to quiet bugs).

Furthermore ``DoubleParallel`` would only work with children bricks that have ``input_dim`` and ``output_dim`` attributes.

dmitriy-serdyuk:
Anyway, the attributes should be the same. I guess one may check if they are initialized and not the same and raise an error. But it is overcomplicated for this tutorial.

You shouldn't use `input/output_dim`. What you need to demostrate is how to extract names of a child brick inputs/outputs using properties `apply.inputs`, `apply.outputs` and how to get dimension information with `get_dim(name)`.

docs/create_your_own_brick.rst
```diff
@@ -0,0 +1,327 @@
...
+   >>>
+   >>> f1 = theano.function([x1, x2], [output1, output2])
+   >>> f1(x1_mat, x2_mat) # doctest: +ELLIPSIS
+   [array([[ 60.,  60.]]...), array([[ 40.],
+          [ 40.]]...)]
+
+One can also create the brick using :class:`Linear` children bricks, which
+
+    >>> class ParallelLinear2(Initializable):
+    ...     def __init__(self, input_dim1, input_dim2, output_dim1, output_dim2,
```
ASalvail: `@lazy` decorator missing?

adbrebs: Thanks for pointing this out!

docs/create_your_own_brick.rst
```diff
@@ -0,0 +1,327 @@
...
+    ...     @application(inputs=['input1_', 'input2_'], outputs=['output1',
+    ...         'output2'])
+    ...     def apply(self, input1_, input2_):
+    ...         output1 = self.linear1.apply(input1_)
+    ...         output2 = self.linear2.apply(input2_)
+    ...         return output1, output2
+    ...
+    ...     def get_dim(self, name):
+    ...         if name in ['input1_', 'output1']:
+    ...             return self.linear1.get_dim(name)
```
adbrebs: @dmitriy-serdyuk this does not make sense.

adbrebs:
@dmitriy-serdyuk 
Something like that?
```python
class DoubleSequential:
    """
    Two sequential Linear bricks.
    """
    def __init__(self, brick1, brick2):
        self.brick1 = brick1
        self.brick2 = brick2

    def _push_allocation_config(self):
        self.brick2.input_dim = self.brick1.get_dim('input_')

    @application(inputs=['input_'], outputs=['output'])
    def apply(self, x):
        out = self.brick1.apply(x)
        return self.brick2.apply(out)
```
I still don't really understand how to do it and the benefits.

rizar:
@adbrebs, `self.brick1.get_dim(self.brick1.apply.inputs[0])` would be completely interface-independent. But in this particular case it does not help much, because one must know interface of `brick2` to set its configuration. For this example a reasonable requirement would be that both `brick1` and `brick2` support `blocks.bricks.Feedforward` interface. With such a requirement `input_dim` and `output_dim` requests become legitimate.

adbrebs:
Ok thanks @rizar.
@dmitriy-serdyuk I have added the ``DoubleSequential`` brick as you wanted.

docs/create_your_own_brick.rst
```diff
@@ -0,0 +1,364 @@
...
+* :class:`.Sequence`: a sequence of bricks.
+* :class:`.Initializable`: a brick that defines a same initialiation scheme
+  (weights and biases) for all its children.
+* :class:`.Feedforward`: declares an interface for bricks with one input and
+  one output.
+* :class:`.Linear`: a linear transformation with optional bias. Inherits from
+  :class:`.Initializable` and :class:`.Feedforward`.
+* :class:`.BaseRecurrent`: the base class for recurrent bricks. Check the
+  :doc:`tutorial about rnns</rnn>` for more information.
+* many mores!
```
dmitriy-serdyuk: Typo: more

docs/create_your_own_brick.rst
```diff
@@ -0,0 +1,364 @@
...
+Let's say that you want to create a brick from scracth, simply inheriting
+from :class:`.Brick`, then you should consider overwriting the following
+methods (strictly speaking, all these methods are optional):
+
+* :meth:`.Brick.__init__`: you should pass by argument the attributes of your
+  brick. It is also in this method that you should create the potential
+  "children bricks" that belongs to your brick (in that case, you have to put
+  the children bricks into ``self.children``). The initialiazation of the
+  attributes can be lazy as described later in the tutorial.
+* :meth:`you_decide_which_name`: you need to implement a method that actually
```
dmitriy-serdyuk: Let's call `apply` here? It's easier and often the brick name already explains what it does (like `Tanh.apply` or `Linear.apply`). What do you think?

docs/create_your_own_brick.rst
```diff
@@ -0,0 +1,364 @@
...
+
+you_decide_which_name method
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+
+The :meth:`you_decide_which_name` method listed above is probably the most
+important method of your brick because it is the one that actually takes
+theano tensors as inputs, process them and return output tensors. You should
+decorate it with the :func:`.application` decorator, which names variables
+and register auxiliary variables of the operation you implement.
+It is used as followed (:meth:`you_decide_which_name` is named :meth:`apply`):
```
dmitriy-serdyuk: typo: as follows

docs/create_your_own_brick.rst
```diff
@@ -0,0 +1,364 @@
...
+argument of your ``apply`` function, as shown in this example:
+
+    >>> class Foo(Brick): # doctest: +SKIP
+    ...     @application
+    ...     def apply(self, x, application_call):
+    ...         application_call.add_auxiliary_variable(x.mean())
+    ...         return x + 1
+
+You can retrieve these annotated variables as usual with the computational
+graph.
```
dmitriy-serdyuk: Can you add links to `ComputationGraph` and `VariableFilter` classes here?

docs/create_your_own_brick.rst
```diff
@@ -0,0 +1,364 @@
...
+    ...     def __init__(self, brick1, brick2, **kwargs):
+    ...         super(Feedforward, self).__init__(**kwargs)
+    ...         self.brick1 = brick1
+    ...         self.brick2 = brick2
+    ...         self.input_dim = brick1.input_dim
+    ...         self.output_dim = brick2.output_dim
+    ...         self.children = [self.brick1, self.brick2]
+    ...
+    ...     def _push_allocation_config(self):
+    ...         self.brick2.input_dim = self.brick1.get_dim('input_')
```
dmitriy-serdyuk: It should be `get_dim('output')`.

docs/create_your_own_brick.rst
```diff
@@ -0,0 +1,364 @@
...
+    ...     def apply(self, x):
+    ...         return self.brick2.apply(self.brick1.apply(x))
+
+Note how ``get_dim`` is used to retrieve the ``input_dim`` of ``brick1``. You
+can now create a DoubleSeuential brick as follows.
+
+    >>> brick1 = Linear(input_dim=3, output_dim=2)
+    >>> brick2 = Linear(output_dim=4)
+    >>>
+    >>> seq = DoubleSequential(brick1, brick2)
```
dmitriy-serdyuk: It worth adding initialization and demonstrate that the input dimension of brick2 was set:
```
seq.initialize()
brick2.input_dim
```

docs/create_your_own_brick.rst
adbrebs: Ok

docs/create_your_own_brick.rst
adbrebs: Sorry, I don't get it.

docs/create_your_own_brick.rst
adbrebs: up

docs/create_your_own_brick.rst
dmitriy-serdyuk: You had `W1` and `W2` properties earlier. Now it's okay.

docs/create_your_own_brick.rst
dmitriy-serdyuk: This seems totally fine. If the user asked for dimensions of the first brick, ask brick1, otherwise ask brick2.

docs/create_your_own_brick.rst
```diff
@@ -0,0 +1,371 @@
...
+  of the brick and returning its outputs. It can have any name and for simple
+  bricks is often named ``apply``. You should decorate it with the
+  :func:`.application` decorator, as explained in the next section. If you
+  design a recurrent brick, you should instead decorate it with the
+  :func:`.recurrent` decorator as explained in the
+  :doc:`tutorial about rnns</rnn>`.
+* :meth:`.Brick._allocate`: you should implement this method to allocate the
+  shared variables (often representing parameters) of the brick. In blocks,
+  by convention, the built-in bricks allocate their shared variables with nan
+  values and we recommand you to do the same.
```
rizar: recommend

docs/create_your_own_brick.rst
```diff
@@ -0,0 +1,371 @@
...
+* :meth:`.Brick._initialize`: you should implement this method to initialize
+  the shared variables of your brick. This method is called after the
+  allocation.
+* :meth:`.Brick._push_allocation_config`: you should consider overwriting
+  this method if you want to allocate the children bricks in a specific way.
```
rizar: I would say "... if you want to change configuration of the children bricks before they allocate their parameters"

docs/create_your_own_brick.rst
```diff
@@ -0,0 +1,371 @@
...
+* :meth:`.Brick._push_initialization_config`: you should consider
+  overwriting this method if you want to initialize the children bricks in a
```
rizar: The same rephrasal as for `_push_allocation_config` is proposed. Also, I would elaborate on what `Initializable` does (propagates `weights_init` and `biases_init`).

docs/create_your_own_brick.rst
```diff
@@ -0,0 +1,371 @@
...
+  specific way. You should inherit from :class:`.Initializable` to initialize
+  the potential children bricks recursively.
+* :meth:`.Brick.get_dim`: implementing this function is useful if you want
+  to provide a simple way to get the dimensions of the inputs and outputs of
+  the brick.
+
+If you want to inherit from a specific brick, check its docstring to
+identify the particular methods to overwrite and the attributes to define.
+
+apply method
```
rizar: Application methods?

docs/create_your_own_brick.rst
```diff
@@ -0,0 +1,371 @@
...
+    ...         ...
+    ...         return something
+
+In the case above, it will automatically label the theano tensor variable
+``input1`` to ``Foo_apply_input1``, idem for ``input2`` and the output of the
+method.
+
+Under the hood, the ``@application`` decorator creates an object of class
+:class:`.Application`, named ``apply``, which becomes an attribute of the
+brick.
```
rizar: I would emphasize that it is an attribute of the brick class, not brick object.

docs/create_your_own_brick.rst
```diff
@@ -0,0 +1,371 @@
...
+
+Application properties
+""""""""""""""""""""""
+
+In the previous examples, variables were named with strings. If you need to
```
rizar: Variable names are always strings, the only difference is that sometimes they can not be specified at the time when the class is defined, but only when the object is created. This subsection has to be rewritten.

docs/create_your_own_brick.rst
```diff
@@ -0,0 +1,371 @@
...
+    ...         ...
+    ...     @apply.property('inputs')
+    ...     def apply_inputs(self):
+    ...         # Note that you can use any python code to define the name
+    ...         return self.fancy_name
+
+Using application calls
+"""""""""""""""""""""""
+
+You can also annotate specific variables by passing ``application_call`` as
```
rizar: The purpose of providing `application_call` is not "to annotate specific variables", but to allow adding auxiliary variables to the computation graph.

docs/create_your_own_brick.rst
```diff
@@ -0,0 +1,371 @@
...
+batch inputs and multiply them respectively by two matrices, resulting in two
+outputs.
+
+The first step is to identify which brick to inherit from. Clearly we are
+implementing a variant of the :class:`.Linear` brick. Contrary to
+:class:`.Linear`, ours has two inputs and two outputs, which means that we can
+not inherit from :class:`.Feedforward`, which requires a single input and a
+single output. Our brick will have to manage two shared variables
+representing the matrices to multiply the inputs with and thus, inheriting from
+:class:`.Initializable` makes perfectly sense as we will let the user decide
```
rizar: Inheriting from `Initializable` just adds `weights_init` and `biases_init` in this case, I am not sure I understand what "let the user decide which initialization scheme to use" means.

docs/create_your_own_brick.rst
```diff
@@ -0,0 +1,371 @@
...
+    ...         return output1, output2
+    ...
+    ...     def get_dim(self, name):
+    ...         if name in ['input1_', 'output1']:
+    ...             return self.linear1.get_dim(name)
+    ...         if name in ['input2_', 'output2']:
+    ...             return self.linear2.get_dim(name)
+    ...         super(ParallelLinear2, self).get_dim(name)
+
+You can test this new version as follows:
```
rizar: Here we should definitely pay attention to the fact that `weights_init` and `biases_init` are propagated behind the scenes to the linear children, and that this is in effect the main reason to use `Initializable`.

rizar:
@dmitriy-serdyuk asked me to take a look. I like the tutorial a lot, but in some cases it deviates from the truth a little bit.

adbrebs:
Yes I'm not yet very familiar with many blocks features, so I am not always sure I cover every points very well.

adbrebs:
@dmitriy-serdyuk @rizar Thanks for reviewing, it should be better now.

adbrebs:
I don't understand why these tests fail: ``tests.bricks.test_conv.FunctionTestCase (test_convolutional_sequence)``

rizar:
Sorry, out builds currently fail and we are trying to figure out why, see #836 

But since you can be sure that your PR does not break the convolutional bricks, you don't have to worry about them :)

adbrebs:
No risk that the documentation breaks the conv bricks ;).
@dmitriy-serdyuk is this good enough for merging? If yes, I can clean up the commit log before the merge.

docs/create_your_own_brick.rst
```diff
@@ -0,0 +1,387 @@
...
+  :class:`.Initializable` and :class:`.Feedforward`.
+* :class:`.BaseRecurrent`: the base class for recurrent bricks. Check the
+  :doc:`tutorial about rnns</rnn>` for more information.
+* many more!
+
+Let's say that you want to create a brick from scracth, simply inheriting
+from :class:`.Brick`, then you should consider overwriting the following
+methods (strictly speaking, all these methods are optional):
+
+* :meth:`.Brick.__init__`: you should pass by argument the attributes of your
```
rizar: Here is a question: I do not remember exactly if we have now a documentation that explains the life-cycle of the brick: creation, allocation of parameters, initialization of parameters, application. One can infer them from this document, but I think it must be stated clearly somewhere...

docs/create_your_own_brick.rst
```diff
@@ -0,0 +1,387 @@
...
+  before they allocate their parameters.
+* :meth:`.Brick._push_initialization_config`: you should consider
+  overwriting this method if you want to change the initialization schemes of
+  the children before they get initialized.
+  If the children bricks need to be initialized with the same scheme, then you
+  should inherit your brick from :class:`.Initializable`, which
+  automatically push the initialization schemes of your brick (provided as
+  arguments ``weights_init`` and ``biases_init`` of the constructor) to the
+  children bricks.
+* :meth:`.Brick.get_dim`: implementing this function is useful if you want
```
rizar: I would like to also explain here why using `get_dim` is preferable to accessing the attributes of the brick. The reason is that `get_dim` is the interface that is the same for all bricks, whereas the exact way allocation configuration is stored in a brick object might vary. For instance, `Feedforward` bricks have `input_dim` and `output_dim` attributes, whereas `Parallel` bricks have `input_dims` and `output_dims` (note plural instead of singular). When one wants to change the configuration of a brick, knowing how the attributes are called and what they mean is necessary, but when only read-only access required, `get_dim` should be sufficient. Thus, it was introduced to facilitate writing bricks that use other bricks without making too many assumptions about their interface.

Please let me know if this explanation was helpful and if you can integrate it in a concise way into your tutorial.

docs/create_your_own_brick.rst
```diff
@@ -0,0 +1,387 @@
...
+This tutorial assumes that you are already familiar with
+:doc:`bricks <bricks_overview>` and how to use them from a user point of view.
+
+
+Bricks ingredients and recipe
+-----------------------------
+
+All the bricks in blocks inherit directly or indirectly from the
+:class:`.Brick`. There is already a rich inheritance hierarchy of
+bricks implemented in blocks and thus, you should consider which brick level
```
rizar: Blocks, because it is a proper name

docs/create_your_own_brick.rst
```diff
@@ -0,0 +1,387 @@
...
+* :meth:`apply`: you need to implement a method that actually
+  implements the operation of the brick, taking as arguments the inputs
+  of the brick and returning its outputs. It can have any name and for simple
+  bricks is often named ``apply``. You should decorate it with the
+  :func:`.application` decorator, as explained in the next section. If you
+  design a recurrent brick, you should instead decorate it with the
+  :func:`.recurrent` decorator as explained in the
+  :doc:`tutorial about rnns</rnn>`.
+* :meth:`.Brick._allocate`: you should implement this method to allocate the
+  shared variables (often representing parameters) of the brick. In blocks,
```
rizar: **B**locks

docs/create_your_own_brick.rst
```diff
@@ -0,0 +1,387 @@
...
+  allocation.
+* :meth:`.Brick._push_allocation_config`: you should consider overwriting
+  this method if you want to change configuration of the children bricks
+  before they allocate their parameters.
+* :meth:`.Brick._push_initialization_config`: you should consider
+  overwriting this method if you want to change the initialization schemes of
+  the children before they get initialized.
+  If the children bricks need to be initialized with the same scheme, then you
+  should inherit your brick from :class:`.Initializable`, which
+  automatically push the initialization schemes of your brick (provided as
```
rizar: push*es*

docs/create_your_own_brick.rst
```diff
@@ -0,0 +1,387 @@
...
+and register auxiliary variables of the operation you implement.
+It is used as follows:
+
+    >>> class Foo(Brick): # doctest: +SKIP
+    ...     @application(inputs=['input1', 'input2'], outputs=['output'])
+    ...     def apply(self, input1, input2)
+    ...         ...
+    ...         return something
+
+In the case above, it will automatically label the theano tensor variable
```
rizar: Explanation should be accompanied by a code sample which shows what exactly you mean. Also, we should mention here not only how `.name` is changed, but also `.tag.names` and `.tag.roles`.

docs/create_your_own_brick.rst
```diff
@@ -0,0 +1,387 @@
...
+``input1`` to ``Foo_apply_input1``, idem for ``input2`` and the output of the
+method.
+
+Under the hood, the ``@application`` decorator creates an object of class
```
rizar: Also, a small code fragment would be helpful (which queries `Foo.apply` and shows what kind of object it is).

docs/create_your_own_brick.rst
```diff
@@ -0,0 +1,387 @@
...
+of your ``apply`` function, as shown in this example:
+
+    >>> class Foo(Brick): # doctest: +SKIP
+    ...     @application
+    ...     def apply(self, x, application_call):
+    ...         application_call.add_auxiliary_variable(x.mean())
+    ...         return x + 1
+
+You can retrieve these annotated variables as usual with the computational
+graph :class:`.ComputationGraph` and filters :class:`.VariableFilter`.
```
rizar: Please add an example and please say one sentence why one might need auxiliary variables at all.

docs/create_your_own_brick.rst
```diff
@@ -0,0 +1,387 @@
...
+two :class:`.Feedforward` bricks for which you do not need to
+specify the ``input_dim`` of ``brick2`` directly at its creation.
+
+    >>> class DoubleSequential(Feedforward):
+    ...     """Two sequential Feedforward bricks."""
+    ...     def __init__(self, brick1, brick2, **kwargs):
+    ...         super(Feedforward, self).__init__(**kwargs)
+    ...         self.brick1 = brick1
+    ...         self.brick2 = brick2
+    ...         self.input_dim = brick1.input_dim
```
rizar: That's actually an example of a bad behaviour, because `brick1` might not have `input_dim` at that stage. You can see `Sequence` for a more robust example of implementing `Feedforward` interface: https://github.com/mila-udem/blocks/blob/master/blocks/bricks/__init__.py#L641

docs/create_your_own_brick.rst
```diff
@@ -0,0 +1,387 @@
...
+    ... def __init__(self, attr1, attr1)
+    ...     ...
+
+This allows the user to specify ``attr1`` and ``attr2`` after the creation of 
+the brick. For example, the following ``DoubleSequential`` brick is composed of
+two :class:`.Feedforward` bricks for which you do not need to
+specify the ``input_dim`` of ``brick2`` directly at its creation.
+
+    >>> class DoubleSequential(Feedforward):
+    ...     """Two sequential Feedforward bricks."""
```
rizar: The word 'sequential' is a bit confusing here. I would rather call the class `ChainOfTwoFeedforward`, this is lengthy but to the point.

docs/create_your_own_brick.rst
```diff
@@ -0,0 +1,387 @@
...
+Note how ``get_dim`` is used to retrieve the ``input_dim`` of ``brick1``. You
+can now use a ``DoubleSeuential`` brick as follows.
+
+    >>> brick1 = Linear(input_dim=3, output_dim=2, use_bias=False,
+    ...                 weights_init=Constant(2))
+    >>> brick2 = Linear(output_dim=4, use_bias=False, weights_init=Constant(2))
+    >>>
+    >>> seq = DoubleSequential(brick1, brick2)
+    >>> seq.initialize()
+    >>> brick2.input_dim
```
rizar: I would appreciate if you also show here that setting `weights_init` for `DoubleSequential` would have the same effect as setting it for each of the bricks.

rizar: Oh I see, you kind of have it in your next example, and you don't inherit from `Initializable` here. All right!

docs/create_your_own_brick.rst
```diff
@@ -0,0 +1,387 @@
...
+    ...
+    ...     def __allocate(self, input_dim, output_dim, number):
+    ...         W = shared_floatx_nans((input_dim, output_dim),
+    ...                                name='W'+number)
+    ...         add_role(W, WEIGHT)
+    ...         self.parameters.append(W)
+    ...         self.add_auxiliary_variable(W.norm(2), name='W'+number+'_norm')
+    ...
+    ...     def _allocate(self):
+    ...         self.__allocate(self.input_dim1, self.output_dim1, '1')
```
rizar: `_allocate`?

docs/create_your_own_brick.rst
```diff
@@ -0,0 +1,387 @@
...
+   ...                         dtype=theano.config.floatX)
+   >>> x2_mat = 4 * numpy.ones((batch_size2, input_dim2),
+   ...                         dtype=theano.config.floatX)
+   >>>
+   >>> x1 = theano.tensor.matrix('x1')
+   >>> x2 = theano.tensor.matrix('x2')
+   >>> parallel1 = ParallelLinear(input_dim1, input_dim2, output_dim1,
+   ...                            output_dim2, weights_init=Constant(2))
+   >>> parallel1.initialize()
+   >>> # The weights_init initialization scheme is pushed to the children
```
rizar: This statement should be illustrated by actually checking `weight_init` of the children.

rizar: No, hold on, you don't  have children here! Then there is no pushing...

docs/create_your_own_brick.rst
```diff
@@ -0,0 +1,387 @@
...
+    ...         if name in ['input2_', 'output2']:
+    ...             return self.linear2.get_dim(name)
+    ...         super(ParallelLinear2, self).get_dim(name)
+
+You can test this new version as follows:
+
+   >>> parallel2 = ParallelLinear2(input_dim1, input_dim2, output_dim1,
+   ...                             output_dim2, weights_init=Constant(2))
+   >>> parallel2.initialize()
+   >>> output1, output2 = parallel2.apply(x1, x2)
```
rizar: Ok, and you here you should actually show that initialization scheme pushing took place.

docs/create_your_own_brick.rst
adbrebs: Could we point to the docstring of ``Brick``?

docs/create_your_own_brick.rst
adbrebs: Yes I agree and actually I don't see how lazy initialization can be "robust" in general. In the case of ``Sequence``, don't we have the same problem if the ``input_dim`` of the first child is not defined?

docs/create_your_own_brick.rst
adbrebs: __allocate is another function with a different signature

adbrebs:
@rizar I have updated the tutorial to take into account most of your comments.

docs/create_your_own_brick.rst
```diff
@@ -0,0 +1,419 @@
...
+    >>>
+    >>> from blocks.bricks import Brick, Initializable, Linear, Feedforward
+    >>> from blocks.bricks.base import lazy, application
+    >>> from blocks.bricks.parallel import Parallel
+    >>> from blocks.initialization import Constant
+    >>> from blocks.roles import add_role, WEIGHT
+    >>> from blocks.utils import shared_floatx_nans
+
+This tutorial explains how to create a custom brick, which is useful if you
+want to factorize a specific several combined operations (which can be made of
```
rizar: I would rephrase it as ```if you want to group several specific operations (which can be bricks themselves) into a single...```

docs/create_your_own_brick.rst
rizar: Yes, this is a good idea.

docs/create_your_own_brick.rst
rizar: We do have the same problem, but only upon requesting the `input_dim` property, not in the constructor. So you can create a sequence of bricks first and only then set all the dimensions.

docs/create_your_own_brick.rst
```diff
@@ -0,0 +1,419 @@
...
+    ...
+    ...     def _push_allocation_config(self):
+    ...         self.brick2.input_dim = self.brick1.get_dim('output')
+    ...
+    ...     @application
+    ...     def apply(self, x):
+    ...         return self.brick2.apply(self.brick1.apply(x))
+
+Note how ``get_dim`` is used to retrieve the ``input_dim`` of ``brick1``. You
+can now use a ``DoubleSeuential`` brick as follows.
```
rizar: It is not `DoubleSequential` anymore, you have changed the naming.

docs/create_your_own_brick.rst
```diff
@@ -0,0 +1,419 @@
...
+    >>> seq.initialize()
+    >>> brick2.input_dim
+    2
+
+
+Example
+-------
+
+For the sake of the tutorial, let's consider a toy operation that takes two
+batch inputs and multiply them respectively by two matrices, resulting in two
```
rizar: multiplies

adbrebs:
I have updated the PR. doctest should pass now.

rizar:
Good job, if the tests pass I will merge the PR right after we are finally done with the release (PR #883 ).

PR #779, opened 2015-08-02
allow the use of a subset of datasets sources

janchorowski:
The GradientDescent class forces that the iterator provides exactly the same sources as those demanded by the gradient graph. 

This is too restrictive - it is enough that all sources are provided, also one may want to use e.g. an example-id source for monitoring, but not for gradient computations.

### DIFF
--- a/blocks/algorithms/__init__.py
+++ b/blocks/algorithms/__init__.py
@@ -139,7 +139,14 @@ variable_mismatch_error = """
 Blocks tried to match the sources ({sources}) of the training dataset to \
 the names of the Theano variables ({variables}), but failed to do so. \
 If you want to train on a subset of the sources that your dataset provides, \
-pass the `sources` keyword argument to its constructor. """
+pass the `sources` keyword argument to its constructor. Or pass \
+on_unused_sources='warn' or on_unused_sources='ignore' to \
+the GradientDescent algorithm."""
+
+source_missing_error = """
+
+Blocks didn't find all the sources ({sources}) of the training dataset \
+that match the names of the Theano variables ({variables})."""
 
 
 class GradientDescent(DifferentiableCostMinimizer):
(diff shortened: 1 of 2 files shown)


dmitriy-serdyuk:
I also found this limitation inconvinient sometimes. But it may cause problems for novice users. Can you print a warning if the sources are not the same?

rizar:
Why not just [FilterSources](https://github.com/mila-udem/fuel/blob/master/fuel/transformers/__init__.py#L895)?

janchorowski:
Because it is an unnecessary complication.

I don't like it because I index stuff by utterance id. I have utterance 
ids everywhere. I don't want to filter them out.


Dmitry Bogdanov wrote:

rizar:
I see. What do you think about making this behaviour optional and switchable by a flag, e.g.

```
algorithm = GradientDescent(..., on_unknown_sources='ignore')
```

?

janchorowski:
I added that flag. Raises by default, can be configured to warn or ignore.

blocks/algorithms/__init__.py
```diff
@@ -192,7 +202,9 @@ class GradientDescent(DifferentiableCostMinimizer):
 
     """
     def __init__(self, step_rule=None, gradients=None, known_grads=None,
-                 consider_constant=None, **kwargs):
+                 consider_constant=None,
+                 on_unused_sources='warn',
```
rizar: You said it raises by default, but the code suggests that it just warns. I would raise by default, since this is the safest option and because this is the way it was done before. What do you think?

blocks/algorithms/__init__.py
```diff
@@ -234,12 +247,35 @@ def initialize(self):
...
+                                 variables=in_names))
+        if not set(batch.keys()).issubset(set(in_names)):
+            if self.on_unused_sources == 'ignore':
+                pass
+            elif self.on_unused_sources == 'warn':
+                if not hasattr(self, '_unused_sources_warning'):
+                    logger.warn(variable_mismatch_error.format(
+                        sources=batch.keys(),
+                        variables=in_names))
+                self._unused_sources_warning = True
```
rizar: Naming: maybe `_unused_source_warned`? Right now it seems like it is name of a flag, which it is not.

blocks/algorithms/__init__.py
janchorowski: Oh, my bad. Sorry

blocks/algorithms/__init__.py
janchorowski: Fixed

PR #780, opened 2015-08-02
Named aggregation schemes, Sum scheme and more debug information

janchorowski:
Small fixes to aggregations

### DIFF
--- a/blocks/monitoring/aggregation.py
+++ b/blocks/monitoring/aggregation.py
@@ -170,12 +170,11 @@ class Sum_(AggregationScheme):
         summand_zeros = tensor.as_tensor(self.summand).zeros_like()
 
         conditional_update_sum = self.summand + tensor.switch(initialized,
-                                                       summand_acc,
-                                                       summand_zeros)
+                                                              summand_acc,
+                                                              summand_zeros)
 
         initialization_updates = [(summand_acc,
                                    tensor.zeros_like(summand_acc)),
-
                                   (initialized, 0.)]
         accumulation_updates = [(summand_acc,
                                  conditional_update_sum),
--- a/blocks/monitoring/aggregation.py
+++ b/blocks/monitoring/aggregation.py
@@ -138,11 +138,63 @@ class Mean(AggregationScheme):
         return aggregator
 
 
-def mean(numerator, denominator=1.):
+def mean(numerator, denominator=1., name=None):
     """Mean of quantity (numerator) over a number (denominator) values."""
     variable = numerator / denominator
     variable.tag.aggregation_scheme = Mean(numerator, denominator)
-    variable.name = numerator.name
+    if name is not None:
+        variable.name = name
+    else:
+        variable.name = numerator.name
+    return variable
+
+
+class Sum_(AggregationScheme):
+    """Aggregation scheme which computes the sum.
+
+    Parameters
+    ----------
+    summand : :class:`~tensor.TensorVariable`
+        Theano variable to be summed
+
+    """
+    def __init__(self, summand):
+        self.summand = summand
+
+    def get_aggregator(self):
+        initialized = shared_like(0.)
+        summand_acc = shared_like(self.summand)
+
+        # Dummy default expression to use as the previously-accumulated
+        # value, that has the same shape as the new result
+        summand_zeros = tensor.as_tensor(self.summand).zeros_like()
+
+        conditional_update_sum = self.summand + tensor.switch(initialized,
+                                                       summand_acc,
+                                                       summand_zeros)
+
+        initialization_updates = [(summand_acc,
+                                   tensor.zeros_like(summand_acc)),
+
+                                  (initialized, 0.)]
+        accumulation_updates = [(summand_acc,
+                                 conditional_update_sum),
+                                (initialized, 1.)]
+        aggregator = Aggregator(aggregation_scheme=self,
+                                initialization_updates=initialization_updates,
+                                accumulation_updates=accumulation_updates,
+                                readout_variable=summand_acc)
+        return aggregator
+
+
+def sum_(summand, name=None):
+    """Sum of a variable over batches."""
+    variable = summand.copy()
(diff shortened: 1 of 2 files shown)


blocks/monitoring/aggregation.py
```diff
@@ -138,11 +138,62 @@ def get_aggregator(self):
         return aggregator
 
 
-def mean(numerator, denominator=1.):
+def mean(numerator, denominator=1., name=None):
     """Mean of quantity (numerator) over a number (denominator) values."""
     variable = numerator / denominator
     variable.tag.aggregation_scheme = Mean(numerator, denominator)
-    variable.name = numerator.name
+    if name is not None:
```
rizar: Not sure if I like it. One can set name to the variable returned by `mean`, which is almost as simple. Otherwise you have to add these 4 lines to every `mean`-like function, e.g. `sum_` below.

janchorowski: There are only 2 of those functions right now - mean and sum. And most 
functions which create stuff in Theano take a name argument even though 
you can set the name later via tha name tag.



Dmitry Bogdanov wrote:

rizar: I am not sure they do, e.g. `mean` doesn't.

```
In [5]: theano.tensor.mean([1, 2, 3], name='abc')
<ipython-input-5-ddb6db7659e4> in <module>()
----> 1 theano.tensor.mean([1, 2, 3], name='abc')

TypeError: mean() got an unexpected keyword argument 'name'
```

janchorowski: Yes, but matrix, shared etc do add them.

The variable after aggregation will typically be named - because it will be used for reports and such. I added the fix because I first used named_copy to set the name, which removed the aggregation scheme. Adding the name like that is IMHO helpful.

rizar: `matrix` and `shared` are a bit different because they create input variables from scratch. This way it is natural for them to accept names. Correct me I am wrong, but no Theano expression (i.e. a function that takes input variables and creates output variables) supports `name` keyword, and if we want to mimic Theano here, we should not have it as well.

I am quite disappointed in `named_copy`, to be honest, I think it only deteriorates readability.

Also now we have `mean` and `sum`, but I also anticipate `min` and `max`, and this little piece of code that sets the name will have to be copied even more times. I think that the ratio between the convenience this feature brings and the code bloat it brings is not favorable. 

I would be maybe a bit less skeptical if the `name` argument was added by a decorator, reducing the need for copy-paste. This would however decrease the readability compared to straight-forward adding the name to the returned variable, because one would have to read the code of the decorator to see how `name` is handled.

janchorowski: You are correct about theano behavior, still I guess that since adding 
an explicit aggregation scheme is a strong sign of creating an 
end-product which should have a meaningful name, it is the readability 
of the code that uses these constructs that matters.

We can meet in the middle by delegating the name handling to an 
auxiliary function and calling it in all the aggregation schemes.

I find it much nicer to write in the end code:
```
extension.monitor(mean(smth, denom), name="cost-per-token"))
```

rather than:
```
per_token_cost = mean(smth, denom)
per_token_cost.name = "cost-per-token"
extension.monitor(per_token_cost)
```

and this is the code bloat I want to get rid of.

Dmitry Bogdanov wrote:

rizar: We don't really have `extension.monitor(...)`  in Blocks now. One creates `TrainingDataMonitoring` or `DataStreamMonitoring` with the list of all variables to be monitored. 

There is also a homogeneity argument: if you create a variable for monitoring without adding an aggregation scheme, you have to give it a name in a separate line. 

Maybe we should make sure that `named_copy` doesn't lose the aggregation scheme and whatever else it loses?

janchorowski: I know about TrainingDataMonitoring but the problem is the same - you 
want to pass a named variable.

I was thinking about fixing named_copy but I guess it will be too error 
prone - we either stay compatible with variable.copy() which removes 
annotations, or we will introduce very subtle bugs - trust me, It took 
me half a day to figure that I lost my aggregation schemes and I 
discovered it only because I didn't rely on the monitoring to do the 
final scoring.

So we either work with Theano guys making sure that normal copies made 
by theano preserve the annotations, or we name the variables in a 
convenient way and discourage making copies.

Dmitry Bogdanov wrote:

rizar: Without opening a can of worms by asking ourselves whether or not `variable.copy()` should preserve `variable.tag`, we could just use a new utility function `rename` instead of `named_copy` that would change the name of its argument and return it. Moreover, it is a shame that `setattr` doesn't  return its first operand, otherwise we could just write `setattr(mean(x, y), 'name', 'per-token-cost')`.

janchorowski: Then maybe just add a `setattr_` utility that would return its argument? 
So that is is obvious what it does and that it mutates its argument from 
the name itself?

Dmitry Bogdanov wrote:

rizar: I am tempted to agree that this is a reasonable solution, but I am 99% sure @bartvm will hate us for that. Because this is a non-standard extension of Python core, and our need for it is barely justifiable by this additional single line `per_token_cost.name = 'per-token-cost'` one has to write each time.

janchorowski: We can wait and see what he says. There is no rush with this one.

rizar: He is doing an internship now and answers only very occasionally.

For now, I would not mind merging the new aggregation scheme.

janchorowski:
Sorry - I relied on GH' web editor to remove whitespace, it does not do so. Fixed

rizar:
Now that `.copy` method of Theano variables support `name` argument, this PR should be significantly refreshed. @janchorowski , do you plan to do this?

rizar:
Closed due to inactivity

PR #781, opened 2015-08-02
Performance

janchorowski:
A few speed fixes:

replace ifelse with switch (ifelse has no GPU implementation, kills performance in monitors),
prevent the creation of a large temp array in l2norm and defer the isnan/isinf check to only after the arrays are reduced in RemoveNotFinite (inf and nan will propagete with the sum anyway).

Note that Theano has a very poor reduction implementation and gradient norm monitoring. checking for nans can easily take more that gradient computation itself.

### DIFF
--- a/blocks/theano_expressions.py
+++ b/blocks/theano_expressions.py
@@ -14,11 +14,9 @@ def l2_norm(tensors):
         The tensors.
 
     """
-    flattened = [tensor.as_tensor_variable(t).flatten() for t in tensors]
-    flattened = [(t if t.ndim > 0 else t.dimshuffle('x'))
-                 for t in flattened]
-    joined = tensor.join(0, *flattened)
-    return tensor.sqrt(tensor.sqr(joined).sum())
+    summed = [tensor.sqr(tensor.as_tensor_variable(t)).sum() for t in tensors]
+    joined = tensor.stack(*summed)
+    return tensor.sqrt(joined.sum())
 
 
 def hessian_times_vector(gradient, parameter, vector, r_op=False):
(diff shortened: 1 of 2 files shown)


blocks/theano_expressions.py
```diff
@@ -14,11 +14,11 @@ def l2_norm(tensors):
         The tensors.
 
     """
-    flattened = [tensor.as_tensor_variable(t).flatten() for t in tensors]
-    flattened = [(t if t.ndim > 0 else t.dimshuffle('x'))
-                 for t in flattened]
-    joined = tensor.join(0, *flattened)
-    return tensor.sqrt(tensor.sqr(joined).sum())
+    grand_sum = 0.0
```
rizar: We had quite a discussion once with @bartvm about the right way to sum here. The method you propose has a disadvantage that the computation graph will be as deep as many tensor are given as arguments. This might hurt at compile time and also makes it harder to investigate debugprints.

On the other hand, I agree that the old implementation is memory inefficient. 

It seems like one can combine the advantages of two by creating a vector of scalars in the way you do it and joining these scalars like we did. What do you think?

janchorowski: Hi,

1. I need to fix a bug with the ifelse -> tensor.switch change 
(something about broadcasting).

2. What you propose is also OK, the main problem here is poor CAReduce 
performance. This improves things a little bit, but the real fix is 
here: 
https://github.com/janchorowski/Theano/commit/2e85535b36c90547bfa89bf954db8223dc935ed2

Which gets this 20 time faster.

Dmitry Bogdanov wrote:

rizar: Wow! You should probably get in touch with Theano devs if you have such an amazing fix for them.

janchorowski: I wrote to the mailing list. Maybe they will pick it up.

Dmitry Bogdanov wrote:

janchorowski:
OK, apparently getting rid of ifelse in evaluators is not as easy as I thought, so I reverted it. I also added stacking of all partial sums.

PR #782, opened 2015-08-02
Gain in orthogonal initialization

janchorowski:


dmitriy-serdyuk:
Is it a common name, `gain`?

janchorowski:
I was referring to: http://arxiv.org/pdf/1312.6120v3.pdf though he introduces another gain term. Maybe scale would be better. Anyway, it is a useful parameter.

dmitriy-serdyuk:
Yeah, it is useful, I would call it `scale`.

janchorowski:
renamed

blocks/initialization.py
```diff
@@ -159,6 +159,9 @@ class Orthogonal(NdarrayInitialization):
     Only works for 2D arrays.
 
     """
+    def __init__(self, scale=1):
```
rizar: Please add the new parameters to the docstring. Literature reference would be appreciated as well.

janchorowski:
Added doc

rizar:
I am sorry for not spotting everything from the first look, but it also needs a unit-test. Could you please make one?

janchorowski:
Sorry for forgetting about them in the first place. Here they are

janchorowski:
The failing test is unrelated to this change - what do I do about it?

rizar:
Unfortunately we seems to have a bit of stochasticity in our tests :)

PR #783, opened 2015-08-02
Named loggers

janchorowski:
Added __name__ for loggers to make it easier to know where the messages come from.


dmitriy-serdyuk:
LGTM.

dmitriy-serdyuk:
I remember that @bartvm mentioned some kind of problems with logger in Python2. I'll try to find that discussion.

PR #785, opened 2015-08-02
More feedforward layers: PReLU and Liear + Activation

janchorowski:
1. Maxout doesn't play nicely with the MLP and thus we have LinearMaxout. This makes swapping activations difficult. It is easier to make a Sequence of Linear+Activation and then swap individual ones

2. Implementation of  PReLU (ReLU with learned negative slope)

### DIFF
--- a/blocks/bricks/__init__.py
+++ b/blocks/bricks/__init__.py
@@ -512,6 +512,8 @@ class Softmax(Activation):
         x : :class:`~tensor.TensorVariable`
             Each slice along axis represents energies of a distribution,
             that is pre-softmax values.
+        average : bool
+            Whether to compute the mean cost, or return individual costs
 
         Returns
         -------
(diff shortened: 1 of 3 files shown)


blocks/bricks/__init__.py
```diff
@@ -314,6 +315,75 @@ def _set_dim(self, value):
     input_dim = output_dim = property(_get_dim, _set_dim)
 
 
+class PReLU(Feedforward, Initializable):
```
rizar: I would say that cutting-edge research bits and pieces belong to `blocks-extras`, not `blocks`.

rizar:
Maybe I am not getting something, but I don't feel `LinearMaxout` was a good idea in the first place. Why not directly create a `Sequence` with `Linear` and `Maxout` as  its elements?

janchorowski:
Because from the user's perspective what you care about is the number of units and the activation function. For a sequence of LinearActivations you just have to match their input and output dims, and this is easy. It is the LinearMaxout brick that remembers about the multiplication by the number of maxout pieces.

So it is very easy to say that you want a net of 5 layers with 1024 ReLU units (Sequence 5*LinearRelu(input_dim=output_dim=1024)), then swap the ReLUs for 5-piece Maxouts (Sequence 5*LinearMaxout(input_dim=output_dim=1024)). If you decouple the linear brick from the activation, you need to pepper your code with checks for the activation and its influence on the dimensionality. And this doesn't work - in fact you can't easily use Maxout in the MLP brick.

So personally I would get rid of MLP brick, or implement it as a thin wrapper over a sequence of LinearActivation bricks (since there should be an MLP in a neural network library). This is easier and more consistent to use.

janchorowski:
> Maybe I am not getting something, but I don't feel LinearMaxout was a good idea in the first place. Why not directly create a Sequence with Linear and Maxout as its elements?

I guess your question was even more specific - LinearMaxout is the sequence of Linear + Maxout and maybe it should be implemented as such. However, it needs the extra logic to set proper dimensions based on the requested intput and outptu dims. 

Then the LinearActivation is just a convenience that takes care of the documentation.

rizar:
Okay, I see the motivation. Not sure if `MLP` should be rewritten this way though, because as it is written now it supports user-defined activations, whereas your way of chaining requires `LinearXXX` to be available in Blocks.

I will read the code more attentively then.

janchorowski:
The removal of mlp was a little far fetched. But it is not good that there
is no obvoius way of using the mlp with maxout.
3 sie 2015 14:05 "Dmitry Bogdanov" <notifications@github.com> napisał(a):

blocks/bricks/__init__.py
```diff
@@ -533,6 +605,108 @@ def categorical_cross_entropy(self, y, x):
...
+    Notes
+    -----
+    See http://bugs.python.org/issue12773.
+
+    """
+    def __new__(cls, name, bases, classdict):
+        classdict['__doc__'] = \
+            """Linear transformation followed by elementwise application of \
+            {0} function.""".format(name[6:].lower())
+        if 'apply' in classdict:
```
rizar: Should we raise if `'apply' not in classdict`?

blocks/bricks/__init__.py
```diff
@@ -533,6 +605,108 @@ def categorical_cross_entropy(self, y, x):
...
+
+                """.format(name[6:].lower())
+        return super(LinearActivationDocumentation, cls).__new__(cls, name,
+                                                                 bases,
+                                                                 classdict)
+
+
+@add_metaclass(LinearActivationDocumentation)
+class LinearActivation(Initializable, Feedforward):
+    """Base class that ads documentation and has all the logic."""
```
rizar: adds

rizar:
It looks good, but please remove PReLU from this PR. I would be glad to commit them to blocks-extras though.

blocks/bricks/__init__.py
janchorowski: This should not happen - similarly to the `ActivationDocumentation` class the `LinearActivationDocumentation` is supposed to only work with `LinearActivation` which provides an `apply` method. But a check will not hurt.

janchorowski: On a second though - we shouldn't do any checks. All `LinearActivationDocumentation` does is to add documentation. If there is no apply, it will just not document it.

janchorowski:
OK, I pushed the changes

blocks/bricks/__init__.py
```diff
@@ -512,6 +512,8 @@ def categorical_cross_entropy(self, y, x):
         x : :class:`~tensor.TensorVariable`
             Each slice along axis represents energies of a distribution,
             that is pre-softmax values.
+        average : bool
```
rizar: Is it here just by accident?

tests/bricks/test_bricks.py
```diff
@@ -293,6 +294,36 @@ def test_linear():
     assert_allclose(y.eval({x: x_val}), x_val.dot(2 * numpy.ones((16, 8))))
 
 
+def test_linear_activations_with_prelu():
```
rizar: Can you please change the name of this test to get rid of the 'prelu' part?

blocks/bricks/__init__.py
janchorowski: Yes, I'll reread all the code- I was splitting many small changes.

rizar:
@janchorowski , do you intend to finish this?

rizar:
Closed due to inactivity.

Issue #758, opened 2015-07-10
Sensible behaviour when the main loop is not picklable.


rizar:
Currently Blocks does not offer and serialization method for the case when for some reason the whole main loop is not picklable. I guess it is a bit too tough in the light of recent discussions on the mailing list: https://groups.google.com/forum/#!topic/blocks-users/5PXu9-LBZIM and https://groups.google.com/forum/#!topic/blocks-users/jns-KKWTtko

While pickling should remain the main serialization method, it would be nice to optionally not do it. E.g. the current zip file to which we save everything could simply not contain the `pkl` field.

dwf:
I feel like restoring the `Dump` extension under `blocks.extras` would help with this.

rizar:
I think we can do better. For one, instead of saving all shared variables separately and referring them by persistent id, we should save only the parameters I think. This way the `main_loop.zip` will be very close to the old `.npz` files created by `Dump`. Furthermore, we can pickle the log and/or the iteration state separately and also store them as fields of `zip` file (see #746). Finally, all these should be referred from the main loop pickle by the persistent id mechanism as it is done now, with the difference that pickling the main loop should be optional.

Roughly that way we can get an ultimate hybrid of old `Dump` and `Checkpoint`, I guess.

memimo:
The easiest solution in my opinion is to add an option in the checkpoint e.g. save_main_loop, and if it's disabled it will pass main_loop.model.parameters to secure_dump instead of main_loop
It does not solve all the issues you raised above, but it's a quick fix and I can make a PR.

rizar:
Well, this will further increase the entropy we currently have in `Checkpoint`. If `save_main_loop` were a zip file would be saved, and if not, a pickle file. In addition `save_separately` saves pickles. It gets really complicated. Quick fixes are great, but now with `Checkpoint` a complete solution is within reach and I wouldn't like introducing yet another flag in this case. 

The complete solution I propose is the following: zip files are _the_ serialization format supported by Blocks. The model parameters are always saved as numpy arrays in the zip file. Optionally other components of the main loop are also saved as fields in this file, with some name mangling involved to distinguish them from parameters. Optionally a pickle of the whole main loop is saved, with the persistent id mechanism being used to refer to the parameters and other components in the same zip file. This way we (a) store everything in one place and in one format (b) use disk space efficiently and don't waste time pickling things many times.

BTW, @janchorowski, here is a plan for new `Checkpoint` which will be in all aspects better then the old GH one.

Issue #746, opened 2015-07-02
Log should be pickled separately from the main loop


rizar:
It could be stored as another field in the `.zip` file created by `Checkpoint`. This will greatly improve serialization robustness. Upon loading it can be put to its usual `main_loop.log` place.

PR #786, opened 2015-08-04
WIP: New Serialization

rizar:
The current serialization used in Blocks has the following shortcomings:

- although the parameters are indeed saved robustly as numpy arrays, they are contained in a zip file with a lot of rubbish, which makes it in practice not possible to fetch them

- only parameters are saved as additional fields in the dump, for all the rest one can only use `save_separately` which is less convenient, since it produces separate files, #746 

- if main loop is itself not picklable, nothing will be saved, #758 

This PR proposes a new serialization, which does not have these issues. It is so far only a working draft, but I put some effort into docstring of the new `blocks/serialization.py` module and kindly invite you to read it and tell, if you like it and if you think that this PR should be finished and merged.

TODO:
- [ ]  documentation is missing for many functions
- [ ] handle usual shared variables (not created by bricks)
- [ ]  handle shared variables without names
- [ ]  handle variables with duplicate names 
- [ ]  support CUDA

### DIFF
--- a/blocks/extensions/saveload.py
+++ b/blocks/extensions/saveload.py
@@ -6,7 +6,7 @@ from six.moves import cPickle
 
 from blocks.extensions import SimpleExtension, TrainingExtension
 from blocks.utils import reraise_as
-from blocks.serialization import secure_dump, load, load_parameter_values
+from blocks.serialization import secure_dump, load, load_parameters
 
 logger = logging.getLogger(__name__)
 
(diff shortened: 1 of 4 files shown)


vdumoulin:
Does this also solve stack limit issues when retrieving parameters from a really deep theano graph?

rizar:
Partially. That is the parameters will be always 100% saved without crash, period. But if you also pickle the whole thing, you will still get a crash. On the bright side, with this new serialization pickling the whole thing can be switched off.

rizar:
@vdumoulin , I  updated the doctest to explain what I mean.

vdumoulin:
Great, thanks!

dmitriy-serdyuk:
Great work! Looks very good.

I'll look more carefully later. Some parameters can not be present in `model.parameters`, for example if they are not used in the training cost computation. As I can see, they are not going to be pickled with the persistent id.

rizar:
>  Some parameters can not be present in model.parameters, for example if they are not used in the training cost computation.

I will find this situation rather unlikely, but I agree that there should be full control of what is being called parameters, and I plan to have at the following  levels: 

- `model.parameters` is mutable, that is the parameters set can be changed for all extensions at once

- `Checkpoint` should accept `parameters` keyword to directly control what is being saved

rizar:
A note: make sure that it works without CUDA installed.

PR #790, opened 2015-08-05
Make Linear docstring clearer.

### LINKED ISSUE #789: Linear docstring makes no mention of use_bias
The convention in Blocks has been to document class arguments only in the class where the constructor defines them. In the case of the Linear brick, I feel like this is a bad policy, since it makes explicit mention of the *optional* nature of the bias, but provides no guidance whatsoever as to how to set it!

We should at *least* direct the user to the Initializable docstring for information on how to turn off the bias.

dwf:
Fixes #789.

* Mentions that bias is on by default.
* Explicitly points to `Initializable` docstring re: disabling bias.

### DIFF
--- a/blocks/bricks/__init__.py
+++ b/blocks/bricks/__init__.py
@@ -181,8 +181,9 @@ class Feedforward(Brick):
 class Linear(Initializable, Feedforward):
     r"""A linear transformation with optional bias.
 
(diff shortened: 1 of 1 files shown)


dwf:
Merging with @vdumoulin's blessing. I don't know what is up with coveralls, the decrease makes no sense.

PR #796, opened 2015-08-11
MLP & LinearMaxout: child brick names.

### LINKED ISSUE #795: LinearMaxout and MLP bricks do not name their Linear bricks appropriately
It makes it very hard to filter for the inputs to a particular Linear brick if you have multiple MLPs, etc. on the go.

Specifically, {LinearMaxout, MLP} should incorporate the name of the {LinearMaxout, MLP} brick into its child bricks.

dwf:
Incorporate the parent brick name into the names of any automatically-created bricks in MLP and LinearMaxout (this means the Linear bricks for MLP and the Linear and Maxout bricks for LinearMaxout).

Fixes #795.

### DIFF
--- a/blocks/bricks/__init__.py
+++ b/blocks/bricks/__init__.py
@@ -390,8 +390,12 @@ class LinearMaxout(Initializable, Feedforward):
     @lazy(allocation=['input_dim', 'output_dim', 'num_pieces'])
     def __init__(self, input_dim, output_dim, num_pieces, **kwargs):
         super(LinearMaxout, self).__init__(**kwargs)
-        self.linear = Linear()
-        self.maxout = Maxout()
+        linear_kwargs, maxout_kwargs = {}, {}
+        if 'name' in kwargs:
+            linear_kwargs['name'] = '_'.join([kwargs['name'], 'linear'])
+            maxout_kwargs['name'] = '_'.join([kwargs['name'], 'maxout'])
+        self.linear = Linear(**linear_kwargs)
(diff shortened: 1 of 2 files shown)


dwf:
I seem to have stumbled upon an interesting GitHub bug whereby I force-pushed to a branch after I had started the pull request creation process, but before I'd created the pull request. Note that 2d969ae is actually the parent commit of 8b07a8.

dwf:
@rizar I notice I'm breaking some Model tests. Is Model particularly well-supported, is it hanging around, and should I be using it somehow instead of the contents of this pull request.

dwf:
In fact I think the right place for this sort of thing is either `Selector` or (preferably, in my opinion) `VariableFilter`, so I'll close this and the ticket it references.

rizar:
Actually, @janchorowski , had since long proposed a solution for that. `Selector` should be able to select bricks by a regular expression for their hierarchical name, e.g.

```
# suppose you built your computation graph and `cost` is the output
model = Model(cost)
linear_of_linearmaxout = Selector(model.get_top_bricks()).select('*/linear_maxout/linear')
```

Moreover, it was planned to  refactor `Model` to make it simply a `ComputationGraph` that knows about bricks and e.g. can create `Selector`'s under the hood, in which case the code above should become a one-liner

```
linear_of_linearmaxout = Model('cost').select_brick('*/linear_maxout/linear')
```

The model refactoring is on my schedule after #786 is in, the work on which I plan to resume this week, after yet another paper rush is over.

PR #798, opened 2015-08-12
Added a Softplus brick.

ASalvail:
This brick only interfaces with the softplus function from theano.

### DIFF
--- a/blocks/bricks/__init__.py
+++ b/blocks/bricks/__init__.py
@@ -491,7 +491,7 @@ class Logistic(Activation):
 
 
 class Softplus(Activation):
-    """ Softplus brick
+    r""" Softplus brick.
 
     The softplus is defined as :math:`\zeta(x) = \log(1+e^x)`.
 
--- a/blocks/bricks/__init__.py
+++ b/blocks/bricks/__init__.py
@@ -491,6 +491,15 @@ class Logistic(Activation):
 
 
 class Softplus(Activation):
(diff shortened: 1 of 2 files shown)


rizar:
A paper reference in the docs and a test added to `tests/bricks/test_bricks.py:test_activations` would be appreciated. Except that, looks good!

ASalvail:
Forgot the test. Will do.
As for the paper, I have no idea who invented it.

rizar:
[this paper](http://machinelearning.wustl.edu/mlpapers/paper_files/AISTATS2011_GlorotBB11.pdf) cites some other paper as the softplus origin, I guess we should follow the example.

rizar:
Sorry for failing tests, it is because of https://github.com/mila-udem/blocks-examples/issues/34

ASalvail:
Alright, I added the reference and test, plus some bonus tests for the Softmax and Logistic bricks.

ASalvail:
Is test coverage computed according to the test functions name? If it is, I could split the `test_activations` in 5 separate methods.

ASalvail:
Isn't the doc overwritten by `ActivationDocumentation`? If so, how do you propose the referenced paper should be added?

rizar:
You are right, unfortunately it is. For now, let's leave it as is.

rizar:
Sorry for delay, forgot about this PR.

Issue #768, opened 2015-07-21
Cost and CostMatrix


dwf:
The `Cost` base class is insufficiently general (relies on `y` and `y_hat` being required arguments, and the only ones, to `apply`, where one can imagine costs that do not use either, or use more arguments). Honestly it doesn't seem terribly useful to begin with: it enforces that a Cost is a Brick that has an apply method. Not very interesting.

`CostMatrix` is somewhat more useful, because it lets you reuse the common trope where costs are a sum over columns/mean over rows. It would be nice if `CostMatrix` did not enforce that there be exactly two arguments, and just passed on the *args and **kwargs received by `apply` as-is to `cost_matrix`.

dwf:
Closed by #805.

PR #805, opened 2015-08-15
Make Cost and CostMatrix more general.

dwf:
There's no reason that I can see to constrain them to having `y` and `y_hat` arguments.

### DIFF
--- a/blocks/bricks/cost.py
+++ b/blocks/bricks/cost.py
@@ -11,7 +11,7 @@ from blocks.bricks.base import application, Brick
 class Cost(Brick):
     @abstractmethod
     @application
-    def apply(self, y, y_hat):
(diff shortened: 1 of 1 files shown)


rizar:
+1

In general seems like `Cost` brick carries no semantic load what-so-ever. Do we need it?

dwf:
Could be useful with the VariableFilter, I guess.

rizar:
Can `VariableFilter` filter by base class?

Anyway, this PR looks OK.

dwf:
It uses isinstance if it encounters a Brick class
https://github.com/mila-udem/blocks/blob/master/blocks/filter.py#L140-L141
so I think so.

dwf:
Woops, this closes #768.

PR #808, opened 2015-08-17
Improve Selector.get_parameter documentation.

### LINKED ISSUE #771: Better document selection of descendant parameters of bricks in Selector
It would be nice to, for example, select all of the parameters in a nested, composite brick using the VariableFilter. A good example would be encoders and decoders, each of which might consist of many internal bricks. Another good example is the generator/discriminator in adversarial nets. This should probably be implemented separately from the `bricks` argument, perhaps with an `ancestor_bricks` argument?

dwf:
Fixes #771.

### DIFF
--- a/blocks/select.py
+++ b/blocks/select.py
@@ -158,20 +158,45 @@ class Selector(object):
         return Selector(current_bricks)
 
     def get_parameters(self, parameter_name=None):
-        """Returns parameters the selected bricks and their ancestors.
+        r"""Returns parameters from selected bricks and their descendants.
 
         Parameters
         ----------
-        parameter_name : :class:`Path.ParameterName`
-            If given, only parameters with the name `parameter_name` are
-            returned.
+        parameter_name : :class:`Path.ParameterName`, optional
+            If given, only parameters with the assigned variable name
+            `parameter_name` are returned.
 
         Returns
         -------
         parameters : OrderedDict
             A dictionary of (`path`, `parameter`) pairs, where `path` is
-            the string representation of the part to the parameter,
-            `parameter` is the parameter.
+            a string representation of the path in the brick hierarchy
+            to the parameter (i.e. the slash-delimited path to the brick
+            that owns the parameter, followed by a dot, followed by the
+            parameter's name), and `parameter` is the Theano variable
+            representing the parameter.
+
+        Examples
+        --------
+        >>> from blocks.bricks import MLP, Tanh
+        >>> from blocks.initialization import Constant, Uniform
+        >>> mlp = MLP([Tanh(), Tanh(), Tanh()], [5, 7, 11, 2],
+        ...           weights_init=Uniform(width=0.01),
+        ...           biases_init=Constant(0))
+        >>> mlp.initialize()
+        >>> selector = Selector([mlp])
+        >>> selector.get_parameters()  # doctest: +NORMALIZE_WHITESPACE
+        OrderedDict([('/mlp/linear_0.W', W), ('/mlp/linear_0.b', b),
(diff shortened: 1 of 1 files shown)


dwf:
@rizar Please have a look when you get a chance and merge if you have no criticisms.

blocks/select.py
```diff
@@ -158,20 +158,45 @@ def select(self, path):
         return Selector(current_bricks)
 
     def get_parameters(self, parameter_name=None):
-        """Returns parameters the selected bricks and their ancestors.
+        r"""Returns parameters from selected bricks and their descendants.
```
rizar: Sorry, I often confuse the words "ancestor" and "descendant", my bad!

blocks/select.py
```diff
@@ -158,20 +158,45 @@ def select(self, path):
...
 
         Parameters
         ----------
-        parameter_name : :class:`Path.ParameterName`
-            If given, only parameters with the name `parameter_name` are
-            returned.
+        parameter_name : :class:`Path.ParameterName`, optional
+            If given, only parameters with the assigned variable name
```
rizar: "assigned attribute"?

blocks/select.py
```diff
@@ -158,20 +158,45 @@ def select(self, path):
...
+            representing the parameter.
+
+        Examples
+        --------
+        >>> from blocks.bricks import MLP, Tanh
+        >>> from blocks.initialization import Constant, Uniform
+        >>> mlp = MLP([Tanh(), Tanh(), Tanh()], [5, 7, 11, 2],
+        ...           weights_init=Uniform(width=0.01),
+        ...           biases_init=Constant(0))
+        >>> mlp.initialize()
```
rizar: Initialization isn't necessary for this example, isn't it?

rizar:
Looks good except for my very minor remarks in the code.

blocks/select.py
dwf: If I omit it I'm told that MLP doesn't have a parameters member.

blocks/select.py
dwf: I think the wording in 4f3a075 is maximally clear.

blocks/select.py
rizar: Yeah, but you can do `allocate` which will create the parameters but not initialize it. In addition you can omit specification of `weights_init` and `biases_init`.

dwf: Hmm. Do you think it's wise to do such contrived things just to minimize the dependency on other code? It makes the example less realistic and IMHO less useful.

rizar: I do not find that much contrived, perhaps because for me optional initialization is one of Blocks basic features. It origins from the frustration I experienced when GroundHog spent 2 minutes generating huge orthogonal matrices for me which were later replaced by parameters loaded from the disk.

Also optional initialization is great for simple tests like the one you want to write there. I don't think we should hide this feature from the user.

dwf: Okay, you've convinced me. 45c400a should be more to your liking.

rizar:
Great, love merging PRs that improve the docs so radically! Feel free to merge when the tests pass.

blocks/select.py
```diff
@@ -158,20 +158,43 @@ def select(self, path):
...
+            a string representation of the path in the brick hierarchy
+            to the parameter (i.e. the slash-delimited path to the brick
+            that owns the parameter, followed by a dot, followed by the
+            parameter's name), and `parameter` is the Theano variable
+            representing the parameter.
+
+        Examples
+        --------
+        >>> from blocks.bricks import MLP, Tanh
+        >>> from blocks.initialization import Constant, Uniform
```
dwf: Just realized I still import these things. D'oh.

dwf:
Tests pass.

PR #810, opened 2015-08-19
sgd function profile option

### LINKED ISSUE #804: Pass keyword arguments to 'theano.function' call in GradientDescent
`THEANO_FLAGS=mode=ProfileMode` is deprecated in favour of `profile=True` passed to `theano.function`.

orhanf:
fix #804

### DIFF
--- a/blocks/algorithms/__init__.py
+++ b/blocks/algorithms/__init__.py
@@ -191,6 +191,10 @@ class GradientDescent(DifferentiableCostMinimizer):
         be backpropagated. Only makes sense when `gradients` is `None`.
     on_unused_sources : str, one of 'raise' (default), 'ignore', 'warn'
         Controls behavior when not all sources are used.
+    theano_func_kwargs : dict, optional
+        A passthrough to `theano.function` for additional arguments.
+        Useful for passing `profile` or `mode` arguments to the theano
+        function that will be compiled for the algorithm.
 
     Attributes
     ----------
@@ -201,9 +205,8 @@ class GradientDescent(DifferentiableCostMinimizer):
 
     """
     def __init__(self, step_rule=None, gradients=None, known_grads=None,
-                 consider_constant=None,
-                 on_unused_sources='raise',
-                 **kwargs):
+                 consider_constant=None, on_unused_sources='raise',
+                 theano_func_kwargs=None, **kwargs):
         if gradients:
             kwargs.setdefault("parameters", gradients.keys())
         super(GradientDescent, self).__init__(**kwargs)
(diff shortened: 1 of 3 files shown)


blocks/algorithms/__init__.py
```diff
@@ -243,7 +243,8 @@ def initialize(self):
         for parameter in self.parameters:
             all_updates.append((parameter, parameter - self.steps[parameter]))
         all_updates += self.step_rule_updates
-        self._function = theano.function(self.inputs, [], updates=all_updates)
+        self._function = theano.function(self.inputs, [], updates=all_updates,
+                                         profile=kwargs.get('profile', None))
```
dwf: I'd just pass **kwargs.

blocks/main_loop.py
```diff
@@ -168,7 +168,7 @@ def run(self):
                         extension.main_loop = self
                     self._run_extensions('before_training')
                     with Timer('initialization', self.profile):
-                        self.algorithm.initialize()
+                        self.algorithm.initialize(profile=self.profile)
```
dwf: No need for this I don't think --- we have two different notions of "profiling", one is the Theano profiler which profiles calls to compiled Theano functions, and one that is the Blocks-level profiler.

The question is how you would pass **kwargs to `theano.function` via a call to `run()`. We could just add `**kwargs` to `run()` and pass them along to `initialize()` on this line, but... I don't know. 

@rizar curious to hear your opinion of how this should work, you usually have good design ideas about things like this.

rizar: One way to go would be to accept `theano_function_kwargs` dictionary in `GradientDescent.__init__` which would be further used in `initialize` when it is called in `run`. 

Another way to go would be to make `initialize` accept keyword arguments and have the user call it before `main_loop.run()`. But this requires a check for double initialization to be done inside `GradientDescent`. I would like to have it there instead of `MainLoop`, let alone for the reason that `blocks/algorithms.py` is almost a separate library, but it is not yet there. 

What do you think about the first proposal?

orhanf: First option is cleaner i guess, so i made the changes according to that one

rizar:
Could you please add a test for it? You can check e.g. that profiling works.

blocks/algorithms/__init__.py
```diff
@@ -233,6 +236,8 @@ def __init__(self, step_rule=None, gradients=None, known_grads=None,
         self.total_step_norm = named_copy(l2_norm(self.steps.values()),
                                           "total_step_norm")
         self.on_unused_sources = on_unused_sources
+        self.theano_func_kwargs = theano_func_kwargs \
```
dwf: I think using backslash rather than parens violates our style guide but I'm inclined to let it slide.

rizar:
I am closing this PR  to open a new, rebased one, and merged it. Thank you @orhanf!

PR #813, opened 2015-08-20
Fix Parallel.apply routing logic.

### LINKED ISSUE #812: Merge *requires* keyword arguments to apply
[This](https://github.com/mila-udem/blocks/blob/master/blocks/bricks/parallel.py#L85-L86) looks quite broken.

dwf:
Fixes #812.

Basically, the way we handle `*args` and `**kwargs` in Parallel.apply is broken. It essentially forces you to use kwargs. It seems like the `*args` was hastily added to make a test pass or something.

The correct thing to do is treat the `*args` as the inputs for the first `len(args)` things in `self.input_names` and then try to fill the rest with `**kwargs`. If a keyword is there that doesn't match anything in `self.input_names`, we should say so. If we get the same input specified as a positional and keyword argument, we should make like Python does:

```
>>> def foo(a, **kwargs):
...     pass
...
>>> foo(5, a=3)
  File "<stdin>", line 1, in <module>
TypeError: foo() got multiple values for argument 'a'
```

If after accounting for all `*args` and `**kwargs`, we are missing inputs for some child bricks, we should say so.

@vdumoulin You were the last one to edit this line, could you look at my change?

### DIFF
--- a/blocks/bricks/parallel.py
+++ b/blocks/bricks/parallel.py
@@ -82,10 +82,23 @@ class Parallel(Initializable):
 
     @application
     def apply(self, *args, **kwargs):
-        args = args + tuple(kwargs[name] for name in
-                            self.input_names[len(args):])
-        return [child.apply(arg)
-                for arg, child in equizip(args, self.children)]
+        # Use of zip() rather than equizip() intentional here.
+        routed_args = dict(zip(self.input_names, args))
+        for name in kwargs:
+            if name not in self.input_names:
+                raise KeyError('invalid input name: {}'.format(name))
+            elif name in routed_args:
+                raise TypeError("{}.apply got multiple values for "
+                                "argument '{}'".format(self.__class__.__name__,
+                                                       name))
+            else:
+                routed_args[name] = kwargs[name]
+        if set(self.input_names) != set(routed_args):
+            raise ValueError('missing values for inputs: {}'.format(
+                             [name for name in self.input_names
+                              if name not in routed_args]))
+        return [child.apply(routed_args[name])
+                for name, child in equizip(self.input_names, self.children)]
 
     @apply.property('inputs')
     def apply_inputs(self):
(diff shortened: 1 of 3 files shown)


dwf:
(WIP because it needs tests.)

dwf:
Okay, this is done. Please review, whoever has time.

blocks/bricks/parallel.py
```diff
@@ -82,10 +82,23 @@ def _push_allocation_config(self):
 
     @application
     def apply(self, *args, **kwargs):
-        args = args + tuple(kwargs[name] for name in
-                            self.input_names[len(args):])
-        return [child.apply(arg)
-                for arg, child in equizip(args, self.children)]
+        # Use of zip() rather than equizip() intentional here.
+        routed_args = dict(zip(self.input_names, args))
```
rizar: I think that  this logic should be factored out and moved to `/blocks/utils`. E.g.

```python
def extract_args(self, expected_args, args, kwargs):
   """Makes some keyword arguments positional. 

   A frequent situation is that a method of the class gets to 
 
```

dwf: Something like b439144?

blocks/utils/__init__.py
```diff
@@ -460,3 +460,62 @@ def change_recursion_limit(limit):
...
+        A list of strings denoting names for the expected arguments,
+        in order.
+    args : iterable
+        Positional arguments that have been passed.
+    kwargs : Mapping
+        Keyword arguments that have been passed.
+
+    Returns
+    -------
+    routed_args : dict
```
rizar: I think it makes sense to return an ordered dict here.

rizar:
Test crash cause you forgot to import `extract_args`.

I don't mind if you merge it as is (after the tests pass of course), but I think `extract_args` should rather return an `OrderedDict`.

dwf:
Any idea why the hooks aren't running now?

vdumoulin:
LGTM.

dwf:
I don't know why Travis and Scrutinizer aren't running, but I'm sure that the current branch is good, so I'll merge and deal with the consequences if there are any.

PR #815, opened 2015-08-21
Override base path for temporary file creation

yingzha:


yingzha:
@dwf , could you help me check my ticket if you have time? I didn't find where the problem is.

dwf:
Weird, it looks like it's getting called with the string None rather than the value None. Just a second.

dwf:
I do not understand why coverage dropped. That doesn't make a lot of sense, and the Coveralls interface is unhelpful in determining why...

yingzha:
Did I do something wrong :-(

dwf:
No I don't think so, the coverage checker is sometimes a little unreliable... I restarted Travis build.

dwf:
Okay, 0.05% drop is not worth worrying about. It was like -1.4% before, weird.

I think we can merge, I just want @vdumoulin to review the code that _I_ wrote so that I am not accepting my own changes. :smile:

yingzha:
:-) Thanks for your guide and sorry again for my several weeks delay.

dwf:
@rizar @vdumoulin Since I have now contributed to this PR I shouldn't also be merging it without a second pair of eyes. Please merge at will if you think it's ready.

PR #818, opened 2015-08-25
Elaborate LSTM inputs argument to apply method

ASalvail:
Since I lost an hour trying to understand why the inputs should be 4 times as large as the output and because my first thought was to tile the inputs, I made this small addition to the docs of the LSTM `apply` method.


dwf:
LGTM, thanks.

Issue #817, opened 2015-08-25
Bug with the iterate keywords of RNN


ASalvail:
When using `iterate=False` as arguments to the apply function of a `SimpleRecurrent` network, the initial state is not setup automatically.

To reproduce : 
```python
import theano
from blocks import initialization
from blocks.bricks import Identity
from blocks.bricks.recurrent import SimpleRecurrent

x = theano.tensor.tensor3('x')
rnn = SimpleRecurrent(dim=3, activation=Identity(), weights_init=initialization.Identity())
rnn.apply(x, iterate=False)
```
will result in `AsTensorError: ('Cannot convert None to TensorType', <class 'NoneType'>)` which is a result of `states` being uninitialized. In all cases, this should be caught earlier.

From the sources in `blocks/bricks/recurrent.py` we find on lines 130-132
```python
iterate = kwargs.pop('iterate', True)
if not iterate:
    return application_function(brick, *args, **kwargs)
```

which stops from setup of `states` layer further in the same function. This seems to be a bug, especially considering line 160-161 of the same function which checks if `iterate is set to `False`. 

```python
if not iterate:
    batch_size = shape[0]
```

Since I don't know why the function returns early, I don't wan't to risk a patch.

Any comment that would help me patch this would be appreciated.

rizar:
Thanks for reporting this bug!

I think that when `iterate=False` is used, it should be mandatory to provide all states as inputs. It makes quite little sense to build a computation graph of the first step of the RNN, at least I do not know an application.

That being said, I am very surprised that `inputs` and `states` are not mandatory for `SimpleRecurrent`. I have created #819 to fix it.

rizar:
Also, I have deleted lines 160-161, this is old code which is not necessary now since `iterate` is handled in the very beginning of `recurrent_apply`.

ASalvail:
> I think that when `iterate=False` is used, it should be mandatory to provide all states as inputs.

Personally, I think it would be far better if the brick took care of it. I guess this was set this way to give the user the choice to train (with initial_states being parameters) or give (with initial_states given as input to the compiled function) the initial state. That being said, since I want them as parameters, I need to create the shared variables, then assign them roles, then give them to the apply function (instead of the constructor, which is a surprising design choice).

> It makes quite little sense to build a computation graph of the first step of the RNN, at least I do not know an application.

[...]

PR #819, opened 2015-08-26
Arguments for SimpleRecurrent should be required

rizar:
See #817

### DIFF
--- a/blocks/bricks/recurrent.py
+++ b/blocks/bricks/recurrent.py
@@ -157,11 +157,8 @@ def recurrent(*args, **kwargs):
             if len(sequences_given):
                 # TODO Assumes 1 time dim!
                 shape = list(sequences_given.values())[0].shape
-                if not iterate:
-                    batch_size = shape[0]
-                else:
-                    n_steps = shape[0]
-                    batch_size = shape[1]
+                n_steps = shape[0]
+                batch_size = shape[1]
             else:
                 # TODO Raise error if n_steps and batch_size not found?
                 n_steps = kwargs.pop('n_steps')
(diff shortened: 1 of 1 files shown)


ASalvail:
If you intend to make the `states` mandatory, the docstrings of `SimpleRecurrent` and `LSTM`  (at the very least) will need to be changed. Furthermore, the docs of the RNN won't run anymore (see the very first block of code) and will need to be modified.

rizar:
I looked at the docs and it is not said there that the parameters are optional. Moreover, for `GatedRecurrent` and `LSTM` the states are already mandatory. And the tests  pass just fine, including doctests.

Perhaps I should clarify that the states are mandatory only for `iterate=False`. Does this help?

ASalvail:
> Perhaps I should clarify that the states are mandatory only for iterate=False. Does this help?

Indeed, I thought you wanted to make them mandatory for all `apply` calls since this is what the patch of the PR is doing by removing the default argument of `state`.

ASalvail:
AFAIK, this piece of code in the docs on recurrent bricks would fail with this patch on the `rnn.apply(x)`.

```python
>>> import numpy
>>> import theano
>>> from theano import tensor
>>> from blocks import initialization
>>> from blocks.bricks import Identity
>>> from blocks.bricks.recurrent import SimpleRecurrent
>>> x = tensor.tensor3('x')
>>> rnn = SimpleRecurrent(
...     dim=3, activation=Identity(), weights_init=initialization.Identity())
>>> rnn.initialize()
>>> h = rnn.apply(x)
>>> f = theano.function([x], h)
>>> print(f(numpy.ones((3, 1, 3), dtype=theano.config.floatX))) 
```

Are the doctests testing these too?

rizar:
I tried the piece of code you suggested and it did not fail, because the behaviour when `iterate=True` is not changed by this PR.

ASalvail:
Ah, my bad. I just learned how decorator works. Since the recurrent 
decorator provides the default values, it all turns out to work.

Thanks for the info!

PR #822, opened 2015-09-01
Bricks design docs

rizar:
I am trying to write docs explains the design decision behind bricks (since this seems to be the most controversial part of Blocks so far). The approach I have taken is to develop something like bricks and explain every step on the way. My current progress is in this PR. I am especially interested in general feedback: do you think it is a good way to write docs? Is the tutorial overloaded with copy-pasted code? Do I expect too much from the reader or maybe on the opposite, am I explaining obvious things?

@bartvm , @dmitriy-serdyuk , @dwf, @vdumoulin , @fvisin , @janchorowski

### DIFF
--- a/docs/bricks_design.rst
+++ b/docs/bricks_design.rst
@@ -1,96 +1,94 @@
 
-0 Imports
-^^^^^^^^^
+Bricks design explained
+=======================
 
-This is a runnable tutorial. To run the examples, please import all the
-required modules first.
+This is a runnable tutorial, which means you can copy its fragments 
+to an IPython notebook and run. The import section follows:
 
-.. code:: python
+    >>> import theano
+    >>> theano.config.compute_test_value = 'raise'
+    >>> from theano import tensor
 
-    import theano
-    theano.config.compute_test_value = 'raise'
-    from theano import tensor
+Please keep in mind that the code in this tutorial was written for illustration
+purposes only. Often only the common case is handled and corner cases are
+ignored.
 
-Please keep in mind that simplicity and brevity were priorities for
-writing the code in this tutorial. Often corner cases are not handled
-and sketchy assumptions are made.
-
-1. Introduction
-===============
+Introduction
+------------
 
 Deep learning (DL) models implementing using Theano share numerous
 building blocks which become more and more standartized over time.
 Factoring out such components is a very lucrative idea, and many people
 have fallen victims to the temptation to develop a library of DL
 building blocks. The list of rather well-known projects includes
-PyLearn2, Groundhog, Lasagne, Blocks, Keras, but also lots of people and
-groups have smaller and more specialized frameworks.
+`PyLearn2`_, `Groundhog`_, `Lasagne`_, `Keras`_ and of course Blocks. Many people and
+research groups have smaller and more specialized frameworks.
 
-This document contains my speculations on how such libraries should be
+This document contains speculations on how such libraries should be
 designed. We will start with the simplest approach possible and show
-step by step what benefits more advanced methods bring. My final goal is
-to explain the ideas at the core of the respective part of Blocks
-framework. In Blocks the components assisting to build the computation
+step by step what benefits more advanced methods bring. The final goal is
+to explain the ideas at the core of the respective part of Blocks. 
+In Blocks the components assisting to build the computation
 graph are called "bricks". Bricks are a unified representation for
 components of all levels: basic ones, like activation functions, and
 complex ones, like attention mechanism and a generic sequence generator.
-I hope that reading this document will answer many question that people
+Reading this document should answer many questions that people
 often ask about bricks, such as "why so many decorators", "why is not my
 brick initialized in the constructor" and others, which are asked not so
 often, but still daunt people' minds.
 
-2. Design for Basic Components
-==============================
+.. _PyLearn2: https://github.com/lisa-lab/pylearn2
+.. _Groundhog: https://github.com/lisa-groundhog/GroundHog
+.. _Lasagne: https://github.com/Lasagne/Lasagne
+.. _Keras: https://github.com/fchollet/keras
 
-2.1. The Toy Task
-~~~~~~~~~~~~~~~~~
+Design for basic components
+---------------------------
 
-To make our discussion less abstract, let us consider an example. Dima,
+The toy task
+~~~~~~~~~~~~
+
+To make our discussion less abstract, let us consider an example. D,
 a novice in deep learning, has learnt about fully-connected feed-forward
 and recurrent networks (FFN and RNN respectively). He plans to use them
 in a number of projects, so he decides to start with coding a reference
 implementation first. Few hours of hard work and here it is:
 
-.. code:: python
+    >>> # D's FFN, D2L2v0.1
+    >>> inputs = tensor.matrix('x')
+    >>> inputs.tag.test_value = numpy.random.uniform(size=(10, 754))
+    >>> dims = [754, 100, 100, 10]
+    >>> activations = [tensor.tanh, tensor.tanh, tensor.nnet.softmax]
+    >>> weight_matrices = []
+    >>> bias_vectors = []
+    >>> for i in range(len(dims) - 1):
+    >>>     weight_matrices.append(theano.shared(numpy.random.uniform(
+    >>>         size=(dims[i], dims[i + 1]))))
+    >>>     bias_vectors.append(theano.shared(numpy.zeros(dims[i + 1])))
+    >>> outputs = inputs
+    >>> for dim, activation, weight_matrix, bias_vector in zip(
+    >>>         dims[1:], activations, weight_matrices, bias_vectors):
+    >>>     outputs = activation(outputs.dot(weight_matrix) + bias_vector)
+    >>> 
+    >>> # D's RNN, D2L2v0.1
+    >>> inputs = tensor.tensor3('x')
+    >>> inputs.tag.test_value = numpy.random.uniform(size=(5, 10, 100))
+    >>> dim = 100
+    >>> weight_matrix = theano.shared(numpy.random.uniform(size=(dim, dim)))
+    >>> bias_vector = theano.shared(numpy.random.uniform(size=dim))
+    >>> outputs, _ = theano.scan(
+    >>>     lambda states, _inputs, _weight_matrix: states.dot(_weight_matrix) + _inputs,
+    >>>     sequences=[inputs], non_sequences=[weight_matrix],
+    >>>     outputs_info=tensor.zeros_like(inputs[0]))
 
-    # Dima's FFN, D2L2v0.1
-    inputs = tensor.matrix('x')
-    inputs.tag.test_value = numpy.random.uniform(size=(10, 754))
-    dims = [754, 100, 100, 10]
-    activations = [tensor.tanh, tensor.tanh, tensor.nnet.softmax]
-    weight_matrices = []
-    bias_vectors = []
-    for i in range(len(dims) - 1):
-        weight_matrices.append(theano.shared(numpy.random.uniform(
-            size=(dims[i], dims[i + 1]))))
-        bias_vectors.append(theano.shared(numpy.zeros(dims[i + 1])))
-    
-    outputs = inputs
-    for dim, activation, weight_matrix, bias_vector in zip(
-            dims[1:], activations, weight_matrices, bias_vectors):
-        outputs = activation(outputs.dot(weight_matrix) + bias_vector)
-
-.. code:: python
-
-    # Dima's RNN, D2L2v0.1
-    inputs = tensor.tensor3('x')
-    inputs.tag.test_value = numpy.random.uniform(size=(5, 10, 100))
-    dim = 100
-    weight_matrix = theano.shared(numpy.random.uniform(size=(dim, dim)))
-    bias_vector = theano.shared(numpy.random.uniform(size=dim))
-    outputs, _ = theano.scan(
-        lambda states, _inputs, _weight_matrix: states.dot(_weight_matrix) + _inputs,
-        sequences=[inputs], non_sequences=[weight_matrix],
-        outputs_info=tensor.zeros_like(inputs[0]))
-
-The two code snippets above has formed Dima's brand-new DL toolkit D2L2
-(Dima Deep Learning Library). He picks 0.1 as the initial version
+The two code snippets above has formed D's brand-new DL toolkit D2L2
+(D Deep Learning Library). He picks 0.1 as the initial version
 identifies. Hooray, an new project has been born!
 
-2.2. Well, why not just copy-paste?!
+Well, why not just copy-paste?!
 ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
 
-Indeed, why not? For every new project Dima can copy the snippets above,
+Indeed, why not? For every new project D can copy the snippets above,
 change dimensions, activation functions, input variables. In fact
 copy-pasting is a viable way of reusing code. In has a big advantage:
 debugging is the easiest when your model is built by a single chunk of
@@ -106,7 +104,8 @@ no means specific to deep learning and Python, is much longer:
 1. Any change in the original snippet has to be duplicated in each of
    dozens (hundreds?) where it was pasted. This is a lot of repetitive
    work, which means wasted time and high probability of doing it wrong
-   at some point. rh
+   at some point. 
+
 2. If the pasted snippet was modified and then the original snippet was
    improved, merging the changes in both becomes a non-trivial
    procedure. Of course, merging is also unavoidable even when a more
@@ -123,8 +122,8 @@ no means specific to deep learning and Python, is much longer:
 
 To make the long story short, when you copy-paste you create multiple
 versions of code and synchronizing them is a nightmare. This applies not
-only to brutal squashing (?) of a snippet into your code, but also to
-copying functions and classes to reuse them with modifications (these
+only to pasting snippet into your script, but also to
+copying functions and classes to reuse them with modifications (such
 practices were one of the main reasons for Groundhog to fall into
 oblivion). This is why we are so insistive when a Blocks PR introduces
 code duplication.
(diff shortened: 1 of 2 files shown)


docs/bricks_design.rst
```diff
@@ -0,0 +1,602 @@
+
+Bricks design explained
+=======================
+
+This is a runnable tutorial, which means you can copy its fragments 
+to an IPython notebook and run. The import section follows:
```
janchorowski: Just an idea - maybe make it an ipython notebook? They are easy to view right now, and no copy-paste is necessary.

docs/bricks_design.rst
```diff
@@ -0,0 +1,602 @@
...
+
+Design for basic components
+---------------------------
+
+The toy task
+~~~~~~~~~~~~
+
+To make our discussion less abstract, let us consider an example. D,
+a novice in deep learning, has learnt about fully-connected feed-forward
+and recurrent networks (FFN and RNN respectively). He plans to use them
```
janchorowski: FNN for Feed-forward Neural Networks and RNN for Recurrent Neural Networks?

docs/bricks_design.rst
```diff
@@ -0,0 +1,602 @@
...
+modern DL models can hardly be represented as sequences of components
+(consider Neural Turing Machines, Memory Networks, attention-equipped
+Encoder-Decoders with and others).
+
+Similarly, we avoid the wide-spread concept of **the** input of a layer.
+Many components of the systems mentioned in the previous paragraph take
+many inputs, e.g. an attention mechanism often uses both the state of
+the decoder and the encoder input.
+
+Step by step, we have arrived to something virtually indistinguishable
```
janchorowski: This is not needed in a tutorial introduction - lets assume people don't know about other projects, unless they are very popular (like Caffe, or PL2)

docs/bricks_design.rst
```diff
@@ -0,0 +1,602 @@
...
+------------
+
+Deep learning (DL) models implementing using Theano share numerous
+building blocks which become more and more standartized over time.
+Factoring out such components is a very lucrative idea, and many people
+have fallen victims to the temptation to develop a library of DL
+building blocks. The list of rather well-known projects includes
+`PyLearn2`_, `Groundhog`_, `Lasagne`_, `Keras`_ and of course Blocks. Many people and
+research groups have smaller and more specialized frameworks.
+
```
janchorowski: I would add:
At the first sight, building a DL framework is easy with tools like Theano - after all one only needs to connect a few pieces to get a first, working multilayer feed-forward net, the cornerstone of deep learning. However, in our experience a DL framework needs much more than just being able to implement small models. In particular we would like to:
1. Reuse parts of models in other models without copy-pasting code and manual model conversion. This implies hierarchical model design
2. Be able to quickly inspect what happens inside of the model, allow for ad-hoc monitoring of intermediate computations.
3. Exchange parts of models  - you want to compare a GRU layer with an LSTM one in a large machine translation network - no problem.

Many aspects of Bricks - the pieces one builds models in Blocks were added to handle our needs. We want blocks to be a library that scales and allows easy prototyping of even the most complicated models you can imagine.

janchorowski:
Great work!

dmitriy-serdyuk:
Great work, Dima! Maybe I'm a bit conservative, but it looks much more like a blog post, the style is a bit too casual for a doc page... Although this write-up is great for a blog post! It might be better to publish it somewhere else and give a link in the documentation. 

See my further remarks below.

docs/bricks_design.rst
```diff
@@ -0,0 +1,602 @@
...
+    >>> from theano import tensor
+
+Please keep in mind that the code in this tutorial was written for illustration
+purposes only. Often only the common case is handled and corner cases are
+ignored.
+
+Introduction
+------------
+
+Deep learning (DL) models implementing using Theano share numerous
```
dmitriy-serdyuk: Can you add a disclaimer that you assume the reader to know and understand Theano?

docs/bricks_design.rst
```diff
@@ -0,0 +1,602 @@
...
+    >>> outputs, _ = theano.scan(
+    >>>     lambda states, _inputs, _weight_matrix: states.dot(_weight_matrix) + _inputs,
+    >>>     sequences=[inputs], non_sequences=[weight_matrix],
+    >>>     outputs_info=tensor.zeros_like(inputs[0]))
+
+The two code snippets above has formed D's brand-new DL toolkit D2L2
+(D Deep Learning Library). He picks 0.1 as the initial version
+identifies. Hooray, an new project has been born!
+
+Well, why not just copy-paste?!
```
dmitriy-serdyuk: I would say, this paragraph is quite obvious for everyone at least a bit familiar with software engeneering. Another one disclaimer?

docs/bricks_design.rst
```diff
@@ -0,0 +1,602 @@
...
+    >>>         self.weight_matrices = []
+    >>>         self.bias_vectors = []
+    >>>         for i in range(len(dims) - 1):
+    >>>             self.weight_matrices.append(
+    >>>                 theano.shared(weight_initializers[i]((dims[i], dims[i + 1]))))
+    >>>         for i in range(len(dims) - 1):
+    >>>             self.bias_vectors.append(theano.shared(
+    >>>                 bias_initializers[i](dims[i + 1])))
+    >>>             
+    >>>     def apply(self, inputs):
```
dmitriy-serdyuk: Many people call this kind of methods `fprop`.

docs/bricks_design.rst
```diff
@@ -0,0 +1,602 @@
...
+    >>>         y.tag.stone = self
+    >>>         y.tag.role = 'output'
+    >>>         return y    
+
+It seems that the process of annotating could be largely automated.
+Arguably, the most convenient way to proceed is to use decorators.
+Decorators belong to the realm of somewhat less-known Python features,
+but nevertheless, they are very simple:
+
+    >>> def application(wrapped_method):
```
dmitriy-serdyuk: This piece of code may be not *very simple* for some people. I may suggest to use a funciton with one argument for tanh example and add more comments.

docs/bricks_design.rst
bartvm: I've never seen FFN or FNN used much at all, and would just avoid them altogether.

bartvm:
Only got to skim over it really quickly. Looks great! I don't have a problem with it being more conversational, I think it's okay for documentation to not make people fall asleep. I wouldn't fragment Blocks documentation further between blog posts, documentation pages, and API specifications.

I agree that the copy-pasting part might be superfluous for people who are familiar with basic engineering principles but just want to get a quick overview of the engineering principles behind Blocks in particular. Maybe just start by pointing these people to the relevant sections (and make sure that those can be read independently)?

The only other overall comment I have is that it's worth keeping in mind that people tend to very often skim through documentation (everyone does this with frameworks to get a quick feel for things). Few people actually sit down and work their way through dense paragraphs of text and code. With that in mind, I would try and keep paragraphs short (even if that means dropping some analogies, recaps, etc.), use lots of headers and sub-headers (sub-sub-headers!), use bullet points and break up the code snippets in small pieces.

docs/bricks_design.rst
fvisin: I agree with Dima, maybe the software engineering related part could be moved into a separate chapter, so that who is familiar with it can just skip it.

docs/bricks_design.rst
```diff
@@ -0,0 +1,602 @@
...
+    >>> weight_matrices = []
+    >>> bias_vectors = []
+    >>> for i in range(len(dims) - 1):
+    >>>     weight_matrices.append(theano.shared(numpy.random.uniform(
+    >>>         size=(dims[i], dims[i + 1]))))
+    >>>     bias_vectors.append(theano.shared(numpy.zeros(dims[i + 1])))
+    >>> outputs = inputs
+    >>> for dim, activation, weight_matrix, bias_vector in zip(
+    >>>         dims[1:], activations, weight_matrices, bias_vectors):
+    >>>     outputs = activation(outputs.dot(weight_matrix) + bias_vector)
```
fvisin: you don't need dim here

docs/bricks_design.rst
```diff
@@ -0,0 +1,602 @@
...
+The toy task
+~~~~~~~~~~~~
+
+To make our discussion less abstract, let us consider an example. D,
+a novice in deep learning, has learnt about fully-connected feed-forward
+and recurrent networks (FFN and RNN respectively). He plans to use them
+in a number of projects, so he decides to start with coding a reference
+implementation first. Few hours of hard work and here it is:
+
+    >>> # D's FFN, D2L2v0.1
```
fvisin: This line can be confusing, I would just put Feed-forward Neural Network here.

docs/bricks_design.rst
```diff
@@ -0,0 +1,602 @@
...
+    >>> for i in range(len(dims) - 1):
+    >>>     weight_matrices.append(theano.shared(numpy.random.uniform(
+    >>>         size=(dims[i], dims[i + 1]))))
+    >>>     bias_vectors.append(theano.shared(numpy.zeros(dims[i + 1])))
+    >>> outputs = inputs
+    >>> for dim, activation, weight_matrix, bias_vector in zip(
+    >>>         dims[1:], activations, weight_matrices, bias_vectors):
+    >>>     outputs = activation(outputs.dot(weight_matrix) + bias_vector)
+    >>> 
+    >>> # D's RNN, D2L2v0.1
```
fvisin: Dicto

docs/bricks_design.rst
```diff
@@ -0,0 +1,602 @@
...
+   principled then copy-pasting approach is used, but in the latter case
+   you are assisted by the version control system.
+
+3. The only way to tell if the snippet versions used in two scripts are
+   identical is to read the code. Carefully.
+
+4. The namespaces of the snippet and the place where it is applied
+   become one, which is a potential source of horrible bugs.
+
+5. ...
```
fvisin: "Other people might fix a bug in one of the copies of the snippet and not be aware that there are many others that should be fixed in the rest of the code" 
I think this is an important one in a open-source collaborative project.

docs/bricks_design.rst
```diff
@@ -0,0 +1,602 @@
...
+
+To make the long story short, when you copy-paste you create multiple
+versions of code and synchronizing them is a nightmare. This applies not
+only to pasting snippet into your script, but also to
+copying functions and classes to reuse them with modifications (such
+practices were one of the main reasons for Groundhog to fall into
+oblivion). This is why we are so insistive when a Blocks PR introduces
```
fvisin: You should definitely add PyLearn2 to the list, but I wouldn't "criticize" other frameworks. I would just make it generic saying that many good open source projects fall into oblivion because these engineering problems scale exponentially with the dimension of the framework and at some point they become just impossible to cope with.

docs/bricks_design.rst
```diff
@@ -0,0 +1,602 @@
...
+times. At an any larger scale it should be avoided at any cost.
+
+Functions creating Theano graphs
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+
+Having read `The Practice of Programming`_, D decides to never
+copy-paste in his entire life. Instead, he packs his two snippets into
+functions:
+
+    >>> # D's FFN, D2L2v0.2
```
fvisin: Ditto

docs/bricks_design.rst
```diff
@@ -0,0 +1,602 @@
...
+    >>> inputs.tag.test_value = numpy.random.uniform(size=(10, 754))
+    >>> dims = [754, 100, 100, 10]
+    >>> activations = [tensor.tanh, tensor.tanh, tensor.nnet.softmax]
+    >>> weight_initializers = [
+    >>>     lambda shape: numpy.random.uniform(size=shape)] * len(activations)
+    >>> bias_initializers = weight_initializers
+    >>> outputs = build_feedforward_network(
+    >>>     inputs, dims, activations, weight_initializers, bias_initializers)
+    >>>
+    >>> # D's RNN, D2L2v0.2
```
fvisin: Ditto

docs/bricks_design.rst
```diff
@@ -0,0 +1,602 @@
...
+
+We do not list the hypothetical sampling code here for brevity, but
+hopefully people that have tried to implement such things see that it
+can be implemented using the ``apply_one_step`` method of the ``stone``
+object. It is not such a big deal to save the user from reimplementing
+``apply_one_step`` for a simple recurrent network, but this could just
+as well be LSTM or GRU or even Clockwork RNN, for which
+``apply_one_step`` would be much less trivial.
+
+Note, how that we deliberately avoid the name _layer_ for our
```
fvisin: "how that" typo

docs/bricks_design.rst
vdumoulin: The problem is they're harder to test automatically (see [this discussion](https://github.com/mila-udem/fuel/issues/195)).

rizar: I started writing it as IPython notebook, but it is not convenient to keep IPython notebook under version control. Outputs have to be deleted every time before commit and so on.

docs/bricks_design.rst
```diff
@@ -0,0 +1,602 @@
...
+components. A layer is an element of a sequence, and many of interesting
+modern DL models can hardly be represented as sequences of components
+(consider Neural Turing Machines, Memory Networks, attention-equipped
+Encoder-Decoders with and others).
```
fvisin: Either I don't understand the sentence or there is a typo

docs/bricks_design.rst
```diff
@@ -0,0 +1,602 @@
...
+step by step what benefits more advanced methods bring. The final goal is
+to explain the ideas at the core of the respective part of Blocks. 
+In Blocks the components assisting to build the computation
+graph are called "bricks". Bricks are a unified representation for
+components of all levels: basic ones, like activation functions, and
+complex ones, like attention mechanism and a generic sequence generator.
+Reading this document should answer many questions that people
+often ask about bricks, such as "why so many decorators", "why is not my
+brick initialized in the constructor" and others, which are asked not so
+often, but still daunt people' minds.
```
vdumoulin: I think

    which are asked not so often, but still daunt people' minds

could be removed; the first part of the sentence is true in itself and doesn't need to be nuanced, IMO.

docs/bricks_design.rst
```diff
@@ -0,0 +1,602 @@
...
+Arguably, the most convenient way to proceed is to use decorators.
+Decorators belong to the realm of somewhat less-known Python features,
+but nevertheless, they are very simple:
+
+    >>> def application(wrapped_method):
+    >>>     # A decorator takes a function and returns a function.
+    >>>     # Note, that a method is just a function, and `self` is just an argument.
+    >>>     # The returned function is what you actually use later. 
+    >>>     # It typically makes a number of calls of the wrapped function, 
+    >>>     # in addition doing some other stuff, pre- and post- processing
```
fvisin: in addition \*to\* doing

docs/bricks_design.rst
rizar: Honestly, I feel like these high level explanations are very hard to understand to people that have never worked with a large deep learning system. My intention was to stick to the examples.

Perhaps a text similar to the one above can be added in the end of the tutorial instead, when people have gone through heavy-weight things like lazy initialization and so on.

docs/bricks_design.rst
rizar: OK, I will try to add more comments. I would still support multiple arguments, because otherwise in the later sections I will have to rewrite `application` once again, which means even more copy-paste.

docs/bricks_design.rst
rizar: Also copy-paste would still be there, even with IPython notebook. It is just that we have to show how the implementation evolves over time, and even though it is often changed only a little bit, I have to make a copy and edit it.

docs/bricks_design.rst
fvisin: I agree with Dima: this is one of the most obscure parts of Blocks (Python, actually), where people get puzzled and move to something simpler to understand. It's very important to comment it clearly and in depth IMO.

docs/bricks_design.rst
```diff
@@ -0,0 +1,602 @@
...
+    >>>         return y    
+
+It seems that the process of annotating could be largely automated.
+Arguably, the most convenient way to proceed is to use decorators.
+Decorators belong to the realm of somewhat less-known Python features,
+but nevertheless, they are very simple:
+
+    >>> def application(wrapped_method):
+    >>>     # A decorator takes a function and returns a function.
+    >>>     # Note, that a method is just a function, and `self` is just an argument.
```
fvisin: I separate the generic explanation of decorators to the previous paragraph, maybe expanding it a little, and keep here just the comments that relate strictly to this function. Also, it might help to explain more clearly what's the difference between `def returned_function(...)` and `def application(...)`

docs/bricks_design.rst
```diff
@@ -0,0 +1,602 @@
...
+    >>> x.tag.test_value = numpy.zeros((10, 20))
+    >>> theano.printing.debugprint(Tanh().apply(x))
+    Elemwise{identity} [@A] 'Tanh_output'   
+     |Elemwise{tanh,no_inplace} [@B] ''   
+       |Elemwise{identity} [@C] 'Tanh_input'   
+         |x [@D]
+
+
+Now, ``MultiLayer`` can annotate its inputs and outputs usign the same
+decorator! Note, that now activations have a different interface: they are
```
fvisin: I would flip the sentence in "Now, we can annotate the inputs and outputs of ``MultiLayer`` using the same `@application` decorator!"

fvisin: Also, remove the comma after note in "Note, that now .."

docs/bricks_design.rst
```diff
@@ -0,0 +1,602 @@
...
+    >>> stone = MultiLayer(
+    >>>     dims, activations, weight_initializers, bias_initializers)
+    >>> outputs_multilayer = stone.apply(inputs)
+    >>>     
+    >>> inputs = tensor.tensor3('x')
+    >>> inputs.tag.test_value = numpy.random.uniform(size=(5, 10, 100))
+    >>> stone = RecurrentNetwork(100, lambda shape: numpy.random.uniform(size=shape))
+    >>> outputs_recurrent = stone.apply(inputs)
+
+We do not list the hypothetical sampling code here for brevity, but
```
fvisin: Since later you use dropout as an example, why don't you introduce it here and already show the implementation, so that who reads can appreciate the usefulness of `@application`?

docs/bricks_design.rst
```diff
@@ -0,0 +1,602 @@
...
+
+    >>> rng = theano.tensor.shared_randomstreams.RandomStreams()
+    >>> outputs_multilayer_regularized = theano.clone(
+    >>>     outputs_multilayer, 
+    >>>     replace={layer0_outputs:
+    >>>              layer0_outputs * rng.binomial(layer0_outputs.shape,
+    >>>                                            p=0.5)})
+
+The reader might have noticed that in the example above we applied
+dropout regularization. Many other typical deep learning regularization
```
fvisin: Remember to modify this line if you introduce Dropout in the previous part of the tutorial

rizar:
Thank you guys for your feedback!

If I have not answered your comment, it means that I agree and will fix it in the next version, except for the remarks below.

Similarly to Bart I think that it is better when documentation reads easily, and we do not have to stick to formal style when writing docs. 

Copy-paste section should indeed be quite boring for people with a bit of experience. However, all lab repositories to which I was added after they had been created had lots of copy-pasted scripts, which makes me think that the lab would benefit if a starting student read this text. I will try to figure out how to hide this section from the eyes of people who do not need it. Maybe it should just be shortened and merged into development guidelines.

I will see what I can do regarding smaller subsections and bullet points. Also, I would rather merge the thing when it is good enough and before it is perfect to let you guys make changes you want.

It would make sense to use numbering for section and subsections, but then the title of the document shows up with a number in the documentation index, which looks awful.

docs/bricks_design.rst
rizar: Do you mean that I should show that the cost of supporting dropout would be much higher but for annotations? The only way I see to add dropout to `MultiLayer` would be to support it inside its `__init__` and `apply`, and this is ugly and requires quite a lot of code. Although it would be educative to show how ugly it is, I am afraid of amount of extra code required.

Did I understand your proposal?

fvisin:
Great job!

I like it conversational and especially incremental, it is really useful to understand how all the pieces come together.

I strongly agree with @bartvm on the fact that most of the people won't read a long text. The best in my opinion would be to split the tutorial in relatively independent sections (the same as, e.g., the `Building with bricks` tutorial). This should be quite easy as the current tutorial already adds one piece of information at the time, so it is just a matter of splitting the text and adding a title to each section. I would just begin every new section with a quick overview of what you said so far, so that they are somewhat independent one to each other.

docs/bricks_design.rst
```diff
@@ -0,0 +1,602 @@
...
+change dimensions, activation functions, input variables. In fact
+copy-pasting is a viable way of reusing code. In has a big advantage:
+debugging is the easiest when your model is built by a single chunk of
+Theano code. A crash with ``exception_verbosity=high`` will show you
+exactly at which stage the error occured. Likewise, compilation-time
+crashes induced by test values are also very informative when you work
+with monolith models.
+
+Unfortunately, the ease of debugging is the first and the last in the
+list of copy-pasting advantages. The list of disadvantages, is which by
```
vdumoulin: `is which -> which is`

docs/bricks_design.rst
```diff
@@ -0,0 +1,602 @@
...
+    >>>     for i in range(len(dims) - 1):
+    >>>         weight_matrices.append(
+    >>>             theano.shared(weight_initializers[i]((dims[i], dims[i + 1]))))
+    >>>     for i in range(len(dims) - 1):
+    >>>         bias_vectors.append(theano.shared(
+    >>>         bias_initializers[i](dims[i + 1])))
+    >>>
+    >>> outputs = inputs
+    >>> for dim, activation, weight_matrix, bias_vector in zip(
+    >>>         dims[1:], activations, weight_matrices, bias_vectors):
```
vdumoulin: `dims` is not used here either.

docs/bricks_design.rst
janchorowski: Still, I feel that we may need to showcase more the design assumptions. If you look into Lasagne, Keras, https://github.com/Ivaylo-Popov/Theano-Lights they all make a point of being simple. Blocks are more complicated. We need to justify it - Blocks are complicated, because it allows you to do X, Y, and Z. Because they allow you to do complicated stuff.

However, you are right that this may not be the best place for it. My comment was aimed at someone with some programming experience who wants to select a library for his project. You aim to explain Blocks to someone who has chosen them, but doesn't understand them (like a new lab member).

So maybe it is best to say to whom you address this tutorial?

rizar:
Thank you guys for a lot of feedback, but now that we have #611 I do not think is worth continuing. In fact the style I chose here requires way too much effort. I will try instead to incorporate information from https://github.com/dmitriy-serdyuk/blocks_tutorial.

PR #826, opened 2015-09-05
Add role and names to algorithm shared variables

lukemetz:
Currently blocks strives to annotate the computation graph with names and roles for useful variables. The algorithms code does not seem to do any of this. This pull request adds an `AlgorithmStateRole` and annotates algorithm's shared variables with it. I am not sold on the name AlgorithmState, if others have a better suggestion I am all ears.

My motivation for this is I wanted to access these and it proved vary difficult to do in a generic way. I would have to construct a `ComputationGraph` of the steps and somehow get the correct shared variables. Now one can use a `VariableFilter`

Let me know if this is something that would be good to add, if not I will close.

Thanks!

### DIFF
--- a/blocks/algorithms/__init__.py
+++ b/blocks/algorithms/__init__.py
@@ -12,8 +12,9 @@ from six import add_metaclass
 from theano import tensor
 
 from blocks.graph import ComputationGraph
-from blocks.utils import dict_subset, named_copy, pack, shared_floatx
+from blocks.roles import add_role, ALGORITHM_HYPERPARAMETER, ALGORITHM_BUFFER
 from blocks.theano_expressions import l2_norm
+from blocks.utils import dict_subset, named_copy, pack, shared_floatx
 
 logger = logging.getLogger(__name__)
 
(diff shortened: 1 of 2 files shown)


rizar:
Idea-wise it is perfect! It perfects matches Blocks design principles. I will look at the code as soon as I can.

blocks/algorithms/__init__.py
```diff
@@ -392,7 +393,8 @@ class Scale(StepRule):
 
     """
     def __init__(self, learning_rate=1.0):
-        self.learning_rate = shared_floatx(learning_rate)
+        self.learning_rate = shared_floatx(learning_rate, "learning_rate")
+        add_role(self.learning_rate, ALGORITHM_STATE)
```
rizar: It seems to me that one role is not sufficient. One may want to distinguish between algorithm accumulators, like velocity in SGD with momentum, and algorithm hyperparameters, like the momentum it self. Arguably, these should be two different roles, but they might share a common ancestor. E.g. `ALGORITHM_HYPERPARAMETER` and `ALGORITHM_BUFFER` can be subroles of `ALGORITHM_STATE`.

lukemetz:
Completely agree with the differentiations between roles. Updated.

rizar:
Thanks for this contribution!

I guess it would also make sense to store a reference to the step rule that created the variable as an annotation.

Issue #825, opened 2015-09-04
Cannot infer shape of Convolutional brick


dmitriy-serdyuk:
See discussion [here](https://groups.google.com/forum/#!topic/blocks-users/H8vjXOLbRFw).

vzhong:
hey @dmitriy-serdyuk , I attempted a simple fix according to your suggestions. Another problem arises when `image_size` is defaulted to `(None, None)` in that allocation for `b` will fail due to `self.get_dim('output')` returning a shape of `(batch_size, None, None)`. This seems to make sense because I think variable input shapes probably requires tying the biases (right?). I've added an error message in the event we reach allocation and the input shapes are still not known. Let me know what you think :)

dmitriy-serdyuk:
Thank you, this should work. You are right, one cannot use untied biases with variable input.

Can you make a pull request?

vzhong:
Made one at #827

PR #827, opened 2015-09-06
fix for case in which no image size is specified for Convolution #825

vzhong:
This fixes the bug in which allocation fails due to default `image_size` not being set to a tuple.


dmitriy-serdyuk:
Thank you! Fixes #825

PR #828, opened 2015-09-06
Update tutorial.rst

volumetric:
keyword argument `parameters` is giving the following error, which is fixed if the keyword argument is changed to `params`

```
  File "mnist_training.py", line 44, in <module>
    algorithm = GradientDescent(cost=cost, parameters=cg.parameters, step_rule=Scale(learning_rate=0.1))
  File "/usr/local/lib/python2.7/dist-packages/blocks/algorithms/__init__.py", line 193, in __init__
TypeError: __init__() got an unexpected keyword argument 'parameters'
```

### DIFF
--- a/docs/tutorial.rst
+++ b/docs/tutorial.rst
@@ -180,7 +180,7 @@ The training algorithm we will use is straightforward SGD with a fixed
 learning rate.
 
 >>> from blocks.algorithms import GradientDescent, Scale
->>> algorithm = GradientDescent(cost=cost, parameters=cg.parameters,
+>>> algorithm = GradientDescent(cost=cost, params=cg.parameters,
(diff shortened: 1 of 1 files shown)


rizar:
Which version of Blocks are using? It is `parameters` that we use in the most recent one.

volumetric:
0.0.1
Isn't this the version on github

```
"""The blocks library for parametrized Theano ops."""
# Scary warning: Adding code to this file can break namespace packages
# See https://pythonhosted.org/setuptools/setuptools.html#namespace-packages
__import__("pkg_resources").declare_namespace(__name__)
__version__ = '0.0.1'
```

rizar:
It is, but in fact he have already advanced quite a bit from 0.0.1 release. We should change the number, but for now just use the latest master branch.

PR #834, opened 2015-09-15
Replace uses of named_copy with var.copy(name=...)

### LINKED ISSUE #833: Replace uses of named_copy with var.copy(name=...)
Follow-on to #784. Now that Theano/Theano#3369 was merged, we should get rid of `named_copy`.

yingzha:
Fixes #833

### DIFF
--- a/blocks/algorithms/__init__.py
+++ b/blocks/algorithms/__init__.py
@@ -14,7 +14,7 @@ from theano import tensor
 from blocks.graph import ComputationGraph
 from blocks.roles import add_role, ALGORITHM_HYPERPARAMETER, ALGORITHM_BUFFER
 from blocks.theano_expressions import l2_norm
-from blocks.utils import dict_subset, named_copy, pack, shared_floatx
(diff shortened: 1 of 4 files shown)


yingzha:
@dmitriy-serdyuk error comes from mila-udem branch, do I need to make change to that branch or just ignore it?

dmitriy-serdyuk:
It is because of blocks-examples. Can you do replacements in blocks-examples repository as well?

dmitriy-serdyuk:
Waiting for mila-udem/blocks-examples#40

dmitriy-serdyuk:
Ok, this one seems fine as well.

dwf:
@dmitriy-serdyuk you might want to post a message to the mailing list
noting this change and that errors about either named_copy not existing or
var.copy not taking arguments should be resolved by updating blocks,
blocks-examples, and Theano.

dmitriy-serdyuk:
Thanks, I almost forgot about it.

PR #835, opened 2015-09-17
Allow uint type for lookup indexing

johnarevalo:
include uint prefix in dtype checking for lookup indexing.


rizar:
The test failure is weird, I am restarting the tests.

rizar:
Test might stay broken for a while because of #836 , so I will merge this seemingly harmless PR.

PR #839, opened 2015-09-22
Update deps

bartvm:


rizar:
I am lost. Why updating dependencies fixes Blocks builds but does not help for Fuel?

bartvm:
Merge this one and https://github.com/mila-udem/fuel/pull/237 simultaneously, and everything should work :)

bartvm:
Fuel is fixed with https://github.com/mila-udem/fuel/pull/237, but it's status is failing because it runs the tests on Blocks master, which are broken, and will only be fixed after merging this

rizar:
OK, let's try!

PR #840, opened 2015-09-28
Adding a positive bias to the LSTM forget gate

ablavatski:
According to the "Rafal Jozefowicz, Wojciech Zaremba, Ilya Sutskever, An Empirical Exploration of Recurrent Network Architectures, JMLR 2015" the LSTM with large forget bias outperformed all available recurrent units in almost all tasks: "But most importantly, we determined that adding a positive bias to the forget gate greatly improves the performance of the LSTM. Given that this technique the simplest to implement, we recommend it for every LSTM implementation". The pull request contains implementation of the technique.

### DIFF
--- a/blocks/bricks/recurrent.py
+++ b/blocks/bricks/recurrent.py
@@ -354,9 +354,9 @@ class LSTM(BaseRecurrent, Initializable):
         networks*, arXiv preprint arXiv:1308.0850 (2013).
     .. [HS97] Sepp Hochreiter, and Jürgen Schmidhuber, *Long Short-Term
         Memory*, Neural Computation 9(8) (1997), pp. 1735-1780.
-    .. [Jozefowicz15] Jozefowicz R., Zaremba W. and Sutskever I., *An Empirical
-        Exploration of Recurrent Network Architectures*, Journal of Machine
-        Learning Research 37 (2015).
+    .. [Jozefowicz15] Jozefowicz R., Zaremba W. and Sutskever I., *An
+        Empirical Exploration of Recurrent Network Architectures*, Journal
+        of Machine Learning Research 37 (2015).
 
     Parameters
     ----------
@@ -395,7 +395,7 @@ class LSTM(BaseRecurrent, Initializable):
         self.W_cell_to_in = shared_floatx_nans((self.dim,),
                                                name='W_cell_to_in')
         self.W_cell_to_forget = shared_floatx_nans((self.dim,),
-                                                   name='W_cell_to_forget')
+                                                name='W_cell_to_forget')
         self.W_cell_to_out = shared_floatx_nans((self.dim,),
                                                 name='W_cell_to_out')
         # The underscore is required to prevent collision with
@@ -417,7 +417,7 @@ class LSTM(BaseRecurrent, Initializable):
 
         if self.use_bias:
             self.b_cell_to_forget = shared_floatx_nans((self.dim,),
-                                                       name='b_cell_to_forget')
+                                                name='b_cell_to_forget')
             add_role(self.b_cell_to_forget, BIAS)
             self.parameters.append(self.b_cell_to_forget)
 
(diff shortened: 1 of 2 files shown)


blocks/bricks/recurrent.py
```diff
@@ -354,6 +354,9 @@ class LSTM(BaseRecurrent, Initializable):
         networks*, arXiv preprint arXiv:1308.0850 (2013).
     .. [HS97] Sepp Hochreiter, and Jürgen Schmidhuber, *Long Short-Term
         Memory*, Neural Computation 9(8) (1997), pp. 1735-1780.
+    .. [Jozefowicz15] Jozefowicz R., Zaremba W. and Sutskever I., *An
+        Empirical Exploration of Recurrent Network Architectures*, Journal
+        of Machine Learning Research 37 (2015).
```
dmitriy-serdyuk: Sphynx doesn't like citations which were not referenced from somewhere. And actually, I see no reason of mentioning this paper, we cannot mention everyone who worked with LSTMs.

Can you whether add some documentation concerning this paper or delete the reference?

blocks/bricks/recurrent.py
```diff
@@ -459,7 +472,8 @@ def slice_last(x, no):
         in_gate = tensor.nnet.sigmoid(slice_last(activation, 0) +
                                       cells * self.W_cell_to_in)
         forget_gate = tensor.nnet.sigmoid(slice_last(activation, 1) +
-                                          cells * self.W_cell_to_forget)
+                                          cells * self.W_cell_to_forget +
+                                          self.b_cell_to_forget)
```
dmitriy-serdyuk: This is going to fail if `not self.use_bias`.

dmitriy-serdyuk:
Thank you! Have you tested this implementation? Does it outperform the standard one?

ablavatski:
Thank you Dmitriy for comprehensive review of my code. I fixed all your issues you mentioned. I haven't tested this implementation of LSTM yet, but, I think, I will use it. As soon as I have some results, I will let you know.

rizar:
If this is really helpful, we should make it mandatory, I think. Just as currently all the initial states of recurrent networks are trainable, which is not a completely standard technique, but at least never hurts.

ablavatski:
So, what is the next steps for me for this pull request?To my mind, if we set "has_bias" to False, the LSTM will work as previously (without bias to forget) and you can merge this pull request. And as soon we have strong (practical) evidence that with bias it works better, we can set "has_bias" to True and new functionality will be in the framework. What do you think?

rizar:
My opinion is that we can just add these trainable biases without `use_bias` option. However, I have a bad feeling about using `biases_init` initialization scheme for that, because what we want is to set biases for forget gates, not reset gates, e.g. 

But it would really help if you try how it works in practice.

rizar:
Any results?

ablavatski:
Sorry, Dzmitry, but due to the busy conditions I didn't have enough time to fully test the proposed approach. On the task I tried, it didn't show large outperformance in comparison with traditional  LSTM. However, it doesn't mean that LSTM with bias doesn't outperform the traditional one. Most probable, my experiment wasn't good enough and reliable. To really prove something, some we need some experiments with real metrics. Also, if you really want to merge this changes your valuable comment about <code>biases_init</code> should be implemented before. As soon as I will have some free time I will try to check to versions atleast on some examples and we will see. I really apologize.

ablavatski:
Sorry, closed by mistake...reopen

rizar:
No rush, if you are still planning to test this change, I will not close
the PR.

cooijmanstim:
This is a bit of a random comment, but the biases introduced here will be redundant; LSTM takes its biases from the inputs. However, the improvement reported in the paper is due to *initial* bias. We don't need a new trainable parameter to implement that. All we need is an extra term in the expression for the forget gate:

```python
forget_gate = tensor.nnet.sigmoid(slice_last(activation, 1) +
                                  cells * self.W_cell_to_forget +
                                  self.initial_forget_gate_bias)
```

where `self.initial_forget_gate_bias` is a constant that can be specified by the user in the constructor (or lazily assigned).

rizar:
Thanks for this insight Tim! Do you have an experience with this trick,
does it actually help?

cooijmanstim:
I use it and it seems to work, but I don't have a simple experiment to back it up and I may be confounding it with the effect of identity initialization for the h-to-h matrix. It's worth noting that the originator of forget gates used positive initialization for the bias ("Learning to Forget: Continual Prediction with LSTM", ftp://ftp.idsia.ch/pub/felix/GersFA-NIPS.ps).

Issue #830, opened 2015-09-11
Improve visibility of Blocks-examples and add comments


rizar:
Currently Blocks README does not even mention Blocks-examples. The same goes for the documentation. Also, the examples should be more documented, because in fact they are a very (the most?) important part of all the help we provides for the users.

@fvisin , did I understand you right that you would be eager to take care of it?

fvisin:
I went through part of the documentation of Blocks and I took notes on what I would modify or improve. I volunteer to go through the rest of the documentation on www.blocks.readthedocs.org and improve it. I can also skim through the examples in Blocks-examples and add some comments, but it was not my main goal right now.

dwf:
I think there's already an open issue on blocks-examples to document them
better.

I think the first thing to do is to document your problems with the
documentation in the form of issues. It's fine if you don't do it all
yourself, in fact these make good beginner CCW tickets (for people with a
good command of English, anyway).

fvisin:
I created https://github.com/mila-udem/blocks/issues/831

@dwf I would also like to propose the rewording of a couple of sentences that I find ambiguous or not informative enough, but for that if you agree I would directly work on a PR (as soon as I have time), as it will take the same time to write what I would change and actually change it. Also, I think it will be easier to discuss my proposal in the PR in that case, as the modification will be sparse and local.

dmitriy-serdyuk:
Is it fixed?

rizar:
This can be closed, because no we have links to Blocks-examples from Blocks.

PR #844, opened 2015-09-29
Mention extras and examples

rizar:
Addresses #830 (does not solve this issue, because examples need more comments). 

Please let me know where else we should refer to blocks-examples.

### DIFF
--- a/docs/index.rst
+++ b/docs/index.rst
@@ -5,7 +5,8 @@ using Theano.
 
 Want to get try it out? Start by :doc:`installing <setup>` Blocks and having a
 look at the :ref:`quickstart <quickstart>` further down this page. Once you're
-hooked, try your hand at the :ref:`tutorials <tutorials>`.
+hooked, try your hand at the :ref:`tutorials <tutorials>` and the
+examples_.
 
 Blocks is developed in parallel with Fuel_, a dataset processing framework.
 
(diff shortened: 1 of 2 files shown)


docs/index.rst
```diff
@@ -23,6 +24,7 @@ Blocks is developed in parallel with Fuel_, a dataset processing framework.
 .. _mailing list: https://groups.google.com/forum/#!forum/blocks-users
 .. _making a GitHub issue: https://github.com/mila-udem/blocks/issues/new
 .. _Fuel: https://github.com/mila-udem/fuel
+.. _examples: https://github.com/mila-udem/blocks-examples
```
dmitriy-serdyuk: You don't need to define this reference again, Shynx doesn't like it. https://github.com/mila-udem/fuel/pull/232#discussion_r39973585

rizar:
@nke001, can you please check if referring to Blocks-examples this way solves the documentation issue that you came across yesterday?

docs/index.rst
rizar: But how I do two different link texts then?

nke001:
Hi, thanks for modifying the doc. It looks a lot more clear to me this way.

rizar:
@dmitriy-serdyuk  is away for a few days, a new reviewer is needed. @lamblin , could you please help?

rizar:
Thanks for reviewing it quickly!

PR #849, opened 2015-10-01
Refactor model and write a bit of documentation

### LINKED ISSUE #613: Model rationale
I'm still a bit hazy on why one needs the `Model` abstraction at all. It would be good to try and document this somewhere in the narrative documentation, and probably say a little something about it in the docstring.

rizar:
Fixes  #613.

This PR tries to make the concept of Blocks model very simple. It basically follows what @bartvm proposed half-a-year ago. The generality will kind of drop if it is merged, because e.g. an extension that requests `model.parameters` will not work with pure Theano graphs, for which `model.shared_variables` should be used. 

In the future development, this lack of generality can be addressed by introducing `ExtensionConfig` class, an instance of which would be an attribute of every main loop object. Then, a good extension would request `model.extension_config.parameters`, and at this point the user can intercept the request and make `ExtensionConfig` return `shared_variables`. However, I think it is reasonable to make an incremental improvement first by merging this PR and significantly reducing WTF-level for people trying to understand the current `Model`, which basically combines `ExtensionConfig` and a `ComputationGraph` extension. This is especially important before the tutorial scheduled for October 15.

@dwf, could you please review it?

### DIFF
--- a/blocks/model.py
+++ b/blocks/model.py
@@ -40,11 +40,12 @@ class Model(ComputationGraph):
     >>> y = mlp.apply(x)
     >>> model = Model(y)
 
-    With :class:`Model` you can get access to the brick hierarchy. The brick
-    hierarchy is defined by ``children`` attributes that every brick has.
-    The bricks that are not children of other bricks are called top bricks.
-    It is often useful to have access to top bricks of a brick hierarchy
-    used to build a computation graph, and here is how you can do it:
+    With :class:`Model` you can get access to the brick hierarchy. The
+    brick hierarchy is defined by ``children`` attributes that every brick
+    has.  The bricks that are not children of other bricks are called top
+    bricks.  It is often useful to have access to top bricks of a brick
+    hierarchy used to build a computation graph, and here is how you can do
+    it:
 
     >>> model.get_top_bricks() #doctest: +ELLIPSIS
     [<blocks.bricks.MLP object at ...]
--- a/blocks/model.py
+++ b/blocks/model.py
@@ -1,17 +1,20 @@
-"""Defines models.
+"""Model - heavily annotated computation graph.
 
-A model is a thin layer of abstraction between the user-defined computation
-graph, bricks, parameters and main loop extensions. This module provides
-the basic :class:`AbstractModel` interface as well as its implementations
-(currently only :class:`Model`).
+A model in Blocks is simply an annotated computation graph.  The class
+:class:`Model` extends :class:`blocks.graph.ComputationGraph` :class:,
+which is able to handle annotations and roles in general, but is
+deliberately made unaware of specific annotations that a Theano graph
+created by Blocks typically has, such as bricks and application calls.  The
+:class:`Model` adds this functionality. Using :class:`Model` you can do
+things like quering all the bricks used to build the computation graph,
+requesting "hierarhical names" of the parameters, etc.
+
+For more information, see :class:`Model` docstring.
 
 """
 import logging
-from abc import ABCMeta, abstractmethod
 from collections import OrderedDict, Counter
 from itertools import chain
-
-from six import add_metaclass
 
 from blocks.graph import ComputationGraph
 from blocks.select import Selector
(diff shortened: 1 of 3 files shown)


blocks/model.py
```diff
@@ -1,75 +1,105 @@
...
-graph, bricks, parameters and main loop extensions. This module provides
-the basic :class:`AbstractModel` interface as well as its implementations
-(currently only :class:`Model`).
+A model in Blocks is simply an annotated computation graph.  The class
+:class:`Model` extends :class:`blocks.graph.ComputationGraph` :class:,
+which is able to handle annotations and roles in general, but is
+deliberately made unaware of specific annotations that a Theano graph
+created by Blocks typically has, such as bricks and application calls.  The
+:class:`Model` adds this functionality. Using :class:`Model` you can do
+things like quering all the bricks used to build the computation graph,
```
dwf: query*

blocks/model.py
```diff
@@ -1,75 +1,105 @@
...
+requesting "hierarhical names" of the parameters, etc.
```
dwf: request*, to match the tense of query above; hierarchical*, also, not really clear from context what this means

blocks/model.py
```diff
@@ -1,75 +1,105 @@
...
+
+    Examples
+    --------
+    >>> from theano import tensor
+    >>> from blocks.bricks import MLP, Tanh
+    >>> x = tensor.matrix('x')
+    >>> mlp = MLP([Tanh(), Tanh()], [10, 10, 10])
+    >>> y = mlp.apply(x)
+    >>> model = Model(y)
+    >>> # With model you can get access to the brick hierarchy
```
dwf: You should break out of the example and go inti more detail here about what you mean by Brick hierarchy, what counts as a "top brick", etc.

blocks/model.py
```diff
@@ -1,75 +1,105 @@
...
     def get_parameter_dict(self):
-        """Return the model parameters.
+        """Returns parameters with their hierarchical names.
 
-        Returns
+        The parameter names are formed from positions of their owner bricks
+        in the bricks hierarchy. The variable names are used for the
+        parameters that do not belong to any brick.
+
+        Returns:
```
dwf: no colon

blocks/model.py
```diff
@@ -101,97 +132,12 @@ def set_parameter_values(self, parameter_values):
...
-    def get_objective(self):
-        """Return the optimization objective."""
-        pass
-
     def get_top_bricks(self):
-        """Return the top-level bricks that are used in the model.
+        """Get the bricks that do not have parents.
 
-        Returns
+        Returns:
```
dwf: no colon

blocks/model.py
```diff
@@ -1,75 +1,105 @@
...
+    >>> model.get_parameter_dict() #doctest: +NORMALIZE_WHITESPACE
+    OrderedDict([('/mlp/linear_1.b', b), ('/mlp/linear_0.b', b),
+    ('/mlp/linear_0.W', W), ('/mlp/linear_1.W', W)])
 
     """
-    @abstractmethod
+    def __init__(self, *args, **kwargs):
+        bricks = [get_brick(var) for var
+                  in self.variables + self.scan_variables if get_brick(var)]
+        children = set(chain(*(brick.children for brick in bricks)))
```
dwf: Huh, didn't know you could pass an iterator as *args. Does this work on Python 2?

blocks/model.py
```diff
@@ -1,75 +1,105 @@
...
+            if brick not in children and brick not in self.top_bricks:
+                self.top_bricks.append(brick)
+        names = Counter([brick.name for brick in self.top_bricks])
+        repeated_names = [name for name, count in names.items() if count > 1]
+        if repeated_names:
+            raise ValueError("top bricks with the same name:"
+                             " {}".format(', '.join(repeated_names)))
+        brick_parameter_names = {
+            v: k for k, v in Selector(
+                self.top_bricks).get_parameters().items()}
```
dwf: six.iteritems(...)

blocks/model.py
```diff
@@ -1,75 +1,105 @@
...
+        parameter_list = []
+        for parameter in self.parameters:
+            if parameter in brick_parameter_names:
+                parameter_list.append((brick_parameter_names[parameter],
+                                       parameter))
+            else:
+                parameter_list.append((parameter.name, parameter))
+        self._parameter_dict = OrderedDict(parameter_list)
+
     def get_parameter_dict(self):
```
dwf: Not really crazy about this method's name. Not as descriptive as it could be, tells you the type but not semantics.

dwf: I also don't like that it's a method and not an attribute/property. When something is returned by a method, I do not expect that mutating it mutates the object whose method returned it, yet that is what happens here.

blocks/model.py
rizar: Apparently it does, since tests pass and nobody complains.

blocks/model.py
rizar: This is really not a performance bottleneck, are you sure we need `iteritems()` here?

rizar:
I think I addressed all your docstring concerns @dwf. I am not really willing to change the interface in this PR. I think it would be great to merge it before release of 0.1, to make a better-documented stable version, which is however compatible with the current Blocks code people have. We can break interface as much as we want after the release, of course.

I am shocked by the coverage drop...

blocks/model.py
dwf: If you have a model with lots and lots of parameters then this needlessly allocates a bunch of memory on Python 2. But I guess it's not a big deal.

rizar: Hmm... it depends on how shared variables are copied. If a new cuda array
is allocated for the copy, then we should really change the code here. And
probably, it is better not to rely on them being shallow copied even if
they are and use iteritems. I will update the PR.

dwf: Sorry, I just meant the memory needed for the list of pairs created by items(), rather than the iterator of pairs you get on Python 3.

rizar: Oh sure, the elements are actually not copied... I reverted it back to items() :)

dwf:
Normally I'd say -0.4% coverage drop is nothing to worry about but looking at the coverage report, I'd sort of like to see [this](https://coveralls.io/builds/3756309/source?filename=blocks%2Fmodel.py#L89) line covered by at least one test.

rizar:
Ditto.

rizar:
I think the PR is ready.

rizar:
PING!

rizar:
Ping.

dwf:
Yep, on it, sorry, had things to do this evening in the real world :)

rizar:
Great! Sorry for pushing, I just would like to have it in before the tutorial (13:30 tomorrow)

tests/test_model.py
```diff
@@ -50,3 +52,12 @@ def test_model():
     def helper():
         Model(mlp4.apply(mlp3.apply(x)))
     assert_raises(ValueError, helper)
+
+
+def test_model_handles_brickless_parameteres():
```
dwf: Spelling error, but I'll add this as a softball ticket for someone's first CCW PR.

PR #851, opened 2015-10-01
WIP: Batch normalization

cooijmanstim:
This continues the work done on #513. Current status:

Gammas/betas no longer need to be provided by the user, but the default initialization can be overridden with the given_gammas/given_betas kwargs. The initial population statistics can be similarly specified. These keyword arguments should be dicts with variables for keys and initial values for values.

The previous version of apply_batch_normalization had a use_population keyword argument that allowed per-variable switching between batch/population statistics. If necessary, this can now be done by multiple calls to apply_batch_normalization.

To get from here to a fully functional BN implementation, we need to think about how to save/load the BN parameters. The gammas and betas are already recognized by blocks as parameters, and so the user could construct the given_gammas/given_betas argument from a saved model's parameter dictionary (maybe a helper function for this would be convenient). The population statistics however aren't saved/loaded yet when they definitely should be as they are part of the knowledge the model has acquired. I am open to suggestions on how to do this.

The gist at https://gist.github.com/cooijmanstim/8b570f38cbb174ab0438 shows an example of the new usage.


rizar:
Thanks for making a new PR. The gist link does not open :(

Does this code change the means and variances during inference? I only see one set of updates returned and two graph computation graphs.

dwf:
@cooijmanstim Thanks for picking this up.

> we need to think about how to save/load the BN parameters. 

By "saved" you mean serialized with a checkpoint? It seems like where to save that entire test-time graph is more the issue, isn't it? One would like to be able to get access to and evaluate it.

One thing I think we should leave up to the user is whether to adapt these via some additional squared error objective function passed as a term of the cost function to GradientDescent (which seems like a straightfoward and easy way to do it) or just via some simple moving average updates added with `add_updates`. In either case I don't think it's a problem to consider these quantities PARAMETERs as they don't appear in the train-time graph and therefore won't be adapted if you do gradient descent on the train-time graph's (single) output.

There's a larger arc to this story which is whether the graph substitution framework is really appropriate for this situation. Other frameworks are tackling this by adding a "Brick" equivalent into the original model object hierarchy, which makes it a little easier to figure out what's going on. In that kind of a setup, from a  Blocks perspective, we'd essentially have two apply methods, one for train time and one for test time. In that situation, however, we'd need some way of conveniently excluding the learned means and variances from gradient descent on the real objective function, which probably means a new parent role for PARAMETER, like ADAPTABLE or something.

cooijmanstim:
@rizar I fixed the gist link.

The means and variances change through the updates. In the example the updates are added to GradientDescent, and so they are only changed during training. They could be updated during inference by e.g. passing the updates to the monitoring extension.

@dwf Your suggestion of adding PARAMETER roles seems like a good solution to me, at least for now. The user would get an unhappy surprise though if they ever use one of the inference graph outputs in the training graph.

I like the idea of allowing the user to control the strategy of estimating the statistics. I'll think about how that should work.

Personally I like the graph substitution approach much better because it is more generally applicable. Also as a user I would very much like to outsource the complex and confusing burden of constructing the two different graphs to a simple function call at the end of my model construction code.

rizar:
Correct me if I am wrong, but if we go for the "brick" solution, one will never be able to add batch norm to existing bricks, including MLP? If it's true, I think it is a strong argument to use replacements.

cooijmanstim:
I have been working on an improvement suggested by @rizar that would attach the updates to the graph using annotations. The code is basically done but I'm running into some problems. Using annotations requires giving up the distinction between `initialization_updates` and `accumulation_updates`, which can be done using ifelse with a shared variable denoting whether the initialization has occurred or not. Unfortunately it turns out that there are some issues with the gradient of ifelse that can cause its branches to be evaluated unconditionally. This is a problem for the `gamma` and `beta` parameters, which do need a conditional to avoid using them until we know their shapes.

rizar:
Thanks for letting us know the status. I am not sure I understand though
why the gradient of the updates is taken.

cooijmanstim:
It isn't, rather the gammas and betas appear in the training graph as ifelse(initialized, gamma, T.ones_like(batch_mean)), and funny things happen when the gradient of that is taken.

cooijmanstim:
To be clear, by the chain rule (f(g(x))' = f'(g(x))g'(x)) the gradient with respect to gamma of that expression is  `grad(ifelse)(initialized, gamma, T.ones_like(batch_mean)) * grad(gamma)`. The shape of `grad(gamma)` is incorrect initially, and the code breaks.

dwf:
An ifelse in the training graph is less than ideal... can it be avoided?

cooijmanstim:
Yes, by returning initialization updates and having the user make sure they are performed before anything else, as done in the version this pull request is at.

rizar:
But to make sure that these updates are done before anything else you have
to go to great lengths like writing a special extension to do that. I do
not have a better idea for now :(

cooijmanstim:
There's another problem when the user has to deal with updates, which is that any further replacements the user performs in the graph (e.g. dropout or additional batch normalization) are not performed on the updates. If you applied batch normalization and then dropout, the batch normalization updates would be in terms of the non-dropped-out variables. This is okay in this case for the initialization updates because we only need the shape, but in general it's incorrect and slow.

Maybe we should go with the brick approach after all. If we make one brick that carries out the batch normalization and another that is basically a FeedforwardSequence with batch normalization followed by some user-specified activation function. This could be used as the activation function in e.g. MLP. This is how I've been doing batch normalization prior to picking up this PR.

rizar:
Without expressing a preference for either of the choices, I would like to
point out that the fact that replacements do not act on auxiliary variables
and updates is a known major design issue:
https://github.com/mila-udem/blocks/issues/514.

cooijmanstim:
I'm abandoning this in favor of whatever comes out of #920.

PR #852, opened 2015-10-01
Move testing utilities to make them importable from outside

rizar:
I just want to use them for testing in blocks-extras. This PR should be super-easy to review.


rizar:
The coverage drop is not a big surprise: some of the moved code is not used _inside_ tests, only to wrap them. The scrutinizer crash is mysterious.

rizar:
Tests pass, impatiently waiting for an LGTM.

rizar:
@dwf , @dmitriy-serdyuk , can anybody authorize me to merge this?

dmitriy-serdyuk:
Looks good, but do we want it to be available after setup?

rizar:
What do you mean?

dmitriy-serdyuk:
A user who installed blocks with `setup.py install` doesn't need all these testing utilities in their PYTHONPATH. Maybe it's not a big problem.

rizar:
I am not sure I agree. The whole point here is to make these utilities available as soon as Blocks is installed, so that other packages using Blocks could use them for their tests. So they should be copied to the system (or local) directories upon installation together with the rest of the code.

dmitriy-serdyuk:
Ok, agreed.

rizar:
Is it LGTM?

PR #859, opened 2015-10-06
Add link to fuel installation

dmitriy-serdyuk:
This fixes installation before we have a version at pypi. It should also fix Read the docs builds.

The problem was that `setup.py install` tried to search fuel at PyPI, didn't find it and crashed. So the API documentation cannot be built, since blocks cannot be imported...


dmitriy-serdyuk:
@rizar , please review, this is very simple change but it should solve many problems. After we place fuel and blocks to PyPI we should revert it.

dmitriy-serdyuk:
Probably it won't fix the problem with documentation, I'll try to do the same as in mila-udem/fuel#245.

dmitriy-serdyuk:
Oh, fuel is already wrapped into mock.

PR #861, opened 2015-10-07
RTD/installation fixes

### LINKED ISSUE #752: all subsections of "API Reference" in readthedocs are missing
except for Variable roles

### LINKED ISSUE #846: Extensions docs not showing up on RTD
https://blocks.readthedocs.org/en/latest/api/extensions.html

bartvm:
Analogous to https://github.com/mila-udem/fuel/pull/247

Fix #846, fix #752.


bartvm:
All tests pass (https://travis-ci.org/mila-udem/blocks/builds/84013503), reverting the last commit that was using Fuel's `rtd` branch so that things can be merged. Needs to be merged simultaneously with https://github.com/mila-udem/fuel/pull/247.

bartvm:
@dmitriy-serdyuk @vdumoulin I know you guys were trying to fix RTD; these two PRs do that, so PTAL :)

dmitriy-serdyuk:
Cool! Let's try it! LGTM

dmitriy-serdyuk:
I'm a bit concerned that RTD runs `./setup.py install` instead of `pip install` and it's not configurable. So it's still going to crush when check if fuel is installed.

bartvm:
I disabled the virtual environment, so it doesn't run `setup.py install` at all anymore (see https://readthedocs.org/projects/blocks/builds/3380159/)

rizar:
Looking at the [latest of build of the master branch]( https://readthedocs.org/projects/blocks/builds/3382267/) I see all kinds of crashes of import commands like `import theano` and `import picklable_itertools`. And the sections that used to be empty apparently remained empty. @bartvm , are you sure that the current Sphinx version does not need mocks?

bartvm:
That's interesting... It's giving import errors, but in the virtual environment that caused errors I think, and now it's happily green. Anyway, I'll add the mocked modules back in!

dmitriy-serdyuk:
I tried to return mocks and it is still doesn't work. You can take a look at my build: https://readthedocs.org/projects/blocks-test/builds/, it's okay, but the documents http://blocks-test.readthedocs.org/en/latest/ still Extentions section is empty.

dmitriy-serdyuk:
Oops, it was my browser cache.

bartvm:
I've got the Blocks one building as well (https://readthedocs.org/projects/blocks/builds/3382460/, not sure if you can see it) with a few less dependencies in `req-rtd.txt`, but Fuel still chokes because that Cython module doesn't get compiled.

dmitriy-serdyuk:
I'll build as well and see if all the sections are there. I don't like RTD because you cannot see the docs built for a PR.

bartvm:
@dmitriy-serdyuk Sorry for this broken PR, #864 should fix things (for real this time!)

bartvm:
https://blocks.readthedocs.org/en/rtd2/

PR #864, opened 2015-10-07
Add mocking back in

bartvm:
Follows up on https://github.com/mila-udem/blocks/pull/861.

So, at first I tried to minimize the number of dependencies in `req-rtd.txt` and mock everything instead, since I figured that was faster and less error-prone than managing dependencies in RTD.

However, there's some strange bug that causes mocked modules to crash when `from * import *` or relative imports are used... (see [this SO question](https://stackoverflow.com/questions/32579380/mock-with-submodules-for-readthedocs)). Hence, mocking modules might just break the docs the moment someone uses a `from * import *` statement... So, I went the other direction and tried minimizing the number of modules that are mocked and am installing as many of the requirements as possible (so that the docs won't crash when someone adds imports in the future).

Everything builds without errors now I believe.


dmitriy-serdyuk:
I double checked, it runs without errors but with several warnings (I guess they were always there). Let's wait for tests and merge.

bartvm:
Yeah, those errors are all from the actual docstrings, unrelated to my PRs. They tend to sneak in because there's no tests that will fail when they appear.

dmitriy-serdyuk:
Future improvement: we should build documents when testing with Travis.

rizar:
Well, it will still not be the same environment as RTD.

It would be great to automatically run RTD for all pull requests though..

rizar:
@bartvm , can you please add us as maintainers of readthedocs project for Blocks?

dmitriy-serdyuk:
Congrats, we finally have API docs: http://blocks.readthedocs.org/en/latest/api/extensions.html.

bartvm:
Added @rizar and @vdumoulin, but @dmitriy-serdyuk doesn't exist according to RTD. I guess it ignores the existence of people with hyphens in their GitHub usernames?

dmitriy-serdyuk:
Try serdyuk.dmitriy@gmail.com

bartvm:
That seems to have worked!

Getting Travis to trigger RTD builds... Will probably require some trickery. The API of RTD is too limited, but there's a webhook we could use. First thing that comes to mind is something like:

* Generate an `rtd-build` branch, activate this branch on RTD
* Generate a [GitHub personal access token](https://help.github.com/articles/creating-an-access-token-for-command-line-use/)
* [Encrypt](http://docs.travis-ci.com/user/encryption-keys/) this token so that we can use it on Travis
* Make Travis use the token to push the current PR commit to the `rtd-build` branch
* Use the post-commit hook (`curl -X POST https://readthedocs.org/build/blocks`) to trigger builds
* Get the [latest builds using the API](https://read-the-docs.readthedocs.org/en/latest/api.html#method-get-api-v1-build-id-), match the current commit of the PR to the commit in the `setup` field of one of the builds of `rtd-build`
* Loop and wait until `state` reaches `finished`
* Return the value of the `success` to determine pass/failure of the test (optionally: parse `output` to determine if new warnings were introduced)

PR #867, opened 2015-10-08
Make Convolutional's conv implementation hackable.

dwf:
This enables a workaround for using some experimental Theano functionality not yet available in the mainline convolution ops. It's a small change that simplifies user code greatly in this respect.

### DIFF
--- a/blocks/bricks/conv.py
+++ b/blocks/bricks/conv.py
@@ -44,7 +44,6 @@ class Convolutional(Initializable):
         Defaults to ``False``.
 
     """
-
     # Make it possible to override the implementation of conv2d that gets
     # used, i.e. to use theano.sandbox.cuda.dnn.dnn_conv directly in order
     # to leverage features not yet available in Theano's standard conv2d.
--- a/blocks/bricks/conv.py
+++ b/blocks/bricks/conv.py
@@ -44,6 +44,13 @@ class Convolutional(Initializable):
         Defaults to ``False``.
 
     """
+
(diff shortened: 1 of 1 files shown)


dwf:
Oops, it wasn't passing tests before this, I had made the necessary change locally but not on the PR.

@rizar Can you have a look at this?

dmitriy-serdyuk:
Good idea! Do you know if all the implementations have a similar interface?

dwf:
Sadly the `filter_shape` keyword is not supported, by `dnn_conv`. But I think an interim solution would be to make `dnn_conv` ignore the `filter_shape` and `image_shape` keywords.

dwf:
(Another option would just be to pass in a function that wraps `dnn_conv` and discards these.)

dmitriy-serdyuk:
And also different implementations may have different arguments for changing behaviour of the Op. Yes, wrapping is a solution, can you write in a doc about it?

dwf:
Where's the appropriate place for such documentation? In the # comments I've already written? (I don't think this is quite "user-level" functionality, but an unsupported hack that is there for those who want to get dirty.)

dmitriy-serdyuk:
Yes, I agree, the comment is an appropriate place. Although, generally there is a chapter like "Development" or "Extending library", but I don't know how to separate this piece of documentation and still have it here in the code. 

After this change, LGTM.

dwf:
@dmitriy-serdyuk 620bc7b

PR #868, opened 2015-10-08
warn if setting wrong shape parameters

lukemetz:
Setting the wrong weight shape for a model is surprisingly hard to debug as it generally results in a runtime error. This pull request warns the user in this case.

### DIFF
--- a/blocks/model.py
+++ b/blocks/model.py
@@ -99,6 +99,11 @@ class AbstractModel(object):
 
         for name, value in parameter_values.items():
             if name in parameters:
+                model_shape = parameters[name].container.data.shape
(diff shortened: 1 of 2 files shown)


bartvm:
`X.container.data.shape` works

lukemetz:
I was worried about that. `X.container.data.shape` is a good idea. Updated now.

rizar:
Sorry for the delay! Please add test that checks that the exception is raised and we will merge this PR.

lukemetz:
Done. Also, changed the logger.error to an exception as this seems to make more sense for this kind of error.

Issue #773, opened 2015-07-24
Initializable _push_initialization_config


lukemetz:
A frequent use case I have when using blocks involves creating some graph of bricks, (most of which recursive), and needing to fine tune only a few of the children bricks weights. The workflow suggested by the docs is to set the top level initialization and then call `push_initialization_config` then manually go in and set the required weights. This seems like spreading out the configuration across multiple different locations in a code base as well as leaving no easy way to create prefab bricks with initializations already set inside them. In addition to this, the write over nature seems unintuitive, as there is no warning when a user manually sets weights_init and sees no effect. For example:
    
    class MultiBrick(Initializable):
        def __init__(self, **kwargs):
             self.lin1 = Linear(weights_init=Constant(1))
             self.lin2 = Linear()
             ......
    b = MultiBrick(self, weights_init=Constant(0), bias=Constant(0))
    b.initialize()

I would expect b.lin1.weights_init to be Constant(1) but this is not the case. In this case its easy enough to set it manually post `push_initialization_config` but this is not true the larger the graphs get.

As a solution, could the `Initializable._push_initialization_config` the first check if the `child.weights_init` exists and if it does not overwrite and leave the original value? I quick look at the source says `Initialization` is the only class using `_push_initialization_config`.

Thanks for your consideration.

sotelo:
+1

rizar:
Hi Luke,

it seems to me that what you that the following should work for you:

class MultiBrick(Initializable):

      def _push_initialization_config(self):
           super(self, MultiBrick)._push_initialization_config(self)
           self.lin1.weights_init = Constant(1)
           self.lin1._push_initialization_config()

Does this make sense?

Dima

lukemetz:
@rizar That was the missing link. Should have really thought of that. I retract my original idea.

Thanks for the help.

koningrobot:
@rizar's solution works in this case but in general I would prefer @lukemetz's proposal.  If you didn't have control over MultiBrick's constructor you would still want to be able to specialize its children's initialization configuration.

rizar:
@koningrobot , definitely, but in this case you just need to push `MultiBrick` induced initialization first and override it:

```
brick = MultiBrick(weights_init=Constant(0))
brick.push_initialization_config()
brick.lin1.weights_init = Constant(1)
```

koningrobot:
[...]

PR #873, opened 2015-10-15
ConvolutionalSequence: Don't auto push border_mode.

### LINKED ISSUE #681: Give ConvolutionalSequence a saner default for border_mode
It forces all convolutional layers it wraps to have the same border mode, which is a nightmare to fix if you happen to have layers with 'full' border mode and others with 'valid' border mode.

dwf:
Only push a `border_mode` when the `border_mode` is explicitly specified. Otherwise rely on child defaults/manual specification.

Fixes #681. Props to @vdumoulin and @bartvm for already figuring out and agreeing on the solution, this just goes and implements it.

### DIFF
--- a/blocks/bricks/conv.py
+++ b/blocks/bricks/conv.py
@@ -356,6 +356,14 @@ class ConvolutionalSequence(Sequence, Initializable, Feedforward):
         Width and height of the input (image/featuremap). If given,
         will be passed to theano's convolution operator resulting in
         possibly faster execution.
+    border_mode : 'valid', 'full' or None, optional
+        The border mode to use, see :func:`scipy.signal.convolve2d` for
+        details. Unlike with :class:`Convolutional`, this defaults to
+        None, in which case no default value is pushed down to child
+        bricks at allocation time. Child bricks will in this case
+        need to rely on either a default border mode (usually valid)
+        or one provided at construction and/or after construction
+        (but before allocation).
 
     Notes
     -----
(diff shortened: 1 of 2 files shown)


dwf:
@rizar @dmitriy-serdyuk Any thoughts on this one? See @vdumoulin's [rationale](https://github.com/mila-udem/blocks/issues/681#issuecomment-106912877) for why this feature is necessary (I also spelled it out in docs in the diff).

dmitriy-serdyuk:
I think, it's a general issue, that `_push_allocation_config()` frequently overwrites stuff you defined in the constructor. I remember I was a bit confused about it some time ago.

dwf:
It is, and we can have a discussion about what the semantics of `_push_allocation_config` ought to be (I think there already may be a ticket). However this is one instance where *absolutely unambiguously* the behaviour of overwriting stuff creates a lot of busywork.

dmitriy-serdyuk:
I agree, that it might be crucial for convolutions.

I have a vague idea how to fix it in general. It should be something like a default dictionary of allocation parameters which doesn't allow to set a value second time.

dwf:
In general, we want the user to be able to reassign them manually, but want to prevent `push_allocation_config` from doing so. We can do this by requiring push_allocation_config to go through a separate interface which validates that the only thing it's overriding is `NoneAllocation` Anyway, can we merge this as an interim fix as both Vincent and I have run into it?

(@bartvm I vaguely recall having this discussion with you before.)

dwf:
@dmitriy-serdyuk See discussion on #773 (it was about initialization config rather than allocation config, but same issue). Clearly this is contentious and won't be solved very soon, but I'd like to get this rather slight improvement in before that.

rizar:
I am by no means an expert in convnets, but I think I understand the problem, and your solution is reasonable. I am in favour of merging it.

In the long term we should come up with clearer semantics for configuration in the tree, I think.

dwf:
@rizar the example in @vdumoulin's comment on #681 makes it clearer. Shall we merge then?

dmitriy-serdyuk:
Yes, let's merge.

PR #875, opened 2015-10-15
Make Convolutional{Activation,Layer,Sequence} honour use_bias

### LINKED ISSUE #847: ConvolutionalActivation, ConvolutionalSequence do not honor use_bias=False
ConvolutionalActivation does not push use_bias to its Convolutional child, nor does ConvolutionalSequence:

    import theano.tensor as T

    from blocks.bricks import conv, Rectifier
    from blocks import initialization, graph

    cnn = conv.ConvolutionalSequence([
        conv.ConvolutionalActivation(
            activation=Rectifier().apply,
            filter_size=(1, 1),
            num_filters=1,
            use_bias=False)
        for _ in xrange(3)],
        num_channels=1, image_size=(1, 1),
        use_bias=False,
        weights_init=initialization.IsotropicGaussian(1e-3),
        biases_init=initialization.Constant(0))
    cnn.initialize()

    x = T.tensor4()
    y = cnn.apply(x)

    print graph.ComputationGraph(y).parameters

Removing biases (e.g. for batch normalization) requires reaching into internals.

dwf:
Fixes #847 (which I had said was a CCW ticket but turned out to be a little bit involved).

@cooijmanstim Can you give this a look?

@dmitriy-serdyuk You too? Should be quick.

### DIFF
--- a/blocks/bricks/conv.py
+++ b/blocks/bricks/conv.py
@@ -399,6 +399,7 @@ class ConvolutionalSequence(Sequence, Initializable, Feedforward):
             layer.image_size = image_size
             layer.num_channels = num_channels
             layer.batch_size = self.batch_size
+            layer.use_bias = self.use_bias
 
             # Push input dimensions to children
             layer._push_allocation_config()
--- a/blocks/bricks/conv.py
+++ b/blocks/bricks/conv.py
@@ -223,7 +223,7 @@ class _AllocationMixin(object):
     def _push_allocation_config(self):
         for attr in ['filter_size', 'num_filters', 'border_mode',
                      'batch_size', 'num_channels', 'image_size',
-                     'tied_biases']:
+                     'tied_biases', 'use_bias']:
             setattr(self.convolution, attr, getattr(self, attr))
 
 
--- a/blocks/bricks/conv.py
+++ b/blocks/bricks/conv.py
@@ -219,7 +219,15 @@ class MaxPooling(Initializable, Feedforward):
                                                        st=self.step))
 
 
-class ConvolutionalActivation(Sequence, Initializable):
+class _AllocationMixin(object):
+    def _push_allocation_config(self):
+        for attr in ['filter_size', 'num_filters', 'border_mode',
+                     'batch_size', 'num_channels', 'image_size',
+                     'tied_biases']:
+            setattr(self.convolution, attr, getattr(self, attr))
+
+
+class ConvolutionalActivation(_AllocationMixin, Sequence, Initializable):
     """A convolution followed by an activation function.
 
     Parameters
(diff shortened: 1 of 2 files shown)


cooijmanstim:
Sorry, I see now that I posted my comment on the issue rather than here on the pull request. Looks good to me, thanks!

dwf:
Thanks, I'd like another person with merge rights to sign off on it before I hit the big green button, but it should be in soon.

dwf:
By the way, is this useful/critical for #851? The ticket seemed to sound like it.

cooijmanstim:
No, I intend to touch that PR again this weekend. It's an "interesting" complication though, that users will still have to worry about redundant biases (which may not be redundant depending on tied_bias settings).

dwf:
Okay, that's kinda why I tackled it (and because I was digging around in that code anyway with #873). But glad to know it isn't a blocker.

rizar:
Hey there, I saw recently a proposal to add `_push_allocation_config` to `Initializable` which would handle `use_bias` option. Is there a particular reason why you do no go this way?

Alternatively, what if we just let go of `use_bias`? It is not that much harm to have `biases_ini` propagated to bricks that do not have biases.

dwf:
@rizar I made that proposal at the same time I was doing this, because
ultimately that's a more elegant solution. I think that is probably the
right direction in the long term, but in the short term this fixes what is
clearly a bug in the simplest way possible. We should not let the perfect
be the enemy of the good.

rizar:
I agree that fixing things makes sense. But let me go through all recent bias-related issues for the record:

https://github.com/mila-udem/blocks/issues/847 - `use_bias` is forgotten in the list of the settings pushed downward in `ConvolutionalLayer`, the code sort of relies on it be pushed by `Initializable` but `Initializable` does not do that.

https://github.com/mila-udem/blocks/issues/789 - `use_bias` is often never mentioned in the docstring of the bricks that actually support this option, e.g. `Linear`

https://github.com/mila-udem/blocks/issues/697 - it might be not very to get why `Initializable` brick have both `use_bias` and `has_biases`

So it seems they all have a single reason behind: `use_bias` is declared in `Initializable`, but it is the actual brick where it is handled. I find this strange and I think we should remove `use_bias` from `Initializable` and only have it in the bricks that actually have biases. If you agree with that, I will create an issue.

dwf:
@rizar Maybe a simple way to handle this, presuming basically every brick that has a bias has it on by default anyway:

```python
class HasBias(object):
    @property
    def use_bias(self):
        return getattr(self, '_use_bias', True)

    @use_bias.setter
    def use_bias(self, value):
        self._use_bias = value
```

Inherit from this mixin in every brick that should have a bias.

rizar:
Your code example misses a constructor, and having a constructor means that
also an additional super call will be necessary. Which makes me wonder if
having it as a mixin is better then simply add the related code to every
class. Perhaps it is, because this is a way to declare the interface.

dwf:
Just to be clear, as written, it doesn't actually need a constructor.

rizar:
Okay, but assuming `Linear` inherits from `HasBias`, will it accept
`use_bias=False` as a constructor parameter?

dwf:
I see what you're saying now. Yes, for that we'd need to either add it to
every class or else add a constructor and add it to the chain of super()
calls.

rizar:
Right. But as I said before, maybe it does make sense to have it as a mixin
so that all classes that have biases definitely had the same interface.

PR #877, opened 2015-10-21
WIP: New serialization

### LINKED ISSUE #829: Finish new serialization
Finish ~~#786~~ #877

@dmitriy-serdyuk

Thrandis:
(Fixes #829)

Regarding the previous todo list:

- [x] documentation is missing for many functions
- [x] handle usual shared variables (not created by bricks)
- [x] handle shared variables without names
- [x] handle variables with duplicate names
- [x] support CUDA

What I still need to do:

- [ ] write the tests
- [ ] complete the docstring
- [x] fix the hack in add_to_dump
- [x] add the protocol to pickle

Finally, when we move to the current blocks version:

- [ ] adapt existing interface
- [ ] adapt the tempdir

Finally, I don't really like the name `parameters`, because it is not general enough. It is true that in practice we will use it to save the parameters of our models separately, but I think it would be better to use something like `separated_parts` and `separated_shared_variables`. The name is long though.


blocks/serialization.py
```diff
@@ -1,31 +1,167 @@
...
+as drop-in replacement for the respective functions from the standard
+:mod:`pickle` module. The main differences between them and the standard
+ones are:
+
+    - The dump is physically a tarball, in which the pickle is stored
+      as '_pkl' file.
+
+    - Parts of the dumped object can be pickled separately but be
+      referenced from '_pkl' file using the persisent id mechanism (see
+      :mod:`pickle`) docs. Such parts are stored as arbitrarily named files
```
rizar: The closing parenthesis should be placed after the word _docs_.

blocks/serialization.py
```diff
@@ -1,31 +1,167 @@
...
+      of a selected set of Theano shared variables. Again, this data is
+      referenced from `_pkl` using persistent id mechanism, which means
+      that no duplication takes place. The goal here is to save the values
+      of the parameters (this is what these shared variables are in most
+      cases) in the most robust way possible. The actual format for
+      '_parameters' file is the one used by :func:`numpy.savez`, i.e. a zip
+      file of numpy arrays.
+
+    - Pickling of the whole object in fact can be bypassed if pickling of
+      some parts and parameters is sufficient.
```
rizar: and -> and/or

blocks/serialization.py
```diff
@@ -1,31 +1,167 @@
...
+
+>>> import numpy
+>>> ps = numpy.load(tarball.extractfile(tarball.getmember('_parameters')))
+>>> sorted(ps.keys()) # doctest: +ELLIPSIS
+['|mlp|linear_0.W', '|mlp|linear_0.b', '|mlp|linear_1.W', '|mlp|lin...]
+>>> ps.close()
+
+The names for parameters are chosen intellegently to reflect their
+position in the brick hierarchy, if they belong to bricks, and by
+simpling using the `.name` attribute, if they do not.
```
rizar: simpling -> simply

blocks/serialization.py
```diff
@@ -33,142 +169,93 @@
...
+
+            for name, p in named_parameters.items():
+                array_ = p.container.storage[0]
+                external_objects[id(array_)] = _mangle_parameter_name(
+                    type(array_), name)
+        if parameters:
+            _taradd(_save_parameters, tar_file, '_parameters')
+        if pickle_separately:
+            for name, component in six.iteritems(pickle_separately):
+                def _pickle_separately(f):
```
rizar: As we discussed, the fact that we do not add ids of the objects to  `external_objects` here means they will be pickled together with the whole object. We should decide if we are fine with that, and if yes, change the docstring of the module.

rizar:
So my early impression is that you did a very nice work, but I stopped reading when I hit the first big open design issue. Let's take a decision first.

Thrandis:
@rizar Following what we discussed, I'll make the serialization without persistent ID, ie what is pickled separately will be copied.

rizar:
That's fine with me. It seems like this is exactly what the code does now,
though the docs should be updated to reflect it.

Thrandis:
@rizar the bad thing is that now we have a complicated code that basically pickles (and unpickles) objects into (from) a tar file. And we have a code to rename parameters. That's it.
I feel that without the persistent ID this code is not that great!

rizar:
True, but as long as we have persistent ids for parameters it is still kind of great, I think :)

I thought a bit more about your latest proposal, and I have questions, so we should probably talk at the lab.

Thrandis:
@rizar all right! I'll wait for you then!

Thrandis:
@rizar here is the new serialization where you have a `dump` and a `load` function that are similar to the ones from `pickle`. The main difference is that the parameters can be dumped separately.
Then there is the function `add_to_dump` that allows you to add more objects to the existing archive. The nice feature is that if the parameters are the same as the ones already in the archive, then you can pass them to the parameter `parameters` and they are not serialized a second time!

rizar:
Great, thanks Cesar! I will try to review it tomorrow.

blocks/serialization.py
```diff
@@ -34,141 +142,65 @@
...
     r"""Robust serialization - does not corrupt your files when failed.
 
     Parameters
     ----------
     object_ : object
         The object to be saved to the disk.
-    path : str
-        The destination path.
+    file_ : str
+        The destination for saving.
```
rizar: What is the reasoning behind this renaming?

blocks/serialization.py
```diff
@@ -180,19 +212,150 @@ def secure_dump(object_, path, dump_function=dump, **kwargs):
...
+        _taradd(_save_object, tar_file, name)
+
+
+def continue_training(file_):
     """Continues training using checkpoint.
 
     Parameters
     ----------
-    path : str
+    file_ : str
```
rizar: The same question here, I think `path` is a more informative name when a string is expected.

blocks/serialization.py
```diff
@@ -180,19 +212,150 @@ def secure_dump(object_, path, dump_function=dump, **kwargs):
...
+                parameters = numpy.load(
+                    tar_file.extractfile(tar_file.getmember('_parameters')))
+                s1 = set(parameters.keys())
+                s2 = set(external_parameters.keys())
+                if s1.issuperset(s2):
+                    raise ValueError('The set of parameters is different' \
+                                     ' from the one in the archive.')
+
+    # TODO: How to avoid this hack?
+    file_.close()
```
rizar: Maybe `file.seek(0)`.

Thrandis:
@rizar I fixed the naming issues and the bug in `add_to_dump` using your proposed trick!

blocks/serialization.py
```diff
@@ -34,130 +142,54 @@
...
 
-class PersistentParameterID(PersistentSharedVariableID):
-    """Persist the names of parameter arrays in the zip file.
-
-    Only Theano shared variables are persisted to the zip file using this
-    method. Names are determined using the brick hierarchy, or the shared
-    variable name.
+def dump(object_, file_, parameters=None, use_cpickle=False,
+         protocol=DEFAULT_PROTOCOL, **kwargs):
+    r"""Pickles an object saving optionnaly its parameters separately.
```
rizar: optionally

blocks/serialization.py
```diff
@@ -187,6 +219,135 @@ def secure_dump(object_, path, dump_function=dump, **kwargs):
...
+        Name of the object to load. Default is `_pkl`, meaning that it is
+        the original object which have been dumped that is loaded.
+
+    Returns
+    -------
+    The object saved in file_.
+
+    """
+    with tarfile.open(fileobj=file_, mode='r') as tar_file:
+        p = pickle.Unpickler(
```
rizar: Can we allow choice of the unpickler? We also have `cPickle.Unpickler`.

blocks/serialization.py
```diff
@@ -187,6 +219,135 @@ def secure_dump(object_, path, dump_function=dump, **kwargs):
...
+        return {name.replace(BRICK_DELIMITER, '/'): value
+                for name, value in npz_file.items()}
+
+
+def add_to_dump(object_, file_, name, parameters=None, use_cpickle=False,
+                protocol=DEFAULT_PROTOCOL, **kwargs):
+    r"""Pickles an object to an existing tar archive.
+
+    This function allows to dump more objects to an existing archive. If
+    the object you want to dump posess the same set of shared variables as
```
rizar: posseses

blocks/serialization.py
```diff
@@ -34,130 +142,54 @@
...
+                array_ = p.container.storage[0]
+                external_objects[id(array_)] = _mangle_parameter_name(
+                    type(array_), n)
+        if parameters:
+            _taradd(_save_parameters, tar_file, '_parameters')
+        def _save_object(f):
+            p = pickler(f, protocol=protocol, **kwargs)
+            p.persistent_id = _PersistentID(external_objects)
+            p.dump(object_)
+        _taradd(_save_object, tar_file, '_pkl')
```
rizar: There is one very important thing that we forgot here. Pickling the whole object should be optional, because sometimes it is actually the most painful thing. Can you please add a flag for that?

blocks/serialization.py
```diff
@@ -187,6 +219,135 @@ def secure_dump(object_, path, dump_function=dump, **kwargs):
...
+                                     ' from the one in the archive.')
+
+    file_.seek(0)
+
+    if use_cpickle:
+        pickler = cPickle.Pickler
+    else:
+        pickler = _PicklerWithWarning
+    with closing(tarfile.TarFile(fileobj=file_, mode='a')) as tar_file:
+        def _save_object(f):
```
rizar: I have just realized that save object is defined twice in two different routines, which is a kind of weird. We can try to move it to the global namespace, though it would have to become a functor, i.e. `class _SaveObject`.

blocks/serialization.py
```diff
@@ -220,23 +381,143 @@ def continue_training(path):
...
+    collisions. It also provides names for variables without
+    names.
+
+    Attributes
     ----------
-    path : str or file
-        The source for loading from.
+    used_names : set
+        The set of names already taken.
+    default_name :
```
rizar: default_name : str ?

blocks/serialization.py
```diff
@@ -220,23 +381,143 @@ def continue_training(path):
...
+        The name to use if a parameter doesn't have a name. Default:
+        'PARAMETER'.
 
-    Returns
-    -------
-    A dictionary of (parameter name, numpy array) pairs.
+    """
+    def __init__(self):
+        self.used_names = set()
+        self.default_name = 'PARAMETER'
```
rizar: Names are usually lowercase, maybe 'parameter'?

blocks/serialization.py
```diff
@@ -220,23 +381,143 @@ def continue_training(path):
...
+                i += 1
+                new_name = '_'.join([name, str(i)])
+            name = new_name
+        self.used_names.add(name)
+        return name
+
+
+_ARRAY_TYPE_MAP = {numpy.ndarray: 'numpy_ndarray'}
+_INVERSE_ARRAY_TYPE_MAP = {'numpy_ndarray': numpy.array}
+if cuda_ndarray:
```
rizar: Just to make sure: did you test the code on GPU?

rizar:
Overall, it looks very good! I find a few small things to be done, see the comments in the code.

I think we can merge it next week!

blocks/serialization.py
Thrandis: Yes!

Thrandis:
@rizar I added the changes you wanted! Now I will write the tests to see if everything is working correctly!

Thrandis:
@rizar I added the first set of tests and corrected 2 small issues. Let me know what you think of it!
What else should I test?

rizar:
Your changes look good, I haven't starting reading the tests yet. What about rebasing the PR before writing the tests?

Thrandis:
@rizar Allright, so I need to adapt the existing interface for `Checkpoint` and the other functions for saving. I have however a design question:
Shall we add a parameter `pickle_separately` to the dump function? It would be a list of objects to save at the same time, that could share the parameters of the main object?
Otherwise, we can directly call the new function `add_to_dump`. The current main drawbacks of `add_to_dump` are:
 1. No secure version;
 2. We have to call it for every separate object we want to save.
We can solve point 2 by passing a list of object (instead of just one object) to `add_to_dump`. What do you think about it?

Thrandis:
Ahah damn you rebase

dwf:
I really like the "pickle_separately" idea, by the way. The save_separately
argument seems pretty unnecessarily restrictive. Passing in a dict of name
-> object pairs seems a lot nicer?

>

Thrandis:
@dwf that is what I did before, but the thing is, since it is an extension, you can't pass parts of the main loop to it (since the main loop object is not already created when you create the checkpoint extension). So I don't know how to suppress this problem.

rizar:
@dwf, do you have a usecase in mind, when want you want to pickle is not an attribute of `MainLoop`?

PR #878, opened 2015-10-21
Bump version number to 0.1

rizar:
A companion for https://github.com/mila-udem/fuel/pull/257.

Unlike Fuel, I would postpone refactoring the version handling (see https://github.com/mila-udem/fuel/pull/256) until after the release. The reason is that to do it in the same nice way as I did in Fuel we should first get rid of namespace package support, and I guess it is not a good idea to get such a big change right before the release.


dmitriy-serdyuk:
I agree. 

It also means, that `master` should be merged into `stable`.

PR #879, opened 2015-10-21
Added documentation about the emitter cost in sequence generators.

sotelo:
I added a comment on the AbstractEmitter class saying that the emitter cost has to be implemented such that it will handle inputs of different dimensions.

### DIFF
--- a/blocks/bricks/sequence_generators.py
+++ b/blocks/bricks/sequence_generators.py
@@ -582,6 +582,18 @@ class AbstractEmitter(Brick):
 
     :class:`SoftmaxEmitter` : for integer outputs
 
+    Notes
+    -----
+    An important detail about the emitter cost is that it will be evaluated
+    with inputs of different dimensions so it has to be flexible enough to
+    handle this. The two ways in which it can be applied are:
+
+        1. Inside the cost_matrix function of :class:`BaseSequenceGenerator`
+        where it will be applied to the whole sequence at once.
+        
+        2. Inside the generate function of :class:`BaseSequenceGenerator`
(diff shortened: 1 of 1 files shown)


blocks/bricks/sequence_generators.py
```diff
@@ -582,6 +582,19 @@ class AbstractEmitter(Brick):
...
     :class:`SoftmaxEmitter` : for integer outputs
 
+    Notes
+    -----
+    An important detail about the emitter cost is that it will be
+    evaluated with inputs of different dimensions so it has to be
+    flexible enough to handle this. The two ways in which it can be
+    applied are:
+
+        1. In the cost_matrix function of :class:`BaseSequenceGenerator`
```
dmitriy-serdyuk: Can you use `:meth:\`BaseSequenceGenerator\.cost_matrix`` since it will be rendered and create links to these methods.

dmitriy-serdyuk:
Thanks, it's an important note!

rizar:
Good job!

happygds:
Hi, I have a few questions when reading 'blocks/bricks/sequence_generators.py'. 
    Firstly, I don't understand why use the 'self.readout.initial_outputs(batch_size))' as feedback[0] after feedback was time delayed inside the cost_matrix function of :class:`BaseSequenceGenerator`. 
    Suppose we trained one LSTM layer, the inputs are the first five words and the targets are the following five words in the same sentence. Then my cost should be the cost between the predictive following five words and the targets. Does blocks.recurrent.LSTM() layer supports return the whole sequence instead of the last output?

PR #881, opened 2015-10-23
Correct main loop message for the interrupt

rizar:


dmitriy-serdyuk:
Yes, I will be clearer.

rizar:
I will merge it after we are done with the release.

PR #882, opened 2015-10-23
Release 0.1 version

rizar:


dmitriy-serdyuk:
It's strange, that we have conflicts.

Don't forget to create a new tag.

rizar:
The tag will be created automatically when I create a release using GitHub.

We indeed have conflicts, that's because `stable` branch contains some changes which are supposed to be its unique features: it links `stable` from Fuel and a particular Theano version. I will resolve the conflicts locally and create a new pull request.

PR #886, opened 2015-10-27
Reuse the final values of states to initialize states in a sequence generator

sotelo:
Added the final values of states as auxiliary variables in a sequence generator so they can be reused.


rizar:
Oops! Forgot about it!

rizar:
Thanks!

sotelo:
No prob. Thanks for merging!

PR #888, opened 2015-10-27
Add get_dim to Bidirectional

akhti:
Required to use in conjunction with RecurrentStack


rizar:
Thanks! Please let me know if one can actually use `Bidirectional` in `RecurrentStack`, because at the development time we did not really think about.

By the way, I guess was a bit too fast merging your PR because the new `get_dim` function is not tested. Would be nice if you make an additional PR with a test.

PR #892, opened 2015-10-28
Add test for a stack of bidirectional layers

akhti:
Covers fix from #888


rizar:
Good to know that actually works! Sorry for the delay.

PR #896, opened 2015-10-29
Add the original for SequenceGenerator picture

rizar:
I guess we should have the source .svg in the repository, shouldn't we?

@akhti , you can use this.


rizar:
I don't think this release need a peer review, so...

PR #897, opened 2015-11-01
Custom scaling for dropout

### LINKED ISSUE #807: drop_prob and rescaling should be set independently
Currently, ```apply_dropout``` rescales the replacement variables by ```(1-drop_prob)```, a more general case is to rescale by an arbitrary amount and use ```(1-drop_prob)``` as the default. I've found (in agreement with others' results) that setting them independently can lead to better results.

### LINKED ISSUE #848: apply_dropout docstring issues
* The optional parameters should be marked optional. Default values/behaviour should be explained.
* Missing Returns section.

dwf:
Fixes #807.
Fixes #848.

### DIFF
--- a/blocks/graph.py
+++ b/blocks/graph.py
@@ -560,7 +560,9 @@ def apply_dropout(computation_graph, variables, drop_prob, rng=None,
     -------
     dropped_computation_graph : instance of :class:`ComputationGraph`
         A new computation graph with dropout applied to the specified
-        variables.
+        variables. In order to train with, or monitor, the outputs
+        of the original computation graph with dropout applies, use
+        the variables contained in `dropped_computation_graph.outputs`.
 
     Notes
     -----
--- a/blocks/graph.py
+++ b/blocks/graph.py
@@ -533,7 +533,7 @@ def collect_parameters(computation_graph, parameters):
 
 
 def apply_dropout(computation_graph, variables, drop_prob, rng=None,
-    """Returns a graph to variables in a computational graph.
+                  seed=None, custom_divisor=None):
     """Apply dropout to specified variables in a graph.
 
     Parameters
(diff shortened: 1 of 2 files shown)


dwf:
@jesselivezey FYI

@dmitriy-serdyuk You wrote this code, mind reviewing it? (See #807, which asks for a feature that was previously available in pylearn2.)

blocks/graph.py
```diff
@@ -533,8 +533,8 @@ def collect_parameters(computation_graph, parameters):
 
 
 def apply_dropout(computation_graph, variables, drop_prob, rng=None,
-                  seed=None):
-    """Returns a graph to variables in a computational graph.
+                  seed=None, custom_divisor=None):
```
dmitriy-serdyuk: I would place `custom_divisor` argument before `rng` and `seed` so that's easier to pass it as a not named argument.

blocks/graph.py
```diff
@@ -550,6 +550,17 @@ def apply_dropout(computation_graph, variables, drop_prob, rng=None,
...
+        Divide dropped variables by a given scalar value. If `None`,
+        (default) dropped variables will be divided by `(1 - drop_prob)`
+        which is equivalent to scaling by `(1 - drop_prob)` at test
+        time as recommended in [DROPOUT]_.
+
+    Returns
+    -------
+    dropped_computation_graph : instance of :class:`ComputationGraph`
+        A new computation graph with dropout applied to the specified
+        variables.
```
dmitriy-serdyuk: Since you are modifing dropout, can you mention, that one needs to extract new costs from a modified graph. Seems, sometimes people do not understand it.

blocks/graph.py
dwf: I added to the end so as to not break existing user code/tests.

dmitriy-serdyuk: Ok, maybe it's fine this way, since people rarely set a custom devisor.

blocks/graph.py
dwf: @dmitriy-serdyuk c4426c3

dmitriy-serdyuk:
LGTM!

Issue #257, opened 2015-02-12
Convolutional bricks; enhancements


bartvm:
Merged #221! 

Remaining tasks that popped up/came to mind:

* Analogous to `MLP` it would be nice to have an easy interface to string a sequence of convolutional layers together.
* Document and possibly move/rename the `Flattener` brick in `conv.py`.
* Non-tied biases, 
* Mean-pooling
* Local response normalization

@fvisin, @harmdevries89, @dmitriy-serdyuk, @adbrebs and whoever else showed interest in convnets

fvisin:
The Local response normalization can be "imported" from: pylearn2.expr.normalize.CrossChannelNormalizationBC01

dwf:
I updated the text of this, and am about to tackle mean pooling in #899.

PR #899, opened 2015-11-02
Convolutional network pooling improvements

### LINKED ISSUE #894: Bare MaxPooling can't be used inside a ConvolutionalSequence
It would be really handy, since bare Convolutionals can.

Specifically what's holding it back is:

* A (settable) `image_size` attribute
* A (settable) `num_channels` attribute
* A (gettable) `num_filters` attribute (which is a misnomer for the purpose that it's used inside `ConvolutionalSequence, it really should be `num_output_channels` or something like that)

We can ditch the awkward `input_dim` constructor argument (or else hack these on as properties that actually mutate `input_dim` (which is what I've done in a test and it seems to work).

Changing to `num_output_channels` as the interface queried for eliminates the need for a semantically ill-justified `num_filters` attribute/property at the expense of adding this as a property to Convolutional, ConvolutionalActivation, and ConvolutionalLayer.

dwf:
This pull request accomplishes a few things:
* Most notably, an AveragePooling Brick (mentioned in #257), accomplished by refactoring MaxPooling into a base Pooling Brick and MaxPooling and AveragePooling descendants.
  * Tests for AveragePooling's two modes w.r.t. padding.
* More complete wrapping of Theano's `max_pool_2d` function (which, bizarrely, also handles average pooling)
  * Basic tests for these.
* Improvements to ConvolutionalSequence that let you use MaxPooling (and AveragePooling) without ConvolutionalLayer, and make the interface a bit more coherent (i.e. not directly querying `num_filters` for the number of output channels, which is not always the case e.g. with a convolutional maxout layer or something)
* **Change the default `ignore_border` mode to `False`**, while exposing it in an option, with a warning in the docs that modifying it (with certain step configurations) can force computation onto the CPU.

Fixes #894, and an item in #257 that should be crossed off there.

### DIFF
--- a/blocks/bricks/conv.py
+++ b/blocks/bricks/conv.py
@@ -167,7 +167,6 @@ class Convolutional(Initializable):
         return self.num_filters
 
 
-
 class Pooling(Initializable, Feedforward):
     """Base Brick for pooling operations.
 
@@ -278,6 +277,34 @@ class MaxPooling(Pooling):
                                          step=step, input_dim=input_dim,
                                          ignore_border=ignore_border,
                                          padding=padding, **kwargs)
+
+
+class AveragePooling(Pooling):
+    """Average pooling layer.
+
+    Parameters
+    ----------
+    include_padding : bool, optional
+        When calculating an average, include zeros that are the
+        result of zero padding added by the `padding` argument.
+        A value of `True` is only accepted if `ignore_border`
+        is also `True`. `False` by default.
+
+    Notes
+    -----
+    For documentation on the remainder of the arguments to this
+    class, see :class:`MaxPooling`.
+
+    """
+    @lazy(allocation=['pooling_size'])
+    def __init__(self, pooling_size, step=None, input_dim=None,
+                 ignore_border=False, padding=(0, 0),
+                 include_padding=False, **kwargs):
+        mode = 'average_inc_pad' if include_padding else 'average_exc_pad'
+        super(AveragePooling, self).__init__(mode, pooling_size,
+                                             step=step, input_dim=input_dim,
+                                             ignore_border=ignore_border,
+                                             padding=padding, **kwargs)
 
 
 class _AllocationMixin(object):
--- a/blocks/bricks/conv.py
+++ b/blocks/bricks/conv.py
@@ -163,30 +163,39 @@ class Convolutional(Initializable):
         return super(Convolutional, self).get_dim(name)
 
 
-class MaxPooling(Initializable, Feedforward):
-    """Max pooling layer.
+class Pooling(Initializable, Feedforward):
+    """Base Brick for pooling operations.
 
-    Parameters
-    ----------
-    pooling_size : tuple
-        The height and width of the pooling region i.e. this is the factor
-        by which your input's last two dimensions will be downscaled.
-    step : tuple, optional
-        The vertical and horizontal shift (stride) between pooling regions.
-        By default this is equal to `pooling_size`. Setting this to a lower
-        number results in overlapping pooling regions.
-    input_dim : tuple, optional
-        A tuple of integers representing the shape of the input. The last
-        two dimensions will be used to calculate the output dimension.
+    This should generally not be instantiated directly; see
+    :class:`MaxPooling`.
 
     """
-    @lazy(allocation=['pooling_size'])
-    def __init__(self, pooling_size, step=None, input_dim=None, **kwargs):
-        super(MaxPooling, self).__init__(**kwargs)
-
-        self.input_dim = input_dim
+    @lazy(allocation=['mode', 'pooling_size'])
+    def __init__(self, mode, pooling_size, step=None, input_dim=None,
+                 ignore_border=False, padding=(0, 0), **kwargs):
+        super(Pooling, self).__init__(**kwargs)
         self.pooling_size = pooling_size
+        self.mode = mode
         self.step = step
+        self.input_dim = input_dim if input_dim is not None else (None,) * 3
+        self.ignore_border = ignore_border
+        self.padding = padding
+
+    @property
+    def image_size(self):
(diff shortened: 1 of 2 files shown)


dwf:
@dmitriy-serdyuk or @rizar this is ready, I think.

No behaviour changes (aside from previously unavailable functionality), only about 65 lines of actual code... the rest docs and (short, easy to read) tests.

dwf:
After Theano/Theano#3591 and Theano/Theano#3596 I'm thinking it's probably best if we change the default in master to `ignore_border=True`, since that's the only way that it gets replaced with CuDNN ops (and, if using overlapping pools, the only way that it gets run on GPU at all).

This is a change we should announce to the mailing list, however the potential for slowdowns is so great and so frustrating (I've wasted almost an entire a day figuring this out, not including the extra computation time itself) that I think it's worth it.

We now give the user access to the padding argument, which they can use to achieve almost the same thing (in a slightly more coherent way: rather than having this weird region in the bottom right that might be a max or average over less elements than usual, you have regions around the entire border that all get partial coverage). 

The magical incantation for getting pooling to use CuDNN is currently documented precisely nowhere, so I also think we should add a big fat warning to our own docs.

Thoughts? (@dmitriy-serdyuk @fvisin @bartvm @vdumoulin)

bartvm:
Having things run on GPU instead of CPU when possible is good enough reason
IMHO to change the defaults. +1 on this needing documentation.

On Tue, Nov 3, 2015 at 8:39 PM, David Warde-Farley <notifications@github.com

dwf:
3b01a8d changes the default and adds the warning.

Aaaand Travis is failing for reasons unrelated to this PR.

fvisin:
I agree with you, the fastest option should be the default *and* the documentation should be very clear on the impact that modifying this option can have on the performances.

dmitriy-serdyuk:
I agree, that we should use the fastest version by default. And we never cared too much about backward compatibility.

blocks/bricks/conv.py
```diff
@@ -207,16 +220,105 @@ def apply(self, input_):
             with the last two dimensions downsampled.
 
         """
-        output = max_pool_2d(input_, self.pooling_size, st=self.step)
+        output = max_pool_2d(input_, self.pooling_size, st=self.step,
```
dmitriy-serdyuk: Should we make polling hackable like convolution? As I can see, cudnn also has an implementation of max pooling.

dmitriy-serdyuk: Ah, wait. Why do you implement `apply` for an abstract class?

dmitriy-serdyuk: It's very misleading that `max_poop_2d

dwf: On Wed, Nov 4, 2015 at 3:08 PM, dmitriy-serdyuk <notifications@github.com>
wrote:

> max_poop_2d
>

max poop indeed :) Theano/Theano#3592

dwf: I think it's less necessary to make it hackable since all options are currently available with the basic implementation, which is not the case for convolution (but will be when we move to the abstract_conv2d stuff). However it is important to have good defaults and a big fat warning about `ignore_border`.

dwf: (It seems that the eventual solution in Theano will be replacing `max_pool_2d` with something named appropriately)

dwf: As for why I went the route of "base class" rather than "abstract base class" with an abstract apply method... I think the implementation makes it clear. Basically MaxPooling and AveragePooling just need different constructors that call different the superclass constructor with different arguments, and everything else remains the same.

dmitriy-serdyuk: Yes, I see now, it makes sense.

tests/bricks/test_conv.py
```diff
@@ -117,13 +118,82 @@ def test_max_pooling():
...
+    x = tensor.tensor4('x')
+    brick = AveragePooling((2, 2), ignore_border=True, padding=(1, 1),
+                           include_padding=False)
+    y = brick.apply(x)
+    x_ = 3 * numpy.ones((1, 1, 2, 2), dtype=theano.config.floatX)
+    output = y.eval({x: x_})
+    assert_allclose(x_, output)
+
+
+def test_pooling_works_in_convolutional_sequence():
```
dmitriy-serdyuk: Maybe we should add pretty docstrings to tests, and nosetests will print `Test pooling works...`.

dmitriy-serdyuk:
Have you tried to run LeNet example or your convnet with these changes?

dwf:
Just did, and I don't think anything has changed. The pools are size (2, 2) and the stride was the same, so I think as long as the input sizes were even, nothing would have changed. `convnet.top_mlp.input_dim` stays 5000, so I think that's probably an indicator that it's the same throughout, and the same layer dimensions are reported in the log messages.

Almost the same test error and test cost are achieved after two (a bit off in the 3rd decimal place), but that's probably a result of the graph switching max pooling implementations and doing rounding differently or something. Nothing to worry about, I don't think.

tests/bricks/test_conv.py
dwf: CCW ticket? :)

dmitriy-serdyuk:
Ok, LGTM.

dwf:
@dmitriy-serdyuk I added one more commit to be less annoying to people with previously pickled models, 6177388. Can you give it a quick look? (5 lines + a test).

dwf:
(and, relatedly, 769eb70)

dmitriy-serdyuk:
Good idea, but we sould remove  https://github.com/mila-udem/blocks/commit/6177388bcdec4b858d8ca051ef0bb40d23f91c83 after a while.  I'll create an issue. 

I think, this PR is ready, let's merge.

dwf:
Great, thanks!

PR #903, opened 2015-11-04
Fix Travis

dwf:
The Miniconda installer is, for some reason, now defaulting to $HOME/miniconda2 instead of $HOME/miniconda. Hardcode a prefix to be robust to this.


dwf:
I can't imagine anyone having an objection to a one-liner that fixes the CI bot, so I'm going to go ahead and merge.

PR #907, opened 2015-11-05
Add API sections for two log backends

dmitriy-serdyuk:
I've heard many complaints that log is not documented, here is the fix.


rizar:
Tried on my machine, works!

PR #921, opened 2015-11-24
Get rid of ConvolutionalLayer.

### LINKED ISSUE #912: Do we still need ConvolutionalLayer?
After #899, now that you can stick pooling bricks inside a `ConvolutionalSequence`, it makes `ConvolutionalLayer` a (more restricted) way of doing the same thing that can be accomplished a different way; with `ConvolutionalLayer`, you're currently restricted to MaxPooling, and can't use `padding` or `ignore_border`.

The three options: 
 1. Maintain the status quo (somewhat unsatisfying).
 2. Add these features to `ConvolutionalLayer`.
 3. Remove `ConvolutionalLayer` altogether.

Personally I'm in favour of option 3, because the less classes the better, and both hardcoding the possible choices of pooling class and passing a class name as an argument both leave bad tastes in my mouth.

Obligatory PEP20 reference:
```
There should be one-- and preferably only one --obvious way to do it.
```

dwf:
For all the reasons mentioned in #912. Fixes #912.

Requires a sister pull request for blocks-examples.


bartvm:
Reduces entropy significantly, so I'm all for it.

bartvm:
+1 for the branch name

dwf:
Restarted Travis.

Note: blocks-examples is expected to fail until mila-udem/blocks-examples#64 is merged.

dwf:
Failure is unrelated, see #924.

dwf:
@rizar If you're looking for something to really LGTM, this (and https://github.com/mila-udem/blocks-examples/pull/64 which is the reason Travis fails) would be helpful. :smile:

tests/bricks/test_conv.py
```diff
@@ -259,14 +229,14 @@ def test_convolutional_sequence():
     batch_size = 5
     activation = Rectifier().apply
 
-    conv = ConvolutionalLayer(activation, (3, 3), 5,
-                              (pooling_size, pooling_size),
-                              weights_init=Constant(1.),
-                              biases_init=Constant(5.))
+    conv = ConvolutionalActivation(activation, (3, 3), 5,
```
rizar: I do not understand why an activation needs `weights_init` and `biases_init`. Moreover, I do not understand why it accepts these arguments and does not crash.

dwf: A `ConvolutionalActivation` has always been a `Convolutional` and and `Activation` sandwiched together; therefore the `Convolutional` has parameters that need to be set. This was not my doing or my preference. (In fact, I'd be in favour of just allowing `Activation`s to be put inside of `ConvolutionalSequence`s, since the input-to-output shape inference is trivial, and removing this class too).

dwf: This is mentioned in the [docs](https://github.com/mila-udem/blocks/blob/master/blocks/bricks/conv.py#L347-L348), by the way. I'm not crazy about the name (or this class at all), but it is what it is.

dwf: Incidentally, it turns out it's completely trivial to do what I said, to allow actual elementwise `Activations` inside a `ConvolutionalSequence`. So this could be a way of reducing entropy further and eliminating `ConvolutionalActivation` as well... See https://github.com/dwf/blocks/commit/3522f51c2491d5c56cc7345443e87aa24ae959b6

But for now, let's focus on this PR.

rizar: OK, thanks for the explanation. It was not immediately clear that `ConvolutionalActivation` is in fact `Convolution`  + `Activation`.

rizar:
Good job!

PR #924, opened 2015-11-25
Fix debugprint output for Theano/Theano#1953.

dwf:


dwf:
This is a purely mechanical (and non-optional, given the contents of the tutorial) change, so I hope nobody will mind if I just merge it.

rizar:
A late LGTM for you :)

Issue #856, opened 2015-10-05
Update Convolutional to use new conv2d interface when available


dwf:
Currently in progress in Theano/Theano#2665.

I'm most interested in the explicit padding specification currently available in `dnn_conv`.

dwf:
This should now be addressable.

dwf:
Once we are fully over to the new conv interface we can maybe remove the "hackability" of the Convolutional class (the static methods) as it really won't be necessary anymore.

PR #925, opened 2015-11-25
Replace deprecated Theano getOutputShape interface.

dwf:
Related to #856, but does not quite address it. Just addresses DeprecationWarnings in the tests.

### DIFF
--- a/blocks/bricks/conv.py
+++ b/blocks/bricks/conv.py
@@ -166,8 +166,10 @@ class Convolutional(Initializable):
             image_shape = (None, self.num_channels) + self.image_size
             kernel_shape = ((self.num_filters, self.num_channels) +
                             self.filter_size)
-            return self.get_output_shape(image_shape, kernel_shape,
-                                         self.border_mode, self.step)[-3:]
+            out_shape = self.get_output_shape(image_shape, kernel_shape,
+                                              self.border_mode, self.step)
+            assert len(out_shape) == 4
+            return out_shape[1:]
         return super(Convolutional, self).get_dim(name)
 
     @property
--- a/blocks/bricks/conv.py
+++ b/blocks/bricks/conv.py
@@ -1,4 +1,4 @@
-from theano.tensor.nnet.conv import conv2d, ConvOp
+from theano.tensor.nnet.conv import conv2d, get_conv_output_shape
 from theano.tensor.signal.downsample import max_pool_2d, DownsampleFactorMax
 
 from blocks.bricks import Initializable, Feedforward, Sequence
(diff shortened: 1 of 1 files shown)


dwf:
This is ready for merge, if someone would like to review it.

blocks/bricks/conv.py
```diff
@@ -55,9 +55,12 @@ class Convolutional(Initializable):
...
 
     # Used to override the output shape computation for a given value of
-    # conv2d_impl. Should accept 4 positional arguments: the image size,
-    # the filter size, the step (strides), and the border mode.
-    get_output_shape = staticmethod(ConvOp.getOutputShape)
+    # conv2d_impl. Should accept 4 positional arguments: the shape of an
+    # image minibatch (with 4 elements: batch size, number of channels,
+    # height, and width), the shape of the filter bank (number of filters,
+    # number of output channels, filter height, filter width), the border
+    # mode, and the step (vertical and horizontal strides).
```
dmitriy-serdyuk: What should it return?

blocks/bricks/conv.py
```diff
@@ -157,9 +160,11 @@ def get_dim(self, name):
...
             return (self.num_channels,) + self.image_size
         if name == 'output':
-            return ((self.num_filters,) +
-                    self.get_output_shape(self.image_size, self.filter_size,
-                                          self.step, self.border_mode))
+            image_shape = (None, self.num_channels) + self.image_size
+            kernel_shape = ((self.num_filters, self.num_channels) +
+                            self.filter_size)
+            return self.get_output_shape(image_shape, kernel_shape,
+                                         self.border_mode, self.step)[-3:]
```
dmitriy-serdyuk: Why do you need this `-3`? It can conceal errors, I would just let it crash.

dwf: Because the new thing returns a 4-tuple, which can be None. Our interfaces revolve around 3-tuples.

blocks/bricks/conv.py
dwf: Good point, it should return a 4-tuple.

blocks/bricks/conv.py
dwf: @dmitriy-serdyuk 928dcbb

blocks/bricks/conv.py
dwf: @dmitriy-serdyuk  928dcbb

blocks/bricks/conv.py
dwf: @dmitriy-serdyuk Not sure if I addressed this to your satisfaction, but my attempt is in ff70906.

dmitriy-serdyuk:
Forgot about this one, LGTM.

PR #928, opened 2015-12-01
Pass shared_floatx kwargs to theano.shared.

dwf:
Not sure why this wasn't done. This lets you e.g. set broadcastable flags on the created shared variable.

### DIFF
--- a/blocks/utils/__init__.py
+++ b/blocks/utils/__init__.py
@@ -138,8 +138,8 @@ def shared_floatx(value, name=None, borrow=False, dtype=None, **kwargs):
                          name=name, borrow=borrow, **kwargs)
 
 
-def shared_like(variable, name=None):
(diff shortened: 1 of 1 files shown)


blocks/utils/__init__.py
```diff
@@ -133,8 +135,7 @@ def shared_floatx(value, name=None, borrow=False, dtype=None):
     if dtype is None:
         dtype = theano.config.floatX
     return theano.shared(theano._asarray(value, dtype=dtype),
-                         name=name,
-                         borrow=borrow)
+                         name=name, borrow=borrow, **kwargs)
 
 
 def shared_like(variable, name=None):
```
dmitriy-serdyuk: What about this one? Let's make similar interface for all `shared_*` functions.

dwf: @dmitriy-serdyuk b7f5df3

PR #940, opened 2016-01-13
Disable namespace packages suppport

rizar:
@bartvm , it's your call


rizar:
The tests fail because of SQLite backend, it's is also the case for master. @bartvm , can I merge ?

dwf:
Don't forget to announce on blocks-users

rizar:
Sure, in progress

Issue #885, opened 2015-10-26
Flatten namespaces


dwf:
I propose that we make the public API available no more than 2 layers deep.

I spend a lot of time trying to remember whether things are in `blocks.bricks` or `blocks.bricks.base`, and so on.

I tend to agree, in general,  with the sentiment about imports expressed by Raymond Hettinger in [this talk (link to 24:45 where the import/namespace comments start)](https://www.youtube.com/watch?v=wf-BqAjZb8M#t=24m45s). We don't have a good use case for multi-level namespaces. `blocks.bricks, blocks.initialization` is plenty of quarantining.

dwf:
There are added benefits for refactoring: if we make it such that user imports are always done from the `blocks.bricks`, `blocks.initialization` etc. levels, we are able to move stuff around more easily. User imports don't change, regardless of where the code is defined.

rizar:
+100!!!

We should move all the code from __init__ files and only have user-oriented
imports there.

PR #941, opened 2016-01-16
Brick-based batch normalization.

### LINKED ISSUE #920: Brick-based batch normalization solution
#513 and #851 have now been in progress for 8 months. I think this is a testament to the fact that the current approach is proving hard to get right.

Below is my plan to accomplish this by explicitly introducing batch normalization bricks into the graph at construction.  I think this somewhat simpler, and this solves the issue of how BN population parameters get serialized. Indeed, the abstract of Ioffe & Szegedy (2015) refers to their method as making normalization "part of the model". It seems only appropriate that this be reflected in the graph that is constructed.

## New brick

`BatchNormalization` will house the learned shift and scale parameters (betas and gammas), as well as population mean and standard deviation statistics shared variables. The latter can't be tagged as `PARAMETER` because code that takes all `PARAMETER`s in a graph and blindly adapts them according to a cost will do entirely the wrong thing. I thus suggest a superclass role `ADAPTABLE` (I also like `LEARNED` or `LEARNED_QUANTITY`), with the tree looking like this.

* `ADAPTABLE`
  * `PARAMETER`
  * `BATCH_NORMALIZATION_PARAMETER`
     * `BATCH_NORMALIZATION_POPULATION_MEAN`
     * `BATCH_NORMALIZATION_POPULATION_STDEV`

Serialization code that looks for `PARAMETER`s to serialize should instead look for `ADAPTABLE`s.

BatchNormalization will have an apply method that uses the [Theano `batch_normalization` function](https://github.com/Theano/Theano/blob/master/theano/tensor/nnet/bn.py). **By default, the graph generated will use the population mean and standard deviations in BatchNormalization's apply method**, for two reasons:

* It's kind of icky to produce a graph which, by default upon construction, has the output for each example depending on the entire minibatch it arrived in.
* If the population means and variances are initialized to zeros and ones respectively, this makes it a no-op relative to what the graph would do without these additional bricks.

## Modifications to existing Bricks

As batch normalization becomes more an more ubiquitous, we want to make it easy to add it to models anywhere an affine transformation takes place. The best way to accomplish this is something I'm not *entirely* sure about but I think I have a pretty good idea.

I think the cleanest way is to introduce `BatchNormalizedLinear` and `BatchNormalizedConvolutional` classes. These would be drop-in replacements that include batch normalization as a child brick. In order to use these within an `MLP` (or a `Convolutional*`), these could grow keyword arguments not unlike the "prototype" keyword in the parallel bricks; you could either pass a lazily-allocated `BatchNormalized*` brick or simply the class object itself, e.g. `ConvolutionalActivation(..., convolution_prototype=BatchNormalizedConvolutional())` vs. `ConvolutionalActivation(..., convolutional_class=BatchNormalizedConvolutional)`.

Note that you could use this scheme with `Parallel` bricks by passing `BatchNormalizedLinear(use_bias=False)` as the prototype.

## Support functions

To actually train with batch normalization, two support functions are provided:

* `batch_normalization_training_graph(graph)` replaces `BATCH_NORMALIZATION_POPULATION_MEAN`s with the minibatch-wise means from the application method of the corresponding bricks, ditto for standard deviations.
 
* `batch_normalization_population_updates(graph, alpha)` returns an `OrderedDict` of updates which perform moving average updates on the population statistics for all the `BATCH_NORMALIZATION_PARAMETERS` in the graph, with lag parameter alpha (i.e. `(1 - alpha) * old_value + alpha * value_estimated_from_minibatch`). You pass this to `GradientDescent.add_updates` and you're on your way. Users are free not to use this, or to do something fancier if they please, but this is the easy thing that Christian Szegedy told me works just fine. We can also provide a lower-level function that simply returns pairs of `(population, minibatch_estimate)` so that users can apply a LearningRule of their choice or do whatever manner of sophisticated nonsense they like, but we should have one obvious way to adapt these things during training.

## Conclusion

I think that all of these things are doable with the framework we have *today*, and don't depend on the resolution of fundamental design issues like like #514. I'd appreciate folks' thoughts.

dwf:
Tries to distill the best ideas from @Thrandis, @cooijmanstim, and my own proposal into something that fits into Blocks.

To avoid circular imports, the bricks and graph packages had to be split up a bit. I think this is long overdue anyway, and will aid in #885.

* [ ] ~~~Utility code for inference network updates: provide at least a simple function to calculate moving averages, and maybe some analog to César's `BatchNormExtension`.~~~ save for later!
* [x] Examples of usage, at least via docstrings. Narrative docs can be a CCW ticket.
* [x] Some code for easily accessing the population <-> minibatch relationships when you create the graph using the context manager.

Fixes #920.


dwf:
@rizar @dmitriy-serdyuk If this interests either of you, I think it's ready for an initial look. Mainly needs finishing touches now.

dwf:
Tests occasionally fail with sqlite backend (#942) but it's unrelated to this PR.

aam-at:
@dwf I forked your implementation before and had one problem: graph replacements in blocks using theano.clone will create multiple copies of the same path.
For multiple bn layers, you get a combinatorial explosion of paths in the graph. During function compilation, they will be optimized away. However, it drastically increases compilation time.

```python
import time

from blocks.bricks import Rectifier, BatchNormalizedMLP, Softmax
from blocks.graph.bn import batch_normalize
from theano import tensor, function

x = tensor.matrix('features')
y = BatchNormalizedMLP(activations=[Rectifier(), Rectifier(), Rectifier(),
                                    Rectifier(), Rectifier(), Softmax()],
                       dims=[784, 1000, 1000, 1000, 1000, 1000, 10]).apply(x)
cg = ComputationGraph(y)
new_cg, _ = batch_normalize(cg)
y_hat = new_cg.outputs[0]

start = time.time()
function(inputs=[x], outputs=y)
end = time.time()
print 'inference graph:', end - start
start = time.time()
function(inputs=[x], outputs=y_hat)
end = time.time()
print 'training graph:', end - start
```
Compilation timings for 6 layer MLP:
```
inference graph: 0.382107973099
training graph: 7.72867417336
```
For 8 layer MLP:
```
inference graph: 0.478291988373
training graph: 370.310412884
```

As a workaround, I used to create a copy of graph and worked directly with ApplyNode to update graph. Though, as I understand, it's not in the spirit of blocks.

I am also attaching simple pdf to show difference between graph before and after batch_normalize for three layer MLP.

[y.pdf](https://github.com/mila-udem/blocks/files/94323/y.pdf)
[y_hat.pdf](https://github.com/mila-udem/blocks/files/94324/y_hat.pdf)

cooijmanstim:
+1 for refactoring the `__init__.py`s.

Re combinatorial cloning, I believe a cheap way to merge clones before compilation would be to put a special identity attribute on each node at construction time (i.e. `self.tag.original_id = id(self)`) and use that to identify clones. This will be faster than traversing the graph like `equal_computations`. I guess this is outside the scope of blocks though.

I've been working on an alternative graph substitution feature for Theano with which I've successfully been able to avoid combinatorial explosions due to cloning, but it's not ready for general use.

Alternatively it may be possible to do one big clone/replacement for all BN bricks rather than many small ones.

dwf:
I would think that cg.replace would do that if it were possible, no?

Regardless I think the implementation details of cg.replace can improve
underneath this implementation once we figure out the best way to go about
it. It may involve complementary features in Theano and Blocks.

dwf:
@aam-at I've made the implementation a bit more efficient (only does one replace per brick, rather than two) and also added another facility for creating the training graph via a context manager.

aam-at:
@dwf After your update both ```apply_batch_normalization``` and ```batch_normalization``` produces identical graphs and does not suffer from the compilation problem. Nice!
And I like your context manager idea. Essentially we can pass additional ```kwargs``` to apply like in Lasagne but without the problem who gets what.

janchorowski:
I once had another idea for a context magager: use the tags of input theano variables (to get the context you would walk the graph from one of the function inputs, with caching results in the tags of inputs this is linear with graph size). The advantage is that you don't have a private field in a brick that you toggle to change its operation.

pseudocode:

   def find_context(*vars):
      ctx = {}
      for var in vars:
          if hasattr(var.tag, 'context'):
               ctxv = var.tag.context
               # maybe do something smart with name clashes in the context?
               ctx.update(ctxv)
          else:

               for v1 in var.owners.inputs:
                     ctxv1 = find_context(v1)
   def apply(in1, in2):
       ctx = find_context(in1, in2)
       if ctx['train']:
            do this
       else
            do that

aam-at:
@janchorowski Having a top-level apply call to pass context is the same as Lasagne ```kwargs``` in ```get_output```. And the problems with this approach are the same: mistyping of strings, unused arguments.
The advantage of a private field in brick is that you get autocompletion for setting and getting ```training_mode```. So code is not prone to mistyping strings, e.g. ```'trian'```. One more advantage, you can change context only for some bricks in the hierarchy.

janchorowski:
@aam-at Thanks for pointing out the issues.

rizar:
@dwf, I gave it the first quick look and it looks nice! Do I understand it right that you let the user to choose whether they want to use replacements or context manager to obtain batch normalized graph for training?

dwf:
@rizar Yes, that's the idea. It sounds like doing outputs replacement rather than replacement of quantities used in the internally created graph has solved the problem @aam-at was seeing with graph size explosions, but it might be worth keeping it just in case, I don't know. It'd be worth comparing compile times with, say, a 100 layer MLP.

One thing I'm having trouble with: something you'll want to do in either case is get a list of (population shared variable, corresponding minibatch estimate) pairs, with which you do *something* -- either a moving average, or something like what @Thradis does in his extension (though this is prohibitively expensive for large datasets). But the crucial step is getting those pairs. I return a second value from `apply_batch_normalization` currently, but there's no easy facility for this if you use the 

Writing such a function is not hard, but the question is *what should it be called*? Also, if we are tending towards #885 and asking that namespaces be flattened, I'd like to give it a descriptive name that isn't prohibitively long. `gather_batch_normalization_statistics` or something is getting pretty long.

One thought is to make it a static method of `BatchNormalization`, like `BatchNormalization.gather_statistics`. Static methods aren't very common but I think.

Another problem is what to do when a given `BatchNormalization` brick appears twice in a graph, and you may have two different minibatch estimates for the mean/standard deviation. I sort of consider this the user's problem, but I think I should probably have an `allow_duplicates` keyword that is `False` by default (with the `False` behaviour being to raise an exception).

dwf:
So after `sys.setrecursionlimit(100000)` to take care of the interpreter's silly stack issues, a 100 layer batch-normalized net (with graph replacement) takes 18 seconds to compile, so I think it's officially Not A Big Deal anymore.

dwf:
Buildbot currently failing due to Theano/Theano#3894. Sigh...

blocks/bricks/simple.py
```diff
@@ -0,0 +1,395 @@
+"""Some of the simplest individual bricks."""
```
rizar: BTW, I would rename `simple.py` into `basic.py`.

rizar:
Given that the problem seems to be solved, do we still need the context manager?

dwf:
The context manager gives you a bit more flexibility, i.e. you can choose to only activate it for portions of your graph depending on the bricks you pass.

More than that, I know I'll be using `find_bricks` for other things (it scratches an itch I often have, and it might even be useful in speeding up the Selector as it does a breadth first search without use of the stack), and the additional maintenance burden on top of that to provide the context manager is 7 lines, plus docs and tests.

vdumoulin:
@dwf by the way, the import issue in Theano has been resolved.

rizar:
@dwf, can you please add an example to Blocks-examples, as we discussed?

@cooijmanstim ,  can you please review this PR?

blocks/bricks/bn.py
```diff
@@ -0,0 +1,345 @@
...
+from ..utils import (shared_floatx_zeros, shared_floatx,
+                     shared_floatx_nans)
+from .base import lazy, application
+from .sequences import Sequence, Feedforward, MLP
+from .interfaces import RNGMixin
+
+
+def _add_batch_axis(var, name=None):
+    """Prepend a singleton axis to a TensorVariable."""
+    new_var = var.dimshuffle('x', *list(range(var.ndim)))
```
cooijmanstim: I think you can just do `new_var = tensor.shape_padleft(var)` here.

blocks/bricks/bn.py
```diff
@@ -0,0 +1,345 @@
...
+
+    """
+    @lazy(allocation=['input_dim'])
+    def __init__(self, input_dim, broadcastable=None,
+                 save_memory=True, epsilon=1e-4, weights_init=None,
+                 biases_init=None, **kwargs):
+        self.input_dim = input_dim
+        self.broadcastable = broadcastable
+        self.save_memory = save_memory
+        self.epsilon = 1e-4
```
cooijmanstim: Should be `self.epsilon = epsilon`.

blocks/bricks/bn.py
```diff
@@ -0,0 +1,345 @@
...
+    def __enter__(self):
+        self._training_mode = True
+
+    def __exit__(self, *exc_info):
+        self._training_mode = False
+
+    def _compute_training_statistics(self, input_):
+        axes = (0,) + tuple((i + 1) for i, b in
+                            enumerate(self.population_mean.broadcastable)
+                            if b)
```
cooijmanstim: It's correct as is but I would prefer you use `self.broadcastable` to determine the axes.

dwf: That actually won't work, as self.broadcastable can be `None` as per the current `_allocate`.

blocks/bricks/bn.py
```diff
@@ -0,0 +1,345 @@
...
+    def _prepare_population_statistics(self):
+        mean = _add_batch_axis(self.population_mean, 'population_offset')
+        stdev = _add_batch_axis(self.population_stdev, 'population_divisor')
+        return mean, stdev
+
+    def _allocate(self):
+        input_dim = ((self.input_dim,)
+                     if not isinstance(self.input_dim, collections.Sequence)
+                     else self.input_dim)
+        broadcastable = (tuple(False for _ in range(len(input_dim)))
```
cooijmanstim: `for _ in range(len(x))` == `for _ in x`

blocks/bricks/bn.py
```diff
@@ -0,0 +1,345 @@
...
+        add_role(self.population_mean, BATCH_NORM_POPULATION_MEAN)
+        add_role(self.population_stdev, BATCH_NORM_POPULATION_STDEV)
+
+    @property
+    def W(self):
+        return self._W
+
+    @property
+    def b(self):
+        return self._b
```
cooijmanstim: It looks like these properties are unnecessary.

blocks/bricks/bn.py
```diff
@@ -0,0 +1,345 @@
...
+        self.biases_init.initialize(self.b, self.rng)
+        self.weights_init.initialize(self.W, self.rng)
+
+    # Needed for the Feedforward interface.
+    @property
+    def output_dim(self):
+        return self.input_dim
+
+    # The following properties allow for BatchNormalization bricks
+    # to be used directly inside of a ConvolutionalSequence.
```
cooijmanstim: These look like they belong in `SpatialBatchNormalization`.

blocks/bricks/bn.py
```diff
@@ -0,0 +1,345 @@
...
+    input_dim : int or tuple
+        Shape of a single input example. It is assumed that a batch axis
+        will be prepended to this.
+    broadcastable : tuple, optional
+        Tuple the same length as `input_dim` which specifies which of the
+        per-example axes should be averaged over to compute means and
+        standard deviations. For example, in order to normalize over all
+        spatial locations in a `(batch_index, channels, height, width)`
+        image, pass `(False, True, True)`.
+    save_memory : bool, optional
```
cooijmanstim: I don't like the use of "save" here as it could be taken to mean something is being saved to somewhere. Not sure I have a better suggestion though. "Conserve" maybe?

blocks/bricks/bn.py
dwf: Actually, you might want a normal `BatchNormalization` in a `ConvolutionalSequence`, specifically this is done at the very end of the chain in the BN paper where you *don't* want to normalize across space.

blocks/bricks/bn.py
```diff
@@ -0,0 +1,345 @@
...
+    See :class:`BatchNormalization` for more details (and additional
+    keyword arguments).
+
+    """
+    @lazy(allocation=['input_dim'])
+    def __init__(self, input_dim, **kwargs):
+        if not isinstance(input_dim,
+                          collections.Sequence) or len(input_dim) < 2:
+            raise ValueError('expected input_dim to be length >= 2 '
+                             '(channels, height, width)')
```
cooijmanstim: The condition is for at least 2 dimensions whereas the error lists 3. Requiring at least 2 dimensions seems right to me as it allows for 1D convolutions if we ever get those. Maybe put an "e.g." in front of the list of dimensions in the error message.

blocks/bricks/bn.py
```diff
@@ -0,0 +1,345 @@
...
+
+    @property
+    def save_memory(self):
+        return self._save_memory
+
+    @save_memory.setter
+    def save_memory(self, value):
+        self._save_memory = value
+        for act in self.activations:
+            assert isinstance(act.children[0], BatchNormalization)
```
cooijmanstim: This assertion would be nice to have in `_push_allocation_config` below as well.

blocks/graph/bn.py
```diff
@@ -0,0 +1,163 @@
...
+    """
+    # Avoid circular imports.
+    from blocks.bricks import BatchNormalization
+
+    bn = find_bricks(bricks, lambda b: isinstance(b, BatchNormalization))
+    # Can't use either nested() (deprecated) nor ExitStack (not available
+    # on Python 2.7). Well, that sucks.
+    for brick in bn:
+        brick.__enter__()
+    yield
```
cooijmanstim: Maybe wrap this in a try/finally to be sure `__exit__` is called. Also the loop below should technically go over the bricks in reverse.

blocks/bricks/bn.py
dwf: I don't properly remember why I added these. Possibly to shadow `Linear`.

blocks/bricks/bn.py
dwf: 6ba3364

blocks/graph/bn.py
```diff
@@ -0,0 +1,163 @@
...
+                               BatchNormalization.apply), inputs.keys())
+    for app_call in remove:
+        for mapping in (inputs, outputs, means, stdevs):
+            del mapping[app_call]
+
+    replacements = []
+    update_pairs = []
+    for app_call in inputs:
+        old_output = outputs[app_call]
+        unpacked = inputs[app_call].owner.inputs[0]
```
cooijmanstim: Why is it necessary to "unpack" this, and is it robust enough?

blocks/bricks/bn.py
dwf: 57e3b0b

blocks/bricks/bn.py
dwf: Good point. 5065fda

blocks/graph/bn.py
```diff
@@ -0,0 +1,163 @@
...
+        with app_call.application.brick:
+            new_output = app_call.application.brick.apply(unpacked)
+        replacements.append((old_output, new_output))
+        new_app_call = get_application_call(new_output)
+        update_pairs.append((app_call.application.brick.population_mean,
+                             new_app_call.metadata['offset']))
+        update_pairs.append((app_call.application.brick.population_stdev,
+                             new_app_call.metadata['divisor']))
```
cooijmanstim: In the BatchNormalization constructor, these metadata are set to the population statistics, and so the updates don't change anything. Am I missing something?

blocks/bricks/bn.py
dwf: f652056

blocks/graph/bn.py
dwf: 76909ab

blocks/graph/bn.py
dwf: Oof. Yeah, that's a bug.

dwf: Actually, no it isn't: in the current version of apply, mean or stdev are set to either the population statistics or the minibatch statistics, and then those are stored in the metadata dict under 'offset' and 'divisor'. We're extracting that metadata from `new_app_call.metadata`, where `new_app_call = get_application_call(new_output)` and `new_output` was generated in training mode (holding the bric context manager). So what's in there is indeed training minibatch statistics.

blocks/bricks/bn.py
cooijmanstim: You could use `SpatialBatchNormalization` with `broadcastable=(False,)*len(input_dim)` for that case though, right?

blocks/graph/bn.py
dwf: The input variable is a clone of whatever was passed into apply, apply creates a copy and tags it "INPUT". I just remove that copy, which also gets rid of the variable that's tagged with the original ApplicationCall, so that it won't show up in VariableFilter results. 

I've added a comment and an assertion in 6885bf2.

blocks/graph/bn.py
cooijmanstim: Yup, I see.

blocks/bricks/bn.py
dwf: Sure, but my thinking for SpatialBatchNormalization was "a shortcut for doing Batch Normalization *across space*", whereas BatchNormalization normalizes per feature on every axis, and hey, if you want to shove that in a ConvolutionalSequence, that's cool too. (I'd kind of eventually like to see that logic be factored out or else obviated by some sort of change like [this](https://github.com/dwf/blocks/commit/3522f51c2491d5c56cc7345443e87aa24ae959b6), probably not relying on Activation isinstance though).

```python
x = ConvolutionalSequence([..., SpatialBatchNormalization(), ..., SpatialBatchNormalization(), ..., BatchNormalization()])
```
versus
```
x = ConvolutionalSequence([..., SpatialBatchNormalization(), ..., SpatialBatchNormalization(), ..., SpatialBatchNormalization(broadcastable=(False,) * 3)])
```
I think the former wins the cognitive load.

cooijmanstim: Agreed.

blocks/bricks/simple.py
dwf: It's a little close to `base`, don't you think?

blocks/bricks/bn.py
```diff
@@ -0,0 +1,349 @@
...
+                                            mode=('low_mem'
+                                                  if self.conserve_memory
+                                                  else 'high_mem'))
+        return normalized
+
+    def __enter__(self):
+        self._training_mode = True
+
+    def __exit__(self, *exc_info):
+        self._training_mode = False
```
cooijmanstim: There's a subtle bug here due to nesting:

```python
brick = BatchNormalization(...)
with brick:
    with brick:
        pass
    assert(brick._training_mode)
```

The assertion will fail because the exit from the nested `with` will turn off training mode. I don't know if this pattern is worth worrying about. The logic in the context manager in graph/bn.py is okay in this respect as it enters all of the bricks at once and then exits all of them at once.

dwf: Good catch, it's easy to fix so I did in bb92fe0.

dwf:
@cooijmanstim As I said above, I'm thinking of factoring the second output argument of `apply_batch_normalization` into a standalone function so that you can use it if you're using context managers. The million dollar question: what do you think this should be called? Keep in mind it'll be plopped into the `blocks.graph` namespace (alternative: `BatchNormalization` static method).

cooijmanstim:
Hmm.. So the example you gave, `gather_batch_normalization_statistics` is long but pretty minimal; if it'll be in `blocks.graph` it's going to have the shape `<verb>_batch_normalization_<noun>s`. I'd suggest to use "updates" instead of "statistics" though as that's what they are, and `batch_normalization_updates` should be unambiguous enough as to what these updates are.

I feel like `BatchNormalization.<verb>_<noun>s` would be a useful name for an *instance* method that gets the updates for a particular instance. That would be a nice place to put the logic that's currently [here](https://github.com/mila-udem/blocks/pull/941/files#diff-38cc460193f54f5dd06fcce5910749d7R201) so that it can be used by the standalone function as well.

blocks/bricks/bn.py
dwf: 169e0cc

blocks/bricks/bn.py
dwf: 11c08bf

blocks/bricks/bn.py
dwf: 4b2302b

Thrandis:
@dwf I went really quickly through the code, and here is a comment:

Why not calling `W` `gamma` and `b` `beta`, as in the paper? It would, for example, help understanding what are those parameters when you are looking at the list of parameters in a 'computation_graph'.

Also, I am wondering the role `WEIGHT` is the right one for gamma. If you think about regularization schemes like L2, weight noise or dropout, you would apply it on the `WEIGHT` of the network. However, it shouldn't be applied to the `gamma` and `beta` parameters of the networks, or maybe not in the same way.

dwf:
@cooijmanstim I've added `batch_normalization_updates`. PTAL.

@Thrandis that's a good point, actually. I mainly eschewed the beta/gamma names because they aren't very descriptive of the function if you haven't read the paper, but I'd be okay with changing to scale/shift, and giving them their own PARAMETER-inheriting roles.

dwf:
Okay, barring any further issues, and assuming the tests pass except for the stupid flaky sqlite one, I'd say I'd personally consider this ready for merge. :joy:

@cooijmanstim Thanks for your very thorough reviewing up to this point!

cooijmanstim:
Thanks, hope the granularity was right. :-P I really like how it turned out, especially having inference be the default makes handling the population statistics much smoother. The context manager is a nice touch and I could see that become the standard way to use this.

There is still a class of bugs in the way the context nesting is carried out though, but it won't happen and it will crash and burn rather than silently do the wrong thing. It's potentially triggered if one of the `__enter__` calls fails such that control goes to the `finally` block before all of the bricks have been entered. All of the bricks will be exited, which is incorrect. However if there are too many `__exit__` calls on a brick, one of them will fail to pop the `training_mode` stack because it is empty, and the code will crash.

dwf:
Yeah, I don't see how `__enter__` could reasonably fail, and as you said, it would die with a pop failure. So I don't think there's a reason to worry too much. 

Thanks again for the timely reviewing. 

@Thrandis @rizar @vdumoulin If there are no objections I'll merge around noon tomorrow.

blocks/bricks/bn.py
```diff
@@ -0,0 +1,350 @@
...
+    def num_output_channels(self):
+        return self.num_channels
+
+
+class SpatialBatchNormalization(BatchNormalization):
+    """Convenient subclass for batch normalization across spatial inputs.
+
+    Parameters
+    ----------
+    input_dim : int or tuple
```
vdumoulin: This says that `input_dim` can be an `int`, but it looks like the code enforces it to be a sequence of length `>= 2`.

vdumoulin:
This is not a PR blocker, but in its current state, your PR doesn't have batch-normalized convnet support, right?

vdumoulin:
Nevermind, I put my foot in my mouth. However, does the current code make it easy to apply batch normalization *before* the nonlinearity is applied in convnets?

blocks/bricks/bn.py
vdumoulin: Also, should we put this check at the allocation stage? Lazy assignation is broken otherwise.

vdumoulin:
There's something else I realized while trying this PR out: step rules that keep track of gradient statistics (e.g. `Momentum`, `RMSProp`) initialize those shared variables without paying attention to the broadcasting pattern of the corresponding shared variable, which makes graph compilation fail.

blocks/bricks/bn.py
dwf: Hmm, interestingly lazy initialization works, but I don't know why. But yes, that should be moved to `_allocate`.

dwf: Oh, I know why. NoneAllocation is not an instance of collections.Sequence (that's also why ints are accepted). But it should still be in `_allocate`.

dwf:
The way you'd currently apply it before the nonlinearity is with a ConvolutionalActivation where Activation is Sequence([SpatialBatchNormalization().apply, Rectifier().apply]). But, this brick also can go right into a ConvolutionalSequence, and there's no reason we can't make the trivial change to allow activation functions to be able to do the same.

blocks/bricks/bn.py
dwf: 2800b3f

dwf: d9171ff actually gets it right.

vdumoulin:
That does the job for me, thanks!

dwf:
@vdumoulin d129aa7

dwf:
Sorry, just amended as 90cc8bb

vdumoulin:
That fixes it for me, thanks!

vdumoulin:
Also, I had a look at the latest commit, LGTM.

dwf:
Okay, good. I'll merge once the tests eventually pass (Travis is having some issues again today according to their status page).

dwf:
Tests pass except for the flaky sqlite build, so I'm merging. Thanks @vdumoulin, @cooijmanstim for the assistance.

PR #946, opened 2016-01-20
ConvolutionTranspose

vdumoulin:
This PR does the following:
* Implement `ConvolutionTranspose`, also colloquially called "deconvolution", like in [Zeiler et al., 2010](http://www.matthewzeiler.com/pubs/cvpr2010/cvpr2010.pdf). Currently implemented in Tensorflow under the name `conv2d_transpose`, which they describe as "the transpose (gradient) of `conv2d`".
* Implement `ConvolutionTransposeActivation`, in analogy to `ConvolutionActivation`.
* Switch to Theano's freshly-[renamed](https://github.com/Theano/Theano/pull/3679) `pool_2d` and `Pool`.
* Switch to Theano's recent `abstract_conv` interface. The convolution ops created via this interface are placeholders that get replaced at graph compilation time with the best option available.

I'm very open to feedback regarding design choices, names, etc. In particular, I'm not sure `Convolutional.conv2d_impl` and `Convolutional.get_output_shape` are needed anymore, but I'm afraid of breaking external code by removing these attributes.

### DIFF
--- a/blocks/bricks/conv.py
+++ b/blocks/bricks/conv.py
@@ -294,7 +294,7 @@ class ConvolutionalTranspose(Initializable):
             W, b = self.parameters
         else:
             W, = self.parameters
-        imshp = tensor.stack(*((input_.shape[0],) + self.get_dim('output')))
+        imshp = (None,) + self.get_dim('output')
         kshp = (self.num_channels, self.num_filters) + self.filter_size
         output = AbstractConv2d_gradInputs(
             imshp=imshp, kshp=kshp, border_mode=self.border_mode,
--- a/blocks/bricks/conv.py
+++ b/blocks/bricks/conv.py
@@ -1,5 +1,5 @@
 from theano.tensor.nnet.conv import conv2d, get_conv_output_shape
-from theano.tensor.signal.downsample import max_pool_2d, DownsampleFactorMax
+from theano.tensor.signal.pool import pool_2d, Pool
 
 from blocks.bricks import Initializable, Feedforward, Sequence
 from blocks.bricks.base import application, Brick, lazy
@@ -230,16 +230,16 @@ class Pooling(Initializable, Feedforward):
             with the last two dimensions downsampled.
 
         """
-        output = max_pool_2d(input_, self.pooling_size, st=self.step,
-                             mode=self.mode, padding=self.padding,
-                             ignore_border=self.ignore_border)
+        output = pool_2d(input_, self.pooling_size, st=self.step,
+                         mode=self.mode, padding=self.padding,
+                         ignore_border=self.ignore_border)
         return output
 
     def get_dim(self, name):
         if name == 'input_':
             return self.input_dim
         if name == 'output':
-            return tuple(DownsampleFactorMax.out_shape(
+            return tuple(Pool.out_shape(
                 self.input_dim, self.pooling_size, st=self.step,
                 ignore_border=self.ignore_border, padding=self.padding))
 
--- a/blocks/bricks/conv.py
+++ b/blocks/bricks/conv.py
@@ -1,6 +1,7 @@
 from theano import tensor
 from theano.tensor.nnet import conv2d
-from theano.tensor.nnet.abstract_conv import get_conv_output_shape
+from theano.tensor.nnet.abstract_conv import (AbstractConv2d_gradInputs,
+                                              get_conv_output_shape)
 from theano.tensor.signal.pool import pool_2d, Pool
 
 from blocks.bricks import Initializable, Feedforward, Sequence
--- a/blocks/bricks/conv.py
+++ b/blocks/bricks/conv.py
@@ -1,4 +1,5 @@
-from theano.tensor.nnet.conv import conv2d, get_conv_output_shape
+from theano.tensor.nnet import conv2d
+from theano.tensor.nnet.abstract_conv import get_conv_output_shape
 from theano.tensor.signal.pool import pool_2d, Pool
 
 from blocks.bricks import Initializable, Feedforward, Sequence
--- a/blocks/bricks/conv.py
+++ b/blocks/bricks/conv.py
@@ -187,8 +187,20 @@ class ConvolutionalTranspose(Convolutional):
     ----------
     image_size : tuple, optional
         Required for tied biases. Defaults to ``None``.
+    num_filters : int
+        Number of filters at the *output* of the transposed convolution,
+        i.e. the number of channels in the corresponding convolution.
+    num_channels : int
+        Number of channels at the *input* of the transposed convolution,
+        i.e. the number of output filters in the corresponding
+        convolution.
     original_image_size : tuple
-        The height and width of the output (image or feature map).
+        The height and width of the image that forms the output of
+        the transpose operation, which is the input of the original
+        (non-transposed) convolution.
+    step : tuple, optional
+        The step (or stride) of the corresponding *convolution*.
+        Defaults to (1, 1).
 
     See Also
     --------
--- a/blocks/bricks/conv.py
+++ b/blocks/bricks/conv.py
@@ -580,7 +580,7 @@ class ConvolutionalTransposeActivation(_AllocationMixin, Sequence,
         return self.convolution.get_dim(name)
 
     def _push_allocation_config(self):
-        super(ConvolutionalActivation, self)._push_allocation_config()
+        super(ConvolutionalTransposeActivation, self)._push_allocation_config()
         self.convolution.step = self.step
         self.convolution.original_image_size = self.original_image_size
 
--- a/blocks/bricks/conv.py
+++ b/blocks/bricks/conv.py
@@ -180,142 +180,45 @@ class Convolutional(Initializable):
         return self.num_filters
 
 
-class ConvolutionalTranspose(Initializable):
+class ConvolutionalTranspose(Convolutional):
     """Performs the transpose of a 2D convolution.
 
     Parameters
     ----------
-    filter_size : tuple
-        The height and width of the filter (also called *kernels*).
-    num_filters : int
-        Number of filters per channel.
-    num_channels : int
-        Number of input channels in the image. For the first layer this is
-        normally 1 for grayscale images and 3 for color (RGB) images. For
-        subsequent layers this is equal to the number of filters output by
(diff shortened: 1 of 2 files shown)


dmitriy-serdyuk:
It looks, like this piece of code copy-pastes a lot from `Convolutional`. Is it possible to inherit from `Convolutional` and redefine the `conv2d_impl` to use the gradient?

According to the name convention: `ConvolutionTranspose`  is not consistent with `Convolutional`. I propose to use `Deconvolutional` and explain in the documentation what it does.

dwf:
"Deconvolution" is a huge misnomer though.

https://en.wikipedia.org/wiki/Deconvolution

vdumoulin:
I agree with @dwf there. The operation implemented in `ConvolutionTranspose` really *is* the transpose of the matrix representation of some convolution.

vdumoulin:
Well this is awkward. It looks like Theano doesn't like our compile flags when using `abstract_conv`.

dwf:
Perhaps we should add a test once that's resolved...

vdumoulin:
Our tests are run in `FAST_COMPILE` mode to save time, but that forces the Python implementation to be used. The problem was that there is no Python implementation of `CorrMM`, which is used on Travis since `cudnn` isn't available.

The solution is to manually force the `FAST_RUN` mode to be used in the `theano.function` call for `blocks.bricks.conv` tests.

dwf:
Okay, can you leave a comment to that effect (even just copy and paste what
you just wrote) if you haven't already?

vdumoulin:
@dwf f588e4b

vdumoulin:
After further investigation, the `CorrMM` implementation should not have been used by Theano, since we disable C code through the `cxx=` flag.

This should be fixed when Theano/Theano#3905 is merged, at which point I'll re-run the tests to confirm that the issue is gone.

vdumoulin:
The CI tests are now passing!

blocks/bricks/conv.py
```diff
@@ -177,6 +180,144 @@ def num_output_channels(self):
...
+                 step=(1, 1), border_mode='valid', tied_biases=False,
+                 **kwargs):
+        super(ConvolutionalTranspose, self).__init__(**kwargs)
+
+        self.filter_size = filter_size
+        self.num_filters = num_filters
+        self.batch_size = batch_size
+        self.num_channels = num_channels
+        self.image_size = image_size
+        self.original_image_size = original_image_size
```
dwf: Is this strictly required? Seems kind of odd. Also looks like it's not required by `AbstractConv2d_gradInputs`.

blocks/bricks/conv.py
```diff
@@ -177,6 +180,144 @@ def num_output_channels(self):
...
+        the previous convolutional layer. The filters are pooled over the
+        channels.
+    original_image_size : tuple
+        The height and width of the output (image or feature map).
+    batch_size : int, optional
+        Number of examples per batch. If given, this will be passed to
+        Theano convolution operator, possibly resulting in faster
+        execution.
+    image_size : tuple, optional
+        The height and width of the input (image or feature map). If given,
```
dwf: I think this description is wrong. It never gets passed, but it's necessary if you want untied biases.

blocks/bricks/conv.py
vdumoulin: Judging from the conversation I had with @lamblin regarding Theano/Theano#3913, it's not required, but there are some cases where the original size can be ambiguous with non-unit strides.

Take for example a `3x3` convolution with strides `2x2`. Both a `9x9` input and a `10x10` input will produce a `4x4` output, so the gradient of the output with respect to the input is undefined if we don't know the original input size.

blocks/bricks/conv.py
vdumoulin: I'll update it. Would you also like me to update the `Convolutional` description as well?

dwf: Isn't it passed for Convolutional?

blocks/bricks/conv.py
dwf: Is it simple to enumerate the cases where it's ambiguous? That, along with an optional argument, would be ideal. If not, a short explanation of why this is required would be good.

blocks/bricks/conv.py
vdumoulin: You're right.

blocks/bricks/conv.py
```diff
@@ -177,6 +180,144 @@ def num_output_channels(self):
...
+    num_filters : int
+        Number of filters per channel.
+    num_channels : int
+        Number of input channels in the image. For the first layer this is
+        normally 1 for grayscale images and 3 for color (RGB) images. For
+        subsequent layers this is equal to the number of filters output by
+        the previous convolutional layer. The filters are pooled over the
+        channels.
+    original_image_size : tuple
+        The height and width of the output (image or feature map).
```
dwf: This could be clearer. Maybe "The height and width of the image that forms the output of the transpose operation, which is the input of the original (non-transposed) convolution."

blocks/bricks/conv.py
```diff
@@ -177,6 +180,144 @@ def num_output_channels(self):
...
+    filter_size : tuple
+        The height and width of the filter (also called *kernels*).
+    num_filters : int
+        Number of filters per channel.
+    num_channels : int
+        Number of input channels in the image. For the first layer this is
+        normally 1 for grayscale images and 3 for color (RGB) images. For
+        subsequent layers this is equal to the number of filters output by
+        the previous convolutional layer. The filters are pooled over the
+        channels.
```
dwf: I'm confused as to whether this refers to the output (of the transpose operation) or the input (of the transpose operation). I assume the former. So we should be clearer.

blocks/bricks/conv.py
```diff
@@ -177,6 +180,144 @@ def num_output_channels(self):
...
+        Number of examples per batch. If given, this will be passed to
+        Theano convolution operator, possibly resulting in faster
+        execution.
+    image_size : tuple, optional
+        The height and width of the input (image or feature map). If given,
+        this will be passed to the Theano convolution operator, resulting
+        in possibly faster execution times.
+    step : tuple, optional
+        The step (or stride) with which to slide the filters over the
+        image. Defaults to (1, 1).
```
dwf: This has a particular strange meaning in the case of this brick. Again, it's the stride of the original un-transposed convolution.

blocks/bricks/conv.py
dwf: Or maybe not, given the first line of apply(). Hmm...

dwf: err _allocate

blocks/bricks/conv.py
```diff
@@ -396,6 +537,54 @@ def _push_allocation_config(self):
...
+            **kwargs)
+
+    def get_dim(self, name):
+        # TODO The name of the activation output doesn't need to be `output`
+        return self.convolution.get_dim(name)
+
+    def _push_allocation_config(self):
+        super(ConvolutionalTransposeActivation, self)._push_allocation_config()
+        self.convolution.step = self.step
+        self.convolution.original_image_size = self.original_image_size
```
dwf: Seems  like more stuff should get pushed?

vdumoulin: It's handled by the `_AllocationMixin` superclass.

blocks/bricks/conv.py
vdumoulin: b3450f4 addresses that as well as @dmitriy-serdyuk 's suggestion of making `ConvolutionalTranspose` a subclass of `Convolutional`.

blocks/bricks/conv.py
vdumoulin: Pinging @lamblin on that.

vdumoulin:
@dwf 0a5f809 addresses your comments regarding the clarity of parameter description.

blocks/bricks/conv.py
dwf: Ok good.

blocks/bricks/conv.py
```diff
@@ -177,6 +180,59 @@ def num_output_channels(self):
         return self.num_filters
 
 
+class ConvolutionalTranspose(Convolutional):
+    """Performs the transpose of a 2D convolution.
+
+    Parameters
+    ----------
+    image_size : tuple, optional
```
dwf: I think this should go last if anywhere. Probably also should note that this is the input to the *transpose*, output of the original convolution.

dwf:
`ConvolutionalTransposeActivation` needs a test.

Other than that and my one comment about the `image_size` documentation, LGTM.

blocks/bricks/conv.py
```diff
@@ -1,5 +1,8 @@
-from theano.tensor.nnet.conv import conv2d, get_conv_output_shape
-from theano.tensor.signal.downsample import max_pool_2d, DownsampleFactorMax
+from theano import tensor
+from theano.tensor.nnet import conv2d
```
dwf: Oh, is this actually using abstract conv2d or just regular old conv2d?

vdumoulin: No, the old `conv2d` is located in `theano.tensor.nnet.conv`. This one aliases to `theano.tensor.nnet.abstract_conv.conv2d`.

dwf: Got it.

vdumoulin:
@dwf 87d143f addresses your constructor arg reordering suggestion and removes more duplicate code, as well as clarifies and moves down the `image_size` description. 648646d adds a test for `ConvolutionalTransposeActivation`.

With that, I think I addressed all your comments. Thank you very much for you feedback, and please let me know if there's anything else I forgot.

dwf:
LGTM

blocks/bricks/conv.py
lamblin: So, maybe I'm too late, but the original size is ambiguous iff the stride is greater than 1.
However, currently the Theano Ops for `grad_wrt_inputs` require it as an input all the time, even when the stride is 1 and it could be inferred from the shapes of "topgrad" and "weights". Ideally, this would be improved in Theano.
Note that this computation when the stride is 1 would be similar to the shape inference for the forward convolution, but not exactly the same, since the padding would have to be handled differently.

PR #947, opened 2016-01-21
change attention.py

wenhuchen:
Change attention score calculation method.

### DIFF
--- a/blocks/bricks/attention.py
+++ b/blocks/bricks/attention.py
@@ -350,6 +350,18 @@ class SequenceContentAttention(GenericSequenceAttention, Initializable):
             match_vectors.shape[:-1], ndim=match_vectors.ndim - 1)
         return energies
 
+    @application
+    def compute_energies_multiply(self, attended, preprocessed_attended, states):
+        if not preprocessed_attended:
+            preprocessed_attended = self.preprocess(attended)
+        transformed_states = self.state_transformers.apply(as_dict=True,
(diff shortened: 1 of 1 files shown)


dwf:
Lots of tests failing, and commit messages like "change <file>" are not useful. We can determine what files are changed by looking at the diff. Commit message should say what the purpose of the change was.

rizar:
An explanation of why you think it is better to multiply than to add would be great.

wenhuchen:
Actually I made this change according to http://www.aclweb.org/anthology/D15-1166, the paper provides three algorithms to calculate the attention. I used the simple dot version and got an increase of about 0.5 BLEU in IWSLT De-En Ted-Talk data set.

dwf:
@wenhuchen Given who wrote it (@rizar), it's probably the case that the method of calculating the attention score currently is already employed in some part of the literature (notably Bahdanau et al, the first paper in the bibliography that manuscript you link). If you'd like to add flags or subclasses in order to make it easy to use the alternate method, that would be great, but simply replacing the score calculation method isn't an option. What works best on one dataset won't necessarily work best for all.

wenhuchen:
Thank you, I will try adding a subclasses and test if it really brings improvement in all dataset.

rizar:
Perhaps the simplest way to add this to Blocks would be to add a flag.
Also, a reference to the paper, where the method originates, would be
required.

PR #950, opened 2016-01-21
Diagnose stupid flaky test [do not merge]

dwf:


dwf:
@bartvm Somehow, this test fails 80% of the time on Travis recently. Can you take a look, since you wrote it and are the most familiar?

The bug is somewhere between at or below the SQLite log backend itself, as an extension attempting to overwrite the value of `saved_to` in the log's current row is failing to do so and retains the original value set (see [this assertion failure](https://travis-ci.org/mila-udem/blocks/jobs/104379032#L895) plus the diff for this PR).

Now that I've isolated the error somewhat I'll close this pull request and create an issue for this particular issue.

PR #953, opened 2016-01-24
Skip test_save_the_best with sqlite backend, restore sanity.

### LINKED ISSUE #942: Tests are flaky with sqlite backend
Both #940 and in-progress #941 have encountered failing tests with the sqlite backend. What's particularly troubling is that they are not failing deterministically, but only sometimes.

dwf:
This test is flaky due to the SQLite backend not always correctly overwriting a value already set on the current row, at least on Travis (#954).

This renders the CI status flag pretty useless, so for the time being, we'll skip this test when the sqlite backend is enabled.

Fixes #942.

### DIFF
--- a/blocks/utils/testing.py
+++ b/blocks/utils/testing.py
@@ -76,6 +76,27 @@ def skip_if_not_available(modules=None, datasets=None, configurations=None):
             raise SkipTest
 
 
+def skip_if_configuration_set(configuration, value, message=None):
+    """Raise SkipTest if a configuration option has a certain value.
+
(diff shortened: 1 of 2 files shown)


dwf:
Expected drop in coverage, but worth it if one doesn't have to dig through the log every single time there's a PR.

Are other people okay with this way of doing business? @dmitriy-serdyuk @vdumoulin @rizar @bartvm

rizar:
I think we have to do that for now, because development is really hindered by these episodic crashes.

PR #957, opened 2016-01-27
Optional bias in ShallowEnergyComputer

rizar:
@dmitriy-serdyuk , this is one in the series of the PRs that I will need to port changes from our fully-neural-lvsr repo.


dmitriy-serdyuk:
Simple change, LGTM.

PR #959, opened 2016-01-28
Lookup table changes

rizar:
Includes two changes: 

- get_dim added
- WEIGHT role added for the parameter


dmitriy-serdyuk:
LGTM

PR #961, opened 2016-01-28
Better error message and test for duplicate names

rizar:


blocks/monitoring/evaluators.py
```diff
@@ -114,7 +114,12 @@ def __init__(self, variables, use_take_last=False):
 
         self.variable_names = [v.name for v in self.variables]
         if len(set(self.variable_names)) < len(self.variables):
-            raise ValueError("variables should have different names")
+            duplicates = []
+            for vname in set(self.variable_names):
+                if self.variable_names.count(vname) > 1:
+                    duplicates.append(vname)
+            raise ValueError("variables should have different names!"
```
dmitriy-serdyuk: Can you capitalize 'v'?

rizar: The message after a colon, so it makes sense to start it with a small letter.

dmitriy-serdyuk:
In this case, LGTM.

PR #963, opened 2016-01-28
Use default protocol for consistency with dump

rizar:
@dwf, actually, I think this is better than #962 that you have just merged.


dwf:
SGTM

dwf:
Why is scrutinizer erroring? Are you rebased on top of current master?

rizar:
But it does not, it show a green check mark, doesn't it?

PR #968, opened 2016-01-30
Allow elementwise Activations inside ConvolutionalSequence.

dwf:
This lets you do things like
```python
ConvolutionalSequence([
    Convolutional(..., use_bias=False),
    SpatialBatchNormalization(),
    Tanh(),
    ...
])
```
Rather than the significantly more awkward
```python
ConvolutionalSequence([
    ConvolutionalActivation([
        SpatialBatchNormalization(input_dim=(...)).apply,
        Tanh().apply
    ],
    use_bias=False)])
```

"Flat is better than nested" quoth PEP20, etc.

In fact this completely eliminates the need for `ConvolutionalActivation` and it's `-Transpose` sibling, reducing codebase entropy quite a bit.

### DIFF
--- a/blocks/bricks/conv.py
+++ b/blocks/bricks/conv.py
@@ -495,6 +495,8 @@ class ConvolutionalSequence(Sequence, Initializable, Feedforward):
     layers : list
         List of convolutional bricks (i.e. :class:`Convolutional`,
         :class:`ConvolutionalActivation`, or :class:`Pooling` bricks).
+        :class:`Activation` bricks that operate elementwise can also
+        be included.
     num_channels : int
         Number of input channels in the image. For the first layer this is
         normally 1 for grayscale images and 3 for color (RGB) images. For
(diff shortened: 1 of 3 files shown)


dwf:
@vdumoulin Can you have a look at this? It's a small diff. (Let's wait until after the ICML deadline to kill ConvolutionalActivation, if that's what we decide to do.)

dwf:
@bartvm this might be interesting for the class if they want to use batch normalization in their conv nets.

vdumoulin:
Looks good, although `ConvolutionalSequence.get_dim` might fail if the last layer is an `Activation` subclass. One way around that would be to call the last non-`Activation`'s `get_dim` method.

dwf:
@vdumoulin Good catch, 84da520 fixes that issue.

dwf:
Oops, make that b2990d2.

Also 3569882 switches to calling the public version of the method which I don't think there was a good reason not to do in the first place.

vdumoulin:
LGTM

dwf:
Great, thanks. Absent any complaints I'll merge tomorrow morning.

PR #973, opened 2016-02-02
`learning_rate` as theano shared variable in Adam

johnarevalo:
So it can be monitored or modified using extensions.

### DIFF
--- a/blocks/algorithms/__init__.py
+++ b/blocks/algorithms/__init__.py
@@ -743,8 +743,10 @@ class AdaGrad(StepRule):
 
     """
     def __init__(self, learning_rate=0.002, epsilon=1e-6):
-        self.learning_rate = learning_rate
-        self.epsilon = epsilon
+        self.learning_rate = shared_floatx(learning_rate, "learning_rate")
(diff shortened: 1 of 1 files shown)


blocks/algorithms/__init__.py
```diff
@@ -789,7 +789,7 @@ class Adam(StepRule):
     def __init__(self, learning_rate=0.002,
                  beta1=0.1, beta2=0.001, epsilon=1e-8,
                  decay_factor=(1 - 1e-8)):
-        self.learning_rate = learning_rate
+        self.learning_rate = shared_floatx(learning_rate, "learning_rate")
```
rizar: I support this change, bet let's be consistence, either all hyperparameters in an algorithm should be a shared variables, or none. Could you please do the same for `beta1`, `beta2`, etc.

dwf: :+1:

johnarevalo:
@rizar, rebase done. It also updates `AdaGrad`

rizar:
Looks good, thanks!

PR #974, opened 2016-02-02
Fix Application pickling

### LINKED ISSUE #958: Subtle issue with Application serialization
This test currently fails with an assertion error in the implementation failing.

```python
from theano import tensor
from blocks.bricks import BatchNormalization
from blocks.graph import ComputationGraph, apply_batch_normalization


def test_batch_normalization_deserialized():
    y = BatchNormalization(input_dim=5).apply(tensor.matrix())
    yprime = cPickle.loads(cPickle.dumps(y))
    apply_batch_normalization(ComputationGraph([yprime]))
```

The issue seems to be that the `Application` wrapper object gets deserialized along with the graph, and the one currently attached to the `BatchNormalization` class does not compare equal to the deserialized one.

I can think of a few ways to solve this:

* The most obvious is to change the assertion to look at `application.name` and 
* Add some `__setstate__` and `__getstate__` or maybe `__new__` magic to make sure that these objects are replaced by their equivalents that are instantiated when the interpreter evaluates the class (this is aesthetically the nicest, since it futureproofs against changes to `Application` and also prevents there being two objects in memory that represent `Foo.apply`).
* Implement `__eq__` on `Application` so that non-identical objects at least compare equal if they *represent* the same application.

rizar:
Fixes #958

### DIFF
--- a/blocks/bricks/base.py
+++ b/blocks/bricks/base.py
@@ -326,6 +326,13 @@ class Application(object):
             return OrderedDict(zip(bound_application.outputs, outputs))
         return unpack(outputs)
 
+    # Application instances are used instead of usual methods in bricks.
+    # The usual methods are not pickled per-se, similarly to classes
+    # and modules. Instead, a reference to the method is put into the pickle.
+    # Here, we ensure the same behaviour for Application instances.
+    def __reduce__(self):
+        return (getattr, (self.brick, self.application_name))
+
 
 class BoundApplication(object):
     """An application method bound to a :class:`Brick` instance."""
--- a/tests/bricks/test_bricks.py
+++ b/tests/bricks/test_bricks.py
@@ -3,6 +3,7 @@ import six
 import theano
 from numpy.testing import assert_allclose, assert_raises
 from theano import tensor
+from six.moves import cPickle
 
 from blocks.bricks import (Identity, Linear, Maxout, LinearMaxout, MLP, Tanh,
                            Sequence, Random, Logistic, Softplus, Softmax)
@@ -247,6 +248,10 @@ def test_application():
     test_brick.delegated_apply.inputs = ['x']
     assert test_brick.delegated_apply.inputs == ['x']
     assert TestBrick.delegated_apply.inputs == ['w']
+
+
+def test_application_serialization():
+    assert id(cPickle.loads(cPickle.dumps(Linear.apply))) == id(Linear.apply)
 
 
 def test_apply():


rizar:
Even if you like, please don't merge, because if would like to use the same method for pickling variable roles.

dwf:
OK, ping me when you want a review.

rizar:
I  want a review, but even if it looks good, don't merge please :)

dwf:
I guess what I meant is, ping me when it's complete and I'll review it all
at once? :)

rizar:
I am just curious if you find the particular method that I applied a good
one. Maybe you are aware of certain issues with `copyreg`, etc.

blocks/bricks/base.py
```diff
@@ -327,6 +328,16 @@ def copy_and_tag(variable, role, name):
         return unpack(outputs)
 
 
+# Ensure that Application instance are not pickled, and that when
+# unpickling, a reference to an existing Application object is
+# substituted instead of creating a new one.
```
dmitriy-serdyuk: Why doesn't it happen with regular pickling?

dmitriy-serdyuk: Sorry, I misunderstood the original issue. But this comment is extremely confusing.

dmitriy-serdyuk:
As for me, this approach is too hacky.  We already use custom picklers, getstate/setstate magic. It becomes a bit messy.

blocks/bricks/base.py
dwf: > Why doesn't it happen with regular pickling?

Because the Application object is created at interpret-time, but when pickled it's serialized like everything else, so when you unpickle there's two.

rizar: @dmitriy-serdyuk : if you have an idea how to do it with `__getstate__` and
`__setstate__`, I would be surely interested to listen to your proposal.
But as far as I understand, __setstate__ is only run *after* the object is
created, and there is no way to avoid creating a new object using
`__setstate__`.

rizar:
I improved the explanation of what is being done in the code. Also, I decided not to use the same technique for variable roles, because there are many variable role classes and a separate `copyreg.pickle` call would have to be done for each of them.

@dwf , I think the PR is ready for review.

dwf:
You can't use `__getstate__` and `__setstate__` but there is one other way to do it.

```python
class Bar(object):
    def __new__(cls, parent, attr):
        existing = getattr(parent, attr, None)
        if existing is None:
            return super(Bar, cls).__new__(cls)
        else:
            return existing
```

And then we have

```python
In [26]: class Foo(object):
    pass
   ....:

In [27]: Foo.bar = Bar(Foo, 'bar')

In [28]: pickle.loads(pickle.dumps(Foo())).bar is Foo.bar
Out[28]: True
```

rizar:
Comparing two solution by the level of obscurity, I prefer mine. Overriding `__new__` not to create a new object sounds like an abuse of this functionality to me.

bartvm:
I'm not sure it's an abuse of functionality, from the data model documentation

>  [`__new__`] is also commonly overridden in custom metaclasses in order to customize class creation.

It's also a common way of implementing a singleton in Python

```python
class Singleton(object):
  _instances = {}
  def __new__(class_, *args, **kwargs):
    if class_ not in class_._instances:
        class_._instances[class_] = super(Singleton, class_).__new__(class_, *args, **kwargs)
    return class_._instances[class_]
```
And this case is kind of like a singleton, no?

rizar:
I did find a recipe of implementing singletons with __new__ on the internet:


http://python-3-patterns-idioms-test.readthedocs.org/en/latest/Singleton.html

One difference from what we are doing here is that it is the Singleton
class in the example that keeps track of the instances. David's code
potentially allows creating two Application for the same brick and with the
same name, given that both are created before any one of them is assigned
as the brick's attribute.

I am also a bit afraid that according to the docs `__init__` will be called
every time after `__new__`.

rizar:
There one more issue with the singleton approach: when `Application` is created, it is not yet known to which brick it will belong...

rizar:
Please check out the new solution with `__reduce__`: I find it pretty simple and transparent.

rizar:
@bartvm gave me oral LGTM for using `__reduce__`. I will squash the commits, fix the docs and merge.

BTW, the singleton approach seems to be applicable to the roles!

dwf:
LGTM too. Thanks.

PR #976, opened 2016-02-06
WIP: New Serialization 2

### LINKED ISSUE #829: Finish new serialization

Thrandis:
Since the rebase failed, I open a new PR here! Fixes #829 and #758
What needs to be done:
- [ ] Some tests are missing (`Load`)
- [x] Fix `Load` extension bug

### DIFF
--- a/blocks/extensions/saveload.py
+++ b/blocks/extensions/saveload.py
@@ -37,6 +37,10 @@ class Checkpoint(SimpleExtension):
         in a separate file in the tar archive. It may be used for example
         to save the log separetely. The name of the attribute will be used
         as name in the tar file.
+    save_main_loop : bool
+        Choose whether to save the main loop or not. This can be useful
+        for example if you are only interested in saving the log, but
+        not the whole main loop. Defaults to `True`.
     use_cpickle : bool
         See documentation of :func:`~blocks.serialization.dump`.
 
@@ -53,12 +57,13 @@ class Checkpoint(SimpleExtension):
 
     """
     def __init__(self, path, parameters=None, save_separately=None,
-                 use_cpickle=False, **kwargs):
+                 save_main_loop=True, use_cpickle=False, **kwargs):
         kwargs.setdefault("after_training", True)
         super(Checkpoint, self).__init__(**kwargs)
         self.path = path
         self.parameters = parameters
         self.save_separately = save_separately
+        self.save_main_loop = save_main_loop
         self.use_cpickle = use_cpickle
 
     def do(self, callback_name, *args):
@@ -81,7 +86,10 @@ class Checkpoint(SimpleExtension):
             if self.parameters is None:
                 if hasattr(self.main_loop, 'model'):
                     self.parameters = self.main_loop.model.parameters
-            secure_dump(self.main_loop, path,
+            object_ = None
+            if self.save_main_loop:
+                object_ = self.main_loop
+            secure_dump(object_, path,
                         dump_function=dump_and_add_to_dump,
                         parameters=self.parameters,
                         to_add=to_add,
--- a/blocks/extensions/saveload.py
+++ b/blocks/extensions/saveload.py
@@ -71,11 +71,11 @@ class Checkpoint(SimpleExtension):
             if from_user:
                 path, = from_user
             if self.save_separately:
-                add_to_dump = {}
+                to_add = {}
                 for attr in self.save_separately:
-                    add_to_dump[attr] = getattr(self.main_loop, attr)
+                    to_add[attr] = getattr(self.main_loop, attr)
             else:
-                add_to_dump = None
+                to_add = None
             if hasattr(self.main_loop, 'model'):
                 parameters = self.main_loop.model.parameters
             else:
@@ -83,7 +83,7 @@ class Checkpoint(SimpleExtension):
             secure_dump(self.main_loop, path,
                         dump_function=_dump_and_add_to_dump,
                         parameters=parameters,
-                        add_to_dump=add_to_dump,
+                        to_add=to_add,
                         use_cpickle=self.use_cpickle)
         except Exception:
             path = None
--- a/blocks/extensions/saveload.py
+++ b/blocks/extensions/saveload.py
@@ -76,9 +76,13 @@ class Checkpoint(SimpleExtension):
                     add_to_dump[attr] = getattr(self.main_loop, attr)
             else:
                 add_to_dump = None
+            if hasattr(self.main_loop, 'model'):
+                parameters = self.main_loop.model.parameters
+            else:
+                parameters = None
             secure_dump(self.main_loop, path,
                         dump_function=_dump_and_add_to_dump,
-                        parameters=self.main_loop.model.parameters,
+                        parameters=parameters,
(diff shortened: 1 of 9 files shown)


Thrandis:
@rizar In the `Checkpoint` extension, it can happen that a `main_loop` doesn't have any `model`, which means that we don't save any parameters (because we extract the parameters from the `model`). In this case, the extension `Load` will crash when it tries to load the parameters.
So, shall I add a `parameter` argument to the `Checkpoint` extension or shall I check if the `main_loop` that we want to load has a `model` (and thus, parameters)?

rizar:
Good question. For now I would probably go for an optional argument `parameters` in `Checkpoint`. If `parameters` are given, it can save them. If not, it can use `model.parameters`.

Thrandis:
@rizar Now the code finally working. There is just one thing to change in blocks-examples, but I still get an error when I run `test_reverse_words`, saying that the parameters haven't been saved.
I'll investigate that more.

blocks/extensions/saveload.py
```diff
@@ -31,12 +29,14 @@ class Checkpoint(SimpleExtension):
     ----------
     path : str
         The destination path for pickling.
+    parameters : list, optional
+        The parameters to save separaetely. If None, the parameters from
```
rizar: A typo, _separately_.

blocks/extensions/saveload.py
```diff
@@ -94,12 +74,20 @@ def do(self, callback_name, *args):
...
             if from_user:
                 path, = from_user
-            secure_dump(self.main_loop, path, use_cpickle=self.use_cpickle)
-            filenames = self.save_separately_filenames(path)
-            for attribute in self.save_separately:
-                secure_dump(getattr(self.main_loop, attribute),
-                            filenames[attribute], cPickle.dump,
-                            protocol=DEFAULT_PROTOCOL)
+            if self.save_separately:
+                to_add = {}
```
rizar: Perhaps a more concise way to write this code would be

```
to_add = {attr: getattr(self.main_loop) for attr in self.save_separately}
```

blocks/extensions/saveload.py
```diff
@@ -94,12 +74,20 @@ def do(self, callback_name, *args):
...
+                for attr in self.save_separately:
+                    to_add[attr] = getattr(self.main_loop, attr)
+            else:
+                to_add = None
```
rizar: You can also put `to_add = None` before `if self.save_separately`, thereby further reducing the number of lines.

blocks/extensions/saveload.py
```diff
@@ -94,12 +74,20 @@ def do(self, callback_name, *args):
...
+            if self.parameters is None:
+                if hasattr(self.main_loop, 'model'):
+                    self.parameters = self.main_loop.model.parameters
+            secure_dump(self.main_loop, path,
+                        dump_function=_dump_and_add_to_dump,
```
rizar: If `_dump_and_add_to_dump` is used outside of `blocks/serialization.py`, it should not be private (i.e. its name should not start with an underscore)

blocks/serialization.py
```diff
@@ -1,9 +1,119 @@
...
+>>> with open('main_loop.tar', 'rb') as src:
+...     parameters = load_parameters(src)
+>>> sorted(parameters.keys()) # doctest: +ELLIPSIS
+['/mlp/linear_0.W', '/mlp/linear_0.b', '/mlp/linear_1.W', '/mlp/line...]
+
+Loading parameters saved by :func:`dump` with :func:`load_parameters`
+ensures that their heirarchical names are compatible with
+:class:`~blocks.model.Model` and :class:`~blocks.select.Selector` classes.
+
+TODO: Add information about :func:`add_to_dump`.
```
rizar: It would be great to also mention here `pickle_object` option, because it is a great new feature that this PR brings. This, together with documenting `add_to_dump`, can be done in the ticket though.

blocks/extensions/saveload.py
```diff
@@ -94,12 +74,20 @@ def do(self, callback_name, *args):
...
+                        parameters=self.parameters,
+                        to_add=to_add,
+                        use_cpickle=self.use_cpickle)
```
rizar: Seems like `pickle_object` can not be used here, which is not nice. Many people are willing to sacrifice pickling the whole main loop in favour of having less restrictions on their code.

rizar:
Looks 99% ready!

I have found a few issues, but they all are not critical and can make your next ticket. I am inclined to merge this PR as soon as you fix `reverse_words`.

rizar:
Oops, one thing I think must be done before we merge: it should be possible to pass `pickle_object` through `Checkpoint`.

dwf:
Also, what happens to people's old checkpoints with the merge of this PR? By now, people may have a lot of checkpoints sitting around that they don't want to lose the ability to deserialize.

rizar:
I understand you concerns, but I don't know what we can do. We can't keep
two serialization implementations because that would be a huge mess. There
is no big deadline coming, so I guess it is a good time for big changes.

We could move the current serialization code to Blocks-extras, and this way
people can still use the old one in those projects, where they have lots of
old models.

sotelo:
I'm having a look at this PR and I have a question. Do you guys think that saving ONLY the parameters it's an option worth considering for the Checkpoint extension? Sometimes my checkpoints are taking a very long time and I'd prefer to only save the parameters.

dwf:
I was thinking the same today but for the save_separately stuff... I really
only need the log in my case

rizar:
Yes we are considering this! `serialization.dump` already has the
`pickle_object` option that can be set to `False`, the only thing left is
to accept it in `Checkpoint`.

rizar:
@Thrandis , can you please give an estimate how much time you need to finish this?

Thrandis:
@rizar as soon as the reverse_words is fixed, it will go pretty fast. But I still don't know what doesn't work with it! I wanted to do this earlier this week, but I got sick. I'll try to work on it this afternoon if I'm ok, otherwise I will do it tomorrow!

rizar:
Get better! I will take a look at the test log, there should be no serious
problem there.

Thrandis:
@rizar I finally managed to fix the bug (I had to add another `file_.seek(0)` in add_to_dump).

Now I'll fix what you commented above!

Thrandis:
@rizar : We still have to do it twice (I just checked). But I can put it before `if parameters is not None:` if you prefer!

Thrandis:
@rizar let me know if everything looks good to you, and I'll change back `.travis.yaml`.

rizar:
@sotelo , @dwf , the features that you request was added by Cesar in one of the latest commits.

@bartvm , @vdumoulin , @dmitriy-serdyuk , @dwf : what do guys think about releasing the 0.3 version before merging this PR?

sotelo:
Thanks a lot @Thrandis 
This PR seems SUPER useful!

rizar:
@Thrandis , looks very good, you can change `.travis.yaml` back.

Thrandis:
@rizar done!

Issue #970, opened 2016-02-01
Metaissue: Improve initialization


dmitriy-serdyuk:
@memimo , this is going to be you CCW task. I don't know what's need to be done yet, I just summarize information.

I list here the issues with some summary because the title doesn't always correspond to the discussion.  
I found several discussions on initialization:
- #773 where many people are complaining about the current way of initialization and pushing the config
- #740 which proposes a new way to initialize stuff
- #384 another thread on initialization
- #31 and yet another

Other threads which concern initialization:
- #347 discussion on where we should check if every brick was initialized
- #463 enhance initialization schemes printing
- #816 argues about noninitializable sequence and raises a question how we should manage several different properties
- #935 figures out that allocation and possibly initialization is broken with several parent bricks

@rizar , I forgot something?

rizar:
@memimo , can you please drop by at  3248 to discuss this ticket?

memimo:
I'm not in the lab today. Still have been finishing my last ticket. Will come by tomorrow.

rizar:
Summary of the discussion of @rizar, @memimo  and @bartvm  

1) Initialization schemes are stored in a role -> scheme dictionary attribute of a brick.
2) Each brick has a class attribute which specifies all the roles that its parameters can possibly have.
3) A method `Brick.collect_roles` collects the set of roles that the brick and its children can have. 
4) The role -> scheme dictionary is constrained to only have the roles from `self.collect_roles`, and this constraint should be enforced when initialization configuration is pushed down the brick hierarchy.
5) We keep attribute based access through `__getattr__`. 
6) We accept `<role>_init` attribute in the constructor as well as the full role -> scheme dictionary.

PR #980, opened 2016-02-08
Make child alloc/init/config push conditional.

### LINKED ISSUE #935: Multiple parents blow away child bricks' existing allocation
Consider this example:

```python
from blocks.bricks import Rectifier, MLP, Sequence, Logistic
from blocks.select import Selector
from theano import tensor
from theano.printing import debugprint

sub_mlp = MLP([Rectifier(), Rectifier()], [5, 4, 3])
a = Sequence([sub_mlp.apply, Logistic().apply])
b = Sequence([sub_mlp.apply, Logistic().apply])
x = tensor.matrix()
y = a.apply(x)
ids_pre = sum([[id(x) for x in lin.parameters] for lin in sub_mlp.linear_transformations], [])
print(ids_pre)
z = b.apply(x)
ids_post = sum([[id(x) for x in lin.parameters] for lin in sub_mlp.linear_transformations], [])
print(ids_post)
```
The two lists of ids will be different, because upon calling `b.apply(...)`, `b` sees that it is not allocated and allocates itself and its children, thus blowing away the allocation that `a` did. Now the parameters of `sub_mlp` from the first allocation are in the computational graph of `y` but an entirely different set of parameters are in the computational graph of `z`. Notably, *both* sets of parameters say they belong to `sub_mlp` in their metadata.

This is a pretty ugly situation. I notice that the `parents` logic in Brick *does* suppose that a brick can have multiple parents, but the allocation situation in this scenario makes that borderline unusable. This could be ameliorated in a couple of (complementary) ways.

* Allocation could be not be automatically performed by `apply(...)`. This would have at least made the problem apparent to me a lot more quickly.
* Re-allocation could raise an error (unless you pass some override flag to `allocate()`) . Not sure how viable this is, but I think it respects the "brick life cycle" outlined in the docs.

A work-around in this particular instance is to set `b.allocated` to `True` (and `b.parameters to []`, if you want to satisfy assumptions elsewhere in the library e.g. you want the Selector to work) but this isn't a very general solution. Better control over (and *documentation* of) the tendency to obliterate existing child allocations seems in order, in any case.

dwf:
@rizar @memimo #970, fixes #935 I think (but maybe not satisfactorily).

### DIFF
--- a/blocks/bricks/base.py
+++ b/blocks/bricks/base.py
@@ -606,7 +606,8 @@ class Brick(Annotation):
         if not self.allocation_config_pushed:
             self.push_allocation_config()
         for child in self.children:
-            child.allocate()
+            if not child.allocated:
+                child.allocate()
         self.parameters = []
         self._allocate()
         self.allocated = True
@@ -646,7 +647,8 @@ class Brick(Annotation):
         if not self.initialization_config_pushed:
             self.push_initialization_config()
         for child in self.children:
-            child.initialize()
+            if not child.initialized:
+                child.initialize()
         self._initialize()
         self.initialized = True
 
@@ -675,11 +677,12 @@ class Brick(Annotation):
         self._push_allocation_config()
         self.allocation_config_pushed = True
         for child in self.children:
-            try:
-                child.push_allocation_config()
-            except Exception:
-                self.allocation_config_pushed = False
-                raise
+            if not child.allocated:
+                try:
+                    child.push_allocation_config()
+                except Exception:
+                    self.allocation_config_pushed = False
+                    raise
 
     def _push_allocation_config(self):
         """Brick implementation of configuring child before allocation.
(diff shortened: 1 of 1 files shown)


blocks/bricks/base.py
```diff
@@ -646,7 +647,8 @@ def initialize(self):
         if not self.initialization_config_pushed:
             self.push_initialization_config()
         for child in self.children:
-            child.initialize()
+            if not child.initialized:
+                child.initialize()
```
rizar: This is too radical of a solution. I can easily imagine that there is a lot of code where bricks are initialized, then initialization schemes are altered, then they are initialized again.

At the same time, I think it would be okay to do it for allocation, because one normally does not want to reallocate a brick.

rizar: On the other hand, maybe we should forbid both reallocation and reinitialization to keep things a bit simpler... needs more thinking.

PR #989, opened 2016-02-16
Add suffix option to MonitoringExtension

bouthilx:
Only prefix is supported so far and I see no reasons why favoring
prefixes over suffixes and vice versa. Also, `Printing` extensions makes
it convenient to add suffixes rather than prefixes so that *quietly*
logged variables with names beginning with `_` keep naturally their *private*
property.

Right now we need to do something like this

```python
to_be_printed_monitor = DataStreamMonitoring(
    variables=[to_be_printed_variable], data_stream=data_stream,
    prefix="prefix")

_prefix_not_to_be_printed_variable = _not_to_be_printed_variable
_prefix_not_to_be_printed_variable.name = \
    "_prefix" + _not_to_be_printed_variable.name
no_to_be_printed_monitor = DataStreamMonitoring(
    variables=[_prefix_not_to_be_printed_variable], data_stream=data_stream)
```

But with suffixes we can create a single DataStreamMonitoring

```python
monitor = DataStreamMonitoring(
    variables=[to_be_printed_variable, _no_to_be_printed_variables],
    data_stream=data_stream, suffix="suffix")
```

### DIFF
--- a/blocks/extensions/monitoring.py
+++ b/blocks/extensions/monitoring.py
@@ -9,7 +9,7 @@ from blocks.monitoring.aggregation import MonitoredQuantity, take_last
 from blocks.monitoring.evaluators import (
     AggregationBuffer, MonitoredQuantityBuffer, DatasetEvaluator)
 
-PREFIX_SEPARATOR = '_'
+SEPARATOR = '_'
 logger = logging.getLogger(__name__)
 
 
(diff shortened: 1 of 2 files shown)


rizar:
Sorry for the delay, I have just noticed this PR and will review it tomorrow!

bouthilx:
No problem!

rizar:
This is a perfect PR, everything is done very accurately! Sorry for not reviewing it earlier, but now it needs a rebase.

bouthilx:
Thank you @rizar. ^_^

PR #996, opened 2016-02-21
Better version handling for Blocks

rizar:
A follow-up of https://github.com/mila-udem/fuel/pull/256


rizar:
@bartvm , this needs a look from a Python guru :)

bartvm:
I believe this is exactly the same as done in https://github.com/mila-udem/fuel/pull/256, right? LGTM.

rizar:
Yes, this is exactly the same change.

PR #1003, opened 2016-02-24
Fix Blocks ignores python's recursionlimit setting

fvisin:
Don't override the user defined sys.setrecursionlimit if greater than the
config's default


dmitriy-serdyuk:
LGTM

Issue #1006, opened 2016-02-25
Monitoring of non-Theano quantities on training data


rizar:
Currently only `DataStreamMonitoring` can accept instance of `MonitoredQuantity`. Ideally, `TrainingDataMonitoring` should also be able to work with them. The implementation would be very similar to what we currently do in `DatasetEvalutor`.

dwf:
First I've heard of MonitoredQuantity. What is it and why does a simple callable not suffice?

rizar:
You can find it [here](https://github.com/mila-udem/blocks/blob/master/blocks/monitoring/aggregation.py#L176). It is more involved than a callback for two reasons:

- it defines a set of Theano variables that it needs to compute its value
- it defines how its values should be aggregated

It it especially useful for all kinds of models that generated structured outputs. One can compute complex statistics of what is generated, e.g. edit distance from the groundtruth. I can imagine it being quite useful for training generative models of images as well, e.g. if you have some code that can evaluates the quality of your samples, which you don't want to reimplement in Theano.

rizar:
Fixed by #1007

PR #1007, opened 2016-02-25
WIP: Monitoring of non-Theano quantities on training data.

rizar:
For #1006.

This is the very first code that worked for me.

### DIFF
--- a/blocks/extensions/monitoring.py
+++ b/blocks/extensions/monitoring.py
@@ -160,7 +160,7 @@ class TrainingDataMonitoring(SimpleExtension, MonitoringExtension):
                 self._required_for_non_variables.accumulation_updates)
             self._variables.initialize_aggregators()
             self._required_for_non_variables.initialize_aggregators()
-            self._non_variables.initialize()
+            self._non_variables.initialize_quantities()
         else:
             # When called first time at any iterations, update
             # monitored non-Theano quantities
@@ -185,4 +185,4 @@ class TrainingDataMonitoring(SimpleExtension, MonitoringExtension):
             self.add_records(
                 self.main_loop.log,
                 self._non_variables.get_aggregated_values().items())
-            self._non_variables.initialize()
+            self._non_variables.initialize_quantities()
--- a/blocks/extensions/monitoring.py
+++ b/blocks/extensions/monitoring.py
@@ -99,7 +99,7 @@ class TrainingDataMonitoring(SimpleExtension, MonitoringExtension):
 
     Parameters
     ----------
-    observables : list of :class:`~tensor.TensorVariable` or
+    variables : list of :class:`~tensor.TensorVariable` or
                   :class:`~blocks.monitoring.aggregation.MonitoredQuantity`
         The variables or non-Theano quantities to monitor.
         The variable names are used as record names in the logs.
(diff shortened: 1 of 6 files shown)


blocks/extensions/monitoring.py
```diff
@@ -122,22 +134,39 @@ def do(self, callback_name, *args):
         aggregation buffer and instructs the training algorithm what
         additional computations should be carried at each step by adding
         corresponding updates to it. In all other cases it writes
-        aggregated values of the monitored variables to the log.
+aggregated values of the monitored variables to the log.
```
dmitriy-serdyuk: Something wrong with indention here.

blocks/extensions/monitoring.py
```diff
@@ -122,22 +134,39 @@ def do(self, callback_name, *args):
...
         if callback_name == 'before_training':
             if not isinstance(self.main_loop.algorithm,
                               DifferentiableCostMinimizer):
                 raise ValueError
             self.main_loop.algorithm.add_updates(
-                self._buffer.accumulation_updates)
-            self._buffer.initialize_aggregators()
+                self._variables.accumulation_updates)
+            self.main_loop.algorithm.add_updates(
+                self._required.accumulation_updates)
```
dmitriy-serdyuk: Shouldn't you add updates for required variables before the variables?

rizar: I don't think that the order matters here. These updates will act on different shared variables.

rizar:
By the way, method names in classes related to monitoring are so random:

```python
class MonitoredQuantity(object):
     def initialize()
     def accumulate()
     def readout()

class MonitoredQuantityBuffer(object):
     def initialize()
     def accumulate_quantities()
     def get_aggregated_values()

class AggregationBuffer(object):
    def initialize_aggregators()
    def get_aggregated_values()
```

I guess a cleanup is needed, e.g. instead of `accumulate` we can always use `aggregate`. `initialize_aggregators` can be just `initialize`, `readout` can become `get_aggregated_value`. Opinions?

rizar:
parity_problem is broken because I renamed the input of `TrainingDataMonitoring` from `variables` to `observables`.  Not sure if that was a good decision, because after all, usually it is variables that is monitored. Opinions?

dwf:
Observables is a weird choice. Sounds like some sort of graphical model
thing. I'd stick with variables.

rizar:
Variables it is. Tests pass. What about the renaming that I proposed above?

rizar:
Assuming that the tests will pass, this is now ready for a final review.

dmitriy-serdyuk:
This looks good to me.

A note for future: `requires` syntax (like [here](https://github.com/mila-udem/blocks/pull/1007/files#diff-9abcd6167516efa783ed70442ad2c666R56)) looks a little bit weird. And the signature of `aggregate` of the abstract `MonitoredQuantity` doesn't suppose to input any additional parameters. Maybe we should refactor it to be able to input a dictionary `{name: variable}` and pass it by name.

rizar:
I think that "requires" argument currently is pretty convenient. The signature of `aggregate` is actually quite horrible, I will improve it a bit before the merge.

PR #1009, opened 2016-02-26
Refactor common stuff from Linear and Convolutional

dwf:
I was hacking out an implementation of [weight normalization](http://arxiv.org/abs/1602.07868) and realized that it would be significantly less code if the `W` and `b` properties that I introduced in `Linear` were a) used in the way I intended and b) existed in `Convolutional` too, so this accomplishes both (and reduces duplication to boot).

### DIFF
--- a/blocks/bricks/sequences.py
+++ b/blocks/bricks/sequences.py
@@ -1,4 +1,5 @@
 """Bricks that compose together other bricks in linear sequences."""
+import copy
 from toolz import interleave
 from picklable_itertools.extras import equizip
 
@@ -86,6 +87,10 @@ class MLP(Sequence, Initializable, Feedforward):
     dims : list of ints
         A list of input dimensions, as well as the output dimension of the
         last layer. Required for :meth:`~.Brick.allocate`.
+    prototype : :class:`.Brick`, optional
+        The transformation prototype. A copy will be created for every
+        activation. If not provided, an instance of :class:`~simple.Linear`
+        will be used.
 
     Notes
     -----
@@ -107,11 +112,15 @@ class MLP(Sequence, Initializable, Feedforward):
 
     """
     @lazy(allocation=['dims'])
-    def __init__(self, activations, dims, **kwargs):
+    def __init__(self, activations, dims, prototype=None, **kwargs):
         self.activations = activations
-
-        self.linear_transformations = [Linear(name='linear_{}'.format(i))
-                                       for i in range(len(activations))]
+        self.prototype = Linear() if prototype is None else prototype
+        self.linear_transformations = []
+        for i in range(len(activations)):
+            linear = copy.deepcopy(self.prototype)
+            name = self.prototype.__class__.__name__.lower()
+            linear.name = '{}_{}'.format(name, i)
+            self.linear_transformations.append(linear)
         # Interleave the transformations and activations
         application_methods = []
         for entity in interleave([self.linear_transformations, activations]):
--- a/blocks/bricks/__init__.py
+++ b/blocks/bricks/__init__.py
@@ -2,7 +2,8 @@
 from .base import application, Brick, lazy
 from .bn import (BatchNormalization, SpatialBatchNormalization,
                  BatchNormalizedMLP)
-from .interfaces import Activation, Feedforward, Initializable, Random
+from .interfaces import (Activation, Feedforward, Initializable, LinearLike,
+                         Random)
 from .simple import (Linear, Bias, Maxout, LinearMaxout, Identity, Tanh,
                      Logistic, Softplus, Rectifier, Softmax,
                      NDimensionalSoftmax)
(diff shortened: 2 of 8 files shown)


dwf:
Tests now pass. I don't understand the coverage drop but it's negligible anyway. @rizar @dmitriy-serdyuk thoughts?

dwf:
@rizar @bartvm Possible to get some eyes on this?

blocks/bricks/sequences.py
```diff
@@ -107,11 +112,15 @@ class MLP(Sequence, Initializable, Feedforward):
...
-    def __init__(self, activations, dims, **kwargs):
+    def __init__(self, activations, dims, prototype=None, **kwargs):
         self.activations = activations
-
-        self.linear_transformations = [Linear(name='linear_{}'.format(i))
-                                       for i in range(len(activations))]
+        self.prototype = Linear() if prototype is None else prototype
+        self.linear_transformations = []
+        for i in range(len(activations)):
+            linear = copy.deepcopy(self.prototype)
```
bartvm: Although I see why you need this, `deepcopy` kind of scares me. Just hoping that it doesn't end up copying things like application methods unnecessarily or something crazy like that.

bartvm:
LGTM, made one comment but it's more of a sidenote, feel free to merge.

blocks/bricks/sequences.py
dwf: I was just aping what the `parallel` bricks do, for the sake of consistency.

Personally I like the idea of *factories* more than prototypes (and if you need to partially configure a class, make your factory `functools.partial(TheClass, some_keyword=this, some_other_keyword=that)`).

dwf: @bartvm For reference, 

```
>>> a = Linear()
>>> b = copy.deepcopy(a)
>>> a.apply.application == b.apply.application
True
```

So I think whatever Dima did in #974 handles deepcopy correctly.

bartvm: Right, I didn't think of asking users to use `partial`, I agree that would be the better option. Should probably do the same for `parallel` bricks then though. We can make an issue for that if you don't feel like making it part of this PR.

bartvm: Cross-post, but cool, fingers crossed then.

dwf: I created an issue, anyway, if you're still interested in that direction (#1014).

rizar:
My humble opinion is that this PR is dangerously close to over-engineering. Sometimes a little bit of copy-paste is much easier to read.

dwf:
The mixin class part of it you mean? Or the whole thing?

bartvm:
The way I understand it it isn't about just avoiding copy-paste, it's about making classes use `self.W` agnostic to the fact whether `self.W` is the actual shared variable (`self.parameters[0]`) or a Theano expression for some sort of regularized version. I guess this makes it easier to write regularized versions of a series of bricks that have a single weight matrix and set of biases.

rizar:
@dwf , I meant `LinearLike`. I understand why one might want to use something different from `Linear` in `MLP`. 

@bartvm , sorry, I didn't really get "regularized versions of a series of bricks" part.

I am not too opinionated about that, but I think that this PR makes the number of people who are brave enough to read the code of the bricks even less. Also, from the viewpoint of OOP, it's kind of hard to say what is a linear-like brick. I think our other mixins, such as `Initializable` and `Random` have more semantics to them. E.g. `Initializable` is a brick, whose parameters can be initialized, and `Random` is a brick whose outputs depend on its inputs in a stochastic way. And what is `LinearLike` on the conceptual level? I don't know..

bartvm:
@dwf mentioned weight normalization, so I'm assuming he wants to do something like this (just made this up):

```python
class WeightNormalized(object):
    def _allocate(self):
        super(WeightNormalized, self)._allocate()
        self.parameters.append(shared_floatx(self.parameters[0].norm(2).eval()))

    @property
    def W(self):
        W = super(NotQuiteLinear, self).W
        return W / tensor.sqrt((W ** 2).sum()) * self.parameters[-1]

class WeightNormalizedLinear(WeightNormalized, Linear):
    pass

class WeightNormalizedConvolutional(WeightNormalized, Convolutional):
    pass
```

rizar:
That does look nice, however my question is: should we compromise simplicity of the very basic classes in our attempts to reproduce results from cutting-edge research papers? I am not sure this is a good idea. If weight normalization indeed proves successful, like batch norm did, then maybe it is worth to pay certain price for supporting it. But AFAIK, it is not yet that popular.

dwf:
This is actually the reason I introduced the properties on `Linear` way back, after (I think) a discussion with @rizar, but apparently it didn't get finished, or it got reverted back.

This way of parameterizing the brick is useful for much more than weight normalization: any instance where you wish to impose some sort of structure or constraint on the parameters (e.g. a mask, a constraint to be block diagonal, for example, or a low rank constraint, or indeed a normalized embedding matrix...).

The "simpler" implementation that did `W, b = self.parameters` inside `apply` also made it impossible to do subclasses where the subclass might have more parameters (this could've also been fixed with `W, b = self.parameters[:2]`, I guess).

rizar:
I see why `W, b = self.parameters` is unsatisfying. There was an idea from Jan Chorowski to have `Parameters.__getattr__` that would look for the parameter using its name, e.g. `self.parameters.W`, `self.parameters.b`. Or perhaps `__getitem__` and `self.parameters['W']`. What do you think about that? This way weight normalization could be implemented by changing the `self.parameters` attribute of `Brick` to create an instance of `NormalizedParameters` instead of `Parameters`. This could in turn be a mixin.

PR #1012, opened 2016-02-28
Support for mean-only batch normalization.

dwf:
As described [here](http://arxiv.org/abs/1602.07868).


dwf:
Tests are passing so I'd say this is ready for a look.

vdumoulin:
LGTM.

PR #1016, opened 2016-03-02
Tag algorithm buffers with parameter they are tracking.

### LINKED ISSUE #986: Statistics on algorithm buffers are hard to monitor
They generally only get a role but no other identifying information.

Often they mirror a specific parameter attached to a specific brick, so I'd recommend:

* Annotating them with the relevant brick
* Adding a `for_parameter` key to the tag similar to `replacement_of`.
* Maybe letting the `Selector` search for them somehow.

dwf:
Fixes #986.

### DIFF
--- a/blocks/algorithms/__init__.py
+++ b/blocks/algorithms/__init__.py
@@ -20,6 +20,13 @@ from blocks.utils import (dict_subset, pack, shared_floatx,
 logger = logging.getLogger(__name__)
 
 
+def _create_algorithm_buffer_for(param, *args, **kwargs):
+    buf = shared_floatx_zeros_matching(param, *args, **kwargs)
+    buf.tag.for_parameter = param
+    add_role(buf, ALGORITHM_BUFFER)
+    return buf
+
+
 @add_metaclass(ABCMeta)
 class TrainingAlgorithm(object):
     """Base class for training algorithms.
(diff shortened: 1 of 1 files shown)


blocks/algorithms/__init__.py
```diff
@@ -801,10 +801,10 @@ def __init__(self, learning_rate=0.002,
             add_role(param, ALGORITHM_HYPERPARAMETER)
 
     def compute_step(self, parameter, previous_step):
-        mean = shared_floatx_zeros_matching(parameter, 'mean')
-        add_role(mean, ALGORITHM_BUFFER)
-        variance = shared_floatx_zeros_matching(parameter, 'variance')
-        add_role(variance, ALGORITHM_BUFFER)
+        mean = _create_algorithm_buffer_for(parameter, 'mean')
+        mean.tag.for_parameter = parameter
```
rizar: This line seems to be not necessary, as well as a similar one below.

dwf: Good point, these predate my refactoring. I will fix them up later this evening (off to dinner).

dwf:
@rizar Fixed the issue you pointed out.

However, it also looks like Travis is shitting its pants (no g++?).

rizar:
LGTM on my side, can't help you with Travis though.

dwf:
So that coverage number is clearly bullshit caused by the Travis incident, since you gave it the LGTM I'll merge.

Issue #1018, opened 2016-03-04
Travis weirdness


bartvm:
As noticed by @dmitriy-serdyuk in https://github.com/mila-udem/blocks/pull/1013#issuecomment-192341151 and @dwf in https://github.com/mila-udem/blocks/pull/1016#issuecomment-192010418, Travis is throwing a tantrum again.

Theano complains about `g++` missing and not being able to load `scipy.linalg.blas`. The issues with `libgfortran` are reminiscent of https://github.com/mila-udem/blocks/issues/836, but the SciPy version hasn't changed since, so I'm not sure why this suddenly resurfaced...

dmitriy-serdyuk:
Seems, it's fixed now.

PR #1019, opened 2016-03-04
Update dependencies

bartvm:
Before investigating #1018 further, time for some updates.


bartvm:
The updates seem to fix `master`, and also fixes another `flake8` error that went unnoticed because of Scrutinizer's failure to set the test status correctly.

Because `master` is currently broken and because there's not really anything to review here, I'll just go ahead and merge.

PR #1020, opened 2016-03-04
Remove warning about g++ missing in Travis

bartvm:


bartvm:
The `g++ not detected` error remains, but I guess that makes sense, since GCC (and hence `g++`) was installed anyway (see https://docs.travis-ci.com/user/ci-environment/#Compilers-%26-Build-toolchain), it's just that Theano can't find it for some reason...

bartvm:
So the error turned out to be very misleading. We actually disabled C code intentionally by setting `THEANO_FLAGS=cxx=` in `.travis.yml`. Theano however interprets `cxx=` as setting it to `None` instead of setting to an empty string, so it complains. There is no way to explicitly pass an empty string using `THEANO_FLAGS`, but if you put `cxx =` in `.theanorc` the complaint disappears...

Issue #697, opened 2015-06-03
Do really need both use_bias and has_biases? 


dwf:
https://github.com/mila-udem/blocks/blob/master/blocks/bricks/__init__.py#L100,#L112 ಠ_ಠ

dwf:
Okay, reading the docs, I think I understand the difference. But I think `has_biases` is somewhat poorly named in that case, since I would colloquially say that `Linear(use_bias=False)` "does not have" a bias.

PR #1021, opened 2016-03-04
[WIP] For changing the initialization mechanism 

### LINKED ISSUE #970: Metaissue: Improve initialization

memimo:
fixes #970
This is just proof of concept. When you guys agree that this is the way to go, I will change all the codes affected by this, plus rebase to the head of the master.
@rizar @dmitriy-serdyuk

### DIFF
--- a/blocks/bricks/interfaces.py
+++ b/blocks/bricks/interfaces.py
@@ -141,31 +141,61 @@ class Initializable(RNGMixin, Brick):
 
     """
     has_biases = True
+    initialization_roles = set([])
 
     @lazy()
-    def __init__(self, weights_init=None, biases_init=None, use_bias=True,
-                 seed=None, **kwargs):
+    def __init__(self,
+                weights_init=None,
+                biases_init=None,
+                use_bias=True,
+                initialization_schemes = None,
+                seed=None,
+                **kwargs):
         super(Initializable, self).__init__(**kwargs)
-        self.weights_init = weights_init
-        if self.has_biases:
-            self.biases_init = biases_init
-        elif biases_init is not None or not use_bias:
-            raise ValueError("This brick does not support biases config")
+
+        self.initialization_schemes = initialization_schemes
+        if self.initialization_schemes is None:
+            self.initialization_schemes = {'weights_init': weights_init}
+            if use_bias:
+                self.initialization_schemes['biases_init'] = biases_init
+        elif weights_init is not None or biases_init is not None:
+            raise ValueError("All initializations are accepted either through"
+                "initialization_schemes or correspodong attribute but not both")
+
+        if not self.has_biases:
+            if initialization_schemes['biases_init'] is not None or not use_bias:
+                raise ValueError("This brick does not support biases config")
         self.use_bias = use_bias
         self.seed = seed
 
     def _push_initialization_config(self):
+        self._collect_roles()
+        for child in self.children:
+                if (isinstance(child, Initializable) and
+                        hasattr(child, 'initialization_schemes')):
+                    for role in child.initialization_schemes:
+                        if role not in self.initialization_roles:
+                            raise ValueError("The initialization role: " +
+                            "{} is not defined in the class ".format(role) +
+                            "initialization roles")
         for child in self.children:
             if isinstance(child, Initializable):
                 child.rng = self.rng
-                if self.weights_init:
-                    child.weights_init = self.weights_init
+                for role in self.initialization_schemes:
+                    if self.initialization_schemes[role]:
+                        child.initialization_schemes[role] = \
+                            self.initialization_schemes[role]
         if hasattr(self, 'biases_init') and self.biases_init:
             for child in self.children:
                 if (isinstance(child, Initializable) and
                         hasattr(child, 'biases_init')):
                     child.biases_init = self.biases_init
+        # import ipdb; ipdb.set_trace()
 
+    def _collect_roles(self):
+        for child in self.children:
+            if isinstance(child, Initializable):
+                self.initialization_roles.update(child.initialization_roles)
 
 class Random(Brick):
     """A mixin class for Bricks which need Theano RNGs.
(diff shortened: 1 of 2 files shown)


dwf:
I thought that part of the the plan was basically to be able to examine
kwargs for anything ending in `_init`, which chops off the `_init` and
looks that string up as a role name?

blocks/bricks/interfaces.py
```diff
@@ -141,31 +141,61 @@ class Initializable(RNGMixin, Brick):
 
     """
     has_biases = True
+    initialization_roles = set([])
 
     @lazy()
-    def __init__(self, weights_init=None, biases_init=None, use_bias=True,
-                 seed=None, **kwargs):
+    def __init__(self,
```
rizar: Please, do not use one-keyword-argument-per-line indentation scheme. It is never used in Blocks, for the sake of maintaining a common style for the codebase, I would like to ask you not to use it as well.

blocks/bricks/interfaces.py
```diff
@@ -141,31 +141,61 @@ class Initializable(RNGMixin, Brick):
 
     """
     has_biases = True
+    initialization_roles = set([])
```
rizar: What do you think about calling this one `parameter_roles`?

blocks/bricks/interfaces.py
```diff
@@ -141,31 +141,61 @@ class Initializable(RNGMixin, Brick):
...
             for child in self.children:
                 if (isinstance(child, Initializable) and
                         hasattr(child, 'biases_init')):
                     child.biases_init = self.biases_init
+        # import ipdb; ipdb.set_trace()
 
+    def _collect_roles(self):
+        for child in self.children:
+            if isinstance(child, Initializable):
+                self.initialization_roles.update(child.initialization_roles)
```
rizar: We've got a big problem here: you defined `initialization_roles` as a class attribute. Now this assignment will affect all the object of the class, which I guess is not what you wanted.

blocks/bricks/interfaces.py
```diff
@@ -141,31 +141,61 @@ class Initializable(RNGMixin, Brick):
 
     """
     has_biases = True
```
rizar: Why do we need `has_biases` when we have `initialization_roles`?

rizar:
@dwf , @bartvm , I guess I need a Python advice here. Is there a general agreement on when one should call `super().__init__`, in the beginning or in the end of `def __init__`? 

I think that in this PR we should do the following: first initialize the list of `children`, then call `super().__` that will eventually call `Initializable.__init__`. In the latter the roles of the children will be collected in order to validate the initialization schemes provided by the user.

This would only work if everybody places their `super().__init__` call in the end of `def __init__`. And I have a bad feeling about the numerous facepalms that this can cause. What do you think?

rizar:
@memimo , nice to see that you started working on the issue, but this PR is definitely very raw. We have to work closer together to come up with a good and implementable design.

blocks/bricks/interfaces.py
memimo: backward compatibility. I know there was a discussion about dealing with biases like rest of parameters but it didn't look as it was conclusive. I can get rid of it, and treat backward compatibility with a replication warning.

blocks/bricks/interfaces.py
memimo: yeah, that could work.

blocks/bricks/interfaces.py
memimo: yeah, will go thorough formatting issues at the end. This was just to open the discussion about the way to solve this issues.

blocks/bricks/interfaces.py
rizar: Backward compatibility is going to kill us, if we go for it. We have stable versions for those who work with their old models. So I think we should care only about getting a new solution for initialization.

rizar:
@dwf , @bartvm , can you please comment on what I wrote above?

blocks/bricks/interfaces.py
dwf: I complained about the name `has_biases` in #697 and would be happy to see it go.

dwf:
Whether you put `super().__init__(...)` in the beginning, end, or middle of the `__init__` definition (applies to any super call in any method, really) depends on what your constructor does and what you are relying on the superclass to do.

Sometimes you need to compute arguments for the `super().__init__` call based on your own arguments, so it's naturally going to come after that. Sometimes you want to let `super().__init__` configure something that you then tinker with slightly, so it's naturally going to come before that.

If there's no interaction between what I'm doing and what the superclass is doing then I typically put the call at the end, but I think this is a matter of taste. There are some [subtleties with respect to super and multiple inheritance](https://rhettinger.wordpress.com/2011/05/26/super-considered-super/) but the way we use it, they usually don't pose a problem as long as everyone calls super.

rizar:
OK, thanks for the feedback. 

I think I understood where the confusion comes from: we do not allow children to be passed as an argument to `Brick`. That's why it always initializes the list of chiden with `[]`, and that's why it is becomes crucial to fill `self.children` with content only after the super call. This should be changed, and the recommended way of setting the children should be by passing them as an argument to `Brick`. If we go this way, `Initializable.__init__` will call `super().__init__` first, which will initialize the list of children, which in turn allows it to check which initialization schemes a brick can have.

blocks/bricks/interfaces.py
rizar: Agreed.

memimo:
should I make a separate PR to fix the children issue first before continuing with this PR?

rizar:
If you agree that this is the way to go, then yes. I will create a ticket.

memimo:
Yeah, I think this makes things more clear and avoid the __init__ order issue.

PR #1022, opened 2016-03-04
Added test dump_and_add_to_dump

### LINKED ISSUE #998: Full test coverage for new serialization
@Thrandis , this will be your new ticket.

Thrandis:
Fixes #998 
I'll check what else is missing in coverage.

TODO:
- [x] use temporary files
- [x] use more specific exception types
- [x] test the data stream (aka iteration state)
- [ ] figure out why the tests are broken

### DIFF
--- a/tests/extensions/test_saveload.py
+++ b/tests/extensions/test_saveload.py
@@ -15,6 +15,7 @@ from blocks.extensions.saveload import Checkpoint, Load
 from blocks.initialization import Constant
 from blocks.main_loop import MainLoop
 from blocks.model import Model
+from blocks.utils.testing import skip_if_configuration_set
 
 
 class TestCheckpoint(unittest.TestCase):
@@ -57,6 +58,8 @@ class TestCheckpoint(unittest.TestCase):
 
     def test_load_log_and_iteration_state(self):
         """Check we can save the log and iteration state separately."""
+        skip_if_configuration_set('log_backend', 'sqlite',
+                                  'Bug with log.status["resumed_from"]')
         new_main_loop = MainLoop(
             model=self.model,
             data_stream=self.data_stream,
@@ -72,8 +75,6 @@ class TestCheckpoint(unittest.TestCase):
             assert new_key == old_key
             assert (new_main_loop.log.status[new_key] ==
                     self.main_loop.log.status[old_key])
-
-        assert new_main_loop.log.status == self.main_loop.log.status
         # Check the iteration state
         new = next(new_main_loop.iteration_state[1])['data']
         old = next(self.main_loop.iteration_state[1])['data']
--- a/tests/extensions/test_saveload.py
+++ b/tests/extensions/test_saveload.py
@@ -1,5 +1,7 @@
 import os
 import numpy
+import tarfile
+import tempfile
 import theano
 import unittest
 from fuel.datasets import IterableDataset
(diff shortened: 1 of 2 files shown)


Thrandis:
@rizar is there a reason why the coverage test is not running? I want to see what is not tested!

Thrandis:
@rizar nevermind, I found a way to see what is missing!

bartvm:
https://coveralls.io/jobs/12531159

Thrandis:
@bartvm yes that is what I'm finally using, because I wasn't able to run the coverage test locally!

Thrandis:
@rizar Do you have a nice way of testing the `load_iteration_state` of the `

rizar:
Sure. You can use `FakeMainLoop` from `blocks/utils/testing`. You can try to run it for a few iterations, serialize, load using `Load`, and check that it resumes from where it stopped. I don't know anybody using this feature though. Perhaps because currently `Load` unpickles the main loop, which is very restrictive.

tests/extensions/test_saveload.py
```diff
@@ -14,48 +15,76 @@
...
-        algorithm=GradientDescent(cost=cost, parameters=[W]),
-        extensions=[Load('mynonexisting.tar')]
-    )
-    main_loop.extensions[0].main_loop = main_loop
-    main_loop._run_extensions('before_training')
-
-    # Cleaning
-    if os.path.exists('myweirdmodel.tar'):
-        os.remove('myweirdmodel.tar')
+class TestCheckpoint(unittest.TestCase):
```
rizar: It's OK if you like using classes for your tests and not functions like you did here, but just to make sure we are on the same page, I should note that this is not mandatory. In this case you only have one method in the class, which seems to be equivalent to having a function. 

No need to change anything, this is just for your information.

tests/test_serialization.py
```diff
@@ -148,3 +150,16 @@ def test_secure_dump():
     assert_raises(PicklingError, secure_dump, bar, f.name)
     with open(f.name, 'rb') as f:
         assert type(load(f)) is object
+
+
+def test_dump_and_add_to_dump():
+    x = 3
+    y = 2
+    path = 'test.tar'
```
rizar: Can you please use `NamedTemporaryFile` like we do e.g. [there](https://github.com/mila-udem/blocks/blob/master/tests/test_serialization.py#L135)? Good thing about temporary files is that you don't have care about their deletion that much. They will be automatically deleted at some point anyway.

tests/extensions/test_saveload.py
```diff
@@ -14,48 +15,76 @@
...
+        # Test that we can load the log and the iteration state
+        new_main_loop = MainLoop(
+            model=Model(cost),
+            data_stream=data_stream,
+            algorithm=GradientDescent(cost=cost, parameters=[W]),
+            extensions=[Load('myweirdmodel.tar', True, True)]
+        )
+        new_main_loop.extensions[0].main_loop = new_main_loop
+        new_main_loop._run_extensions('before_training')
+        assert new_main_loop.log.status == main_loop.log.status
```
rizar: You don't test here that the iteration state has been loaded.

tests/extensions/test_saveload.py
```diff
@@ -14,48 +15,76 @@
...
+        main_loop.extensions[0].main_loop = main_loop
+        main_loop._run_extensions('before_training')
+
+        # Test loading exception
+        path = 'test.tar'
+        with open(path, 'wb') as f:
+            f.write('a')
+        load = Load(path)
+        load.main_loop = main_loop
+        self.assertRaises(Exception, load.before_training)
```
rizar: When you check, that an exception is raised, please check a more specific subclass of `Exception`, e.g. `IOError`, `OSError`, `KeyError`, etc.

rizar:
All right, here is your TODO list:

- use temporary files
- use more specific exception types
- test the data stream (aka iteration state)
- figure out why the tests are broken

tests/extensions/test_saveload.py
Thrandis: Yeah I wanted to use `SetUp` to create the first `MainLoop` and the reuse it in different tests. This way I can also use `self.assertRaises` to check the errors.

tests/test_serialization.py
Thrandis: Allright!

tests/extensions/test_saveload.py
Thrandis: Yep, I'll change it now that I know how to do it. Also, `assert new_main_loop.log.status == main_loop.log.status` is not working in some cases (see travis). Do you have a better way to check that the two logs are the same?

rizar: The so-called iteration state is just a tuple of datastream and a data
stream iterator. You can check that the next data provided by the iterator
is correct.

As for the comparison of statuses, it could that this is due to the random
order in the dictionaries. I do not remember what is the exact type of
status and how is comparison implemented for this type. I believe you can
sort it out, let Google be your guide ;)

Thrandis: Allright!

Thrandis:
@rizar everything seems to be fixed except for the sqlite version. Is there something different with the `log` with sqlite?

rizar:
I am not sure, can you please to run it with SQLite locally?

Thrandis:
@rizar so with the `python` backend, you have:
>>> new_main_loop.log.status[resumed_from] = None
>>> self.main_loop.log.status[resumed_from] = None
They should be the same, so it is fine! However with the `sqlite` backend:
>>> new_main_loop.log.status[resumed_from] = None
>>> self.main_loop.log.status[resumed_from] = 873f21e13b9148d084c7f70d90186174
So it seems that there is a problem with this `resumed_from` in the `sqlite` backend. It should be `None` because this `MainLoop` has not be resumed.

rizar:
All right, I will check where it comes from tomorrow.

rizar:
So I am struggling to understand SQLite log and `resumed_from`, and I suggest we skip this test for this backend. You can use [this helpful fuction](https://github.com/mila-udem/blocks/blob/master/blocks/utils/testing.py#L79) to do so.

Thrandis:
@rizar done!

rizar:
Thanks, good job!

PR #1026, opened 2016-03-10
make mlp inherit feedforward-sequence

### LINKED ISSUE #1008: MLP should inherit from FeedforwardSequence
... it inherits from Sequence and Feedforward separately, which is silly. Also it would make MLP's class definition shorter.

sygi:
and reuse its input/output dim properties.

fixes #1008


dwf:
I'm a little surprised that this passes tests, considering `self.dims` is used in `_allocate`.

sygi:
I wasn't sure how it works inside, but if executing `_push_allocation_config` is possible after using input/output dim setter, then should I leave it as it was (passing everything with self.dims) or iterate over `[self.input_dim] + self.dims[1:-1]`, etc. [here](https://github.com/mila-udem/blocks/commit/2d5b958d9f6dc21e401a24106092136f4ceb8536#diff-a7f8ce39f8353f5543e38cd8daed483dL158)?

dwf:
Yeah, I think I may have actually underestimated the degree that this would help the length. 

That solution won't work because it's not actually that often that someone will explicitly set the input_dim and output_dim of an MLP. So that breaks the case where you just do `mlp = MLP(dims=[5, 4, 3], activations=[Tanh(), Tanh(), Tanh()]); mlp.allocate()`, unfortunately.

Sorry for sending you on a bit of a nuisance quest.

dwf:
That one-liner would be good. :) Probably also worth squashing the commits.

dwf:
LGTM, thanks and sorry for the confusion.

PR #1028, opened 2016-03-11
Pass children as argument to Brick

### LINKED ISSUE #1027: It should possible to pass `children` as an argument to a Brick
See https://github.com/mila-udem/blocks/pull/1021#issuecomment-195388088

The PR should also include refactoring of all existing bricks. They should use pass children to the superclass constructor instead of using `self.children =`.

memimo:
fixes #1027

### DIFF
--- a/docs/create_your_own_brick.rst
+++ b/docs/create_your_own_brick.rst
@@ -57,8 +57,8 @@ of :class:`.Brick` for a precise description of the life-cycle of a brick):
 
 * :meth:`.Brick.__init__`: you should pass by argument the attributes of your
   brick. It is also in this method that you should create the potential
-  "children bricks" that belongs to your brick (in that case, you have to put
-  the children bricks into ``self.children``). The initialization of the
+  "children bricks" that belongs to your brick (in that case, you have to pass
+  the children bricks to ``super().__init__``). The initialization of the
   attributes can be lazy as described later in the tutorial.
 * :meth:`apply`: you need to implement a method that actually
   implements the operation of the brick, taking as arguments the inputs
@@ -210,10 +210,11 @@ specify the ``input_dim`` of ``brick2`` directly at its creation.
     >>> class ChainOfTwoFeedforward(Feedforward):
     ...     """Two sequential Feedforward bricks."""
     ...     def __init__(self, brick1, brick2, **kwargs):
-    ...         super(Feedforward, self).__init__(**kwargs)
     ...         self.brick1 = brick1
     ...         self.brick2 = brick2
-    ...         self.children = [self.brick1, self.brick2]
+    ...         children = [self.brick1, self.brick2]
+    ...         children += kwargs.get('children', [])
+    ...         super(Feedforward, self).__init__(children=children, **kwargs)
     ...
     ...     @property
     ...     def input_dim(self):
@@ -370,12 +371,14 @@ One can also create the brick using :class:`Linear` children bricks, which
     >>> class ParallelLinear2(Initializable):
     ...     def __init__(self, input_dim1, input_dim2, output_dim1, output_dim2,
     ...                  **kwargs):
-    ...         super(ParallelLinear2, self).__init__(**kwargs)
     ...         self.linear1 = Linear(input_dim1, output_dim1,
     ...                               use_bias=False, **kwargs)
     ...         self.linear2 = Linear(input_dim2, output_dim2,
     ...                               use_bias=False, **kwargs)
-    ...         self.children = [self.linear1, self.linear2]
+    ...         children = [self.linear1, self.linear2]
+    ...         children += kwargs.get('children', [])
+    ...         super(ParallelLinear2, self).__init__(children=children,
+    ...                                               **kwargs)
     ...
     ...     @application(inputs=['input1_', 'input2_'], outputs=['output1',
     ...         'output2'])
--- a/tests/extensions/test_saveload.py
+++ b/tests/extensions/test_saveload.py
@@ -15,6 +15,7 @@ from blocks.extensions.saveload import Checkpoint, Load
 from blocks.initialization import Constant
 from blocks.main_loop import MainLoop
 from blocks.model import Model
+from blocks.utils.testing import skip_if_configuration_set
 
 
 class TestCheckpoint(unittest.TestCase):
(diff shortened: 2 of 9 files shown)


memimo:
Travis is failing on the docs code. I'm not sure how my changes are causing those errors.

docs/create_your_own_brick.rst
```diff
@@ -375,7 +375,9 @@ One can also create the brick using :class:`Linear` children bricks, which
     ...                               use_bias=False, **kwargs)
     ...         self.linear2 = Linear(input_dim2, output_dim2,
     ...                               use_bias=False, **kwargs)
-    ...         self.children = [self.linear1, self.linear2]
+    ...         children = [self.linear1, self.linear2]
+    ...         super(ParallelLinear2, self).__init__(children=children,
```
bartvm: You're calling `super` twice now, you need to remove the call in line 373.

docs/create_your_own_brick.rst
```diff
@@ -119,11 +119,11 @@ to the tag attributes of the variables, as shown below:
...
     >>> theano.printing.debugprint(y)
-    Elemwise{identity} [id A] 'foo_apply_output'   
-     |Elemwise{add,no_inplace} [id B] ''   
-       |Elemwise{identity} [id C] 'foo_apply_input1'   
+    Elemwise{identity} [id A] 'foo_apply_output'
+     |Elemwise{add,no_inplace} [id B] ''
+       |Elemwise{identity} [id C] 'foo_apply_input1'
        | |i1 [id D]
-       |Elemwise{identity} [id E] 'foo_apply_input2'   
+       |Elemwise{identity} [id E] 'foo_apply_input2'
```
bartvm: Removing this will probably result in the doctests failing because Theano actually prints trailing whitespace.

docs/create_your_own_brick.rst
memimo: right, thanks.

docs/create_your_own_brick.rst
memimo: I see, my editor removed those, will put them back manually.

rizar:
While this is a good change, I guess this might be not the complete solution for the issue that we discussed in #1021. Suppose I create a new brick class B that inherits from a brick A. Suppose A has two children which it creates in the constructor, and B wants to add one more.  If B passes a `children` keyword argument to its `super().__init__` code, and A does the same, additional children added by B will never reach `Brick.__init__` before `Initializable.__init__` is called. 

One workaround is to make sure that all bricks handle children as follows:
```python
class MyBrick(Brick):
  def __init__(self):
      ...
      super(MyBrick, self).__init__(children=[a_child] + kwargs.get('children', []))
```

Another is to relax and postpone the sanity check for initialization schemes until `_push_initialization_config`. But on the other hand, we still need to know the roles of the children in order to fetch the initialization schemes from keyword arguments. What do you think?

memimo:
right, your suggested fix makes life easier for the initialization, but it raised the issue that people might forget doing this when designing a new brick. But that could be fixed with some documentation. Not sure where is the best place though. Perhaps a sentence in create_your_own_brick.rst

dwf:
Honestly, appending to children after a super call is not that big a deal!
It will be inevitable in some situations, better not to engineer
complicated solutions where simple ones will do.

>

memimo:
I did a git rebase to ammend to old commits, and now some old commits are showing up as part of my PR. Any idea how to avoid that?

dwf:
First refresh the page and make sure it's not just the dumb GitHub
javascript.

If yes, check if they show up in git log, if they do, git rebase -i on top
of the upstream master branch and delete the lines corresponding to the
duplicates, I guess.

But it's probably just GitHub dumb javascript.

dwf:
Yeah just checked, looks fine to me.

rizar:
@dwf, it ends up being a pretty big deal to us. If we want to fetch the
keyword arguments of the form '<role>_init' from __init__ keyword
arguments, we have to know what are the roles that we expect! We could of
course look for all possible roles, but that's kind of ugly.

Also, if one forgets to add the children from kwargs to the list of the
children that their brick creates, it will not be a big deal in 99% cases.
It can be an issue only if somebody inherits from that brick, adds a child,
passes an initialization scheme that should be propagated to the child
only, and this initialization scheme is not fetched in
`Initializable.__init__`. Still perhaps we should at least have the bricks
from the library respect this convention.

tests/extensions/test_saveload.py
```diff
@@ -12,50 +15,92 @@
...
-        algorithm=GradientDescent(cost=cost, parameters=[W]),
-        extensions=[Load('mynonexisting.tar')]
-    )
-    main_loop.extensions[0].main_loop = main_loop
-    main_loop._run_extensions('before_training')
-
-    # Cleaning
-    if os.path.exists('myweirdmodel.tar'):
-        os.remove('myweirdmodel.tar')
+class TestCheckpoint(unittest.TestCase):
```
rizar: This should not be a part of your PR.

rizar:
I like the code, but your history is broken. Perhaps the fast would be to start a new branch from master and cherry-pick the two commits that you made.

memimo:
Made a new pr #1029

dwf:
Continuing discussion here out of convenience:

> Also, if one forgets to add the children from kwargs to the list of the
> children that their brick creates, it will not be a big deal in 99% cases.
> It can be an issue only if somebody inherits from that brick, adds a child,
> passes an initialization scheme that should be propagated to the child
> only, and this initialization scheme is not fetched in
> `Initializable.__init__`. Still perhaps we should at least have the bricks
> from the library respect this convention.

Okay, I see the issue now. In this case, can we make sure that `Initializable` makes noise in a very specific, interpretable way if there are `_init` kwargs that don't correspond to anything in the role dict?

rizar:
I am not sure it possible, because some kwargs might be there for some other mixin brick. A usual unexpected argument error will happen if a role is not fetched by `Initializable`. Do you think it is not clear enough?

dwf:
I might still suggest adding adding a log.warning() call in this case. This
will have the dual effect of discouraging the use of `_init` names for
parameters that are not meant for `Initializable` and alerting the user to
the *actual* cause of the problem.

If the exception raised in that case were more specific than `TypeError` we
could simply catch that particular exception from the super() call. What we
could do is put the super call in a try/except TypeError, check if the
exception message contains 'unexpected keyword argument' and the name of
one of the suspicious init (enclosed in single quotes, as the Python
interpreter does, to stop false positives), and if it does, raise a more
helpful error message. If not, reraise the original exception.

But this is pretty involved and I think my original idea of issuing a
warning would be a good compromise.

PR #1029, opened 2016-03-14
Pass children as argument to Brick

### LINKED ISSUE #1027: It should possible to pass `children` as an argument to a Brick

memimo:
fixes #1027 

Same as #1028 with clean history


rizar:
Good job, thanks!

PR #1030, opened 2016-03-15
[WIP] Managing initilizations via a role scheme dictionary

### LINKED ISSUE #970: Metaissue: Improve initialization

memimo:
fixes #970
Re-attempt of #1021 
So far it only changes simple.py will propagate changes to other bricks when the format is approved.


blocks/bricks/interfaces.py
```diff
@@ -127,44 +127,55 @@ class Initializable(RNGMixin, Brick):
...
+        if self.initialization_schemes is None:
+            self.initialization_schemes = {}
+
+        kwargs_ = {}
+        for key in kwargs:
+            if key[-5:] == "_init":
+                if key in self.initialization_schemes:
+                    raise ValueError("All initializations are accepted either"
+                                     "through initialization_schemes or "
+                                     "correspodong attribute but not both")
```
dwf: typo

dwf: also s/attribute/keyword argument/

blocks/bricks/interfaces.py
```diff
@@ -127,44 +127,55 @@ class Initializable(RNGMixin, Brick):
...
         self.seed = seed
+        self.initialization_schemes = initialization_schemes
+        self.parameter_roles = set([])
+        if self.initialization_schemes is None:
+            self.initialization_schemes = {}
+
+        kwargs_ = {}
+        for key in kwargs:
+            if key[-5:] == "_init":
+                if key in self.initialization_schemes:
```
dwf: This looks wrong, unless you're expecting `initialization_schemes` keys to be strings of the form `rolename_init`? (I thought they were role objects).

blocks/bricks/interfaces.py
```diff
@@ -127,44 +127,55 @@ class Initializable(RNGMixin, Brick):
...
-        elif biases_init is not None or not use_bias:
-            raise ValueError("This brick does not support biases config")
         self.use_bias = use_bias
         self.seed = seed
+        self.initialization_schemes = initialization_schemes
+        self.parameter_roles = set([])
+        if self.initialization_schemes is None:
+            self.initialization_schemes = {}
+
+        kwargs_ = {}
```
dwf: I don't think there's any harm in deleting from the original dict as you go and passing it along.

blocks/bricks/interfaces.py
```diff
@@ -127,44 +127,55 @@ class Initializable(RNGMixin, Brick):
...
 
     def _push_initialization_config(self):
         for child in self.children:
+            if (isinstance(child, Initializable) and
+                    hasattr(child, 'initialization_schemes')):
+                for role in child.initialization_schemes:
+                    if role not in self.parameter_roles:
+                        raise ValueError("The parameter role: " +
+                                         "{} is not defined in".format(role) +
+                                         "in the class parameter_roles")
```
dwf: you can start this one at 'the'

Also you can continue strings without a +

```python
                        raise ValueError("The parameter role: "
                                         "{} is not defined in "
                                         "the class parameter_roles".format(role))
```
or something to that effect is valid.

blocks/bricks/interfaces.py
```diff
@@ -127,44 +127,55 @@ class Initializable(RNGMixin, Brick):
...
+        self._collect_roles()
 
     def _push_initialization_config(self):
         for child in self.children:
+            if (isinstance(child, Initializable) and
+                    hasattr(child, 'initialization_schemes')):
+                for role in child.initialization_schemes:
+                    if role not in self.parameter_roles:
+                        raise ValueError("The parameter role: " +
+                                         "{} is not defined in".format(role) +
```
dwf: missing a space after 'in'

blocks/bricks/interfaces.py
```diff
@@ -127,44 +127,55 @@ class Initializable(RNGMixin, Brick):
...
-            for child in self.children:
-                if (isinstance(child, Initializable) and
-                        hasattr(child, 'biases_init')):
-                    child.biases_init = self.biases_init
+                child.initialization_schemes = self.initialization_schemes
+
+    def _collect_roles(self):
+        for child in self.children:
+            if isinstance(child, Initializable):
+                self.parameter_roles.update(child.parameter_roles)
```
dwf: Should this not recurse?

blocks/bricks/interfaces.py
```diff
@@ -127,44 +127,55 @@ class Initializable(RNGMixin, Brick):
...
+                    raise ValueError("All initializations are accepted either"
+                                     "through initialization_schemes or "
+                                     "correspodong attribute but not both")
+                else:
+                    self.initialization_schemes[key] = kwargs[key]
+            else:
+                kwargs_[key] = kwargs[key]
+
+        super(Initializable, self).__init__(**kwargs_)
+        self._collect_roles()
```
dwf: Seems like this could be deferred until the top of `_push_initialization_config` with no ill effects, in which case it would give other code time to add children, etc.

blocks/bricks/interfaces.py
dwf: I guess it's not necessary if called from the constructor (though I think that might be suboptimal)

blocks/bricks/interfaces.py
```diff
@@ -196,8 +207,10 @@ def _initialize(self):
         # Use self.parameters[] references in case W and b are overridden
         # to return non-shared-variables.
         if self.use_bias:
-            self.biases_init.initialize(self.parameters[1], self.rng)
-        self.weights_init.initialize(self.parameters[0], self.rng)
+            self.initialization_schemes['biases_init'].initialize(
```
dwf: Same question about what the keys here ought to be. Also the minor nit that the role names are "BIAS" and "WEIGHT" but for the current init args we use the plural.

blocks/bricks/interfaces.py
memimo: yeah, the kes are rolenames_init and the values the initialization objects. "role objects" are you referring to the blocks.roles? I'm not using those. Roles here are just strings.

dwf:
Technically this is @rizar's review but hopefully my comments from the peanut gallery are a little helpful to at least one of you :stuck_out_tongue:

blocks/bricks/interfaces.py
memimo: If you're looping over dict elements and then delete, then python throws an error the the dict size has changed.

dwf: yeah you need `for key in list(kwargs.keys()):`

blocks/bricks/interfaces.py
memimo: yeah. not sure why put it here :p

memimo:
@rizar could you please take a look at the high level structure of the PR. I'll take care of the comments of dwf when the high-level is approved.

rizar:
Sure! Will work on it today.

rizar:
Sorry for the delay, I am struggling with the cluster and my course project. I will definitely review it tomorrow.

blocks/bricks/interfaces.py
rizar: I think you should use the roles from `blocks.roles` in `initialization_schemes`. Because those are the roles that the parameters have.

blocks/bricks/interfaces.py
rizar: Hmm, not sure that we should allow adding children after creation. I think that it makes development a lot easier if we assume that all the children must be specified in the constructor call. Do you have a usecase in mind in which this can not be done?

blocks/bricks/interfaces.py
rizar: I think that here you should just initialize each parameter with the scheme corresponding to its most specific role. This assumes that the keys in `initialization_schemes` are the roles. In this case for each parameter you can find the most specific role in `initialization_schemes` and use the scheme given for this roles to initialize the parameter. Note, that with such an approach one does not even have to implement `_initialize` method!

rizar:
I have three suggestions:

- use the roles as the keys in `initialization_schemes`
- initialize parameters using the scheme given for their most specific role
- check that the given initialization schemes match the parameter roles of the brick in two places: in the end of constructor and in `_push_initialization_config`. This will allow to detect errors as early as possible.

memimo:
I reorganized the PR a little bit. Again, let's discuss the generality of the PR first again, I will fix formatting stuff later.

Some notes:

1-The problem with having the roles as keys is that, roles are objects, and different objects of same class will be hashed to different values. So instead I use the name of the roles as the keys.
2-One major API change is that, if some one wants to manually change the schemes self.weights_init and self.biases_init is not valid anymore and one has to use self.initializations_schems
3-There is a tiny bit difference in the GRU weight initialization values now. I don't believe that will have any practical impact on performance, unless someone has an evidence against it.

rizar:
1. I think you gave up the idea of a role-indexed dictionary too early. Check https://docs.python.org/2/reference/datamodel.html#object.__hash__ , you should be able to overload this method.

2. I thought we would support `weights_init` and `biases_init` through `__getattr__`, why not?

dwf:
Indeed, 

```python
def __hash__(self):
    return hash(str(self))
```

should work in the case of roles.

blocks/bricks/interfaces.py
```diff
@@ -127,45 +128,97 @@ class Initializable(RNGMixin, Brick):
...
-        if use_bias is not None:
-            self.use_bias = use_bias
+        self.use_bias = use_bias
         self.seed = seed
+        self.initialization_schemes = initialization_schemes
+        self.parameter_roles = set([])
+        if self.initialization_schemes is None:
+            self.initialization_schemes = {}
+
+        initialization_to_role = {"weights_init": WEIGHT, 'biases_init': BIAS,
```
rizar: For the record, it looks like you have chosen to only handle a handful of `<role>_init` keyword arguments, not all of them. This is probably good, because this clearly makes `initialization_schemes` the primary way of defining the initialization schemes.

But the code clearly lacks a check `if key not in initialization_to_role`. Since now we have a clearly defined white list, I would probably raise a warning, that calling the arguments `smth_init` is recommended, but still pass this argument to the `super().__init__` call.

blocks/bricks/interfaces.py
```diff
@@ -127,45 +128,97 @@ class Initializable(RNGMixin, Brick):
...
+                                                                 self.rng)
+
+    def __getattr__(self, name):
+        if name == "weights_init":
+            if WEIGHT in self.initialization_schemes:
+                return self.initialization_schemes[WEIGHT]
+        elif name == "biases_init":
+            if BIAS in self.initialization_schemes:
+                return self.initialization_schemes[BIAS]
+        raise AttributeError("Attribute {} not found".format(name))
```
rizar: Here you should call `super(Initializable, self).__getattr__`.

blocks/bricks/interfaces.py
```diff
@@ -127,45 +128,97 @@ class Initializable(RNGMixin, Brick):
...
+                    hasattr(child, 'initialization_schemes')):
                 child.rng = self.rng
-                if self.weights_init:
-                    child.weights_init = self.weights_init
-        if hasattr(self, 'biases_init') and self.biases_init:
-            for child in self.children:
-                if (isinstance(child, Initializable) and
-                        hasattr(child, 'biases_init')):
-                    child.biases_init = self.biases_init
+                for role, scheme in self.initialization_schemes.items():
```
rizar: We should not propagate all the roles to the children, I think. It would not be nice, if even the bottom-most brick has initialization schemes for all the roles. We should only push those initialization schemes that can be useful for the child brick. Given that `push_initialization_config` can be called before `allocate`, we can not use `self.parameter_roles`, at least given the way you collect them.

rizar:
I think the major design decision that I disagree with is having both `initializable_roles` and `parameter_roles`. Neither of the two is fully suitable for controlling how the initialization schemes are spread down the hierarchy. The former is hardcoded, and the latter is known too late, after the parameters are created. 

Here is a different design which seems better to me

- each `Initilizable` brick has an attribute `parameter_roles` that contains all the roles that the brick's parameters **will have** when constructed
- the value of this attribute is defined as follows:
    - first, `Initializable` sets it to `[WEIGHT, BIAS]` (only `[WEIGHT]` is not `use_bias`)
    - ... unless the user provides `parameter_roles` as a constructor argument
    - then, after `super(Initializable, self).__init__`, `Initializable._collect_roles` is called to add
      the roles of the children to `parameter_roles`
- we are still in `Initializable.__init__`, but we can already check that `initialization_schemes` does not contain schemes for the roles that are not in `parameter_roles`
- furthermore, when the parameters are created (that is in `allocate`), we can check that all the parameters have roles that are present in `parameter_roles`
- when initialization configuration is pushed (`Initializable._push_initialization_config`), only the schemes for the relevant roles should be pushed to the children. One should be careful here to check that sometimes initialization schemes for roles that are parents of the precise parameter roles should also be pushed.
- in `initialize`, we can repeat the sanity check that we did in constructor, to make sure that user did not provide useless initialization schemes 

What do you think?

blocks/bricks/interfaces.py
```diff
@@ -127,45 +128,97 @@ class Initializable(RNGMixin, Brick):
...
+
+    def _collect_roles(self):
+        if hasattr(self, 'parameters'):
+            for param in self.parameters:
+                for role in param.tag.roles:
+                    if role in self.initializable_roles:
+                        self.parameter_roles.update(set([role]))
+
+    def _initialize(self):
+        for param in self.parameters:
```
rizar: This is a dangerous code. First, if a parameter has more than one role, it is not clear at all how it should be initialized. I would rather raise an exception upon encountering such a situation. Also, we should choose the most specific initialization scheme (e.g. the one for `RECURRENT_WEIGHT` and not for just `WEIGHT` if the parameter is `RECURRENT_WEIGHT`).

blocks/bricks/interfaces.py
```diff
@@ -127,45 +128,97 @@ class Initializable(RNGMixin, Brick):
...
+
+        for key in self.initialization_schemes:
+            if key not in self.initializable_roles:
+                raise ValueError("{} is not member of ".format(key) +
+                                 "initializable_roles")
+
+        super(Initializable, self).__init__(**kwargs)
+
+    def _validate_roles_schmes(self):
+        for role in self.parameter_roles:
```
rizar: This validation checks that initialization schemes are given for all parameters. It is very restrictive to run it in `_push_initialization_config` as you do it, because this method can be called in the middle of specifying initialization schemes for bricks. 

Also, we get this kind of validation for free when we actually try to initialize parameters. Instead, I would rather check that there no initialization schemes are provided that are clearly useless. This is often a sign of a bug, when something was useless was given as an argument.

rizar:
@memimo , we should resume working on this, and I really hope we can finish it until  I leave Montreal on June 8.

memimo:
@rizar could you take a look about the high level design choice?
Travis is falling on block_examples code, because some initilization values have changed. I will take care of that later, if the design is good.

blocks/bricks/interfaces.py
```diff
@@ -127,45 +128,105 @@ class Initializable(RNGMixin, Brick):
...
-            self.use_bias = use_bias
+    def __init__(self, initialization_schemes=None, parameter_roles=None,
+                 use_bias=True, seed=None, **kwargs):
+        self.use_bias = use_bias
         self.seed = seed
+        self.initialization_schemes = initialization_schemes
+        if self.initialization_schemes is None:
+            self.initialization_schemes = {}
+
+        if parameter_roles:
```
rizar: Hold on, haven't we decided to not have `parameter_roles` and instead collect roles the actual parameter variables?

memimo: I thought it was the other way :p I guess it was a confusion. I'll give it a try, the other way around

memimo:
@rizar  ok, so this version collect roles from parameter tags. Will rebase later on.

blocks/bricks/interfaces.py
```diff
@@ -127,45 +128,109 @@ class Initializable(RNGMixin, Brick):
...
+                                     "through initialization schemes or "
+                                     "corresponding attribute but not both")
+                else:
+                    self.initialization_schemes[initialization_to_role[
+                                                key]] = kwargs[key]
+                kwargs.pop(key)
+
+        super(Initializable, self).__init__(**kwargs)
+
+    def _validate_roles(self):
```
rizar: I guess here you propagate initialization schemes does the hierarchy of roles. But it seems to me that you do not choose the most specific role out of those for which an initialization scheme is available. E.g. if we have an init. scheme for `PARAMETER` and for `WEIGHT`, and now we are looking for an init. scheme for `RECURRENT_WEIGHT`, then we should always choose the scheme given for `WEIGHT`. In the code
below it seems like the choice will depend on the order of the dictionary.

rizar: Also, I find it a bit weird that you change `initialization_schemes`. This  might make debugging quite a bit hard. I would rather put the search for the most specific init. scheme directly in `initialize`.

blocks/bricks/interfaces.py
```diff
@@ -127,45 +128,109 @@ class Initializable(RNGMixin, Brick):
...
+                    if isinstance(role, type(key)):
+                        self.initialization_schemes[role] = \
+                                            self.initialization_schemes[key]
+                        high_level_roles.append(key)
+
+        for key in high_level_roles:
+            if key not in self.parameter_roles:
+                self.initialization_schemes.pop(key)
+
+        for key in self.initialization_schemes:
```
rizar: This seems to be wrong, because an exception can be raised in perfectly legitimate cases when e.g. `self.parameter_roles` has the only role `RECURRENT_WEIGHTS` and `key` is `WEIGHTS`.

I think the only when we can (and should) raise an exception is when an init. scheme is provided for a roles that is not a parent of any of `parameter_roles`.

memimo:
@rizar I updated based on your latest comments.

blocks/bricks/interfaces.py
```diff
@@ -127,45 +129,121 @@ class Initializable(RNGMixin, Brick):
...
+            return all_roles
+
+        self.parameter_roles = set(get_param_roles(self))
+        for child in self.children:
+            if isinstance(child, Initializable):
+                child._collect_roles()
+                self.parameter_roles.update(child.parameter_roles)
+
+    def _initialize(self):
+        def get_scheme(role, schemes):
```
rizar: It does not seem to me that this code will choose the most specific rolefor which there is an intialization scheme.

Here roughly what I think should work:

```python
def get_scheme(role, schemes):
   candidates = []
   for key in schemes:
       if isinstance(role, key):
           if all([isinstance(key, c) for c in candidates]):
               candidates = [c]
           else:
               candidates.append(c)
    if not candidates:
         raise ValueError("No init scheme")
    if len(candidates) > 1:
         raise ValueError("Multiple candidate init schemes")
```

memimo: I'm not sure your code snippet do that either. Let's assume we have this class hierarchy A->B->C
a = A(), b=B(), c=C()
Then:
isinstance(c, A)  -- True
type(c) == A -- False

rizar: Right, but all I am using is "isinstance", so what's wrong?

memimo: You're not going up the hierarchy. Let's say if we have in schemes both C and B. And ideally we should choose the scheme associated with C not B, your code does not do that.

PR #1035, opened 2016-03-19
Saner use_bias push behaviour (and tied_biases for ConvolutionalSequence)

dwf:
Push use_bias to children only if a value is explicitly specified
in the parent.  This mirrors the behaviour of `weights_init` and
`biases_init`.

This can be accomplished by either Initializable using a value of
`None` (and `Linear` and `ConvolutionalSequence`) special-casing
that value to mean True) or simply by not setting the attribute.
I've chosen the latter since it's more likely to raise an error
if it's not what you expect: `bool(None)` is `False`, and so this might
lead to subtle bugs and unintended behaviour. The relevant classes
instead use `getattr(self, 'use_bias', class_preferred_default)`.

**EDIT 2016/03/20:** It makes a lot of sense to adopt this policy for `tied_biases` in `ConvolutionalSequence` as well, so I've added that to the pull request.

### DIFF
--- a/blocks/bricks/conv.py
+++ b/blocks/bricks/conv.py
@@ -414,6 +414,10 @@ class ConvolutionalSequence(Sequence, Initializable, Feedforward):
         need to rely on either a default border mode (usually valid)
         or one provided at construction and/or after construction
         (but before allocation).
+    tied_biases : bool, optional
+        Same meaning as in :class:`Convolutional`. Defaults to ``None``,
+        in which case no value is pushed to child :class:`Convolutional`
+        bricks.
 
     Notes
     -----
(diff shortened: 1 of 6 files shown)


dwf:
@bartvm @dmitriy-serdyuk @rizar @vdumoulin Input greatly appreciated.

I think this has very little potential to break existing code but a lot of potential to save people from the situation I just found myself in, where for *weeks* I assumed that

```python
ConvolutionalSequence([
    Convolutional(..., use_bias=False),
    ...,
], image_size=(32, 32), num_channels=3)
```
was running without biases in the `Convolutional` bricks.

rizar:
Sorry for the delay, Dima S and I were in China with very crappy internet.

It seems to make a whole lot of sense to do it this way. LGTM

dwf:
Alright, great, thanks.

PR #1041, opened 2016-03-26
ConvolutionalTranspose: infer original_image_size.

dwf:
Transposed (strided) convolution is a one-to-many mapping, with zero padding of border pixels dropped by the original convolution being reconstructed in every case where `original_image_size[i] != filter_size[i] + step[i] * (image_size[i] - 1)`.

In the most common case, you want to produce an upsampled output image of which the transposed convolutional filter touches every pixel. This adds the automatic inference of the `original_image_size` with this case in mind. It also adds documentation on what constitutes a valid `original_image_size` for manual specification.

### DIFF
--- a/blocks/bricks/conv.py
+++ b/blocks/bricks/conv.py
@@ -186,6 +186,13 @@ class ConvolutionalTranspose(Convolutional):
         Image size of the input to the *transposed* convolution, i.e.
         the output of the corresponding convolution. Required for tied
         biases. Defaults to ``None``.
+    unused_edge : tuple, optional
+        Tuple of pixels added to the inferred height and width of the
+        output image, whose values would be ignored in the corresponding
+        forward convolution. Must be such that 0 <= ``unused_edge[i]`` <=
+        ``step[i]``. Note that this parameter is **ignored** if
+        ``original_image_size`` is specified in the constructor or manually
+        set as an attribute.
     original_image_size : tuple, optional
         The height and width of the image that forms the output of
         the transpose operation, which is the input of the original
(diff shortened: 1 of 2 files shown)


blocks/bricks/conv.py
```diff
@@ -190,19 +186,57 @@ class ConvolutionalTranspose(Convolutional):
...
+                 original_image_size=None, **kwargs):
         super(ConvolutionalTranspose, self).__init__(
             filter_size, num_filters, num_channels, **kwargs)
-        self.original_image_size = original_image_size
+        self._original_image_size = original_image_size
+
+    @property
+    def original_image_size(self):
+        if self._original_image_size is None:
+            if all(s is None for s in self.image_size):
```
vdumoulin: Should this be `any`?

blocks/bricks/conv.py
```diff
@@ -190,19 +186,57 @@ class ConvolutionalTranspose(Convolutional):
...
+                raise ValueError("can't infer original_image_size, "
+                                 "no image_size set")
+            last_edge = [d - s for d, s in zip(self.filter_size, self.step)]
+            tups = zip(self.image_size, self.step, last_edge)
+            return tuple(i * s + e for i, s, e in tups)
```
vdumoulin: This assumes a `'valid'` border mode, doesn't it?

blocks/bricks/conv.py
dwf: Well, it shouldn't happen that only one of the dimensions is None...

blocks/bricks/conv.py
dwf: You're right, of course. 9159d91 addresses other border modes (all of them are w.r.t. the original convolution, though, which I think is the most straightforward way).

dwf:
Hmm, it seems I'm in a particular situation where it'd be nice to be able to pad the convolution with the 'a' parameter in your tech report... How would you feel about replacing the `original_image_size` argument with a parameter to that effect? That would eliminate redundancy and give us full functionality in the case where we don't quite know the output shape but we know how far it is from what we want (think parameterized generation of transposed conv layers).

dwf:
@vdumoulin I think this is ready for a second look.

blocks/bricks/conv.py
vdumoulin: Ok

vdumoulin:
LGTM.

PR #1043, opened 2016-03-29
information about dimensionality for apply in rnn tutorial

### LINKED ISSUE #1040: input argument to lstm.apply dimensionality
In the docs to the LSTM's [`apply` function](http://blocks.readthedocs.org/en/latest/api/bricks.html?highlight=LSTM#blocks.bricks.recurrent.LSTM.apply) one can read "inputs (TensorVariable) – The 2 dimensional matrix of inputs in the shape (batch_size, features * 4)".
As far as I understand and see in some examples, it is also possible (and reasonable) to pass 3-dimensional matrix here of shape (batch_size, length_of_the_sequence, features * 4).
1. Am I right? Will it do the expected thing, i.e. pass first element of the sequence (vector of length features) to the LSTM, output something and change hidden state, then pass the second element of the sequence to that changed state, etc.?
2. If so, do you think it's a good idea to put that information to the docs? I spent some time today trying to figure it out, and still am unsure. I can do the change.

sygi:
Currently the dimensionality of the tensor passed to `apply` is not well documented anywhere.

fixes #1040

### DIFF
--- a/docs/rnn.rst
+++ b/docs/rnn.rst
@@ -168,8 +168,13 @@ Iterate (or not)
 ----------------
 
 The ``apply`` method of a recurrent brick accepts an ``iterate`` argument,
-which defaults to ``True``. Setting it to ``False`` causes the ``apply`` method
-to compute only one step in the sequence.
+which defaults to ``True``. It is the reason for passing above a tensor of one
+more dimension than described in :meth:`.recurrent.SimpleRecurrent.apply` - the
+extra first dimension corresponds to the length of the sequence we are iterating
+over.
+
+Setting ``iterate`` to ``False`` causes the ``apply`` method to compute only
+one step in the sequence.
 
 This is very useful when you're trying to combine multiple recurrent layers in
 a network.


docs/rnn.rst
```diff
@@ -168,8 +168,12 @@ Iterate (or not)
 ----------------
 
 The ``apply`` method of a recurrent brick accepts an ``iterate`` argument,
-which defaults to ``True``. Setting it to ``False`` causes the ``apply`` method
-to compute only one step in the sequence.
+which defaults to ``True``. It is the reason for passing a 3-dimensional tensor
```
dmitriy-serdyuk: This is true in this case: for recurrent models which transform states (input and output continuous features), the dimensionality is TxBxF. But recurrent decorator can be used for generative nets, like `SequenceGenerator`, in this case the outputs are integers (indexes) and have only 2 dimensions TxB.

rizar: Yep, the input vector must normally have 1 dimension more that `apply` functions expects, but it does not have to be 3-dimensional.

rizar:
Just noticed that you finished this PR, thanks!

PR #1047, opened 2016-04-01
Fix some None-handling in ConvolutionalSequence for when image size is unspecified

rightaditya:
`ConvolutionalSequence` breaks when using it with layers that don't have the image size specified. The reason is due to a missing `None`-check (prior to setting the image size for a layer) and a `None`-check that checks whether the image size is `None`, whereas the `Convolutional` and `Pooling` classes both use `tuple`s of `None`.

### DIFF
--- a/blocks/bricks/conv.py
+++ b/blocks/bricks/conv.py
@@ -440,8 +440,9 @@ class ConvolutionalSequence(Sequence, Initializable, Feedforward):
 
     """
     @lazy(allocation=['num_channels'])
-    def __init__(self, layers, num_channels, batch_size=None, image_size=None,
-                 border_mode=None, tied_biases=None, **kwargs):
+    def __init__(self, layers, num_channels, batch_size=None,
+                 image_size=(None, None), border_mode=None, tied_biases=None,
+                 **kwargs):
         self.layers = layers
         self.image_size = image_size
         self.num_channels = num_channels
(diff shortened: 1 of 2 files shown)


dwf:
Hmm. Technically, an image size of `(None, None)` should be valid as long as you're using `tied_biases=True`. This gives you a ConvolutionalSequence that can support different sized images (although Theano is breaking when I try that right now, it shouldn't).

There is some funny business though, because a value of `None` is getting pushed. I think the right fix is just to change the default argument for `image_size` in the `ConvolutionalSequence` constructor to `(None, None)`.

rightaditya:
Yeah, I think the problem is that some parts of the code forget that `(None, None)` is what we want to be working with, and assumes `None` instead.

I do prefer your solution of setting the default argument for `image_size` to `None, None)`, especially since that'd be more consistent with how the `Convolutional` constructor does it, but that would only address the issue at line 483. The line at 494 would still need the change.

I've made the relevant change and pushed a new commit.

dwf:
Cool. Only thing missing now is a regression test. Something like

```
def test_convolutional_sequence_no_image_size():
    try:
        # code that used to raise an exception
    except ExceptionThatWasRaised:
        assert False, "This should have succeeded"
```

should do it. May as well rebase out 11b2200 too if you feel like it. Thanks for starting the PR.

rightaditya:
Hmm, it looks like in writing the test I've uncovered another issue that goes a bit deeper, and isn't in ConvolutionalSequence. I left out the particular test line that triggers it in the commit here but I'll open an issue for it as it may require a bit more discussion than this one.

dwf:
Thanks, test looks good! I'll have a look at the other ticket in a bit, it's been a long day and I don't really have the bandwidth to think carefully about it right now. :sleepy:

PR #1050, opened 2016-04-04
Fix Pooling output dimensions w/ unspecified input dimensions

### LINKED ISSUE #1048: Pooling classes can't return output dimension with unspecified image size
The `Pooling` classes should be able to provide output dimensions when the image size is unspecified for use with `ConvolutionalSequence`, but this currently raises an exception:

```python
>>> from blocks.bricks.conv import MaxPooling
>>> m = MaxPooling(1,1)
>>> m.get_dim('output')
  File "<stdin>", line 1, in <module>
  File "/usr/local/lib/python3.5/dist-packages/blocks/bricks/conv.py", line 289, in get_dim
    ignore_border=self.ignore_border, padding=self.padding))
  File "/usr/local/lib/python3.5/dist-packages/theano/tensor/signal/pool.py", line 189, in out_shape
    r += padding[0] * 2
TypeError: unsupported operand type(s) for +=: 'NoneType' and 'int'
```

This is a relatively simple fix---the `Pooling.get_dim()` could just do a check---but it looks like it was written with the same idea as `Convolutional`---that `None` could be handled by just passing it down to Theano. Unfortunately it looks like the `Pool.out_shape()` in Theano doesn't like that. I've opened this issue to see if others would prefer trying to rely on Theano to make the change---if just doing a check in `Pooling.get_dim()` makes more sense, I'd be happy to submit a PR for it or include it in #1047.

rightaditya:
The `Pooling` classes break when trying to provide output dimensions when the input dimensions are unspecified. This is particularly important for use with `ConvolutionalSequence`. This fixes that with a simple check on the input dimensions rather than passing the `None`s down to Theano's `Pool.out_shape()`, which won't handle them. Also included is a regression test, as well as an addition to the `ConvolutionalSequence` regression test.

Fixes #1048.

### DIFF
--- a/blocks/bricks/conv.py
+++ b/blocks/bricks/conv.py
@@ -335,7 +335,7 @@ class Pooling(Initializable, Feedforward):
         if name == 'input_':
             return self.input_dim
         if name == 'output':
-            if None in self.input_dim[-2:]:
+            if self.input_dim[-2:] == (None, None):
                 # if input dimensions are unspecified, return similarly
                 # unspecified output dimensions (Theano's Pool.out_shape()
                 # won't take Nones)
--- a/blocks/bricks/conv.py
+++ b/blocks/bricks/conv.py
@@ -335,9 +335,15 @@ class Pooling(Initializable, Feedforward):
         if name == 'input_':
             return self.input_dim
         if name == 'output':
-            return tuple(Pool.out_shape(
-                self.input_dim, self.pooling_size, st=self.step,
-                ignore_border=self.ignore_border, padding=self.padding))
+            if None in self.input_dim[-2:]:
+                # if input dimensions are unspecified, return similarly
+                # unspecified output dimensions (Theano's Pool.out_shape()
+                # won't take Nones)
+                return self.input_dim
+            else:
+                return tuple(Pool.out_shape(
+                    self.input_dim, self.pooling_size, st=self.step,
+                    ignore_border=self.ignore_border, padding=self.padding))
 
     @property
     def num_output_channels(self):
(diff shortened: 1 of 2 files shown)


blocks/bricks/conv.py
```diff
@@ -335,9 +335,15 @@ def get_dim(self, name):
         if name == 'input_':
             return self.input_dim
         if name == 'output':
-            return tuple(Pool.out_shape(
-                self.input_dim, self.pooling_size, st=self.step,
-                ignore_border=self.ignore_border, padding=self.padding))
+            if None in self.input_dim[-2:]:
```
dwf: Now I'm slightly worried about the case where only one of the dimensions is None. Hm...

rightaditya: That occurred to me too, but when I took a look at Theano's code, it turns out that if either of the dimensions is None, it'll break.

dwf: What happens to subsequent layers if they get only one None in their output_dim though? The thing that concerns me is that the non-None value will be incorrect.

rightaditya: Hmm. Is having only one None in the input_dim a supported case? Taking a quick glance through the code, it looks like the check is always to see if it is equal to `(None, None)`. I can definitely change this code to reflect that. But it looks like if, say, the `Convolutional` class is given input dimensions where exactly one dimension is `None`, it'd break, since it wouldn't pass through the `(None, None)` check, and then would have some operations done on it that wouldn't work with a `None` value. So it seems unlikely that the `Pooling` class would ever be given such an input, but if that case were to be supported, it'd require similar changes to other classes as well, such as `Convolution`.

rightaditya: Okay, I've changed the comparison to check against `(None, None)` for now. If you're considering making some larger changes for how this case is handled (like in `Convolutional`, `ConvolutionalSequence`, etc.), I think it'd still be good to have `Pooling` work in this case for the interim.

Issue #1044, opened 2016-03-29
Add log.debug() statements for push_allocation_config and push_initialization_config


dwf:
@bartvm pointed out that it can be very helpful to have a trace of what is pushing initialization/allocation config where. The best way to do this is to just add DEBUG-level logging statements, that the user can turn on by altering the log level.

This should be split into sub-tickets for various brick modules, probably.

PR #1054, opened 2016-04-06
More logging for checkpointing

rizar:
Checkpointing is so often a source of daunting speed issues, that I think it makes sense to have a few logging statements directly in Blocks source code.

### DIFF
--- a/blocks/serialization.py
+++ b/blocks/serialization.py
@@ -113,6 +113,7 @@ import six
 import tarfile
 import tempfile
 import warnings
+import logging
 
 from contextlib import closing
 from pickle import HIGHEST_PROTOCOL
@@ -132,6 +133,8 @@ from blocks.config import config
 from blocks.filter import get_brick
 from blocks.utils import change_recursion_limit
 
+
+logger = logging.getLogger(__name__)
 
 BRICK_DELIMITER = '|'
 MAIN_MODULE_WARNING = """WARNING: Main loop depends on the function `{}` in \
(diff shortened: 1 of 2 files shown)


rizar:
Review required.

blocks/serialization.py
```diff
@@ -215,10 +218,13 @@ def secure_dump(object_, path, dump_function=dump, **kwargs):
 
     """
     try:
+        logger.debug("Dump object to a temporaty file")
```
dwf: I'd say "Dumping" as a description of what's being done. "Dump" sounds like a command.

blocks/serialization.py
```diff
@@ -215,10 +218,13 @@ def secure_dump(object_, path, dump_function=dump, **kwargs):
...
         with tempfile.NamedTemporaryFile(delete=False,
                                          dir=config.temp_dir) as temp:
             dump_function(object_, temp, **kwargs)
+        logger.debug("Move the temporary file")
```
dwf: Ditto for moving/move.

blocks/serialization.py
dwf: Also, "temporaty" -> "temporary"

dwf:
Would it make sense to make these `log.debug` calls? I've suggested we should use them a lot more liberally for things that you sometimes want to see but often don't (#1044 for example).

dwf:
Actually, I'd leave the checkpointing started/finished ones at INFO level, and the others at DEBUG, personally. The latter are more fine-grained.

rizar:
Aren't the current logging levels the same as you want?

blocks/serialization.py
rizar: http://dictionary.cambridge.org/dictionary/english/dump - an ​act of ​moving ​information from a computer's ​memory to another ​place or ​device. I was also not sure, but the dictionary says it's ok.

rizar: Oh, sorry, here it actually makes more sense to use "dumping". I meant that "dump is finished" is probably proper English.

dwf:
Oops. Yes. Just the typos then.

rizar:
I fixed the typos.

dwf:
Great, thanks.

PR #1055, opened 2016-04-06
Uninfy LSTM arguments

dmitriy-serdyuk:
Allow to change LSTM gate activation.


blocks/bricks/recurrent.py
```diff
@@ -371,12 +371,18 @@ class LSTM(BaseRecurrent, Initializable):
 
     """
     @lazy(allocation=['dim'])
-    def __init__(self, dim, activation=None, **kwargs):
+    def __init__(self, dim, activation=None, gate_activation=None, **kwargs):
```
dwf: Docstring needs an update.

dmitriy-serdyuk: 459024409755b6448d62e85179fa6dedd55dab4f

PR #1057, opened 2016-04-07
Add links to different recurrent models

dmitriy-serdyuk:
I noticed, that some people in the lab do not know about existence of bidirectional recurrent models or recurrent stack implemented in blocks. I hope, adding the link to RNN tutorial will help.

Also, Read the docs very bad at searching. For example, a search for [LSTM](http://blocks.readthedocs.org/en/latest/search.html?q=LSTM&check_keywords=yes&area=default#) doesn't really find it.


rizar:
LGTM, but with the prerequisite that you compile it locally and check that it builds nicely.

dmitriy-serdyuk:
It compiles.

rizar:
And did you try following the links that you added?

dmitriy-serdyuk:
Sure!

PR #1058, opened 2016-04-07
Add links to the code in docs

dmitriy-serdyuk:
Borrowed this piece of code from numpy and modified the library name.

It works when I build it locally.

### DIFF
--- a/docs/conf.py
+++ b/docs/conf.py
@@ -12,6 +12,7 @@
 # All configuration values have a default; values that are commented out
 # serve to show the default.
 
+import inspect
 import sys
 import os
 from mock import Mock as MagicMock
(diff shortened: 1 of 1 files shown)


dmitriy-serdyuk:
This code is mostly copy-paste of [this](https://github.com/numpy/numpy/blob/master/doc/source/conf.py#L286) function in numpy. It creates a link to a corresponding file, line number at github using inspection mechanisms.

rizar:
But I don't understand where and by whom this `linkcode_resolve` is called.

dmitriy-serdyuk:
It is called by `sphinx.ext.linkcode` sphinx [module](http://www.sphinx-doc.org/en/stable/ext/linkcode.html) every time it creates a documentation item for a class/function/etc. `info` argument is a dictionary which contains module name, full path and maybe other stuff which is not very well documented. 

Here, I'm using the path to get inspection information: full name and line number where it is defined in a file.

rizar:
OK, I see. Nice new feature for our docs! If it works for you locally, I do not mind merging.

dmitriy-serdyuk:
Let's try, I'll check read the docs.

dmitriy-serdyuk:
Cool, it works!

PR #1062, opened 2016-04-18
Speed up BatchNormalization training graphs

dwf:
Calculating mean and variance for `BatchNormalization` separately with `.mean()` and `.var()` seems to be causing an explosion in the number of `CAReduce` nodes.

This change takes the following approach:
 * Calculate mean with `.mean()`
 * Calculate variance by subtracting the square of the above-calculated mean from `tensor.sqr(input_).mean()`
 * Add stabilizing constant, take square root.

Apparently this was enough to get @rizar a ~33% speedup, such that full batch norm is now approximately as fast as mean-only batch norm.

@rizar promised to paste his profiles with/without into the comments tomorrow so we can preserve the rationale for this change.

@nouiz @lamblin you might be interested in this (from the perspective of making Theano be less dumb about `mean()` and `var()` calls on the same variable). @cooijmanstim also.

### DIFF
--- a/blocks/bricks/bn.py
+++ b/blocks/bricks/bn.py
@@ -190,8 +190,10 @@ class BatchNormalization(RNGMixin, Feedforward):
         if self.mean_only:
             stdev = tensor.ones_like(mean)
         else:
-            stdev = tensor.sqrt(tensor.var(input_, axis=axes, keepdims=True) +
-                                numpy.cast[theano.config.floatX](self.epsilon))
+            var = (tensor.sqr(input_).mean(axis=axes, keepdims=True) -
+                   tensor.sqr(mean))
+            eps = numpy.cast[theano.config.floatX](self.epsilon)
+            stdev = tensor.sqrt(var + eps)
             assert (stdev.broadcastable[1:] ==
                     self.population_stdev.broadcastable)
             add_role(stdev, BATCH_NORM_MINIBATCH_ESTIMATE)


nouiz:
Was the problem also happening while cloning the graph? I'm pretty sure the problem is related to cloning, I have see it sell optimized in other code without cloning.

dwf:
So @rizar told me that the speedup is not that big after all, unfortunately. I don't know whether he was using cloning or not, as there are two interfaces for batch normalizing a graph, one with cloning and one without.

rizar:
Sorry guys, it was a mistake on my side: David's new code is just as slow as the version from the master branch. I do not use cloning, I use the context manager.

Still, batch normalization is terribly expensive for me. Here is a profile for theano function that does validation for me. All these element-wise computations end up being quite costly for me:

```
Class
---
<% time> <sum %> <apply time> <time per call> <type> <#call> <#apply> <Class name>
  55.6%    55.6%      13.665s       8.54e-05s     C   160000      32   theano.sandbox.cuda.basic_ops.GpuElemwise
  27.6%    83.3%       6.785s       1.70e-04s     C    40000       8   theano.sandbox.cuda.dnn.GpuDnnConv
   5.4%    88.6%       1.319s       2.64e-04s     C     5000       1   theano.sandbox.cuda.basic_ops.GpuJoin
   3.1%    91.8%       0.762s       7.62e-05s     C    10000       2   theano.sandbox.cuda.blas.GpuDot22
   2.6%    94.3%       0.632s       3.16e-05s     C    20000       4   theano.sandbox.cuda.dnn.GpuDnnPool
   1.0%    95.4%       0.257s       1.28e-05s     C    20000       4   theano.sandbox.cuda.basic_ops.GpuFromHost
   0.7%    96.1%       0.179s       5.96e-06s     Py   30000       4   theano.ifelse.IfElse
   0.7%    96.8%       0.160s       1.06e-05s     C    15000       3   theano.sandbox.cuda.basic_ops.HostFromGpu
   0.6%    97.4%       0.157s       1.57e-05s     C    10000       2   theano.compile.ops.DeepCopyOp
   0.5%    97.9%       0.115s       2.30e-05s     Py    5000       1   theano.tensor.nnet.nnet.CrossentropyCategorical1Hot
   0.3%    98.2%       0.084s       3.91e-07s     C   215000      43   theano.sandbox.cuda.basic_ops.GpuDimShuffle
   0.3%    98.5%       0.073s       1.45e-05s     C     5000       1   theano.sandbox.cuda.nnet.GpuSoftmaxWithBias
   0.2%    98.7%       0.059s       1.18e-05s     C     5000       1   theano.sandbox.cuda.basic_ops.GpuCAReduce
   0.2%    99.0%       0.058s       1.16e-05s     C     5000       1   theano.tensor.basic.MaxAndArgmax
   0.2%    99.2%       0.055s       3.03e-07s     C   180000      36   theano.compile.ops.Shape_i
   0.2%    99.4%       0.050s       1.26e-06s     C    40000       8   theano.sandbox.cuda.basic_ops.GpuAllocEmpty
   0.2%    99.6%       0.045s       3.33e-07s     C   135000      27   theano.tensor.elemwise.Elemwise
   0.1%    99.7%       0.026s       1.03e-06s     C    25000       5   theano.sandbox.cuda.basic_ops.GpuSubtensor
   0.1%    99.8%       0.019s       1.86e-07s     C   100000      20   theano.sandbox.cuda.basic_ops.GpuContiguous
   0.1%    99.8%       0.017s       4.31e-07s     C    40000       8   theano.sandbox.cuda.dnn.GpuDnnConvDesc
   ... (remaining 4 Classes account for   0.16%(0.04s) of the runtime)

Ops
---
<% time> <sum %> <apply time> <time per call> <type> <#call> <#apply> <Op name>
  28.7%    28.7%       7.038s       1.76e-04s     C     40000        8   GpuElemwise{Composite{(((((i0 + i1) - i2) / i3) * i4) + i5)}}[(0, 0)]
  27.6%    56.3%       6.785s       1.70e-04s     C     40000        8   GpuDnnConv{algo='small', inplace=True}
  15.4%    71.7%       3.787s       8.42e-05s     C     45000        9   GpuElemwise{Switch}[(0, 0)]
  10.1%    81.8%       2.486s       5.52e-05s     C     45000        9   GpuElemwise{Composite{Cast{float32}(GT(i0, i1))},no_inplace}
   5.4%    87.2%       1.319s       2.64e-04s     C     5000        1   GpuJoin
   3.1%    90.3%       0.762s       7.62e-05s     C     10000        2   GpuDot22
   2.6%    92.9%       0.632s       3.16e-05s     C     20000        4   GpuDnnPool{mode='max'}
   1.0%    93.9%       0.257s       1.28e-05s     C     20000        4   GpuFromHost
   0.7%    94.7%       0.179s       5.96e-06s     Py    30000        4   if{inplace,gpu}
   0.7%    95.3%       0.160s       1.06e-05s     C     15000        3   HostFromGpu
   0.6%    95.9%       0.157s       1.57e-05s     C     10000        2   DeepCopyOp
   0.5%    96.5%       0.128s       1.28e-05s     C     10000        2   GpuElemwise{Add}[(0, 0)]
   0.5%    96.9%       0.115s       2.30e-05s     Py    5000        1   CrossentropyCategorical1Hot
   0.4%    97.4%       0.109s       1.09e-05s     C     10000        2   GpuElemwise{Add}[(0, 1)]
   0.3%    97.7%       0.073s       1.45e-05s     C     5000        1   GpuSoftmaxWithBias
   0.3%    98.0%       0.069s       1.38e-05s     C     5000        1   GpuElemwise{Composite{((i0 / i1) + i2)}}[(0, 0)]
   0.2%    98.2%       0.059s       1.18e-05s     C     5000        1   GpuCAReduce{add}{1,0}
   0.2%    98.4%       0.058s       1.16e-05s     C     5000        1   MaxAndArgmax
   0.2%    98.6%       0.053s       3.29e-07s     C     160000       32   GpuDimShuffle{x,0,1,2}
   0.2%    98.9%       0.050s       1.26e-06s     C     40000        8   GpuAllocEmpty
   ... (remaining 29 Ops account for   1.15%(0.28s) of the runtime)
```

rizar:
For the record, the profiles are in /data/lisatmp4/bahdanau/catsdogs{david_full_batch_norm.prf, master_full_batch_norm.prf,master_mean-only_batch_norm.prf}. David's version has more ops but is slightly faster (<10%).

One very interesting observation is that forward propagation takes the same time for mean-only and full batch norm, whereas the backward pass is quite a bit slower for the full one.

nouiz:
Someone wrapped BN from cudnn v4 in this branch:

https://github.com/Theano/Theano/pull/4237

As cudnn v5 change its interface, I think we should only merge the v5
interface in Theano. Do one of you have time to work on this?

dwf:
@rizar 10% isn't bad for 4 lines. I had an idea for speeding it up a little further, by doing only one reduction. Would you mind running the same code you benchmarked with b589f5d?

If this isn't any better (or is worse) then I say I rollback to 8150b10 and we merge that, since a 10% boost is better than nothing.

The CuDNN ops scare me a little, also I don't totally understand how much gain could be had since it seems like you still need to do the mean/variance reduction outside the op, is that the case @nouiz? The docs are not clear at all.

nouiz:
On Wed, Apr 20, 2016 at 2:30 PM, David Warde-Farley <
notifications@github.com> wrote:

> @rizar <https://github.com/rizar> 10% isn't bad for 4 lines. I had an
> idea for speeding it up a little further, by doing only one reduction.
> Would you mind running the same code you benchmarked with b589f5d
> <https://github.com/mila-udem/blocks/commit/b589f5dcfbbdad56ada1e539e19226a72f130ac3>
> ?
>
> If this isn't any better (or is worse) then I say I rollback to 8150b10
> <https://github.com/mila-udem/blocks/commit/8150b102e3268713c129eaf609b81e7e720b0623>
> and we merge that, since a 10% boost is better than nothing.
>
> The CuDNN ops scare me a little, also I don't totally understand how much
> gain could be had since it seems like you still need to do the
> mean/variance reduction outside the op, is that the case @nouiz
> <https://github.com/nouiz>? The docs are not clear at all.
>

CuDNN ops was more difficulty when we introduced it in Theano. But now it
work well. I'm pretty sure it does the reduction itself. Otherwise it would
miss optimization opportunity.

There is a version working with cudnn v4. So the need is mostly understand
what is the difference with v5 and do the corresponding change.

dwf:
Okay, so it looks like it does a combined

* Estimate current batch mean & variance
* Produce batch-normalized outputs
* Estimate update exponential moving average mean & variance (the current value of which is an input to the Op)

But I think those Theano ops are nowhere near being merged given the current state of the PR, and I don't have the spare time to work on it. In addition, it would be nice to have a high-level interface that produces an ~equivalent CPU graph: we have refrained from directly introducing GPU ops in Blocks so far and it would be a shame if we had to give that up now.

dwf:
My quick timing of the batch norm example:

* master: 0.646s/epoch
* b589f5d: 0.686s/epoch (damn)
* 8150b10: 0.619s/epoch

dwf:
I'm going to guess that the rank ordering of Dima's example would be roughly the same and merge the version from earlier this week, which still represents a modest speedup.

dwf:
@rizar If you want to give it a try anyway, dwf/blocks@possibly_speedup_batch_norm contains the alternative I was trying out.

PR #1065, opened 2016-04-19
Addition of Leaky Rectifier ReLU

isacarnekvist:


blocks/bricks/simple.py
```diff
@@ -284,6 +284,28 @@ def apply(self, input_):
         return tensor.switch(input_ > 0, input_, 0)
 
 
+class LeakyRectifier(Activation):
+    r"""A linear transformation with optional bias.
```
dmitriy-serdyuk: Seems like copy-paste from somewhere else. It's totally not true.

I think, documentation needs some citations.

dmitriy-serdyuk:
Thank you! But I think it might be better to place this to [blocks-extras](https://github.com/mila-udem/blocks-extras/) since it's not so much widely used as ReLU.

I made some comments according to documentation. Also, this code needs some tests.

dwf:
It's becoming quite widely used actually. I'd agree for parametric ReLU,
but I see ReLUs with a fixed leak more and more often.

isacarnekvist:
I will add some tests for activation and gradients during the day as well. @dwf When you say parametric ReLU, do you mean the above simple LeakyReLU or the one where you can include a, and b from 

`f(x) = max(ax, bx)`

in the gradient so that you can learn these during training as well?

blocks/bricks/simple.py
```diff
@@ -284,6 +284,29 @@ def apply(self, input_):
         return tensor.switch(input_ > 0, input_, 0)
 
 
+class LeakyRectifier(Activation):
+    r"""Like Rectifier, but instead does:
```
dmitriy-serdyuk: Please take a look at [numpy docstring format](https://github.com/numpy/numpy/blob/master/doc/HOWTO_DOCUMENT.rst.txt).

It should one line short description, than a long description.

blocks/bricks/simple.py
```diff
@@ -284,6 +284,29 @@ def apply(self, input_):
...
+
+    .. math:: f(x) = \text{max}(x, ax)
+
+    Parameters
+    ----------
+    leak, optional : float
```
dmitriy-serdyuk: See a link above, the syntax for parameters is different.

dmitriy-serdyuk:
Ok, looks good to me.

PR #1066, opened 2016-04-19
Allow empty config file

dmitriy-serdyuk:
`yaml` returns `None` when parses an empty string, standard behaviour should be an empty dict in our case.

### DIFF
--- a/tests/test_config.py
+++ b/tests/test_config.py
@@ -47,14 +47,5 @@ class TestConfig(object):
         config.new_config = 'new_config'
         assert config.new_config == 'new_config'
 
-
-class TestEmptyConfig(object):
-    def setUp(self):
-        self._environ = dict(os.environ)
-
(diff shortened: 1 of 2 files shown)


rizar:
Could you please make a test for that?

dmitriy-serdyuk:
Done.

dmitriy-serdyuk:
@rizar , is it better?

tests/test_config.py
```diff
@@ -36,6 +46,15 @@ def test_config():
         assert config.data_path == 'manual_path'
         config.new_config = 'new_config'
         assert config.new_config == 'new_config'
-    finally:
+
+
+class TestEmptyConfig(object):
```
rizar: But why not merge two tests?

dmitriy-serdyuk:
Oops, forgot about this PR. Fixed.

PR #1069, opened 2016-04-22
Support context_name in serialization.py

rizar:
@abergeron , @nouiz , this is the first version that can load both `GpuArray` and `cuda_ndarray`. Will be continued on Monday (sorry for not finishing today, it took me too long to install libgpuarray)

### DIFF
--- a/blocks/serialization.py
+++ b/blocks/serialization.py
@@ -195,8 +195,8 @@ def dump(object_, file_, parameters=None, use_cpickle=False,
             for name, p in named_parameters.items():
                 array_ = p.container.storage[0]
                 context_name = (p.context_name
-                                if pygpu
-                                and isinstance(p, pygpu.gpuarray.GpuArray)
+                                if pygpu and
+                                isinstance(p, pygpu.gpuarray.GpuArray)
                                 else None)
                 external_objects[id(array_)] = _mangle_parameter_name(
                     type(array_), context_name, name)
@@ -552,11 +552,14 @@ class _Renamer(object):
         self.used_names.add(name)
         return name
 
+
 def _recreate_numpy_ndarray(_, content):
     return numpy.array(content)
 
+
 def _recreate_cuda_ndarray(_, content):
     return cuda_ndarray.cuda_ndarray.CudaNdarray(content)
+
 
 def _recreate_pygpu_array(context_name, content):
     context = theano.sandbox.gpuarray.get_context(context_name)
--- a/blocks/serialization.py
+++ b/blocks/serialization.py
@@ -114,7 +114,6 @@ import tarfile
 import tempfile
 import warnings
 import logging
-
 from contextlib import closing
 from pickle import HIGHEST_PROTOCOL
 try:
@@ -123,12 +122,17 @@ try:
 except ImportError:
     DEFAULT_PROTOCOL = HIGHEST_PROTOCOL
     from pickle import Pickler as _Pickler
+
 from six.moves import cPickle
+import theano
 try:
     from theano.sandbox.cuda import cuda_ndarray
 except ImportError:
     cuda_ndarray = None
-
+try:
+    import pygpu
+except:
+    pygpu = None
 from blocks.config import config
 from blocks.filter import get_brick
 from blocks.utils import change_recursion_limit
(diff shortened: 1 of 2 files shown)


rizar:
This seems to be done. I will run the tests locally with both old and new backend. It would be nice to add a backward compatibility test  with the old serialization format, but  I can not think of an easy way of doing so.

rizar:
@nouiz , here is the pull request.

blocks/serialization.py
```diff
@@ -188,10 +192,13 @@ def _save_parameters(f):
...
                               for n, p in named_parameters.items()})
-            for n, p in named_parameters.items():
+            for name, p in named_parameters.items():
                 array_ = p.container.storage[0]
+                context_name = (p.context_name
+                                if isinstance(p, pygpu.gpuarray.GpuArray)
+                                else None)
                 external_objects[id(array_)] = _mangle_parameter_name(
-                    type(array_), n)
+                    type(array_), context_name, name)
```
nouiz: would it be safer to keep the order as it was to don't break old pickled file?

blocks/serialization.py
```diff
@@ -349,7 +356,7 @@ def add_to_dump(object_, file_, name, parameters=None, use_cpickle=False,
                 parameters = numpy.load(
                     tar_file.extractfile(tar_file.getmember('_parameters')))
                 s1 = set(parameters.keys())
-                s2 = [_unmangle_parameter_name(x)[1] for x in
+                s2 = [_unmangle_parameter_name(x)[2] for x in
```
nouiz: if we do this change, undo this.

blocks/serialization.py
```diff
@@ -544,13 +551,24 @@ def __call__(self, parameter):
         self.used_names.add(name)
         return name
 
+def _recreate_numpy_ndarray(_, content):
```
nouiz: To keep back-ward compatibility, you should swap those and maybe give a default value to _

blocks/serialization.py
```diff
@@ -544,13 +551,24 @@ def __call__(self, parameter):
...
+    return numpy.array(content)
+
+def _recreate_cuda_ndarray(_, content):
```
nouiz: idem

blocks/serialization.py
```diff
@@ -544,13 +551,24 @@ def __call__(self, parameter):
...
+    return cuda_ndarray.cuda_ndarray.CudaNdarray(content)
+
+def _recreate_pygpu_array(context_name, content):
```
nouiz: idem

blocks/serialization.py
```diff
@@ -188,10 +192,9 @@ def _save_parameters(f):
             named_parameters = {renamer(p): p for p in parameters}
             numpy.savez(f, **{n: p.get_value()
                               for n, p in named_parameters.items()})
-            for n, p in named_parameters.items():
+            for name, p in named_parameters.items():
                 array_ = p.container.storage[0]
```
nouiz: this line isn't needed anymore.

blocks/serialization.py
```diff
@@ -577,18 +593,35 @@ def __call__(self, id_):
...
+def _mangle_parameter_name(parameter, name):
+    array_type = type(parameter.container.storage[0])
+    context_name = (parameter.context_name
+                    if pygpu and
+                    isinstance(parameter, pygpu.gpuarray.GpuArray)
+                    else None)
+    if isinstance(context_name, str) and '.' in context_name:
+        raise ValueError("context name must not contain dots")
+    return '#1{}.{}.{}'.format(
+        _ARRAY_TYPE_MAP[array_type], context_name, name)
```
nouiz: Here for safety, I would keep name as the second element of the tuple. Just in case at other place we expect it there.

rizar: There are only two places in Blocks where _unmangle_parameter_name is used.
Both expect the name as the third element of the tuple.

nouiz: As I told, my comment was for safety. I think it is a good habit to not change input order. You don't know if someone else have code outside that reuse it. But as this is function marked as internal, I'm fine to merge this now.

I let you merge as I don't know if you consider blocking that the coverage went down.

PR #1075, opened 2016-05-03
Allow override pattern for SharedVariableModifier.

dwf:
Gets around an annoying limitation of `SharedVariableModifier` on Python 2.x.

cc: @memimo


dwf:
Tests pass.

vdumoulin:
LGTM.

PR #1078, opened 2016-05-05
Handle adding children to kwargs appropriately if a subclass has already added children.

SwordYork:
If we use `kwargs.get('children', [])` and initialize the super class by 

```
super(Class2, self).__init__(children=children,  **kwargs)
```
Then it will result the following error if the `kwargs` contains children, because we have pass the children twice.
```
    super(Class2, self).__init__(children=children, **kwargs)
TypeError: __init__() got multiple values for keyword argument 'children'
```

I'm not sure why using `kwargs.get('children', [])` in #1029 . In my opinion, we should use  `kwargs.pop('children', [])`.

### DIFF
--- a/blocks/bricks/attention.py
+++ b/blocks/bricks/attention.py
@@ -325,9 +325,9 @@ class SequenceContentAttention(GenericSequenceAttention, Initializable):
         self.energy_computer = energy_computer
 
         children = [self.state_transformers, attended_transformer,
-                    energy_computer] + kwargs.get('children', [])
-        super(SequenceContentAttention, self).__init__(children=children,
-                                                       **kwargs)
+                    energy_computer]
+        kwargs.setdefault('children', []).extend(children)
+        super(SequenceContentAttention, self).__init__(**kwargs) 
 
     def _push_allocation_config(self):
         self.state_transformers.input_dims = self.state_dims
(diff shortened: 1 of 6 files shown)


dwf:
Actually kwargs.setdefault is cleaner still.

SwordYork:
Sorry, I don't understand what you mean. If we pass children in `kwargs`, it will cause an error.  How can I deal with this error? I can't figure out how to use `kwargs.setdefault ` to solve this issue in a cleaner way.

dwf:
```
kwargs.setdefault('children', []).extend(children_i_am_adding)
super(ClassName, self).__init__(**kwargs)
```

SwordYork:
Thank you! I got it.

SwordYork:
I have changed the relative code into this form.
```
kwargs.setdefault('children', []).extend(children)
super(ClassName, self).__init__(**kwargs)
```

blocks/bricks/attention.py
```diff
@@ -325,9 +325,9 @@ def __init__(self, match_dim, state_transformer=None,
         self.energy_computer = energy_computer
 
         children = [self.state_transformers, attended_transformer,
-                    energy_computer] + kwargs.get('children', [])
-        super(SequenceContentAttention, self).__init__(children=children,
-                                                       **kwargs)
+                    energy_computer]
+        kwargs.setdefault('children', []).extend(children)
```
rizar: Wow, this is pretty cool :)

SwordYork: Thanks to @dwf , it's really a cleaner and cleverer way.

dwf: Yep, this is a variation on a pretty common idiom with `setdefault`: http://python.net/~goodger/projects/pycon/2007/idiomatic/handout.html#dictionary-setdefault-method-1

PR #1079, opened 2016-05-06
GradientDescent: refactoring, enhancements, bug fixes

### LINKED ISSUE #735: DifferentiableCostMinimizer is a lame duck
Perhaps I am missing something, but it does not look like DifferentiableCostMinimizer does very much (it neither differentiates nor minimizes), nor do we have any subclasses other than GradientDescent. Is there an argument for pushing functionality found there up to TrainingAlgorithm and/or down to GradientDescent?

### LINKED ISSUE #870: Crash when manually initializing GradientDescent
MainLoop initializes the algorithm regardless of whether it has already been initialized, and GradientDescent.initialize() adds the updates to self.updates, resulting in duplicate updates:

    ValueError: ('this shared variable already has an update
    expression, (<TensorType(float64, scalar)>,
    Elemwise{sub,no_inplace}.0)\n\nOriginal exception:\n\tValueError:
    this shared variable already has an update expression,
    (<TensorType(float64, scalar)>, Elemwise{sub,no_inplace}.0)',
    (<TensorType(float64, scalar)>, Elemwise{sub,no_inplace}.0))

To reproduce:

    import theano
    from theano import tensor
    from blocks.algorithms import GradientDescent, Scale
    from blocks.main_loop import MainLoop
    from fuel.datasets import MNIST
    from fuel.streams import DataStream
    from fuel.schemes import SequentialScheme

    a = theano.shared(1.)
    x = tensor.scalar()
    y = x + a

    mnist_train = MNIST(['train'])

    algorithm = GradientDescent(cost=y, parameters=[a])
    algorithm.initialize()

    main_loop = MainLoop(
        algorithm,
        DataStream(
            mnist_train,
            iteration_scheme=SequentialScheme(mnist_train.num_examples, 100)))
    main_loop.run()

### LINKED ISSUE #919: GradientDescent shouldn't need cost when gradients are given
It's not being used for anything important but it's still annoying. Strongly related to #735.

### LINKED ISSUE #938: TrainingDataMonitoring cause algorithm hang at initialization
I recently found that when monitoring an output using TrainingDataMonitoring, but the output requires an input that is not in the computation graph of the cost, main loop will hang at "Initializing the training algorithm". My current solution is to modify _cost_computation_graph in DifferentiableCostMinimizer to be the computation graph of all the outputs.  Is this a problem of the algorithm or of my particular setting?

### LINKED ISSUE #1023: Why DifferentiableCostMinimizer does not accept updates as a keyword argument?
Seems like the only way to add update to an algorithm object currently is to do `algorithm.updates = cg.updates`, which is quite weird. We should definitely support a providing the updates at the construction time.

Related to #735

dwf:
A refactoring plus a variety of improvements and potential bug fixes to `GradientDescent`.

* `DifferentiableCostMinimizer` is gone in favour of a more useful base class, `UpdatesAlgorithm`, which just takes a list of updates and compiles a Theano function for you (as suggested/concurred by @rizar [here](https://github.com/mila-udem/blocks/issues/735#issuecomment-116605243)). (#735)
* `GradientDescent` now no longer requires a `cost` argument, which was superfluous but still if `gradients`. (#919) This also means inputs are now always inferred from the computation graph of the *updates* and not of the *cost*, which is a bit more robust I think (particularly to the case where additional updates may need a source that the training updates don't, e.g. #938)
* `GradientDescent` issues warnings if you've specified gradients as well as a cost (that the cost does nothing in this case), and if you pass parameters and gradients that don't match.
* `GradientDescent` now raises **errors** in situations where `gradients` being a dict (as opposed to an OrderedDict) could lead to non-determinism across runs due to randomized hashing/dictionary iteration order.
* The computation of `GradientDescent`''s `total_step_norm` and `total_gradient_norm` now sums the terms in a fixed order based on `self.parameters`, which should remove a potential source of non-determinism in those computations.
* As a happy by-product of `UpdatesAlgorithm`, `GradientDescent` now accepts an `updates` keyword argument for directly specifying additional updates at construction (#1023).
* I noticed that this fixes #870 and #938, almost by accident, so I added tests for them.

Fixes #735.
Fixes #919.
Fixes #1023.
Fixes #938.
Fixes #870.

### DIFF
--- a/blocks/algorithms/__init__.py
+++ b/blocks/algorithms/__init__.py
@@ -56,93 +56,6 @@ class TrainingAlgorithm(object):
         pass
 
 
-class DifferentiableCostMinimizer(TrainingAlgorithm):
-    """Minimizes a differentiable cost given as a Theano expression.
-
-    Very often the goal of training is to minimize the expected value of a
-    Theano expression. Batch processing in this cases typically consists of
-    running a (or a few) Theano functions.
-    :class:`DifferentiableCostMinimizer` is the base class for such
-    algorithms.
-
-    Parameters
-    ----------
-    cost : :class:`~tensor.TensorVariable`
-        The objective to be minimized.
-    parameters : list of :class:`~tensor.TensorSharedVariable`
-        The parameters to be tuned.
-
-    Attributes
-    ----------
-    updates : list of :class:`~tensor.TensorSharedVariable` updates
-        Updates to be done for every batch. It is required that the
-        updates are done using the old values of optimized parameters.
-    cost : :class:`~tensor.TensorVariable`
-        The objective to be minimized.
-    parameters : list of :class:`~tensor.TensorSharedVariable`
-        The parameters to be tuned.
-
-    Notes
-    -----
-    Changing `updates` attribute or calling `add_updates` after
-    the `initialize` method is called will have no effect.
-
-    .. todo::
-
-       Some shared variables are not parameters (e.g. those created by
-       random streams).
-
-    .. todo::
-
-       Due to a rather premature status of the :class:`ComputationGraph`
-       class the parameter used only inside scans are not fetched
-       currently.
-
-    """
-    def __init__(self, cost, parameters):
-        self.cost = cost
-        self.parameters = parameters
-        self._cost_computation_graph = ComputationGraph(self.cost)
-        self._updates = []
-
-    @property
-    def inputs(self):
-        """Return inputs of the cost computation graph.
-
-        Returns
-        -------
-        inputs : list of :class:`~tensor.TensorVariable`
-            Inputs to this graph.
-
-        """
-        return self._cost_computation_graph.inputs
-
-    @property
-    def updates(self):
-        return self._updates
-
-    @updates.setter
-    def updates(self, value):
-        self._updates = value
-
-    def add_updates(self, updates):
-        """Add updates to the training process.
-
-        The updates will be done _before_ the parameters are changed.
-
-        Parameters
-        ----------
-        updates : list of tuples or :class:`~collections.OrderedDict`
-            The updates to add.
-
-        """
-        if isinstance(updates, OrderedDict):
-            updates = list(updates.items())
-        if not isinstance(updates, list):
-            raise ValueError
-        self.updates.extend(updates)
-
-
 variable_mismatch_error = """
 
 Blocks tried to match the sources ({sources}) of the training dataset to \
@@ -158,7 +71,7 @@ Blocks didn't find all the sources ({sources}) of the training dataset \
 that match the names of the Theano variables ({variables})."""
 
 
-class GradientDescent(DifferentiableCostMinimizer):
+class GradientDescent(TrainingAlgorithm):
     """A base class for all gradient descent algorithms.
 
     By "gradient descent" we mean a training algorithm of the following
@@ -177,6 +90,11 @@ class GradientDescent(DifferentiableCostMinimizer):
 
     Parameters
     ----------
+    cost : :class:`~tensor.TensorVariable`, optional
+        The objective to be minimized.
+    parameters : list of :class:`~tensor.TensorSharedVariable`, optional
+        The parameters to be tuned. If not provided, inferred from the
+        keys of `gradients`.
     step_rule : instance of :class:`StepRule`, optional
         An object encapsulating most of the algorithm's logic. Its
         `compute_steps` method is called to get Theano expression for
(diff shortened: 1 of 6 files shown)


dwf:
@dmitriy-serdyuk Thanks for reviewing mila-udem/fuel#355 so quickly. If you've still got free time on your hands, this one's pretty juicy. I think it represents a big improvement over the current state of things (plus, 5 tickets!).

dwf:
(Scrutinizer errored but this has nothing to do with the PR.)

blocks/algorithms/__init__.py
```diff
@@ -56,66 +56,75 @@ def process_batch(self, batch):
...
 
-    Very often the goal of training is to minimize the expected value of a
-    Theano expression. Batch processing in this cases typically consists of
-    running a (or a few) Theano functions.
-    :class:`DifferentiableCostMinimizer` is the base class for such
-    algorithms.
+Blocks tried to match the sources ({sources}) of the training dataset to \
+the names of the Theano variables ({variables}), but failed to do so. \
+If you want to train on a subset of the sources that your dataset provides, \
+pass the `sources` keyword argument to its constructor. Or pass \
```
dmitriy-serdyuk: It is fuel specific, but one can use `FilterSources` transformer.

blocks/algorithms/__init__.py
```diff
@@ -198,100 +230,130 @@ class GradientDescent(DifferentiableCostMinimizer):
...
+    .. todo::
+
+       Some shared variables are not parameters (e.g. those created by
+       random streams).
+
+    .. todo::
+
+       Due to a rather premature status of the :class:`ComputationGraph`
+       class the parameter used only inside scans are not fetched
+       currently.
```
dmitriy-serdyuk: Do you mean, that updates of scans which are not wrapped into `recurrent` are not provided as updates?

dwf: This warning predates this PR.

blocks/algorithms/__init__.py
dwf: Mentioned in 01918f6.

blocks/algorithms/__init__.py
dwf: Should we remove this warning? I'm not as familiar with issues surrounding scan and `recurrent` so I don't know. I'm also not entirely sure why the other TODO is there either.

dmitriy-serdyuk:
Ok, I finished reading it, very nice changes!

Second initialization doesn't recompute the gradients in the case if the model changed, right? Should we add a warning message?

dwf:
Should we put that in the docstring or in a logger.warn call if initialize() is called twice?

dmitriy-serdyuk:
Or is it better to move taking gradients to the `initialize()`? Usually it takes some time to take gradient, so this change can simplify debugging.

dwf:
@dmitriy-serdyuk See d48422c. Because the updates list can be affected by both the updates kwarg passed in and mutated by `add_updates`, it's too difficult in this scenario to make sure that gradient updates don't get added multiple times, so I just issue a warning on a second (or nth) call to `initialize()` and bail.

dwf:
Again with the spurious Scrutinizer failure.

dmitriy-serdyuk:
I just realized, that the algorithm is initialized after `before_training` extensions in order to allow that extension to add their updates to the algorithm. That's why Travis crashes, one of extensions is trying to fetch `total_grad_norm`.

dwf:
Heh, welp. I'll roll back that commit then and just add the warning.

blocks/algorithms/__init__.py
dmitriy-serdyuk: Yes, it looks like both were fixed long time ago. The computation graph distinguishes between parameters and shared variables and the graph can fetch updates from scan wrapped in @recurrent .

dmitriy-serdyuk:
> I'll roll back that commit then and just add the warning.
Ok

dwf:
@dmitriy-serdyuk 825bfff puts an additional warning in the `GradientDescent` docstring as seems sort of silly to add an `initialize()` method just containing a warning (there might also be good reasons to want to add some updates and then re-call `initialize()` to recompile the function, but this is a bit dodgy and I don't think we should document this as recommended usage).

blocks/algorithms/__init__.py
dwf: 4e7f384 removes these.

dmitriy-serdyuk:
LGTM, feel free to merge after the test passes.

PR #1092, opened 2016-05-20
Fixes #1086: use six.wraps insted of functools

### LINKED ISSUE #1086: Use six.wraps insted of functools.wraps
#### What?
From python 3.2 onward `functools.wraps` adds a `__wrapped__` field to the wrapper.
`six.wraps` enables this functionality in python2

#### Why?
Having a way to refer to the wrapped function can be useful if you e.g. stack multiple
annotations and want to get the original function signature. (The function signature is,
for example used in the `recurrent` annotation to convert args to kwargs, so using
`recurrent` on a wrapped function *could* lead to problems if the parameters are not
passed in by name (but by index))

naicode:
Replaces imports of `functools.wraps` with imports of `six.wraps` for python2 and 3 compatibility.

### DIFF
--- a/blocks/bricks/base.py
+++ b/blocks/bricks/base.py
@@ -1,7 +1,7 @@
 import inspect
 from abc import ABCMeta
 from collections import OrderedDict
-from functools import wraps
+from six import wraps
 from operator import attrgetter
 from types import MethodType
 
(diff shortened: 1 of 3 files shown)


dmitriy-serdyuk:
Thank you!

PR #1096, opened 2016-05-30
Compare scan Ops by their ids

rizar:
This fixes quite a bad bug. By default in Theano scan ops are compared based on the computation that they do. But sometimes one can have two scans which do identical computations, but nevertheless were created using different bricks. In such cases with this bug those bricks could go unnoticed by `Model`.


dmitriy-serdyuk:
Yes, indeed it is a bug. LGTM.

PR #1097, opened 2016-05-30
Unique path for all parameters

rizar:
I have recently noticed that we save parameters using the names constructed by calling `Brick.unique_path` and load parameters (in `Model`) using parameters produced by `Selector`.  This weird discrepancy may cause bug in the case when a brick has multiple parents, whereby different names are used for loading and saving the same parameter.

This PR proposes to settle to a single method of naming parameters based on `get_unique_path`. I think makes sense to hardcode in the very core what is a name of a parameter. Also, `Selector` was never properly finished, and I don't think that the dependency of `Model` on `Selector` is a good thing.

### DIFF
--- a/blocks/serialization.py
+++ b/blocks/serialization.py
@@ -136,12 +136,12 @@ except:
 from blocks.config import config
 from blocks.filter import get_brick
 from blocks.utils import change_recursion_limit
-from blocks.bricks.base import BRICK_PATH_DELIMITER
+from blocks.bricks.base import BRICK_DELIMITER
 
 
 logger = logging.getLogger(__name__)
 
-SERIALIZATION_BRICK_PATH_DELIMITER = '|'
+SERIALIZATION_BRICK_DELIMITER = '|'
 MAIN_MODULE_WARNING = """WARNING: Main loop depends on the function `{}` in \
 `__main__` namespace.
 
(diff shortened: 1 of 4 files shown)


blocks/serialization.py
```diff
@@ -136,11 +136,12 @@
 from blocks.config import config
 from blocks.filter import get_brick
 from blocks.utils import change_recursion_limit
+from blocks.bricks.base import BRICK_PATH_DELIMITER
 
 
 logger = logging.getLogger(__name__)
 
-BRICK_DELIMITER = '|'
+SERIALIZATION_BRICK_PATH_DELIMITER = '|'
```
dmitriy-serdyuk: Oh, that's a bit too much ;) Maybe just `SERIALIZATION_PATH_DELIMITER`?

rizar: Well, it is used only once, so I do not think that this very long name is
an issue. On the other hand it is super clear what it is used for.

dwf:
The failure of the examples test raises an interesting point: this name might be imported in client code, so we should have a compelling reason to change it. Honestly, I thought `BRICK_DELIMITER` was fine.

rizar:
May I merge it?

dwf:
LGTM but @dmitriy-serdyuk did the initial review so I would leave it up to him.

PR #1100, opened 2016-05-31
Fixes #1099, use `.rsplit(..., 1)` instead of split+join

### LINKED ISSUE #1099: split_suffix in RecurrentStack does not handle names with multiple # correctly
Hy, 

the "split_suffix" code:
```python
    @staticmethod
    def split_suffix(name):
        name_level = name.split(RECURRENTSTACK_SEPARATOR)
        if len(name_level) == 2 and name_level[-1].isdigit():
            name = RECURRENTSTACK_SEPARATOR.join(name_level[:-1])
            level = int(name_level[-1])
        else:
            # It must be from bottom layer
            level = 0
        return name, level
```

won't work correctly if the name hase more than one # ("#"==`RECURRENTSTACK_SEPARATOR)`,
it will see that `len(name_level)` is > 2 and set the level to 0 (but on the other hand it reconstructs the name in the first line of the if body as if it could have had a len > 2.

PS:
I will send in a fix using `.rsplit(..., 1)`  (split only one time first occurrence from right)

dathinab:
Because of the `len(name_level) == 2` (insted of `>=2` ) the handling of names with more than on "#" had been wrong, this PR fixes it  (and uses rsplit+split limit instead of split+join)


rizar:
Thanks, seems perfectly legitimate to me. Would you mind adding a test?

dathinab:
np, but will likely take until up to Fr (I don't have anymore time today and the next days, even through it doesn't take long ;) )

rizar:
Sure, no problem, whenever you have time.

dathinab:
Took me longer then anticipated to come back to blocks, but I had finally some time to add two unit tests.

rizar:
Thanks for getting back to this. Unfortunately, our standard practice for updating old dev branches is rebasing on top of the master branch, not merging the master branch.  I believe in this case the simplest solution would be to create a new pull request, because the changes are fairly isolated.

dathinab:
Ok here is the new pull request: https://github.com/mila-udem/blocks/pull/1143

PR #1101, opened 2016-05-31
Bug fix + enhancement to l2_norm.

### LINKED ISSUE #1088: DeprecationWarning raised by l2_norm
[This line](https://github.com/mila-udem/blocks/blob/master/blocks/theano_expressions.py#L18) yields
```
DeprecationWarning: stack(*tensors) interface is deprecated, use stack(tensors, axis=0) instead.
```

dwf:
* Instead of squaring the output of a sqrt(), add a keyword argument to just not do the sqrt in the first place.
* Fix #1088 by using an alternate form of `tensor.stack`.


dmitriy-serdyuk:
LGTM

PR #1103, opened 2016-06-01
Support auto-building ComputationGraphs in VariableFilter.

### LINKED ISSUE #1080: VariableFilter should auto-instantiate a ComputationGraph if given a list or TensorVariable
It's pretty unambiguous what I want if I pass either of these.

dwf:
Fixes #1080.

### DIFF
--- a/blocks/filter.py
+++ b/blocks/filter.py
@@ -3,6 +3,7 @@ import re
 
 from blocks.bricks.base import ApplicationCall, BoundApplication, Brick
 from blocks.roles import has_roles
+from theano.tensor import TensorVariable
 
 
 def get_annotation(var, cls):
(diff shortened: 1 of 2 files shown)


blocks/filter.py
```diff
@@ -120,14 +121,32 @@ def __init__(self, roles=None, bricks=None, each_role=False, name=None,
...
+            built from that variable. If an iterable (e.g. a list,
+            or :class:`~blocks.graph.ComputationGraph`) containing
+            :class:`~tensor.TensorVariable`s, operate on those variables.
+        graph : bool, optional
+            If `True`, build a `~blocks.graph.ComputationGraph` instance
+            with `variables` as the output. Default: `False`. Note that
+            `False` has no effect if `variables` is a
+            :class:`~tensor.TensorVariable` (i.e. the result is the same,
+            an instance of :class:`~blocks.graph.ComputationGraph` is
+            built).
```
dmitriy-serdyuk: Can you give an example from the issue? It might be hard to see a use case for this argument without an explanation.

PR #1104, opened 2016-06-01
Add switches for shift/scale parameters in BatchNormalization.

### LINKED ISSUE #1071: Flags to disable learned shift/scale in BatchNormalization
Seen this a few places, e.g. the code for the Radford et al DCGAN paper. Should not be hard to do.

This subclass clunkily gets the job done:

```python
class ParameterlessBatchNormalization(BatchNormalization):
    def _allocate(self):
        super()._allocate()
        self.parameters.remove(self.scale)
        self.parameters.remove(self.shift)
        self.scale = tensor.constant(1., dtype=theano.config.floatX)
        self.shift = tensor.constant(0., dtype=theano.config.floatX)

    def _initialize(self):
        pass
```

But we could probably support it with optional kwargs.

dwf:
Also make comparable changes in `BatchNormalizedMLP` and change the treatment of `mean_only` in the latter (because it really is an allocation attribute, unlike `conserve_memory`).

Fixes #1071.

### DIFF
--- a/blocks/bricks/bn.py
+++ b/blocks/bricks/bn.py
@@ -266,6 +266,11 @@ class BatchNormalization(RNGMixin, Feedforward):
             add_annotation(self.population_stdev, self)
 
     def _initialize(self):
+        # We gate with is_shared_variable rather than relying on
+        # learn_scale and learn_shift so as to avoid the unlikely but nasty
+        # scenario where those flags are changed post-allocation but
+        # pre-initialization. This ensures that such a change simply has no
+        # effect rather than doing an inconsistent combination of things.
         if is_shared_variable(self.shift):
             self.shift_init.initialize(self.shift, self.rng)
         if is_shared_variable(self.scale):
@@ -363,6 +368,10 @@ class BatchNormalizedMLP(MLP):
     in the :class:`BatchNormalization` bricks being added. Pass
     `use_bias` with a value of `True` if you really want this for some
     reason.
+
+    `mean_only`, `learn_scale` and `learn_shift` are pushed down to
+    all created :class:`BatchNormalization` bricks as allocation
+    config.
 
     """
     @lazy(allocation=['dims'])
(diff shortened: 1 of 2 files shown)


blocks/bricks/bn.py
```diff
@@ -237,23 +248,27 @@ def _allocate(self):
...
             self.population_stdev = shared_floatx(numpy.ones(var_dim),
                                                   name='population_stdev',
                                                   broadcastable=broadcastable)
             add_role(self.population_stdev, BATCH_NORM_POPULATION_STDEV)
             add_annotation(self.population_stdev, self)
 
     def _initialize(self):
-        self.shift_init.initialize(self.shift, self.rng)
-        if not self.mean_only:
+        if is_shared_variable(self.shift):
```
dmitriy-serdyuk: Why so complicated? `if self.learn_shift` should work.

blocks/bricks/bn.py
```diff
@@ -237,23 +248,27 @@ def _allocate(self):
...
+            self.shift_init.initialize(self.shift, self.rng)
+        if is_shared_variable(self.scale):
```
dmitriy-serdyuk: Idem

blocks/bricks/bn.py
dwf: Wanted to avoid the situation where that flag is changed after allocation
but before initialization. Which is exotic, but better to program
defensively, no? This way it simply has no effect on the brick's operation
post-allocation, rather than doing some combination of inconsistent things.

>

dmitriy-serdyuk: Ah, I see.

dmitriy-serdyuk:
Ok, LGTM

blocks/bricks/bn.py
dwf: Yeah, I should probably add a comment about that.

dmitriy-serdyuk: Yes, for future reference.

dwf:
Docstrings tweaked in 03035b2. Tests pass so I'm merging.

PR #1105, opened 2016-06-01
Refactor bricks.recurrent module

dmitriy-serdyuk:
It became more than 1000 lines long and it's hard to navigate now. I propose to split it into 3 submodules.


dwf:
Might be worth adding the exported names in `bricks/recurrent/__init__.py` to `bricks/__init__.py` (ping #885).

dwf:
Actually the fact that `recurrent` is both the name of the decorator and the name of the module may present a bit of an issue with flattening out the namespace. I just tried it and it's actually fine:
```
dwf@skyquake:~$ ls -F a/b
__init__.py
dwf@skyquake:~$ ls -F a/
__init__.py  b/
dwf@skyquake:~$ cat a/b/__init__.py
def b(x):
    pass

dwf@skyquake:~$ cat a/__init__.py
from __future__ import absolute_import

from .b import b
dwf@skyquake:~$ python2.7 -c 'from a import b; print b'
<function b at 0x10a9a7398>
dwf@skyquake:~$ python3.4 -c 'from a import b; print(b)'
<function b at 0x10063c510>
```
but it renders the module version inaccessible via import. I guess putting `recurrent_module = recurrent` before the `from .recurrent import recurrent` would preserve that capability.

dmitriy-serdyuk:
Oops, in this case recurrent decorator conflicts with the module name.

dwf:
Yeah, it's a bit annoying that it will break everybody's imports. I suppose it's fine to leave out the recurrent decorator.

dmitriy-serdyuk:
Sorry, I haven't seen you comment. I'll check it tomorrow, but seems there are still problems with tests and my IDE is extremely confused.

dmitriy-serdyuk:
Actually, it was a different problem: some modules were importing stuff from `blocks.bricks` which created a cyclic import after I added everything from recurrent to the same namespace. So, I'm reverting back my previous commit.

dwf:
Ah, yes, in that case one needs to import directly from `bricks` submodules (preferably with relative imports).

dmitriy-serdyuk:
@dwf, any other comments?

blocks/bricks/recurrent/architectures.py
```diff
@@ -0,0 +1,372 @@
+# -*- coding: utf-8 -*-
+import numpy
+from theano import tensor
+
+from ...bricks.base import application, lazy
```
dwf: Technically only necessary to use `..base` and `..simple here.

blocks/bricks/recurrent/base.py
```diff
@@ -0,0 +1,251 @@
+# -*- coding: utf-8 -*-
+import inspect
+import logging
+from six import wraps
+
+from picklable_itertools.extras import equizip
+import theano
+from theano import tensor, Variable
+
+from ...bricks.base import Application, application, Brick
```
dwf: idem

blocks/bricks/recurrent/misc.py
```diff
@@ -0,0 +1,414 @@
+# -*- coding: utf-8 -*-
+import copy
+
+from picklable_itertools.extras import equizip
+from theano import tensor
+
+from ...bricks.base import application, lazy
```
dwf: idem

dwf:
The '...bricks' nitpick aside, LGTM

dwf:
Maybe shoot an email to the list saying that you should import from `blocks.bricks` and not `blocks.bricks.recurrent` now?

dmitriy-serdyuk:
@dwf , now we have classes documented twice:
http://blocks.readthedocs.io/en/latest/api/bricks.html#blocks.bricks.recurrent.architectures.GatedRecurrent
http://blocks.readthedocs.io/en/latest/api/bricks.html#blocks.bricks.GatedRecurrent

Do you know how to make sphinx to produce only one copy?

dmitriy-serdyuk:
A huge list of `:exclude-members: ` ?

dwf:
Let me have a look when I am back at my computer.

dwf:
@dmitriy-serdyuk So it looks like none of our options are not ideal. We should probably avoid `:automodule:`'ing in the third-level modules as we're encouraging people to import from level 2, which is the whole point of flattening in the first place.

So one way is removing `:automodule:` for second level modules and simply accepting that the third-level module will show up in the names. There may be a way to add a Sphinx directive to rewrite this name.

Another way is multiple `:automodule:` directives for `blocks.bricks`, each with different `:members:` lists. It sort of sucks that we have to maintain these `:members:` lists manually, but...

PR #1107, opened 2016-06-03
Change MLP dims in test_collect to disambiguate falures.

davidbau:
This test fails on GPU, because the order of cg.variables differs between CPU and GPU.  Changing this test in this way makes the cause of the failure more clear.

(Previous it was unclear whether the test failed because the matrix was transposed vs variables reordered.)

### DIFF
--- a/tests/test_graph.py
+++ b/tests/test_graph.py
@@ -186,7 +186,7 @@ def test_snapshot():
 
 def test_collect():
     x = tensor.matrix()
-    mlp = MLP(activations=[Logistic(), Logistic()], dims=[784, 100, 784],
+    mlp = MLP(activations=[Logistic(), Logistic()], dims=[784, 100, 512],
               use_bias=False)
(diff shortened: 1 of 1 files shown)


tests/test_graph.py
```diff
@@ -186,7 +186,7 @@ def test_snapshot():
 
 def test_collect():
     x = tensor.matrix()
-    mlp = MLP(activations=[Logistic(), Logistic()], dims=[784, 100, 784],
+    mlp = MLP(activations=[Logistic(), Logistic()], dims=[784, 100, 512],
               use_bias=False)
     cost = SquaredError().apply(x, mlp.apply(x))
```
dmitriy-serdyuk: Actually, it should be 784 since we compute MSE between original image and the result of MLP. It doesn't make so big difference in this case, but having 512 may lead to confusion.

davidbau:
Ah i see, what the example is.  Should leave it as-is.

PR #1108, opened 2016-06-03
Convolutions should default to tied_biases=True.

davidbau:
Tied biases in convolutions brings bricks into alignment with other
libraries including Caffe, Lasange, and Keras, which tie biases
in convolutional layers by default.  Now tied_biases=False must be
specified explicitly to untie biases.

### DIFF
--- a/tests/bricks/test_conv.py
+++ b/tests/bricks/test_conv.py
@@ -195,7 +195,7 @@ def test_no_input_size():
                          c.initialize)
 
 
-def test_tied_biases():
+def test_untied_biases():
     x = tensor.tensor4('x')
     num_channels = 4
     num_filters = 3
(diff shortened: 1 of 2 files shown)


tests/bricks/test_conv.py
```diff
@@ -195,19 +195,58 @@ def test_no_input_size():
...
+def test_untied_biases():
     x = tensor.tensor4('x')
     num_channels = 4
     num_filters = 3
     batch_size = 5
     filter_size = (3, 3)
     conv = Convolutional(filter_size, num_filters, num_channels,
                          weights_init=Constant(1.), biases_init=Constant(2.),
-                         tied_biases=True)
+                         image_size = (28, 30), tied_biases=False)
```
dwf: Think there's a PEP8 issue there (spaces around the equals sign). We supposedly have a linter to catch this sort of stuff but it doesn't appear to be working.

dwf:
Thanks!

dwf:
P.S. for future reference, if you add the text "Fixes #NNNN" in the text of your commit message or PR description it will auto-close the issue upon merge. :smile:

PR #1109, opened 2016-06-03
EarlyStopping composite extension

### LINKED ISSUE #1085: Early stopping composite extension
I've had this idea for a while, which is that we should have an extension called `EarlyStopping` that incorporates checkpointing (or at least setting up a passed in checkpointing extension), `TrackTheBest` and `FinishIfNoImprovementAfter`, to make it less complicated to set up the most common thing you want to do with this combination of things: checkpoint every time you get a new best and terminate if it's been too long since. Right now I basically have to copy and paste code from the tests every time.

dwf:
Fixes #1085.

@dmitriy-serdyuk I marked #1085 a CCW ticket too...

### DIFF
--- a/blocks/extensions/predicates.py
+++ b/blocks/extensions/predicates.py
@@ -12,3 +12,7 @@ class OnLogRecord(object):
 
     def __call__(self, log):
         return bool(log.current_row.get(self.record_name, False))
+
+    def __eq__(self, other):
+        return (type(other) == type(self) and
+                other.record_name == self.record_name)
--- a/blocks/extensions/stopping.py
+++ b/blocks/extensions/stopping.py
@@ -110,6 +110,11 @@ class EarlyStopping(CompositeExtension):
 
     Notes
     -----
+    .. warning::
+        If you want the best model to be saved, you need to specify
+        a value for the ``checkpoint_extension`` and
+        ``checkpoint_filename`` arguments!
+
     Trigger keyword arguments will affect how often the log is inspected
     for the record name (in order to determine if a new best has been
     found), as well as how often a decision is made about whether to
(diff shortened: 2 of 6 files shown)


dwf:
This is ready for review by the way, tests pass and coverage is good.

dmitriy-serdyuk:
Maybe it makes sense to add another flag which creates a checkpointing extension inside EarlyStopping?

rizar:
Cool, I have been dreaming of `CompositeExtension` for a while. Still looking at your PR, but so far it looks great.

dwf:
I thought about that, and realized that
* I'd have to provide a dummy filename for the first argument of `Checkpoint` (which feels weird) or else I'd have to make it an optional argument (which also feels weird)
* for this to be really useful we'd have to mirror all the arguments of `Checkpoint` in `EarlyStopping`. The constructor signature is already getting rather long.

dmitriy-serdyuk:
Yeah, I see.

So, the use case is like

```python
checkpoint = Checkpoint("my_model", after_epoch=True)
early_stopping = EarlyStopping('valid_misclass_error', checkpoint, epochs=patience)
MainLoop(extensions=[chekpoint, early_stopping])
```
And it will be saved as "my_model_best_so_far".

dwf:
Actually, it will complain at you in that case. See the warning here, which I think is correct: 
https://github.com/mila-udem/blocks/pull/1109/files#diff-20e63697de27783f09d0cc2fd2d2794fR162

blocks/extensions/stopping.py
```diff
@@ -66,8 +73,97 @@ def do(self, which_callback, *args):
...
+                exts = self.main_loop.extensions
+                if exts.index(self.checkpoint_extension) < exts.index(self):
+                    logger.warn('%s: configured checkpointing extension '
+                                'appears after %s in main loop '
+                                'extensions list. This may lead to '
+                                'unwanted results, as the notification '
+                                'that would trigger serialization '
+                                'of a new best will not have been '
+                                'written yet when the checkpointing '
+                                'extension is run.')
```
dmitriy-serdyuk: You forgot to format the string here.

dmitriy-serdyuk:
I'm a little bit confused. Am I supposed to include it as `[early_stopping, checkpoint]` or not include the `checkpoint` at all? Seems, that checkpointing will be called twice with the former option.

dwf:
See the Notes. You can do either. If it's not in the main loop's extensions
list, it'll get added internally.

dmitriy-serdyuk:
I see now, it's cool in this case!

blocks/extensions/predicates.py
```diff
@@ -12,3 +12,7 @@ def __init__(self, record_name):
 
     def __call__(self, log):
         return bool(log.current_row.get(self.record_name, False))
+
+    def __eq__(self, other):
```
rizar: I don't immediately where is this needed for, can you please explain?

dwf: It's used in a test to make sure that the appropriate `Checkpoint.add_condition` call is made.

dwf: Specifically, https://github.com/mila-udem/blocks/pull/1109/files#diff-376409566e985ed805d621c4bf885535R183

I can imagine it being useful in other contexts too...

blocks/extensions/stopping.py
```diff
@@ -66,8 +73,181 @@ def do(self, which_callback, *args):
...
+    use e.g.
+
+    >>> stopping_ext = EarlyStopping('valid_error', iterations=100)
+
+    which would halt training if a new minimum ``valid_error`` has not
+    been achieved in 100 iterations (i.e. minibatches/steps). To measure
+    in terms of epochs (which usually correspond to passes through the
+    training set), you could use
+
+    >>> epoch_stop_ext = EarlyStopping('valid_error', epoch=5)
```
rizar: `epochs` instead of `epoch`?

rizar:
Travis has found quite a few typos in the PR. Also, @dmitriy-serdyuk correctly noticed, that in one of your error message you seem to have forgotten to add format arguments. Otherwise it looks good, I think.

dwf:
Typos fixed, error message formatting also fixed (I've amended the old commits).

PR #1110, opened 2016-06-06
Improve timing extension

### LINKED ISSUE #1098: Timing extension gives misleading information if triggered every n batches
It's due to `every_n_batches` is turned into `after_batch` callback.

The extension should check how many iterations done since the previous call and report average time.

I guess, the same problem exists with `every_n_epochs`.

dmitriy-serdyuk:
- Add prefix argument
- Average timing if it runs every n batches/epochs
- Improve test

Fixes #1098

### DIFF
--- a/blocks/extensions/__init__.py
+++ b/blocks/extensions/__init__.py
@@ -526,8 +526,8 @@ class Timing(SimpleExtension):
     -----
     Add this extension *before* the :class:`Printing` extension.
 
-    Called with callbacks like ``every_n_batches`` this extension this
-    extension averages the time.
+    Created with callbacks like ``every_n_batches`` this extension
+    averages the time.
 
     This extension does *not* enable full profiling information. To see a
     full profile of the main loop at the end of training, use the
--- a/blocks/extensions/__init__.py
+++ b/blocks/extensions/__init__.py
@@ -506,7 +506,7 @@ class ProgressBar(TrainingExtension):
 
 
 class Timing(SimpleExtension):
-    r"""Add timing information to the log.
+    """Add timing information to the log.
 
     This adds data about the time spent in the algorithm's
     :meth:`~.Algorithm.process_batch` method as well as the time spent
@@ -516,11 +516,7 @@ class Timing(SimpleExtension):
     Parameters
     ----------
     prefix : str
-        Prefix to be added to the log record, looks the best if ends with
-        ``_``. Defaults to the empty string.
-    \*\*kwargs
-        Keyword arguments which are passed to the
-        :class:`.SimpleExtension`.
+        Prefix to be added to the log record. Defaults to the empty string.
 
     Notes
     -----
@@ -551,6 +547,8 @@ class Timing(SimpleExtension):
             'epoch': kwargs.get('every_n_epochs', 1)
         }
         self.prefix = prefix
+        if self.prefix:
+            self.prefix += '_'
 
     def do(self, which_callback, *args):
         current_row = self.main_loop.log.current_row
--- a/blocks/extensions/__init__.py
+++ b/blocks/extensions/__init__.py
@@ -506,23 +506,35 @@ class ProgressBar(TrainingExtension):
 
 
 class Timing(SimpleExtension):
-    """Add timing information to the log.
+    r"""Add timing information to the log.
 
     This adds data about the time spent in the algorithm's
     :meth:`~.Algorithm.process_batch` method as well as the time spent
     reading data per batch or epoch. It also reports the time spent
     initializing the algorithm.
 
+    Parameters
+    ----------
+    prefix : str
+        Prefix to be added to the log record, looks the best if ends with
+        ``_``. Defaults to the empty string.
+    \*\*kwargs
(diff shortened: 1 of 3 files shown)


blocks/extensions/__init__.py
```diff
@@ -506,23 +506,35 @@ def before_batch(self, batch):
...
     :meth:`~.Algorithm.process_batch` method as well as the time spent
     reading data per batch or epoch. It also reports the time spent
     initializing the algorithm.
 
+    Parameters
+    ----------
+    prefix : str
+        Prefix to be added to the log record, looks the best if ends with
+        ``_``. Defaults to the empty string.
+    \*\*kwargs
```
dwf: We don't normally document this explicitly, do we?

blocks/extensions/__init__.py
```diff
@@ -506,23 +506,35 @@ def before_batch(self, batch):
...
+        Keyword arguments which are passed to the
+        :class:`.SimpleExtension`.
+
     Notes
     -----
     Add this extension *before* the :class:`Printing` extension.
 
+    Called with callbacks like ``every_n_batches`` this extension this
+    extension averages the time.
```
dwf: "this extension this extension"

blocks/extensions/__init__.py
```diff
@@ -506,23 +506,35 @@ def before_batch(self, batch):
...
     This adds data about the time spent in the algorithm's
     :meth:`~.Algorithm.process_batch` method as well as the time spent
     reading data per batch or epoch. It also reports the time spent
     initializing the algorithm.
 
+    Parameters
+    ----------
+    prefix : str
+        Prefix to be added to the log record, looks the best if ends with
+        ``_``. Defaults to the empty string.
```
dwf: Hmm, I think in other places we add the "_" automatically, e.g. DataStreamMonitoring

blocks/extensions/__init__.py
```diff
@@ -549,7 +566,10 @@ def do(self, which_callback, *args):
...
             self.previous[level][action] = self.current[level][action]
             self.current[level][action] = profile['training', 'epoch', action]
-            current_row['time_{}_this_{}'.format(action, level)] = \
-                self.current[level][action] - self.previous[level][action]
-            current_row['time_{}_total'.format(action)] = \
+            this_time = self.prefix + 'time_{}_this_{}'
+            current_row[this_time.format(action, level)] = (
+                (self.current[level][action] - self.previous[level][action]) /
+                self.divisor[level])
+            total_time = self.prefix + 'time_{}_total'
```
dwf: It seems like we can make weird things happen here if specifying weird things like `every_n_batches=5, after_batch=True`.

dwf: Actually, how does `level` even get set in the case of `every_n_*` callbacks?

dwf:
Here's an idea for a slightly more robust setup: record the previous batch or epoch for each level, then use the difference between the current one and the previous one as the divisor. This naturally handles the issue I brought up, where if someone has boneheadedly specified `after_batch` and `every_n_batches=5` you will see when the proposed divisor is 0 and can fail or spew warnings appropriately.

dmitriy-serdyuk:
Right, it should be more robust. I also forgot, that a user might create an extension with an arbitrary predicate.

dwf:
Ack, I forgot about predicate-modulated conditions too...

blocks/extensions/__init__.py
dmitriy-serdyuk: Yes, `every_n_*` are triggered as `after_*` callbacks, that was the source of my confusion.

dmitriy-serdyuk: See here: https://github.com/mila-udem/blocks/blob/master/blocks/extensions/__init__.py#L249

dmitriy-serdyuk:
I changed it in a way you suggested, tests are passing.

tests/extensions/test_timing.py
```diff
@@ -1,8 +1,20 @@
...
+                             extensions=[Timing(prefix='each'),
+                                         Timing(prefix='each_second',
+                                                every_n_epochs=2),
+                                         FinishAfter(after_n_epochs=epochs)])
     main_loop.run()
+    iterations = int(main_loop.log.status['iterations_done'] / epochs)
+    assert_allclose(
+        (main_loop.log[iterations]['each_time_train_this_epoch'] +
+         main_loop.log[iterations]['each_time_train_this_epoch']) / 2,
+        main_loop.log.current_row['each_second_time_train_this_epoch'],
```
rizar: Our syntax check is currently broken, but I think I see a syntax error here. Can you please run flake8 locally?

dmitriy-serdyuk: I've run it, formatting is fine. Why do you think it's broken?

rizar: OK, my bad.

rizar:
I would say LGTM, but @dwf  was the first to start reviewing this PR.

BTW, I will probably add wall-time monitoring to this extension right after this PR is merged.

dwf:
LGTM!

PR #1111, opened 2016-06-07
__repr__ for initialization schemes

rizar:
Long ago, there was a PR #504, which was never finished. Hopefully this will be merged faster :)

This is just pretty printing for initialization schemes, a very necessary feature, I think.

### DIFF
--- a/blocks/initialization.py
+++ b/blocks/initialization.py
@@ -6,6 +6,8 @@ import numpy
 import theano
 from six import add_metaclass
 
+from blocks.utils import repr_attrs
+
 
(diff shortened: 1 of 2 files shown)


blocks/initialization.py
```diff
@@ -6,6 +6,8 @@
 import theano
 from six import add_metaclass
 
+from blocks.utils import repr_attrs
```
dmitriy-serdyuk: Let's use relative imports! `.utils`

dmitriy-serdyuk:
I like it, merge if you wish, but it would be cool to switch to relative imports.

PR #1113, opened 2016-06-08
support recurrent with no states.

Beronx86:
The recurrent wrapper does not support loop with no states. But this kind of loop may be useful. So I modified the codes.

### DIFF
--- a/blocks/bricks/recurrent/base.py
+++ b/blocks/bricks/recurrent/base.py
@@ -168,28 +168,31 @@ def recurrent(*args, **kwargs):
                     logger.warning("unknown input {}".format(value) +
                                    unknown_scan_input)
 
-            # Ensure that all initial states are available.
-            initial_states = brick.initial_states(batch_size, as_dict=True,
-                                                  *args, **kwargs)
-            for state_name in application.states:
-                dim = brick.get_dim(state_name)
-                if state_name in kwargs:
-                    if isinstance(kwargs[state_name], NdarrayInitialization):
-                        kwargs[state_name] = tensor.alloc(
-                            kwargs[state_name].generate(brick.rng, (1, dim)),
-                            batch_size, dim)
-                    elif isinstance(kwargs[state_name], Application):
-                        kwargs[state_name] = (
-                            kwargs[state_name](state_name, batch_size,
-                                               *args, **kwargs))
-                else:
-                    try:
-                        kwargs[state_name] = initial_states[state_name]
-                    except KeyError:
-                        raise KeyError(
-                            "no initial state for '{}' of the brick {}".format(
-                                state_name, brick.name))
-            states_given = dict_subset(kwargs, application.states)
+            if len(application.states):
+                # Ensure that all initial states are available.
+                initial_states = brick.initial_states(batch_size, as_dict=True,
+                                                      *args, **kwargs)
+                for state_name in application.states:
+                    dim = brick.get_dim(state_name)
+                    if state_name in kwargs:
+                        if isinstance(kwargs[state_name], NdarrayInitialization):
+                            kwargs[state_name] = tensor.alloc(
+                                kwargs[state_name].generate(brick.rng, (1, dim)),
+                                batch_size, dim)
+                        elif isinstance(kwargs[state_name], Application):
+                            kwargs[state_name] = (
+                                kwargs[state_name](state_name, batch_size,
+                                                   *args, **kwargs))
+                    else:
+                        try:
+                            kwargs[state_name] = initial_states[state_name]
+                        except KeyError:
+                            raise KeyError(
+                                "no initial state for '{}' of the brick {}".format(
+                                    state_name, brick.name))
+                states_given = dict_subset(kwargs, application.states)
+            else:
+                states_given = {}
 
             # Theano issue 1772
             for name, state in states_given.items():


dwf:
I'll let someone more familiar with `recurrent` do the review but I know they will ask for you to add a test. :smile:

Beronx86:
Ok, I'll write the test case.

blocks/bricks/recurrent/base.py
```diff
@@ -168,28 +168,31 @@ def recurrent_apply(brick, application, application_call,
...
-                    except KeyError:
-                        raise KeyError(
-                            "no initial state for '{}' of the brick {}".format(
-                                state_name, brick.name))
-            states_given = dict_subset(kwargs, application.states)
+            if len(application.states):
+                # Ensure that all initial states are available.
+                initial_states = brick.initial_states(batch_size, as_dict=True,
+                                                      *args, **kwargs)
+                for state_name in application.states:
```
rizar: It seems like the code starting from this line can be moved out of the `if` clause, and the `else` part is not really necessary. Right now we pay a high price of having an additional level of indentation for this new feature, and it would be great to keep the complexity of the code down.

Beronx86:
In the original code, the `recurrent` method would require `initial_states` function whether there is `recurrent states` or not. If the `recurrent states` is left empty, a error would occur at the time of visiting `initial_states` function.

Beronx86:
You may produce the error with the following code. The error occurs when the class does not contain a recurrent method named `apply`

```Python
import numpy
import theano
from numpy.testing import assert_allclose
from theano import tensor

from blocks.bricks import Brick
from blocks.bricks.recurrent import BaseRecurrent, recurrent
# from recurrent import recurrent


class RecurrentWrapperNoStatesClass(BaseRecurrent):
    def __init__(self, dim, **kwargs):
        super(RecurrentWrapperNoStatesClass, self).__init__(**kwargs)
        self.dim = dim

    def get_dim(self, name):
        if name in ['inputs', 'outputs', 'outputs_2']:
            return self.dim
        if name == 'mask':
            return 0
        return super(RecurrentWrapperNoStatesClass, self).get_dim(name)

    @recurrent(sequences=['inputs', 'mask'], states=[],
               outputs=['outputs', 'outputs_2'], contexts=[])
    def apply2(self, inputs, mask=None):
        outputs = inputs * 10
        outputs_2 = tensor.sqr(inputs)
        if mask:
            outputs *= mask
            outputs_2 *= mask
        return outputs, outputs_2


if __name__ == '__main__':
    recurrent_examples = RecurrentWrapperNoStatesClass(
        dim=11, name='test_example')

    X = tensor.tensor3('X')
    out, out_2 = recurrent_examples.apply2(inputs=X, mask=None)

    x_val = numpy.random.uniform(size=(5, 1, 1))
    x_val = numpy.asarray(x_val, dtype=theano.config.floatX)

    out_eval = out.eval({X: x_val})
    out_2_eval = out_2.eval({X: x_val})

    assert_allclose(x_val * 10, out_eval)
    assert_allclose(numpy.square(x_val), out_2_eval)
```

blocks/bricks/recurrent/base.py
dmitriy-serdyuk: I suggest to add the line before 
```python
else:
    initial_states = OrderedDict()
```

blocks/bricks/recurrent/base.py
```diff
@@ -168,28 +168,31 @@ def recurrent_apply(brick, application, application_call,
...
+                    else:
+                        try:
+                            kwargs[state_name] = initial_states[state_name]
+                        except KeyError:
+                            raise KeyError(
+                                "no initial state for '{}' of the brick {}".format(
+                                    state_name, brick.name))
+                states_given = dict_subset(kwargs, application.states)
+            else:
+                states_given = {}
```
dmitriy-serdyuk: If I remember right, it should be an `OrderedDict`.

blocks/graph/__init__.py
```diff
@@ -104,7 +104,15 @@ def auxiliary_variables(self):
     @property
     def scan_variables(self):
         """Variables of Scan ops."""
-        return list(chain(*[g.variables for g in self._scan_graphs]))
```
Beronx86: This code supposed that no recurrent class is nested. #1115

blocks/bricks/recurrent/base.py
Beronx86: Since states_given in the else clause is never used, it does not matter whether it is a OrderedDict, dict or None.

blocks/graph/__init__.py
```diff
@@ -104,7 +104,15 @@ def auxiliary_variables(self):
...
+        # BFS
+        scan_graphs = self._scan_graphs
```
rizar: You probably want to copy `scan_graphs` here, like e.g. `scan_graphs = list(self._scan_graphs)`.

blocks/graph/__init__.py
```diff
@@ -104,7 +104,15 @@ def auxiliary_variables(self):
...
+        var_list = []
+        while scan_graphs:
+            g = scan_graphs.pop(0)
+            var_list.append(g.variables)
+            if g._scan_graphs:
+                scan_graphs.extend(g._scan_graphs)
+        return list(chain(*var_list))
```
rizar: This is a desirable change, but I have a few concerns:
 - modifying the list that you iterate over is hard to understand and can very likely cause errors. Please try to  simplify the code
- a test is required

blocks/bricks/recurrent/base.py
```diff
@@ -168,28 +168,31 @@ def recurrent_apply(brick, application, application_call,
...
-                    try:
-                        kwargs[state_name] = initial_states[state_name]
-                    except KeyError:
-                        raise KeyError(
-                            "no initial state for '{}' of the brick {}".format(
-                                state_name, brick.name))
-            states_given = dict_subset(kwargs, application.states)
+            if len(application.states):
+                # Ensure that all initial states are available.
+                initial_states = brick.initial_states(batch_size, as_dict=True,
```
rizar: Again, this seems to be the only line that you want to prevent from being called when `application.states` is not given. It does not seem to be necessary to make a very big `if` clauses that damages code readability (and this is already quite complex code).

blocks/bricks/recurrent/base.py
```diff
@@ -46,6 +46,9 @@ def initial_states(self, batch_size, *args, **kwargs):
             The keyword arguments of the application call.
 
         """
+        if not hasattr(self, 'apply') or not self.apply.states:
+            return
+
```
dmitriy-serdyuk: Can you explain how it works? I cannot immediately see it.

Beronx86: when some subclass call the default `initial_states` function in the `BaseRecurrent` class. This line would check whether it is necessary to return the initial states. If the subclass does not have an `apply` method or its `apply` method does not contain `states`, the `initial_states` would not return anything.
This line would make it to support `recurrent` class with no `apply` function or with no `states`.

dmitriy-serdyuk: Why do you want to have a class without `apply`? It's a mistake if a user forgot to define `apply` and the best is to crash soon.

In a case if `apply.states` is empty, `initial_states` would return an empty list before this change, why is it wrong?

Beronx86: You're right about the `apply` method. The `Brick` subclass should follow some design rules. But no code checks whether there is an `apply` method in a `Brick` subclass at present.

rizar: @Beronx86 , checking `apply.states` in `BaseRecurrent.initial_states` is not a solution. There are quite a few places in Blocks-dependent code where `initial_states` method is overloaded. Instead, like in your previous solution, `initial_states` should not be called if `application` does not have `states`. Can you please revert back to the previous version of your fix?

Beronx86: @rizar I think this check could be carried out in `Brick.__init__` method. I reverted back the changes in `BaseRecurrent`.

rizar:
I don't understand, now you have removed your fix, and it is again not supported to have no states property. Why not just implemented like you did in the first place, but with more gentle changes to the code as I suggested?

PR #1120, opened 2016-06-17
Refactor BatchNormalization to make it more extensible

### LINKED ISSUE #1119: Make BatchNormalization more subclassable
Specifically, I'm thinking:

* Factor out allocation of population statistics in a separate method. This lets you disable them for variants that don't want/need them.
* Remove direct dependence on them from the implementation, already partly accomplished by `_prepare_population_statistics`. Move assertions involving them to tests.

dwf:
Fixes #1119.


dwf:
Can I get an Amen? 😛

dmitriy-serdyuk:
Looks good

PR #1121, opened 2016-06-18
Improve stability of batch norm

dwf:
In fact, the thing I've been doing in `bn.py` is... not ideal. Timing suggests that on a rather juicy example (batch size = 1024, 3x256x256) this is only about 1.5% slower.

Memory profiles incoming.

### DIFF
--- a/blocks/bricks/bn.py
+++ b/blocks/bricks/bn.py
@@ -198,8 +198,13 @@ class BatchNormalization(RNGMixin, Feedforward):
         if self.mean_only:
             stdev = tensor.ones_like(mean)
         else:
-            var = (tensor.sqr(input_).mean(axis=axes, keepdims=True) -
-                   tensor.sqr(mean))
+            # We already have the mean; this saves Theano the trouble of
+            # optimizing the graph produced by tensor.var().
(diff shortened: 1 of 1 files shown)


dwf:
Memory profile remains the same, which I guess is not that surprising. Here's the code I used to benchmark.

```python
from blocks.bricks import BatchNormalization
from blocks.utils import shared_floatx
import numpy
import theano
import time

bn = BatchNormalization((3, 256, 256))
inp = shared_floatx(numpy.random.RandomState(1).normal(size=(1024, 3, 256, 256)))

with bn:
    graph = bn.apply(inp)

f = theano.function([], graph)

times = []
N = 1000
for i in range(N):
    t1 = time.clock()
    f()
    t2 = time.clock()
    times.append(t2 - t1)
print(numpy.mean(times), '+/-', numpy.std(times) / numpy.sqrt(N))
```

dmitriy-serdyuk:
I guess, Theano compiles both versions to almost the same code for simple graphs. Problems usually start when you have a crazy graph with thousands of nodes.

dmitriy-serdyuk:
LGTM

PR #1126, opened 2016-07-05
add before_batch to SimpleExtension

pbrakel:
See title.

### DIFF
--- a/tests/extensions/test_extensions.py
+++ b/tests/extensions/test_extensions.py
@@ -130,3 +130,19 @@ def test_composite_extension_different_schedules():
     comp.do.assert_called_once_with('before_training')
     a.do.assert_called_once_with('after_training')
     b.do.assert_called_once_with('after_batch')
+
+
+def test_simple_extension_before_batch_callback():
+
+    class Foo(SimpleExtension):
+        def __init__(self, **kwargs):
+            self.do = Mock()
+            super(Foo, self).__init__(**kwargs)
+
+        def do(self, which_callback, *args):
+            pass
+
+    ext = Foo(before_batch=True)
+    ext.main_loop = Mock()
+    ext.dispatch('before_batch')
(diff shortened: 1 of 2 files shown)


rizar:
This is perfectly reasonable change, but I totally don't understand how fully-neural-lvsr worked before it. And before we figure out, we should not merge it. Mystery...

rizar:
No mystery, I just added it to the subtree: https://github.com/rizar/fully-neural-lvsr/commit/c7d791414d2a9470528da9153d60fe0f932e00fb

Can you please add a test? I will be happy to merge after that.

pbrakel:
Yes, I didn't figure it out from scratch but just saw that it was there in the subtree version. :)

I'll think of a simple test that's compatible with the other callbacks.

pbrakel:
I added a test.

rizar:
Cool, thanks!

tests/extensions/test_extensions.py
```diff
@@ -130,3 +130,19 @@ def do(self, *args):
...
+
+
+def test_simple_extension_before_batch_callback():
+
+    class Foo(SimpleExtension):
+        def __init__(self, **kwargs):
+            self.do = Mock()
+            super(Foo, self).__init__(**kwargs)
+
+        def do(self, which_callback, *args):
```
dmitriy-serdyuk: There is no sense to add do method, it is already defined as a mock.

dmitriy-serdyuk: Too late, but it is a note for future refactoring.

pbrakel: You're right. I actually copied that part from another test and it seems there are a couple of locations in this file where this happens. Again, also for future refactoring :-).

rizar: Oh, crap, I missed this one.

PR #1128, opened 2016-07-05
Get rid of the unnecessary do method

### LINKED ISSUE #1127: Refactor simple extensions in tests
See discussion in #1126 .

rizar:
Fixes #1127


rizar:
Turns out this is not really possible:

```
 Can't instantiate abstract class Foo with abstract methods do
```

PR #1131, opened 2016-07-12
Provide exception to `on_error` extensions

dmitriy-serdyuk:
It make a lot of sense to send an exception to the on_error extensions.

It makes possible to write this cool extension which is understands the traceback:
```python
class IPDB(SimpleExtension):
    """Runs ipdb debugger when an error occurs."""
    def __init__(self, **kwargs):
        kwargs.setdefault('on_error', True)
        super(IPDB, self).__init__(**kwargs)

    def do(self, which_callback, *args):
        if which_callback == 'on_error':
            exception, = args
            import ipdb
            ipdb.post_mortem(exception.__stacktrace__)
```

Simple postmortem debug is not able to navigate the stack trace.

### DIFF
--- a/blocks/extensions/__init__.py
+++ b/blocks/extensions/__init__.py
@@ -72,9 +72,15 @@ class TrainingExtension(object):
         pass
 
     @callback
-    def on_error(self):
-        """The callback invoked when an error occurs."""
-        pass
+    def on_error(self, exception):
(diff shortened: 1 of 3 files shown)


rizar:
Looks good to me! Sorry for a very late reply.

PR #1132, opened 2016-07-26
Make variable filter test insensitive to order.

davidbau:
This test was previously sensitive to the order in which cg.variables is listed; now the test does not assume a particular ordering.

### DIFF
--- a/tests/test_variable_filter.py
+++ b/tests/test_variable_filter.py
@@ -28,16 +28,16 @@ def test_variable_filter():
 
     # Testing filtering by role
     role_filter = VariableFilter(roles=[PARAMETER])
-    assert parameters == role_filter(cg.variables)
+    assert set(parameters) == set(role_filter(cg.variables))
     role_filter = VariableFilter(roles=[FILTER])
     assert [] == role_filter(cg.variables)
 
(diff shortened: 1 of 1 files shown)


davidbau:
This test fails for me because cg.variables seems to be coming back in an order different from the one expected by the test.  Maybe theano changed the order?  I am using Theano 0.9.0.dev2.

rizar:
Are you running the test on GPU? I assume so, because Travis seems to be quite happy with your PR.

Can you please share the error message with us, so that we know in which line the error happens?

davidbau:
Yes, I have gpu enabled.  Not the right way to run tests?

Error message is just this:

(env) davidbau@titan:~/git/blocks (master)$ nose2 tests.test_variable_filter
Using gpu device 0: GeForce GTX TITAN X (CNMeM is disabled, cuDNN 5005)
F..
======================================================================
FAIL: tests.test_variable_filter.FunctionTestCase (test_variable_filter)
  File "/home/davidbau/git/blocks/tests/test_variable_filter.py", line 31, in test_variable_filter
    assert parameters == role_filter(cg.variables)
AssertionError

Ran 3 tests in 0.072s

FAILED (failures=1)

PR #1134, opened 2016-08-01
Fix a bug involving return_initial_states=True and outputs

rizar:
`return_initial_states=True` did not work when not all of the outputs were before all the states.

@dmitriy-serdyuk , this should be an easy PR to review

@pbrakel , this should bring us one step closer towards using Blocks master

### DIFF
--- a/blocks/bricks/recurrent/base.py
+++ b/blocks/bricks/recurrent/base.py
@@ -227,10 +227,11 @@ def recurrent(*args, **kwargs):
             result = pack(result)
             if return_initial_states:
                 # Undo Subtensor
-                for i in range(len(states_given)):
-                    assert isinstance(result[i].owner.op,
-                                      tensor.subtensor.Subtensor)
-                    result[i] = result[i].owner.inputs[0]
+                for i, info in enumerate(outputs_info):
+                    if info is not None:
+                        assert isinstance(result[i].owner.op,
+                                        tensor.subtensor.Subtensor)
+                        result[i] = result[i].owner.inputs[0]
             if updates:
                 application_call.updates = dict_union(application_call.updates,
                                                       updates)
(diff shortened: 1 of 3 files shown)


blocks/bricks/recurrent/base.py
```diff
@@ -227,10 +227,11 @@ def scan_function(*args):
...
             if return_initial_states:
                 # Undo Subtensor
-                for i in range(len(states_given)):
-                    assert isinstance(result[i].owner.op,
-                                      tensor.subtensor.Subtensor)
-                    result[i] = result[i].owner.inputs[0]
+                for i, info in enumerate(outputs_info):
+                    if info is not None:
+                        assert isinstance(result[i].owner.op,
+                                        tensor.subtensor.Subtensor)
```
dmitriy-serdyuk: Indent is wrong here.

blocks/bricks/recurrent/base.py
```diff
@@ -227,10 +227,11 @@ def scan_function(*args):
             result = pack(result)
             if return_initial_states:
                 # Undo Subtensor
-                for i in range(len(states_given)):
-                    assert isinstance(result[i].owner.op,
-                                      tensor.subtensor.Subtensor)
-                    result[i] = result[i].owner.inputs[0]
+                for i, info in enumerate(outputs_info):
+                    if info is not None:
```
dmitriy-serdyuk: This relies on line 216, right? Can you add a comment that None corresponds to an output and shouldn't be included?

rizar: The convention that `None` corresponds to an output is not Blocks-specific. It is a part of `theano.scan` signature.

rizar:
A fixed the syntax error you pointed out and also a few others in the same file.

blocks/bricks/recurrent/base.py
dmitriy-serdyuk: Right, agreed.

dmitriy-serdyuk:
LGTM.

PR #1136, opened 2016-08-01
Fix Timing extension again

dmitriy-serdyuk:
I introduced a bug when was fixing #1098, timing extension didn't report time reading data because it wasn't able to distinguish it from other actions.

### DIFF
--- a/tests/extensions/test_timing.py
+++ b/tests/extensions/test_timing.py
@@ -18,3 +18,5 @@ def test_timing():
          main_loop.log[iterations]['each_time_train_this_epoch']) / 2,
         main_loop.log.current_row['each_second_time_train_this_epoch'],
         atol=1e-2)
+    assert 'each_time_read_data_this_epoch' in main_loop.log[iterations]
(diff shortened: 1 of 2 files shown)


dmitriy-serdyuk:
@dwf can you take a look?

tests/extensions/test_timing.py
```diff
@@ -18,3 +18,5 @@ def test_timing():
          main_loop.log[iterations]['each_time_train_this_epoch']) / 2,
         main_loop.log.current_row['each_second_time_train_this_epoch'],
         atol=1e-2)
+    assert 'each_time_read_data_this_epoch' in main_loop.log[iterations]
```
dwf: These names don't strike me as particularly clear.

dmitriy-serdyuk: This is only for testing purposes. I used "each_time" prefix, so by default it would be "read_data_this_epoch", like before.

dmitriy-serdyuk:
@dwf, should I change naming in that test?

dwf:
Oops, lost track of this one. No, it's fine.

PR #1138, opened 2016-08-18
Rebased requires-io

dmitriy-serdyuk:
See #977


dmitriy-serdyuk:
That's funny. Conda is always a little bit behind => tests are always going to fail => we can never merge requires.io requests => dependencies never updated.

dmitriy-serdyuk:
According to [this](https://requires.io/features/), directives I added makes requires.io think that pillow is up-to-date. I'm not sure if it will be trying to create pull requests.

Who knows how PR from requires works? I guess, I'll have to force push these changes to the `requires-io-master` branch.

dmitriy-serdyuk:
@dwf, could you take a look? @nouiz, do you have experience with requires, what do you think about these problems?

dwf:
LGTM but I don't really know anything about requires.io.

dmitriy-serdyuk:
I'll merge it to be up-to-date and let's see what happens with #977.

nouiz:
I never used require.io. But I think it make PR to have your repo use the
up to date dependencies. So it mostly push you to make sure you support the
newer version of your dependencies. Not a bad idea.

I don't know why there is so much commit. If the repo have travis that test
PR, I guess that making one commit each time a dependencies is updated will
trigger the tests of all updates. So you will more clearly see which
updates your software don't support. Then it is up to you to merge the PR.
Making one PR for each update if you don't merge them isn't interresting,
so for me it make sence if they really work as I would have done it. Don't
forget, there is many speculation in what I wrote, maybe they just had a
bug...

PR #1139, opened 2016-08-18
Move copy-and-tag to bricks.base namespace.

dwf:
This lets advanced users manually tag inputs to an application (if they aren't directly arguments to the apply method, but perhaps contained in a container or attributes of some object). I see no harm in exposing this as a module-level function.


dmitriy-serdyuk:
Looks fine.

PR #1141, opened 2016-08-18
Accept list of pairs for gradients in GradientDescent.

dwf:
Be liberal about what you accept and conservative about what you emit.

### DIFF
--- a/blocks/algorithms/__init__.py
+++ b/blocks/algorithms/__init__.py
@@ -268,6 +268,9 @@ class GradientDescent(UpdatesAlgorithm):
         # Set initial values for cost, parameters, gradients.
         self.cost = cost
         self.parameters = parameters
+        # Coerce lists of tuples to OrderedDict. Do not coerce Mappings,
+        # as we don't want to convert dict -> OrderedDict and give it
+        # an arbitrary, non-deterministic order.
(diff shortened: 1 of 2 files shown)


dwf:
Okay, tests pass now. @dmitriy-serdyuk I think this is pretty uncontroversial.

blocks/algorithms/__init__.py
```diff
@@ -266,6 +268,8 @@ def __init__(self, cost=None, parameters=None, step_rule=None,
         # Set initial values for cost, parameters, gradients.
         self.cost = cost
         self.parameters = parameters
+        if gradients is not None and not isinstance(gradients, Mapping):
```
dmitriy-serdyuk: Do you need to check if it is `Mapping`? Can gradients just always be converted to `OrderedDict`?

dwf: The reason we check if it's not Mapping is that we don't want to create a dict to an OrderedDict and give it an arbitrary order.

dwf: ceeabeb clarifies this in a comment.

dmitriy-serdyuk:
It makes sense, LGTM.

PR #1143, opened 2016-09-02
Fixes #1099, use `.rsplit(..., 1)`

### LINKED ISSUE #1099: split_suffix in RecurrentStack does not handle names with multiple # correctly

dathinab:
The new pull request like requested
(see https://github.com/mila-udem/blocks/pull/1100 for the original request)

### DIFF
--- a/blocks/bricks/recurrent/misc.py
+++ b/blocks/bricks/recurrent/misc.py
@@ -186,9 +186,9 @@ class RecurrentStack(BaseRecurrent, Initializable):
     @staticmethod
     def split_suffix(name):
         # Target name with suffix to the correct layer
-        name_level = name.split(RECURRENTSTACK_SEPARATOR)
+        name_level = name.rsplit(RECURRENTSTACK_SEPARATOR, 1)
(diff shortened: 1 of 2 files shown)


rizar:
Thank you!

PR #1146, opened 2016-09-12
Fix StepClipping return when threshold=None.

### LINKED ISSUE #1145: StepClipping doesn't work when threshold is None
An oversight here: https://github.com/mila-udem/blocks/blob/master/blocks/algorithms/__init__.py#L705

(No empty updates list returned.)

dwf:
Fixes #1145.

N.B. I also consider it supremely weird that the threshold attribute isn't assigned in the case that it is None. I can change that too if asked.

### DIFF
--- a/blocks/algorithms/__init__.py
+++ b/blocks/algorithms/__init__.py
@@ -696,12 +696,13 @@ class StepClipping(StepRule):
 
     """
     def __init__(self, threshold=None):
-        if threshold:
-            self.threshold = shared_floatx(threshold, "threshold")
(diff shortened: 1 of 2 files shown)


dmitriy-serdyuk:
Is there a case when using the clipping with None argument is easier than not using the clipping at all? I mean, the threshold shouldn't be optional.

dwf:
Makes it easier to conditionally disable clipping. I was definitely pleased when I found the option.

dmitriy-serdyuk:
I see. Maybe it makes sense to assign the attribute to None or inf, some extensions might want to access it.

Otherwise, LGTM.

PR #1147, opened 2016-10-03
Fix "read_data" timing

rizar:
I believe that the code that check that this isn't a duplicate call of `Timing` should be outside of `for action in ... ` loop. @dmitriy-serdyuk , do you agree? I think it was you who refactored this extension last.


rizar:
Never mind, I was using an old version of Blocks.

PR #1148, opened 2016-10-11
Environment variable to set config.temp_dir

rizar:

PR #1149, opened 2016-10-11
Fix small bug in serialization.

Thrandis:
Found this small bug in `serialization.py`:

`theano.sandbox.gpuarray` is now in `theano.gpuarray`.

### DIFF
--- a/blocks/serialization.py
+++ b/blocks/serialization.py
@@ -561,7 +561,7 @@ def _recreate_cuda_ndarray(_, content):
 
 
 def _recreate_pygpu_array(context_name, content):
-    context = theano.sandbox.gpuarray.get_context(context_name)
+    context = theano.gpuarray.get_context(context_name)
     return pygpu.gpuarray.array(content, context=context)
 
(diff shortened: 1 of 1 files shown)


dwf:
Failure on 3.4.x is unrelated, same as #1148...

dwf:
@Thrandis can you rebase so the tests run with the updated Travis config?

rizar:
For the record, I had troubles to reproduce the issue that @Thrandis is
talking about. @Thrandis, was this also because of test values?

Thrandis:
@rizar This one is unrelated to the other issue I have.

rizar:
OK, but I remember you were saying that it occupies a GPU not when pygpu is
imported, but at some random line in blocks/filter.py?

Thrandis:
@rizar this is just a bug fix for the wrong path of `gpuarray` which is not in the sandbox anymore. If you use the serialization with the new GPU backend, it will simply crash. For the other problem, it is in issue #1152.

rizar:
Oh, sorry! I confused it with the other issue. Stupid me :(

dwf:
Remaining failure is unrelated.

PR #1150, opened 2016-10-12
Switch to Miniconda3 for Python 3.x Travis runs

dwf:


dwf:
@bartvm Miniconda->Miniconda3 seemed to fix it. Shall I merge?

dwf:
(Rebasing out the debug commit.)

dmitriy-serdyuk:
But it might cause problems with Python 2 tests. I propose to do something like
```
   - |
       if [[ $TRAVIS_PYTHON_VERSION == 'python2' ]]; then
         wget -q http://repo.continuum.io/miniconda/Miniconda-latest-Linux-x86_64.sh -O miniconda.sh
       else
         wget -q http://repo.continuum.io/miniconda/Miniconda3-latest-Linux-x86_64.sh -O miniconda.sh
       fi
```

dwf:
Why would it cause problems with the Python 2 tests if Miniconda2 didn't cause problems with the Python 3 tests?

dmitriy-serdyuk:
It just did. It is possible that in the future some package from the default Miniconda 3 installation will have conflicts when changing to python 2. Just like it happened now but other way around.

dwf:
Seems less likely as I suspect Python2.7 environments willl be officially
supported on Miniconda3 after Miniconda2 is retired. But I've added the
conditional fetch.

dwf:
Sigh, stupid config file formats.

dwf:
Gonna assume that this has your approval @dmitriy-serdyuk since I more or less copied your comment 😛.

PR #1157, opened 2016-10-22
Add a Timestamp extension.

dwf:
This is way more useful than I would have realized, particularly when trying to reconstruct weird things post-mortem.


dmitriy-serdyuk:
This is great, I will definitely use it! LGTM

dwf:
Yeah, one of those things that is obvious in hindsight.

PR #1160, opened 2016-10-27
Let Sequence [subclasses] accept Bricks.

dwf:
To this day I sometimes forget the `.apply` when creating a Sequence brick. This removes the need. It also simplifies the implementations of Sequence and MLP in the process.

### DIFF
--- a/blocks/bricks/conv.py
+++ b/blocks/bricks/conv.py
@@ -441,9 +441,9 @@ class ConvolutionalSequence(Sequence, Initializable, Feedforward):
     ----------
     layers : list
         List of convolutional bricks (i.e. :class:`Convolutional`,
-        :class:`ConvolutionalActivation`, or :class:`Pooling` bricks).
-        :class:`Activation` bricks that operate elementwise can also
-        be included.
+        :class:`ConvolutionalActivation`, or :class:`Pooling` bricks),
+        or application methods from such bricks.  :class:`Activation`
+        bricks that operate elementwise can also be included.
     num_channels : int
         Number of input channels in the image. For the first layer this is
         normally 1 for grayscale images and 3 for color (RGB) images. For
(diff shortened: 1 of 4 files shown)


dwf:
Don't know what Scrutinizer is on about. Ready for review.

dwf:
Coverage decreased because I eliminated a bunch of code. Figures...

rizar:
I am really looking forward to review this, but the combined pressure from
ICLR deadline and a lengthy upcoming talk that I have to give leaves little
room to anything else. I will take a look at it the first thing after the
deadline, I promise!

dwf:
No worries, I have been using it for a week and it has substantially improved my life. 😉

dwf:
Anyone want to review this with ICLR out of the way?

rizar:
I want. Will do it tomorrow. Sorry for the long wait.

rizar:
There a was a lengthy discussion on whether children should be specified by means of giving the list of bricks or the list of application methods. It did not converge, but people mostly leaned to always using application methods.

That said, Blocks de-factor is in maintenance-only mode, and big structural changes need a lot of man-power. So maybe we can just merge your PR and thereby increase the consistency of the interface.

So much for the high-level considerations, now I will read the code.

blocks/bricks/conv.py
```diff
@@ -494,16 +494,15 @@ class ConvolutionalSequence(Sequence, Initializable, Feedforward):
     def __init__(self, layers, num_channels, batch_size=None,
                  image_size=(None, None), border_mode=None, tied_biases=None,
                  **kwargs):
-        self.layers = layers
+        self.layers = [a if isinstance(a, Brick) else a.brick for a in layers]
```
rizar: Instead of creating a new property, we can just return a tuple, can't we?

rizar:
Looks good to me, high-level concerns put apart.

blocks/bricks/conv.py
dwf: See #1163. If `application_methods` gets modified you want that change to be reflected in `layers`. Having the same information in two places is asking for one of them to become stale (we already sort of have that, in that we have to mirror changes to `application_methods` with possible changes to `children`, which is what #1163 addresses in order to make Sequence surgery less of a pain; having to also manage `layers` for `ConvolutionalSequence` is yet another layer).

dwf:
Well I think this preserves the high-level structure of things but adds a layer of convenience. Guess I'll merge.

PR #1162, opened 2016-10-29
AggregationSchemes for Maximum and Minimum.

### LINKED ISSUE #129: More Aggregation Schemes: Min, Max, Concatenate
That would a nice exercise for a newcomer: to add a couple of new aggregation schemes in addition to existing ones. Would be very helpful for debugging purposes.

dwf:
~~Surprised to find these didn't already exist.~~ Fixes #129.

### DIFF
--- a/blocks/monitoring/aggregation.py
+++ b/blocks/monitoring/aggregation.py
@@ -214,6 +214,39 @@ class Maximum(Minimum):
 maximum = partial(_simple_aggregation, Maximum)
 
 
+class Concatenate(Minimum):
+    """Aggregation scheme which remembers values from all batches.
+
+    Parameters
+    ----------
+    variable: :class:`~tensor.TensorVariable`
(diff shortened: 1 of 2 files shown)


dmitriy-serdyuk:
It is issue #129 .

dwf:
So it is.

dwf:
I added a Concatenate scheme for completeness.

dmitriy-serdyuk:
Sorry, forgot about this PR. It looks good. I have one question: do you think that concatenate aggregator should always create an extra dim? For example, in your test `[4, 12, 7, 5]` should be `[[4, 5], [7, 5]]`.

dwf:
That would make things simpler wouldn't it? Changed in 7e8a556.

dmitriy-serdyuk:
Cool, LGTM.

PR #1163, opened 2016-11-11
Make the Sequence brick implement the Sequence ADT.

dwf:
This makes it a lot easier to do surgery on Sequence bricks. Say, for example, you have some code that generates a sequence and you want to customize it by adding or replacing layers; manually managing the `.children` and `.application_methods` attributes is a huge pain.

Marginally related to #1160.

### DIFF
--- a/blocks/bricks/sequences.py
+++ b/blocks/bricks/sequences.py
@@ -59,10 +59,14 @@ class Sequence(Brick, MutableSequence):
         return self.application_methods[index]
 
     def __setitem__(self, index, new_application):
+        if isinstance(new_application, Brick):
+            new_application = new_application.apply
         del self[index]
         self.insert(index, new_application)
 
     def insert(self, index, new_application):
+        if isinstance(new_application, Brick):
+            new_application = new_application.apply
         if new_application.brick not in self.children:
             self.children.append(new_application.brick)
         self.application_methods.insert(index, new_application)
--- a/blocks/bricks/sequences.py
+++ b/blocks/bricks/sequences.py
@@ -1,4 +1,5 @@
 """Bricks that compose together other bricks in linear sequences."""
+from collections import MutableSequence
 import copy
 from toolz import interleave, unique
 from picklable_itertools.extras import equizip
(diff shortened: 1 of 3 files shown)


dwf:
Open question is whether `__setitem__` and `insert` should auto-convert Bricks into `b.apply` to match #1160.

dwf:
[(The list of stuff we get from the `MutableSequence` inherit.)](https://docs.python.org/2/library/collections.html#collections-abstract-base-classes)

rizar:
Hmm, I sort of always assumed that the brick's children are fixed after they brick is created. Do you often modify them in-place?

dwf:
Wouldn't say often, but for example I keep a lot of functions around that
build standard-ish architectures and it's often a little simpler to just
call that and then do some surgery before allocation for a slightly
customized thing i want to try.

I'm surprised to hear you say that actually since deleting from the
children list correctly handles deletion from the child's list of parents.
:)

rizar:
The usecase that you refer to is valid. I didn't know that there is some support in the core for deletion of children, @bartvm wrote that part. 

Would like to make `self.layers` of `ConvolutionalSequence` a property now? As you said, it makes sense if we assume that application methods can be removed.

blocks/bricks/sequences.py
```diff
@@ -45,6 +46,32 @@ def apply_inputs(self):
     def apply_outputs(self):
         return self.application_methods[-1].outputs
 
+    def __delitem__(self, index):
+        application_method = self.application_methods[index]
+        brick = application_method.brick
+        try:
+            del self.application_methods[index]
+        finally:
+            apps_with_brick = [a for a in self.application_methods
```
rizar: This look weird to me. Why do you want to modify the brick if an exception was raised on `del self.application_methods[index]`?

dwf: Good question. Removed in 303725a.

dwf:
df8a57e makes the change to `ConvolutionalSequence`. I also added the ability to use insert, etc. with Bricks, assuming this means "use the `.apply` method" as it now does with the `application_methods` argument to the constructor. `__getitem__` always returns a `BoundApplication` though.

rizar:
LGTM!

rizar:
@dwf, looks like I approved this PR more than a month ago. Do you wish to proceed and merge it or should I close it?

PR #1165, opened 2016-11-14
Warn instead of raising when the hierarchy is not respected

rizar:
As much as I like the original idea that a brick can only call application methods of its direct children, I found it necessary multiple times to override this restriction. After all, Blocks is a framework for research and quick prototyping, and very hard restrictions is probably not the best policy. Therefore, I propose to warn the user instead.

### DIFF
--- a/blocks/bricks/base.py
+++ b/blocks/bricks/base.py
@@ -1,4 +1,5 @@
 import inspect
+import warnings
 from abc import ABCMeta
 from collections import OrderedDict
 from six import wraps
@@ -287,11 +288,11 @@ class Application(object):
         last_brick = self.call_stack[-1] if self.call_stack else None
         if (last_brick and brick is not last_brick and
                 brick not in last_brick.children):
-            raise ValueError('Brick ' + str(self.call_stack[-1]) + ' tries '
-                             'to call brick ' + str(self.brick) + ' which '
-                             'is not in the list of its children. This could '
-                             'be caused because an @application decorator is '
-                             'missing.')
+            warnings.warn('Brick ' + str(self.call_stack[-1]) + ' tries '
+                          'to call brick ' + str(self.brick) + ' which '
+                          'is not in the list of its children. This could '
+                          'be caused because an @application decorator is '
+                          'missing.')
         self.call_stack.append(brick)
         try:
             outputs = self.application_function(brick, *args, **kwargs)


dwf:
I tend to prefer WARNING level log messages, as they'll respect any logging
config the user has made (and reliably fire each time the call is made --
since the brick applications are only called during graph construction this
isn't much of a burden) but I'm not sure which convention we use elsewhere
in Blocks. In any case we should be consistent. Otherwise LGTM.

rizar:
Consistent? Dunno what this means...

```
╰─$ grep --include='*.py' -R warnings.warn .                                                                                                                                                                                               1 ↵
./blocks/graph/__init__.py:                    warnings.warn(
./blocks/graph/__init__.py:            warnings.warn(
./blocks/serialization.py:            warnings.warn(
./blocks/log/sqlite.py:        warnings.warn('large objects stored in SQLite' +
./blocks/bricks/base.py:            warnings.warn('Brick ' + str(self.call_stack[-1]) + ' tries '
╭─bahdanau@eos14  ~/Dist/blocks  ‹lenient_sanity_check*› 
╰─$ grep --include='*.py' -R logger.warning .
./blocks/extensions/saveload.py:            logger.warning("No dump found")
./blocks/main_loop.py:                logger.warning("different parameters for model and algorithm")
./blocks/main_loop.py:        logger.warning('Received epoch interrupt signal.' +
./blocks/main_loop.py:        logger.warning('Received batch interrupt signal.' +
./blocks/algorithms/__init__.py:                logger.warning(('{}: gradients already specified directly; '
./blocks/bricks/recurrent/base.py:                    logger.warning("unknown input {}".format(value) +
```

I will go through the codebase and get rid of `warnings.warn`.

rizar:
Actually, turns out there are [guidelines for that](https://docs.python.org/2/howto/logging.html):

``` 	
warnings.warn() in library code if the issue is avoidable and the client application should be modified to eliminate the warning

logging.warning() if there is nothing the client application can do about the situation, but the event should still be noted
```

In the case of this PR we are telling the client that he should rather not do what they do, which corresponds to `warnings.warn`, AFAIU.

dwf:
Well I think there might be some reason to this madness. Looks like it's
mostly the MainLoop using the logger.

dwf:
Sounds good to me then. We should file an issue to convert relevant logging
calls.

rizar:
OK, I will merge the PR and create an issue.

rizar:
After I merged this PR the test have failed with a weird conda-related error:

```
aceback (most recent call last):

      File "/home/travis/miniconda/lib/python2.7/site-packages/conda/exceptions.py", line 479, in conda_exception_handler

        return_value = func(*args, **kwargs)

      File "/home/travis/miniconda/lib/python2.7/site-packages/conda/cli/main.py", line 145, in _main

        exit_code = args.func(args, p)

      File "/home/travis/miniconda/lib/python2.7/site-packages/conda/cli/main_install.py", line 80, in execute

        install(args, parser, 'install')

      File "/home/travis/miniconda/lib/python2.7/site-packages/conda/cli/install.py", line 420, in install

        raise CondaRuntimeError('RuntimeError: %s' % e)

    CondaRuntimeError: Runtime error: RuntimeError: Runtime error: Could not open u'/home/travis/miniconda/pkgs/jpeg-8d-2.tar.bz2.part' for writing (HTTPSConnectionPool(host='repo.continuum.io', port=443): Read timed out.).
```

dwf:
I encountered this before, just restart the build and it should go away.

rizar:
Still broken for Python 3.4, float64: https://travis-ci.org/mila-udem/blocks/jobs/175750646

PR #1168, opened 2016-11-29
added printing status and tests for it

mnuke:
Printing extension had default print_status=True without the ability to change it, I'm just adding it as a an argument.

Also added tests to a new `test_printing.py`, I found there was a printing test in `test_progressbar` but that felt more like a integration test than a unit test for something so simple as the print_status argument.

### DIFF
--- a/blocks/extensions/__init__.py
+++ b/blocks/extensions/__init__.py
@@ -447,13 +447,14 @@ class FinishAfter(SimpleExtension):
 
 class Printing(SimpleExtension):
     """Prints log messages to the screen."""
-    def __init__(self, **kwargs):
+    def __init__(self, print_status=True, **kwargs):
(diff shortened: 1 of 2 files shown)


rizar:
Sorry for the late reply and thanks for the PR! I am a bit hesitant to merge it though. There is so many ways in which `Printing` could be customized, and it would never fit everybody's needs. This is why we have been keeping it primitive as it is so far.

mnuke:
That makes sense, but the code already has just an unused variable `print_status` that is just never set. This change makes it usable and adds the ability to make normal runs (that don't need the status, which I imagine would be the majority) much cleaner.

rizar:
I think most people prefer to print status, because otherwise one doesn't even know how many iterations of training have been done.

PR #1172, opened 2017-01-04
Fix tests for the convolutional brick

rizar:
Apparently the error message has changed in Theano, see #1144


rizar:
Travis failure seems to be random and I believe that this PR is simple enough to be merged without review.

PR #1174, opened 2017-01-05
fix progressbar's iter_per_epoch

mnuke:
the `ProgressBar` extension's `get_iter_per_epoch` was out of date and failing to get the iterations for most of the iteration schemes

- updates the `get_iter_per_epoch` method to use `indices` for iteration schemes that have it
- adds tests for the `get_iter_per_epoch` method
- fixes a little bit of indenting (check line 528 in `__init__.py` to see if it conforms with your standards)

### DIFF
--- a/blocks/extensions/__init__.py
+++ b/blocks/extensions/__init__.py
@@ -524,10 +524,14 @@ class ProgressBar(TrainingExtension):
     def get_iter_per_epoch(self):
         """Try to infer the number of iterations per epoch."""
         iter_scheme = self.main_loop.data_stream.iteration_scheme
-        if hasattr(iter_scheme, 'num_batches'):
-            return iter_scheme.num_batches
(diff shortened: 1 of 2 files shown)


blocks/extensions/__init__.py
```diff
@@ -524,10 +524,14 @@ def __setstate__(self, state):
     def get_iter_per_epoch(self):
         """Try to infer the number of iterations per epoch."""
         iter_scheme = self.main_loop.data_stream.iteration_scheme
-        if hasattr(iter_scheme, 'num_batches'):
-            return iter_scheme.num_batches
+        if (hasattr(iter_scheme, 'indices') and
```
rizar: What would be nice to have here for the sake of readability is the names of the iterations schemes that you have in mind for each of the branches.

mnuke: good idea!

PR #1176, opened 2017-01-24
Fixes for adam

### LINKED ISSUE #1159: Adam moment parameterization is backwards
beta1 and beta2 are 1-beta1 and 1-beta2, with respect to how the final version of the paper uses them.

I spent way too much time trying to figure out why shit was breaking for me, and it was because I'd put in the customary default of 0.999 for beta2, which of course meant that, once you flipped it around, my second moment estimates decayed faster than an anxious squirrel on speed.

We should probably change this after the ICLR deadline along with a big mailing list announcement.

### LINKED ISSUE #1175: Errors in the Adam implementation
I think I discovered some errors in the current implementation of adam.

Together, they cause the learning to fail (tested on a relatively small training set). First, the average cross entropy decreases, but after some epochs (20 in my case) it increases again and soon (after 29 epochs) only `nan` is reported. 

- The initial `learning_rate` ([0.002](https://github.com/mila-udem/blocks/blob/master/blocks/algorithms/__init__.py#L861)) doesn't match the value provided in the [current version 8 of the paper](https://arxiv.org/pdf/1412.6980v8.pdf) (0.001)
I'm aware of #767, but I don't know about the sources for the new proposed learning rate. In my opinion, the default value should match the values used in the paper.
- Adam moment parameterization is backwards (#1159). 
- [The name of `beta_1t`](https://github.com/mila-udem/blocks/blob/master/blocks/algorithms/__init__.py#L880) is misleading in my eyes. It is the decayed value of `beta_1`, but not to the power of `t`
- The learning rate isn't decayed as described in the paper (above Theorem 4.1)
- The influence of `epsilon` is increased, because it's [implicitly divided](https://github.com/mila-udem/blocks/blob/master/blocks/algorithms/__init__.py#L882) by `sqrt(1-b_2**t)`. This could be fixed by explicitly calculating `variance_t_hat`.

After I fixed all these issues in a local version of adam, I was able to train the network (slowly, but) successfully.

This is my current version of Adam:
```
def __init__(self, learning_rate=0.001,
             beta1=0.9, beta2=0.999, epsilon=1e-8,
             decay_factor=(1 - 1e-8)):
    self.learning_rate = shared_floatx(learning_rate, "learning_rate")
    self.beta1 = shared_floatx(beta1, "beta1")
    self.beta2 = shared_floatx(beta2, "beta2")
    self.epsilon = shared_floatx(epsilon, "epsilon")
    self.decay_factor = shared_floatx(decay_factor, "decay_factor")
    for param in [self.learning_rate, self.beta1, self.beta2, self.epsilon,
                  self.decay_factor]:
        add_role(param, ALGORITHM_HYPERPARAMETER)

def compute_step(self, parameter, previous_step):
    logging.warning("Using custom adam code")
    mean = _create_algorithm_buffer_for(parameter, 'mean')
    variance = _create_algorithm_buffer_for(parameter, 'variance')
    time = shared_floatx(0., 'time')
    add_role(time, ALGORITHM_BUFFER)

    t1 = time + 1
    beta_1_decayed = self.beta1 * self.decay_factor ** (t1 - 1)
    learning_rate_decayed = self.learning_rate / tensor.sqrt(t1)
    mean_t = beta_1_decayed * mean + (1. - beta_1_decayed) * previous_step
    variance_t = self.beta2 * variance + \
                  (1. - self.beta2) * tensor.sqr(previous_step)

    mean_t_hat = mean_t / (1. - beta_1_decayed ** t1)
    variance_t_hat = variance_t / (1. - self.beta2 ** t1)

    step = (learning_rate_decayed * mean_t_hat /
            (tensor.sqrt(variance_t_hat) + self.epsilon))

    updates = [(mean, mean_t),
               (variance, variance_t),
               (time, t1)]

    return step, updates
```

If you agree that my modifications are correct, I'll happily create a pull request so they can be incorporated.

ChristopherBrix:
Fixes #1175, fixed #1159 

This pull request
 - Changes the moment parametrization to reflect the current [version 8 of the adam paper](https://arxiv.org/pdf/1412.6980v8.pdf). All `1 - beta` are changed to just `beta`.
 - Renames the variable `beta_1t` to `beta_1_decayed` because the former looks like it is `beta_1 ** t1`.
 - Explicitly sets the `decay_factor` to 1 because `numpy.float32(1) - numpy.float32(1e-8) == numpy.float32(1.0-1e-8) == 1`

### DIFF
--- a/blocks/algorithms/__init__.py
+++ b/blocks/algorithms/__init__.py
@@ -855,12 +855,12 @@ class Adam(StepRule):
     epsilon : float, optional
         Default value is set to 1e-8.
     decay_factor : float, optional
-        Default value is set to 1 - 1e-8.
+        Default value is set to 1.
 
     """
     def __init__(self, learning_rate=0.002,
                  beta1=0.9, beta2=0.999, epsilon=1e-8,
-                 decay_factor=(1 - 1e-8)):
+                 decay_factor=1):
         self.learning_rate = shared_floatx(learning_rate, "learning_rate")
         self.beta1 = shared_floatx(beta1, "beta1")
         self.beta2 = shared_floatx(beta2, "beta2")
--- a/blocks/algorithms/__init__.py
+++ b/blocks/algorithms/__init__.py
@@ -877,11 +877,11 @@ class Adam(StepRule):
         add_role(time, ALGORITHM_BUFFER)
 
         t1 = time + 1
-        beta_1t = self.beta1 * self.decay_factor ** (t1 - 1)
+        beta_1_decayed = self.beta1 * self.decay_factor ** (t1 - 1)
         learning_rate = (self.learning_rate *
                          tensor.sqrt(1. - self.beta2**t1) /
-                         (1. - beta_1t**t1))
-        mean_t = beta_1t * mean + (1. - beta_1t) * previous_step
+                         (1. - beta_1_decayed**t1))
+        mean_t = beta_1_decayed * mean + (1. - beta_1_decayed) * previous_step
         variance_t = (self.beta2 * variance +
                       (1. - self.beta2) * tensor.sqr(previous_step))
         step = (learning_rate * mean_t /
--- a/blocks/algorithms/__init__.py
+++ b/blocks/algorithms/__init__.py
@@ -848,10 +848,10 @@ class Adam(StepRule):
         Default value is set to 0.002.
     beta1 : float, optional
         Exponential decay rate for the first moment estimates.
-        Default value is set to 0.1.
+        Default value is set to 0.9.
     beta2 : float, optional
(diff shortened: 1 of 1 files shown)


dwf:
Just a reminder to send an announcement to blocks-users when this is merged
as it will probably change behaviour subtly in an extremely well-used bit
of code.

dmitriy-serdyuk:
The test fails with a segfault, some problem with travis again.

I checked the math, it is correct (the same as it was before). The float64 test passes probably because of wide tolerance intervals.

rizar:
The test failure is clearly unrelated. Thanks for the reminder to announce the change, @dwf.

Issue #1179, opened 2017-02-17
VisibleDeprecationWarning now results an error


SwordYork:
Hi,

After upgrading python to 3.5, the following warning results an error. I think it is better to change the type of [mask](https://github.com/mila-udem/blocks/blob/master/blocks/search.py#L275) or change `outputs = [list(output[:mask.sum()])` ([line 323](https://github.com/mila-udem/blocks/blob/master/blocks/search.py#L323)) to `outputs = [list(output[:int(mask.sum())])`.
I'm not sure which way is better.

> python3.4/site-packages/blocks/search.py:324: VisibleDeprecationWarning: using a non-integer number instead of an integer will result in an error in the future
>   for output, mask in equizip(outputs, masks)]

Thanks.

dmitriy-serdyuk:
Thank you! Could you make a PR? Perhaps, we should start testing with 3.5 version.

PR #1180, opened 2017-02-22
cast the non-integer value `mask.sum()` into an integer #1179

SwordYork:
fix VisibleDeprecationWarning before Python 3.4 which will result an error starting from Python 3.5.
See #1179.


dmitriy-serdyuk:
Thank you!

Travis build is fine, I merge.

PR #1183, opened 2017-03-17
Inherit load from SimpleExtension for flexibility

rizar:
Inheriting `Load` from `SimpleExtension` has an advantage that it can be enabled/disabled by calling `SimpleExtension.set_conditions()`. It should be a super easy PR to review.

### DIFF
--- a/blocks/extensions/saveload.py
+++ b/blocks/extensions/saveload.py
@@ -2,7 +2,7 @@
 import os.path
 import logging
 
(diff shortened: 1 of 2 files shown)


dmitriy-serdyuk:
It looks fine but what happened to the python2 tests?

rizar:
I fixed the expected exception in the convolution tests, now should work.

rizar:
Since @dmitriy-serdyuk said it looks find, I am merging.

PR #1185, opened 2017-03-30
Postpost compilation in DatasetEvaluator

rizar:
For debugging it's much better if the compilation in `DatasetEvaluator` happens when it's first called, not when it's created.

### DIFF
--- a/blocks/monitoring/evaluators.py
+++ b/blocks/monitoring/evaluators.py
@@ -255,7 +255,6 @@ class DatasetEvaluator(object):
         self.monitored_quantities_buffer = MonitoredQuantityBuffer(
             monitored_quantities)
         self.updates = updates
-        self._compile()
 
     def _compile(self):
         """Compiles Theano functions.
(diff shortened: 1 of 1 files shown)


dmitriy-serdyuk:
Hmm. As for me, it is better to crash sooner (if something is wrong with evaluation) rather than later.

rizar:
You can then trigger DataStreamMonitoring on before_training. This way you
will first do the monitoring and only then compile the training algorithm.

dmitriy-serdyuk:
Yes, but evaluation might be very slow. It is always frustrating when model compiles for several minutes, then evaluates for half an hour and then crashes with a first batch of training. And if you interrupt it during evaluation you'll have to wait again for compilation. 

I would explicitly compile the evaluator on `before_training` callback.

rizar:
Let's discuss this offline.

On Thu, 30 Mar 2017 at 11:24 dmitriy-serdyuk <notifications@github.com>
wrote:

Yes, but evaluation might be very slow. It is always frustrating when model
compiles for several minutes, then evaluates for half an hour and then
crashes with a first batch of training. And if you interrupt it during
evaluation you'll have to wait again for compilation.

I would explicitly compile the evaluator on before_training callback.

—
You are receiving this because you authored the thread.
Reply to this email directly, view it on GitHub
<https://github.com/mila-udem/blocks/pull/1185#issuecomment-290446089>, or mute
the thread

dmitriy-serdyuk:
LGTM then

PR #1187, opened 2017-04-15
fix the crash with 'iteration_state'

rizar:
This PR fixes the issue with `iteration_state`: one could not access it before training.


dmitriy-serdyuk:
LGTM

PR #1190, opened 2017-05-25
perplexity aggregation scheme

rizar:
This PR does two nice things: (a) improves the test of the `Mean` aggregation scheme to check what actually happens when batch sizes are different (b) adds and tests a `Perplexity` scheme, which in fact was the original motivation for the aggregation schemes 2 years ago..

@dmitriy-serdyuk , can you please review this?

### DIFF
--- a/blocks/monitoring/aggregation.py
+++ b/blocks/monitoring/aggregation.py
@@ -163,8 +163,7 @@ def perplexity(log_likelihood, n_examples):
     """Perplexity for total log_likelihood of n_examples."""
     variable = tensor.exp(-log_likelihood / n_examples)
     variable.tag.aggregation_scheme = Perplexity(log_likelihood, n_examples)
-    # There is no good default name for the output variable,
-    # so we don't try set any.
+    variable.name = 'perplexity'
     return variable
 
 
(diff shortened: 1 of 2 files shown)


blocks/monitoring/aggregation.py
```diff
@@ -151,6 +159,15 @@ def mean(numerator, denominator=1.):
     return variable
 
 
+def perplexity(log_likelihood, n_examples):
+    """Perplexity for total log_likelihood of n_examples."""
+    variable = tensor.exp(-log_likelihood / n_examples)
+    variable.tag.aggregation_scheme = Perplexity(log_likelihood, n_examples)
+    # There is no good default name for the output variable,
+    # so we don't try set any.
```
dmitriy-serdyuk: Why not name it 'perplexity'?

rizar: At first I thought and it would be not nice to return always the same name. But maybe it's an okay default. Will change.

rizar:
Your only comment was addressed, can you please look again?

dmitriy-serdyuk:
Ok, LGTM.

PR #1191, opened 2017-06-05
Make the main loop independent of theano

dmitriy-serdyuk:
Now the main loop doesn't import theano, therefore doesn't reserve a GPU.

Breaking change (probably minor): theano-related utilities are put into a separate file.

@rizar , can you take a look when you have time?


dmitriy-serdyuk:
There is a recommended [hack](https://stackoverflow.com/questions/2447353/getattr-on-a-module) to do `getattr` on a module. So, it is possible to save the API of `utils` module. But I don't think that we should do it.

dmitriy-serdyuk:
Now waiting for #1192

dmitriy-serdyuk:
Ok, test passes. So, what has been changed:

1. The algorithm checks if it is compatible with the model, not the main loop
2. It allows us to make the main loop almost absolutely independent of the algorithms
3. In order to avoid import theano in the main loop, I split the `blocks.utils` into two submodules
4. `blocks.utils.theano_utils` is responsible for shared variables, floatx, etc
5. `blocks.utils.utils` contains everything else
6. It still can be imported from `blocks.utils` for backward compatibility but every submodule is loaded only if needed

tests/algorithms/test_algorithms.py
```diff
@@ -11,7 +11,7 @@
                                Momentum, AdaDelta, BasicRMSProp, RMSProp, Adam,
                                AdaGrad, RemoveNotFinite, Restrict,
                                UpdatesAlgorithm)
-from blocks.utils import shared_floatx, shared_floatx_zeros
+from blocks.utils.theano_utils import shared_floatx, shared_floatx_zeros
```
rizar: As far as I understood you added a hack to `blocks/utils.py` so that one can still write `from blocks.utils import stuff`. If this is correct, why are you accessing `theano_utils` directly here?

blocks/main_loop.py
```diff
@@ -151,12 +149,7 @@ def run(self):
         # reset `profile.current`. Otherwise, it simply does not hurt.
         self.profile.current = []
 
-        # Sanity check for the most common case
-        if (self._model and isinstance(self._model, Model) and
```
rizar: I think that it would be better to move this sanity check to `Model` and not to `GradientDescent`. Algorithms are the cleanest part of Blocks and I would like to avoid code bloat there.

tests/algorithms/test_algorithms.py
dmitriy-serdyuk: Right, it is not necessary.

blocks/main_loop.py
dmitriy-serdyuk: Ok, I can do it

dmitriy-serdyuk: Fixed in 6a2d290

blocks/model.py
```diff
@@ -84,6 +85,14 @@ def __init__(self, *args, **kwargs):
                 parameter_list.append((parameter.name, parameter))
         self._parameter_dict = OrderedDict(parameter_list)
 
+    def check_sanity(self, algorithm):
+        # Sanity check for the most common case
+        if (self and isinstance(self, Model) and
```
rizar: `isinstance(self, Model)` is not needed any more

tests/utils/test_utils.py
```diff
@@ -78,3 +78,10 @@ def test_find_first_and_second_and_third_level(self):
         assert self.mlp.activations[2] in found
         assert self.mlp.activations[1].children[0] in found
         assert self.mlp.activations[3].children[0].children[0]
+
+
+def test_imports():
```
rizar: Would you like to check there that 'import blocks.utils` doesn't load `theano`?

dmitriy-serdyuk: I do, but it is tricky because of the way how nosetests works. There is no guarantee which order the tests run and it uses the same python interpreter. Therefore theano might be still loaded. Not sure how to solve this.

rizar:
Hmm, OK, I see. If there is no way to test then we can just go ahead. But
maybe you can do `del sys.modules['theano']` before running this test?

dmitriy-serdyuk:
I don't think it will work without some black magic:
```python
In [3]: 'theano' in sys.modules
Out[3]: True

In [4]: del sys.modules['theano']

In [5]: 'theano' in sys.modules
Out[5]: False

In [6]: import theano
AttributeError                            Traceback (most recent call last)
<ipython-input-6-3397704bd624> in <module>()
----> 1 import theano

/Users/dima/miniconda2/lib/python3.5/site-packages/theano/__init__.py in <module>()
     98 # needed during that phase.
     99 import theano.tests
--> 100 if hasattr(theano.tests, "TheanoNoseTester"):
    101     test = theano.tests.TheanoNoseTester().test
    102 else:

AttributeError: module 'theano' has no attribute 'tests'
```

rizar:
Okay, screw it.

dmitriy-serdyuk:
Ok, it seems to be working like this.

rizar:
Looks good to me. I can't understand the output of Scrutinizer though...

dmitriy-serdyuk:
It looks like it tries to check binaries and scripts.

PR #1192, opened 2017-06-05
Catch any exception on old backend import

dmitriy-serdyuk:
Theano 9.0 crashes when trying to import the old backend.

### DIFF
--- a/blocks/serialization.py
+++ b/blocks/serialization.py
@@ -127,7 +127,7 @@ from six.moves import cPickle
 import theano
 try:
     from theano.sandbox.cuda import cuda_ndarray
-except ImportError:
+except:
     cuda_ndarray = None
 try:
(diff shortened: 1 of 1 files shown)


dmitriy-serdyuk:
@rizar, this is an important one.

rizar:
Can you please change to "except Exception" in order not to catch `KeyboardInterrupt` and other stuff like that? 

https://stackoverflow.com/questions/730764/try-except-in-python-how-do-you-properly-ignore-exceptions

After that, LGTM

dmitriy-serdyuk:
I wanted to say that `KeyboardInterrupt` never happens in import section, but considering the slowness of import of theano sometimes... I'll change it.

rizar:
That's precisely how I reasoned about it :)

PR #1194, opened 2017-07-07
Filter by unbound application and call_id

rizar:
This PR does two things:

1) It makes it possible to filter by an unbound application, e.g. `Linear.apply`
2) It introduces an optional keyword argument `call_id` that writes a call identifier to the `metadata` attribute of the `ApplicationCall` object. `VariableFilter` is modified to support filtering by this new attribute. The usecase that I have in mind is when an application method is called twice, e.g. an RNN is used to encode both the question the answer. It is useful in such cases to make a distinction between the variables created in the first and in the second call.

### DIFF
--- a/blocks/bricks/base.py
+++ b/blocks/bricks/base.py
@@ -247,6 +247,7 @@ class Application(object):
     def apply(self, bound_application, *args, **kwargs):
         as_dict = kwargs.pop('as_dict', False)
         as_list = kwargs.pop('as_list', False)
+        call_id = kwargs.pop('call_id', None)
         if as_list and as_dict:
             raise ValueError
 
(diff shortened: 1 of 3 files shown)


dmitriy-serdyuk:
I have a question about 2. Does the call id propagate recursively to applications inside the top level application? Do you plan to annotate every brick with the call id?

It seems that this feature needs to be documented. Maybe in some future PR.

rizar:
>  Does the call id propagate recursively to applications inside the top level application? 

Nope.

> Do you plan to annotate every brick with the call id?

Nope, call id can be passed like `brick.apply(x,y,z, call_id='123')` and it will be assigned to _application call_ not to a brick. As for documenting it, it is not very clear where it should be documented: `as_dict` and `as_list`, which are implemented in a very similar way, aren't documented.

rizar:
So, what's your opinion, @dmitriy-serdyuk ?

dmitriy-serdyuk:
It looks good. But I don't understand why the test fails.

rizar:
It does not seem though that failing test are due to this PR. That said, I
will go ahead and merge it.

PR #1198, opened 2018-08-29
Remove --allow-external

dmitriy-serdyuk:
For mila-udem/fuel#405

@nouiz

### DIFF
--- a/.scrutinizer.yml
+++ b/.scrutinizer.yml
@@ -118,7 +118,6 @@ checks:
         classes_non_parent_init_called: true
         classes_non_iterator_returned: true
         classes_no_method_argument: true
-        classes_no_method_argument: true
         classes_no_init: true
         classes_missing_interface_method: true
         classes_method_hidden: true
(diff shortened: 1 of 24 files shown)


dmitriy-serdyuk:
Fixed

nouiz:
2 CI still fail. I think a new version of pep8 is used. That caused a failure in Fuel

dmitriy-serdyuk:
Ok, I figured out the problem. Virtualenv puts a bunch of its things into `bin` directory. Scrutinizer crashes on those files.

dmitriy-serdyuk:
@nouiz I give up. I was tinkering for quite a while. I fixed Travis and Scrutinizer tests. But I have no idea what happens to Scrutinizer analysis.

nouiz:
So, what do we do? As the tests passed, we could merge. This would fix possible security issue, so I would go in that direction.

dmitriy-serdyuk:
OK, let me cleanup history then

